PR #18, opened 2025-04-22
CLOUDP-309402: MEKO MCK migration + rename enterprise resources

Julien-Ben:
# Summary

## Short Summary

This PR contains 2 main set of changes:

- A new upgrade test, from MEKO to MCK. It runs in single and multi cluster mode. The test deploys a replicaset, and perform the upgrade steps recommended for migration.
- A rename of most components created by the operator, such as service accounts, to make the upgrade possible without conflict between old and new resources. When the MCK helm chart is installed, new service accounts are created and the operator updates the stateful sets so that the pods use the new ones.

### Renaming
- Introduced many variables for resource names.
- Service Accounts, RBAC Resources, Services
- Operator name, deployment name
- Selectors and labels stay the same (`mongodb-enterprise-operator`) because for some resources it is impossible to change them without deleting it.
- Some components kept their old names such as config map and secrets for multi-cluster.

### Tests
- New helper functions to install the new MCK operator, to handle the two deployments. 
- Updated all existing upgrade tests to handle both deployments running concurrently (scale down first operator before installing the second one). This means that our existing tests are all testing a MEKO to MCK upgrade, and some specific ones are testing other paths, such as the sharded cluster upgrade which is doing 1.27 -> MCK.
- New MEKO to MCK dedicated test, working in both single and multi cluster setting. Deploys a replica set and ensure the migration happens correctly.

## Note

It is hard to ensure the renaming is completely exhaustive. The upgrade tests ensure we renamed necessary resources to ensure a smooth upgrade with no conflicts. Follow up PRs are/will be opened to rename smaller things like file names.

There is still one e2e test failing that we are currently investigating, but we preferred to open the PR early for review.

## Proof of Work

E2E test

## Checklist

### DIFF
--- a/public/tools/multicluster/cmd/recover.go
+++ b/public/tools/multicluster/cmd/recover.go
@@ -16,7 +16,7 @@ func init() {
 	multiclusterCmd.AddCommand(recoverCmd)
 
 	recoverCmd.Flags().StringVar(&common.MemberClusters, "member-clusters", "", "Comma separated list of member clusters. [required]")
-	recoverCmd.Flags().StringVar(&RecoverFlags.ServiceAccount, "service-account", "mongodb-kubernetes-operator-multi-cluster", "Name of the service account which should be used for the Operator to communicate with the member clusters. [optional, default: mongodb-enterprise-operator-multi-cluster]")
+	recoverCmd.Flags().StringVar(&RecoverFlags.ServiceAccount, "service-account", "mongodb-kubernetes-operator-multi-cluster", "Name of the service account which should be used for the Operator to communicate with the member clusters. [optional, default: mongodb-kubernetes-operator-multi-cluster]")
 	recoverCmd.Flags().StringVar(&RecoverFlags.CentralCluster, "central-cluster", "", "The central cluster the operator will be deployed in. [required]")
 	recoverCmd.Flags().StringVar(&RecoverFlags.MemberClusterNamespace, "member-cluster-namespace", "", "The namespace the member cluster resources will be deployed to. [required]")
 	recoverCmd.Flags().StringVar(&RecoverFlags.CentralClusterNamespace, "central-cluster-namespace", "", "The namespace the Operator will be deployed to. [required]")
--- a/public/tools/multicluster/pkg/common/common.go
+++ b/public/tools/multicluster/pkg/common/common.go
@@ -59,7 +59,7 @@ type Flags struct {
 }
 
 const (
-	// When upgrading to MCK, we didn't want to require customers to rename their secet, so we kept the old
+	// KubeConfigSecretName stays the same when upgrading to MCK, as we didn't want to require customers to rename their secret, so we kept the old one.
 	// -enterprise prefix
 	KubeConfigSecretName       = "mongodb-enterprise-operator-multi-cluster-kubeconfig"
 	KubeConfigSecretKey        = "kubeconfig"
@@ -69,7 +69,7 @@ const (
 	AppdbRole                  = "mongodb-kubernetes-appdb"
 	AppdbRoleBinding           = "mongodb-kubernetes-appdb"
 	DefaultOperatorName        = "mongodb-kubernetes-operator"
-	// When upgrading to MCK, we didn't want to require customers to rename their configmap, so we kept the old
+	// DefaultOperatorConfigMapName is the legacy configmap name which stays the same when upgrading to MCK, we didn't want to require customers to rename their configmap, so we kept the old one
 	// -enterprise prefix
 	DefaultOperatorConfigMapName = "mongodb-enterprise-operator-member-list"
 )


docker/mongodb-enterprise-tests/kubetester/mongodb.py
```diff
@@ -152,7 +152,12 @@ def assert_abandons_phase(self, phase: Phase, timeout=None):
         happens during the time we call this method. If there is not a lot of work, then the phase can already finished
         transitioning during the modification call before calling this method.
         """
-        return self.wait_for(lambda s: s.get_status_phase() != phase, timeout, should_raise=True)
```
nammn, line 155:
```suggestion
            f"Abandoning phase {phase.name} for resource {self.__class__.__name__} took {end_time - start_time}s"
```

+ follow-up ticket: its easier to wrap such things into the tracer e.g.

```
@TRACER.start_as_current_span("assert_abandons_phase")
```
which does all that by default

docker/mongodb-enterprise-tests/tests/multicluster/multi_cluster_recover_clusterwide.py
```diff
@@ -326,7 +326,7 @@ def test_recover_operator_remove_cluster(
     mdbb_ns: str,
     central_cluster_client: kubernetes.client.ApiClient,
 ):
-    return_code = run_multi_cluster_recovery_tool(member_cluster_names[:-1], namespace, namespace, True)
+    return_code = run_multi_cluster_recovery_tool(member_cluster_names[:-1], namespace, mdba_ns, True)
```
nammn, line 329: Note for reviewer; 

we are referencing the `membernamespace` which contains the `rbac` and not the `masternamespace`. It doesn't seem to make sense to reference the normal namespace, since the membernamespace`` runs the workload - so we want to use that one instead.
For some reason this passes on master. There is something fishy going on. It shouldn't pass on master...

docker/mongodb-enterprise-tests/tests/upgrades/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+from kubernetes import client
+from kubernetes.client import ApiException
+from tests import test_logger
+from tests.conftest import log_deployments_info
+
+logger = test_logger.get_test_logger(__name__)
+
+
+# Scale down a deployment to 0. This is useful for upgrade tests as long as they install two different operators in
```
nammn, line 11: follow-up ticket: remove for mck1.1?

helm_chart/templates/operator-configmap.yaml
```diff
@@ -13,5 +13,10 @@ metadata:
   name: mongodb-enterprise-operator-member-list
   labels:
     multi-cluster: "true"
+  # It is safer for helm to never automatically delete the configmap when rendering. Helm is not the only owner of
```
nammn, line 16: note for reviewer: there a ticket to handle it differently; https://jira.mongodb.org/browse/CLOUDP-315085

public/tools/multicluster/cmd/recover.go
```diff
@@ -16,7 +16,7 @@ func init() {
 	multiclusterCmd.AddCommand(recoverCmd)
 
 	recoverCmd.Flags().StringVar(&common.MemberClusters, "member-clusters", "", "Comma separated list of member clusters. [required]")
-	recoverCmd.Flags().StringVar(&RecoverFlags.ServiceAccount, "service-account", "mongodb-enterprise-operator-multi-cluster", "Name of the service account which should be used for the Operator to communicate with the member clusters. [optional, default: mongodb-enterprise-operator-multi-cluster]")
+	recoverCmd.Flags().StringVar(&RecoverFlags.ServiceAccount, "service-account", "mongodb-kubernetes-operator-multi-cluster", "Name of the service account which should be used for the Operator to communicate with the member clusters. [optional, default: mongodb-enterprise-operator-multi-cluster]")
```
nammn:
```suggestion
	recoverCmd.Flags().StringVar(&RecoverFlags.ServiceAccount, "service-account", "mongodb-kubernetes-operator-multi-cluster", "Name of the service account which should be used for the Operator to communicate with the member clusters. [optional, default: mongodb-kubernetes-operator-multi-cluster]")
```

public/tools/multicluster/pkg/common/common.go
```diff
@@ -59,15 +59,19 @@ type Flags struct {
...
-	KubeConfigSecretName         = "mongodb-enterprise-operator-multi-cluster-kubeconfig"
-	KubeConfigSecretKey          = "kubeconfig"
-	AppdbServiceAccount          = "mongodb-enterprise-appdb"
-	DatabasePodsServiceAccount   = "mongodb-enterprise-database-pods"
-	OpsManagerServiceAccount     = "mongodb-enterprise-ops-manager"
-	AppdbRole                    = "mongodb-enterprise-appdb"
-	AppdbRoleBinding             = "mongodb-enterprise-appdb"
-	DefaultOperatorName          = "mongodb-enterprise-operator"
-	DefaultOperatorConfigMapName = DefaultOperatorName + "-member-list"
+	// When upgrading to MCK, we didn't want to require customers to rename their secet, so we kept the old
```
nammn:
```suggestion
	// KubeConfigSecretName stays the same when upgrading to MCK, as we didn't want to require customers to rename their secret, so we kept the old one.
```

public/tools/multicluster/pkg/common/common.go
```diff
@@ -59,15 +59,19 @@ type Flags struct {
...
+	// -enterprise prefix
+	KubeConfigSecretName       = "mongodb-enterprise-operator-multi-cluster-kubeconfig"
+	KubeConfigSecretKey        = "kubeconfig"
+	AppdbServiceAccount        = "mongodb-kubernetes-appdb"
+	DatabasePodsServiceAccount = "mongodb-kubernetes-database-pods"
+	OpsManagerServiceAccount   = "mongodb-kubernetes-ops-manager"
+	AppdbRole                  = "mongodb-kubernetes-appdb"
+	AppdbRoleBinding           = "mongodb-kubernetes-appdb"
+	DefaultOperatorName        = "mongodb-kubernetes-operator"
+	// When upgrading to MCK, we didn't want to require customers to rename their configmap, so we kept the old
```
nammn:
```suggestion
	// DefaultOperatorConfigMapName is the legacy configmap name which stays the same when upgrading to MCK, we didn't want to require customers to rename their configmap, so we kept the old one
```

config/manager/manager.yaml
lucian-tosa, line 1: I see that the binary name is still mongodb-enterprise-operator. Should we change our Docker stuff to the new name as well?

docker/mongodb-enterprise-tests/tests/multicluster/multi_cluster_recover_clusterwide.py
Julien-Ben: Follow up ticket https://jira.mongodb.org/browse/CLOUDP-316084

docker/mongodb-enterprise-tests/tests/upgrades/__init__.py
Julien-Ben: Done: https://jira.mongodb.org/browse/CLOUDP-316400

controllers/operator/construct/appdb_construction.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 const (
-	appDBServiceAccount    = "mongodb-enterprise-appdb"
+	appDBServiceAccount    = "mongodb-kubernetes-appdb"
 	InitAppDbContainerName = "mongodb-enterprise-init-appdb"
```
lucian-tosa, line 36: what about this one?

docker/mongodb-enterprise-tests/kubetester/mongodb.py
nammn: https://jira.mongodb.org/browse/CLOUDP-316401

api/v1/mdbmulti/mongodb_multi_types.go
```diff
@@ -169,7 +169,7 @@ func (m *MongoDBMultiCluster) ObjectKey() client.ObjectKey {
 
 func (m *MongoDBMultiCluster) GetOwnerLabels() map[string]string {
 	return map[string]string{
-		util.OperatorLabelName: util.OperatorName,
+		util.OperatorLabelName: util.OperatorLabelValue,
```
SimonBaeumer, line 172: I understood from the PR description that labels should stay the same, but here the value is changing. What is the reason for this change?

helm_chart/templates/operator-roles.yaml
```diff
@@ -176,15 +176,15 @@ rules:
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: {{ .Values.operator.name }}-{{ include "mongodb-enterprise-operator.namespace" . }}-webhook-binding
+  name: {{ .Values.operator.name }}-{{ include "mongodb-kubernetes-operator.namespace" . }}-webhook-binding
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: mongodb-enterprise-operator-mongodb-webhook
+  name: {{.Values.operator.baseName}}-operator-mongodb-webhook
```
SimonBaeumer, line 183: Is this a correct change? In most other cases `mongodb-enterprise` was changed to `mongodb-kubernetes`, here it is not.
Also the formatting looks off for go templating, the spaces between the braces are missing?

mongodb-community-operator/test/e2e/setup/test_config.go
```diff
@@ -37,7 +37,7 @@ type TestConfig struct {
 func LoadTestConfigFromEnv() TestConfig {
 	return TestConfig{
 		OperatorImage: "mongodb-kubernetes",
-		Namespace:     envvar.GetEnvOrDefault(testNamespaceEnvName, "mongodb"), // nolint:forbidigo
+		Namespace:     envvar.GetEnvOrDefault(testNamespaceEnvName, "mongodb-test"), // nolint:forbidigo
```
SimonBaeumer, line 40: huh? is this intentional?

config/manager/manager.yaml
nammn: i have another pr that does this

https://github.com/mongodb/mongodb-kubernetes/pull/52
+
https://github.com/mongodb/mongodb-kubernetes/pull/59

but we cannot merge them unless the changeset from julien has been merged to not cause merge conflicts

api/v1/mdbmulti/mongodb_multi_types.go
Julien-Ben: It's a bit hard to track, but I created `OperatorLabelValue` constant and assigned it the legacy operator name (mongodb enterprise), so this value ultimately stay the same.
You can see in `constants.go`

docker/mongodb-enterprise-tests/tests/upgrades/meko_mck_upgrade.py
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        if try_load(resource):
+            return resource
+        return resource.create()
+
+
+# Installs the latest officially released version of MEKO, from Quay
+@mark.e2e_meko_mck_upgrade
+def test_install_latest_official_operator(official_operator: Operator, namespace: str):
```
lsierant, line 95: how this test is different from other operator_upgrade_* tests?

pkg/util/constants.go
```diff
@@ -75,9 +75,11 @@ const (
 	EnvVarSSLTrustedMMSServerCertificate = "SSL_TRUSTED_MMS_SERVER_CERTIFICATE"
 
 	// Pod/StatefulSet specific constants
+	OperatorName                   = "mongodb-kubernetes-operator"
+	LegacyOperatorName             = "mongodb-enterprise-operator" // Still used for some selectors and labels
+	MultiClusterOperatorName       = "mongodb-kubernetes-operator-multi-cluster"
 	OperatorLabelName              = "controller"
-	OperatorName                   = "mongodb-enterprise-operator"
-	MultiClusterOperatorName       = "mongodb-enterprise-operator-multi-cluster"
+	OperatorLabelValue             = LegacyOperatorName
```
lsierant, line 82: can we name it OperatorNameLabelValue

docker/mongodb-enterprise-tests/tests/upgrades/meko_mck_upgrade.py
Julien-Ben: It it the same, I wrote it initially to quickly make a POC of the migration, but it does the same things as our existing upgrade tests.

I'd like to keep it because in next release our upgrade tests will test 1.0 -> 1.1 and this test could stay around to test MEKO 1.33 -> MCK Last version (e.g 1.1)

I mentioned it in the follow up ticket related to upgrade tests

helm_chart/templates/operator-roles.yaml
Julien-Ben: Except for the `.namespace` values where I couldn't do it, I tried to replace all mention of the operator name with that variable, which is assigned "mongodb-kubernetes" as value.

For formatting it indeed looks incorrect, I think we should have a linter for that

mongodb-community-operator/test/e2e/setup/test_config.go
Julien-Ben: Yes, our default test namespace is mongodb-test
I needed to update it for community tests

controllers/operator/construct/appdb_construction.go
Julien-Ben: Container naming is independent of this PR, which focuses more on resources installed on the cluster to enable the upgrade (MEKO -> MCK) with no collisions.
I'll check with Nam if this value is correct (but we'll change it in another PR if needed)

I know it's a bit confusing, there's a lot of renaming going on here

PR #19, opened 2025-04-22
flakiness - fix cert rotation tests

nammn:
# Summary
- cherry-pick of https://github.com/10gen/ops-manager-kubernetes/pull/4298 and linked pr will be closed and cherry-picked from here instead

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/docker/mongodb-enterprise-tests/kubetester/certs.py
+++ b/docker/mongodb-enterprise-tests/kubetester/certs.py
@@ -896,7 +896,7 @@ def yield_existing_csrs(csr_names: List[str], timeout: int = 300) -> Generator[s
 
 
 @TRACER.start_as_current_span("assert_certificate_rotation")
-def assert_certificate_rotation(mdb, namespace, certificate_name):
+def rotate_and_assert_certificates(mdb, namespace, certificate_name):
     """
     Verifies certificate rotation completes successfully.
 
--- a/docker/mongodb-enterprise-tests/tests/tls/tls_x509_configure_all_options_sc.py
+++ b/docker/mongodb-enterprise-tests/tests/tls/tls_x509_configure_all_options_sc.py
@@ -2,9 +2,9 @@ import pytest
 from kubetester import find_fixture, try_load
 from kubetester.automation_config_tester import AutomationConfigTester
 from kubetester.certs import (
-    assert_certificate_rotation,
     create_sharded_cluster_certs,
     create_x509_agent_tls_certs,
+    rotate_and_assert_certificates,
 )
 from kubetester.kubetester import KubernetesTester, is_multi_cluster
 from kubetester.mongodb import MongoDB, Phase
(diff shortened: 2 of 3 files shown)


docker/mongodb-enterprise-tests/kubetester/certs.py
```diff
@@ -879,3 +886,54 @@ def yield_existing_csrs(csr_names: List[str], timeout: int = 300) -> Generator[s
...
+
+        return version_increased and all_processes_ready
+
+    timeout = 600
+    KubernetesTester.wait_until(
+        check_version_and_processes,
+        timeout=timeout,
+    )
+
+    from kubetester.mongodb import Phase
```
m1kola: Non-blocking q: any reason not to put this to the top of the file with rest of the imports?

docker/mongodb-enterprise-tests/tests/tls/tls_x509_configure_all_options_sc.py
```diff
@@ -112,32 +100,3 @@ def test_rotate_config_cert_with_sts_restarting(self, sc: MongoDB, namespace: st
...
-            f"{MDB_RESOURCE_NAME}-config-clusterfile",
-        )
-
-    def test_rotate_mongos_certfile_with_sts_restarting(self, sc: MongoDB, namespace: str):
-        sc.trigger_sts_restart("mongos")
-        assert_certificate_rotation(
-            sc,
-            namespace,
-            f"{MDB_RESOURCE_NAME}-mongos-clusterfile",
-        )
```
m1kola, line 138: Could you please explain why we remove all these tests?

docker/mongodb-enterprise-tests/kubetester/certs.py
```diff
@@ -879,3 +886,54 @@ def yield_existing_csrs(csr_names: List[str], timeout: int = 300) -> Generator[s
...
+
+    Rotates the specified certificate and validates that:
+    1. Automation config version increases, as cert changes causes a new ac version
+    2. All MongoDB processes reach the new goal version
+    3. MongoDB instance returns/stays to Running state
+
+    """
+
+    old_ac_version = KubernetesTester.get_automation_config()["version"]
+    rotate_cert(namespace, certificate_name, should_block_until_ready=True)
```
m1kola, line 911: `assert_certificate_rotation` gives an impression that it does not make any changes and only asserts when in fact it does.

What do you think about the following suggestion?
1. Moving these two lines into the caller functions
2. Make `assert_certificate_rotation` accept `old_ac_version` as an argument

docker/mongodb-enterprise-tests/tests/tls/tls_x509_configure_all_options_sc.py
nammn: I should have added a comment! But basically we are testing the same things. We rotate the sts and assert that certs can be rotated at the same time, and also before we test whether we can only rotate the cert. I don't see value in doing those 2 tests after another

docker/mongodb-enterprise-tests/kubetester/certs.py
nammn: renamed! bf2d1833b

m1kola: A bit of context for transparency. We discussed it in DM: while I think it is good to keep mutation (cert rotation) separate from assertions, @nammn wanted to to keep everything together to avoid forgetting to do part of the checks.

At the same time I think it is more important to address the flake than worry about semantics here. We landed on the middle ground: renaming the function to `rotate_and_assert_certificates` to more clearly indicate the purpose.

PR #32, opened 2025-04-23
Recreate kind cluster on every evg host reboot 

lucian-tosa:
# Summary

Evergreen hosts reboot every day or weekend (depending on your configuration). After every reboot, inter-cluster connectivity might be broken for some reason. Recreating the clusters is the only solution we have so far.
This is a systemd service that runs on every boot and recreates all clusters (including the kind-kind for single cluster tests).
Our tunnel command will now also get the kubeconfig from the host otherwise the tunnel won't open to the (new) ports of the recreated clusters.

The systemd service is not created by default. This needs to be explicitly done by running 
```
evg_host.sh configure --auto-reboot
```

Remove the architecture flag, and instead left it to be inferred from `uname`.

## Proof of Work

Tested locally, but it would be nice if someone can checkout this branch and try it for themselves.

## Checklist


anandsyncs (approved):

SimonBaeumer (changes requested):
I am split on this change, it risks loosing data or setups... I understand the desire to fix this, but on the other hand I don't like when my environment gets cleaned up automatically when it was still in use. For example, if I had run patches against deployments or single cluster deployments my environment will be resetted.

Can you add an env var to opt-in for this re-creation? Than each engineer can decide whether they want re-creation or not.

PR #50, opened 2025-04-27
CLOUDP-314901 OIDC CRD changes + validation

MaciejKaras:
# Summary

Adding new `OIDCProviderConfig` struct to `Authentication` struct and new AuthMode `OIDC`.

```go
type OIDCProviderConfig struct {
	// Unique label that identifies this configuration. This label is visible to your Ops Manager users and is used when
	// creating users and roles for authorization. It is case-sensitive and can only contain the following characters:
	//  - alphanumeric characters (combination of a to z and 0 to 9)
	//  - hyphens (-)
	//  - underscores (_)
	// +kubebuilder:validation:Pattern="^[a-zA-Z0-9-_]+$"
	// +kubebuilder:validation:Required
	ConfigurationName string `json:"configurationName"`

	// Issuer value provided by your registered IdP application. Using this URI, MongoDB finds an OpenID Provider
	// Configuration Document, which should be available in the /.wellknown/open-id-configuration endpoint.
	// +kubebuilder:validation:Required
	IssuerURI string `json:"issuerURI"`

	// Entity that your external identity provider intends the token for.
	// Enter the audience value from the app you registered with external Identity Provider.
	// +kubebuilder:validation:Required
	Audience string `json:"audience"`

	// Select GroupMembership to grant authorization based on IdP user group membership, or select UserID to grant
	// an individual user authorization.
	// +kubebuilder:validation:Required
	AuthorizationType OIDCAuthorizationType `json:"authorizationType"`

	// The identifier of the claim that includes the user principal identity.
	// Accept the default value unless your IdP uses a different claim.
	// +kubebuilder:default=sub
	// +kubebuilder:validation:Required
	UserClaim string `json:"userClaim"`

	// The identifier of the claim that includes the principal's IdP user group membership information.
	// Accept the default value unless your IdP uses a different claim, or you need a custom claim.
	// Required when selected GroupMembership as the authorization type, ignored otherwise
	// +kubebuilder:default=groups
	// +kubebuilder:validation:Optional
	GroupsClaim string `json:"groupsClaim,omitempty"`

	// Configure single-sign-on for human user access to Ops Manager deployments with Workforce Identity Federation.
	// For programmatic, application access to Ops Manager deployments use Workload Identity Federation.
	// Only one Workforce Identity Federation IdP can be configured per MongoDB resource
	// +kubebuilder:validation:Required
	AuthorizationMethod OIDCAuthorizationMethod `json:"authorizationMethod"`

	// Unique identifier for your registered application. Enter the clientId value from the app you
	// registered with an external Identity Provider.
	// Required when selected Workforce Identity Federation authorization method
	// +kubebuilder:validation:Optional
	ClientId string `json:"clientId,omitempty"`

	// Tokens that give users permission to request data from the authorization endpoint.
	// Only used for Workforce Identity Federation authorization method
	// +kubebuilder:validation:Optional
	RequestedScopes []string `json:"requestedScopes,omitempty"`
}

// +kubebuilder:validation:Enum=GroupMembership;UserID
type OIDCAuthorizationType string

// +kubebuilder:validation:Enum=WorkforceIdentityFederation;WorkloadIdentityFederation
type OIDCAuthorizationMethod string
```

⚠️ Because `Security.Authentication` struct is reused also in AppDBSpec it will be available there as well. It will be as usual overridden in https://github.com/mongodb/mongodb-kubernetes/blob/f0050b8942545701e8cb9e42d54d14f0cb58ee6a/api/v1/om/opsmanager_types.go#L622
I believe it is worth noting this behaviour, but not change it as part of this project.

## Proof of Work

New Unit test that verify validation and another set of webhook tests are under way.

## Next steps

- [ ] Add release notes
- [ ] Add validation tests based on webhooks
- [ ] Start discussion with docs team

## Checklist

  * **q**: Clarifying question.


api/v1/mdb/mongodb_validation.go
```diff
@@ -105,9 +106,149 @@ func scramSha1AuthValidation(d DbCommonSpec) v1.ValidationResult {
...
+			return v1.ValidationError("%s", msg)
+		}
+
+		return v1.ValidationSuccess()
+	}
+}
+
+func oidcProviderConfigIssuerURIValidator(config OIDCProviderConfig) func(DbCommonSpec) v1.ValidationResult {
+	return func(_ DbCommonSpec) v1.ValidationResult {
+		ok, url := util.IsURL(config.IssuerURI)
```
SimonBaeumer: Why not returning the URL parsing error to the user?

SimonBaeumer: Instead of returning a bool, returning an error with a message would improve the API response.

MaciejKaras: Right, agree!

MaciejKaras: Done!

api/v1/mdb/mongodb_validation.go
```diff
@@ -105,9 +106,149 @@ func scramSha1AuthValidation(d DbCommonSpec) v1.ValidationResult {
...
+		for _, config := range configs {
+			if config.AuthorizationMethod == OIDCAuthorizationMethodWorkforceIdentityFederation {
+				workforceIdentityFederationConfigs = append(workforceIdentityFederationConfigs, config.ConfigurationName)
+			}
+		}
+
+		if len(workforceIdentityFederationConfigs) > 1 {
+			msg := fmt.Sprintf("Only one OIDC provider config can be configured with Workforce Identity Federation. "+
+				"The following configs are configured with Workforce Identity Federation: %s", strings.Join(workforceIdentityFederationConfigs, ", "))
+			return v1.ValidationError("%s", msg)
```
lucian-tosa: Why not combined?

pkg/util/constants.go
```diff
@@ -147,6 +147,7 @@ const (
 	MONGODBCR                         = "MONGODB-CR"
 	SCRAMSHA256                       = "SCRAM-SHA-256"
 	LDAP                              = "LDAP"
+	OIDC                              = "OIDC"
```
lucian-tosa, line 155: I thought it was MONGODB-OIDC

MaciejKaras: It is used in CRD, but for automation config we send `MONGODB-OIDC`. I think it doesn't make sense to use full name `MONGODB-OIDC` in OIDC, because users don't care how it is translated underneath

api/v1/mdb/mongodb_validation.go
MaciejKaras: Yeah, something must have been wrong with the Intellij cmd for extracting variables. Fixing

pkg/util/constants.go
lucian-tosa: I think we should keep it MONGODB-OIDC. When they attempt to use mongosh, or pymongo or any other driver, OIDC shows up as MONGODB-OIDC. It might be confusing for customers

MaciejKaras: We already have a convenience to use different values than in mongosh and automation config. For example we use `PLAIN` for Ldap in mongosh and AC, but in our CRD we use `LDAP`. Same with [MONGODB-X509](https://www.mongodb.com/docs/manual/core/security-x.509/#std-label-security-auth-x509) - in CRD we use X509.

https://www.mongodb.com/docs/mongodb-shell/reference/options/#std-option-mongosh.--authenticationMechanism

lucian-tosa: Ah, fair enough

api/v1/mdb/mongodb_types.go
```diff
@@ -801,6 +807,22 @@ func (s *Security) IsTLSEnabled() bool {
 	return s.CertificatesSecretsPrefix != ""
 }
 
+func (s *Security) IsOIDCEnabled() bool {
```
anandsyncs, line 810: nit, also just an opinion:

```suggestion
func (s *Security) IsOIDCEnabled() bool {
	if s == nil || s.Authentication == nil || !s.Authentication.Enabled {
		return false
	}
	return s.Authentication.IsOIDCEnabled()
}
```

anandsyncs (approved):

vinilage (approved):

api/v1/mdb/mongodb_validation.go
```diff
@@ -105,9 +105,149 @@ func scramSha1AuthValidation(d DbCommonSpec) v1.ValidationResult {
...
+			if config.GroupsClaim != "" {
+				return v1.ValidationWarning("GroupsClaim will be ignored in OIDC provider config %q when using User ID authorization", config.ConfigurationName)
+			}
+		}
+
+		return v1.ValidationSuccess()
+	}
+}
+
 func ldapAuthRequiresEnterprise(d DbCommonSpec) v1.ValidationResult {
```
anandsyncs, line 257: Do we need this check for OIDC as well?

lucian-tosa: I think we do

PR #51, opened 2025-04-27
Authorization package refactor

MaciejKaras:
# Summary

This refactor was influenced by OIDC integration. It was really hard to understand the code and it was error prone. List of issues this refactor solves:
 - authentication mechanisms were constructed with `Options`, `om.Connection` and `om.AutomationConfig` values. Additionally `Options` and `om.AutomationConfig` was provided through interface methods thus leading to multiple ways of updating the AutomationConfig. I have unified all mechanisms and they now only use what is provided through interface methods. Mechanism should only posses logic not the state.
 ```go
// Mechanism is an interface that needs to be implemented for any Ops Manager authentication mechanism
type Mechanism interface {
	EnableAgentAuthentication(conn om.Connection, opts Options, log *zap.SugaredLogger) error
	DisableAgentAuthentication(conn om.Connection, log *zap.SugaredLogger) error
	EnableDeploymentAuthentication(conn om.Connection, opts Options, log *zap.SugaredLogger) error
	DisableDeploymentAuthentication(conn om.Connection, log *zap.SugaredLogger) error
	// IsAgentAuthenticationConfigured should not rely on util.MergoDelete since the method is always
	// called directly after deserializing the response from OM which should not contain the util.MergoDelete value in any field.
	IsAgentAuthenticationConfigured(ac *om.AutomationConfig, opts Options) bool
	IsDeploymentAuthenticationConfigured(ac *om.AutomationConfig, opts Options) bool
	GetName() MechanismName
}
```
- all `Mechanism` interface methods have to utilise `om.Connection` to retrieve `om.AutomationConfig`. There is no magic that allows to modify internal `om.AutomationConfig` stored in `Mechanism` struct
- moved all Mechanism related code to separate file `authorization_mechanism.go`
- code was sometimes depending on the `MechanismName` and sometimes on `Mechanism` itself. Because `Mechanism` is now just a logic holder we use it instead of `MechansmName` in all places
- cleanup lots of code
- fixed spelling and naming i.e. `unrequired` mechanism

## Checklist

  * **q**: Clarifying question.

### CODE REVIEW

File: controllers/operator/authentication/ldap.go [hunk 1]
```diff
@@ -74,39 +68,40 @@ func (l *ldapAuthMechanism) DisableAgentAuthentication(log *zap.SugaredLogger) e
...
 	}, log)
 }
 
-func (l *ldapAuthMechanism) EnableDeploymentAuthentication(opts Options) error {
-	ac := l.AutomationConfig
-	ac.Ldap = opts.Ldap
-	if !stringutil.Contains(ac.Auth.DeploymentAuthMechanisms, string(LDAPPlain)) {
-		ac.Auth.DeploymentAuthMechanisms = append(ac.Auth.DeploymentAuthMechanisms, string(LDAPPlain))
-	}
+func (l *ldapAuthMechanism) EnableDeploymentAuthentication(conn om.Connection, opts Options, log *zap.SugaredLogger) error {
```
anandsyncs, line 83: Why is it better to pass `conn` and `log` as parameters instead of using composition?

MaciejKaras: Could you show me the example in the code with your changes? I don't fully follow how to utilise composition here

controllers/operator/authentication/authentication_mechanism.go
```diff
@@ -0,0 +1,118 @@
...
+		// if no authentication has been configured, the default value for "AutoAuthMechanism" is "MONGODB-CR"
+		// even if authentication is disabled, so we need to ensure that auth has been enabled.
+		if ac.Auth.AutoAuthMechanism == string(MongoDBCR) && ac.Auth.IsEnabled() {
+			return MongoDBCRMechanism
+		}
+		return ScramSha256Mechanism
+	}
+
+	// this should never be reached as validation of this string happens at the CR level
+	panic(xerrors.Errorf("unknown mechanism name %s", mechanismModeInCR))
```
SimonBaeumer, line 117: Why panic-ing and not returning an error? A panic will crash the application and turn into a down time.

controllers/operator/authentication/configure_authentication_test.go
```diff
@@ -200,17 +202,29 @@ func assertAuthenticationMechanism(t *testing.T, auth *om.Auth, mechanism string
...
+	require.NoError(t, err)
+	assert.True(t, authMechanism.IsDeploymentAuthenticationConfigured(ac, opts))
 }
 
-func assertAgentAuthenticationDisabled(t *testing.T, authMechanism Mechanism, opts Options) {
-	_ = authMechanism.EnableAgentAuthentication(opts, zap.S())
-	assert.True(t, authMechanism.IsAgentAuthenticationConfigured())
+func assertAgentAuthenticationDisabled(t *testing.T, authMechanism Mechanism, conn om.Connection, opts Options) {
+	err := authMechanism.EnableAgentAuthentication(conn, opts, zap.S())
+	require.NoError(t, err)
```
SimonBaeumer, line 216: props for using `require`!

controllers/operator/authentication/authentication_mechanism.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"go.uber.org/zap"
+	"golang.org/x/xerrors"
+
+	"github.com/mongodb/mongodb-kubernetes/controllers/om"
+	"github.com/mongodb/mongodb-kubernetes/pkg/util"
+)
+
+// Mechanism is an interface that needs to be implemented for any Ops Manager authentication mechanism
+type Mechanism interface {
```
SimonBaeumer, line 15: +1 for introducing an abstraction interface here. I think this makes it simpler for following engineers to understand how to add more auth mechanisms!

With regards to the connection. I think the downside of passing it down to each function is that a caller always needs access to the `om.Connection` instance which can lead to passing down connections often in the stack. Potentially CloudManager and OpsManager could diverge and the connection becomes an implementation detail.

Not a concern though for now but saw this in other code bases if connections were passed in func arguments.

controllers/operator/authentication/ldap.go
```diff
@@ -9,23 +9,17 @@ import (
...
+var LDAPPlainMechanism = &ldapAuthMechanism{}
 
-func NewLdap(conn om.Connection, ac *om.AutomationConfig, opts Options) Mechanism {
-	return &ldapAuthMechanism{
-		AutomationConfig: ac,
-		Conn:             conn,
-		Options:          opts,
-	}
+type ldapAuthMechanism struct{}
+
```
SimonBaeumer: To ensure at compile time that a struct implements an interface you can add `var _ foo.RequiredInterface = myType{}` as an assertion in the code.

https://stackoverflow.com/questions/10498547/ensure-a-type-implements-an-interface-at-compile-time-in-go

controllers/operator/authentication/ldap.go
```diff
@@ -9,23 +9,17 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/stringutil"
 )
 
-type ldapAuthMechanism struct {
-	AutomationConfig *om.AutomationConfig
-	Conn             om.Connection
-	Options          Options
-}
+var LDAPPlainMechanism = &ldapAuthMechanism{}
```
SimonBaeumer: Hm, this looks an anti-pattern to me to introduce a global variable. Now I also understand why you are extracted the connection from these structs to avoid state.

I am split tbh, I can understand your motivation but on the other side I think these could be handled in a registry to access all instances. 
We can explore this in an offline call later, I think this PR is already a great improvement, so not blocking this for this one.

controllers/operator/authentication/scramsha.go
```diff
@@ -8,101 +8,72 @@ import (
...
-func NewAutomationConfigScramSha256(ac *om.AutomationConfig) AutomationConfigScramSha {
-	return AutomationConfigScramSha{
-		automationConfig: ac,
-		mechanismName:    ScramSha256,
-	}
-}
+var (
+	MongoDBCRMechanism   = AutomationConfigScramSha{MechanismName: MongoDBCR}
+	ScramSha1Mechanism   = AutomationConfigScramSha{MechanismName: ScramSha1}
+	ScramSha256Mechanism = AutomationConfigScramSha{MechanismName: ScramSha256}
```
SimonBaeumer: Adding the interface as a type assertion?

```suggestion
	MongoDBCRMechanism Mechanism   = AutomationConfigScramSha{MechanismName: MongoDBCR}
	ScramSha1Mechanism Mechanism   = AutomationConfigScramSha{MechanismName: ScramSha1}
	ScramSha256Mechanism Mechanism = AutomationConfigScramSha{MechanismName: ScramSha256}
```

controllers/operator/authentication/x509.go
```diff
@@ -10,25 +10,17 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/stringutil"
 )
 
-const ExternalDB = "$external"
+var MongoDBX509Mechanism = ConnectionX509{}
```
SimonBaeumer:
```suggestion
var MongoDBX509Mechanism Mechanism = ConnectionX509{}
```

controllers/operator/authentication/authentication_mechanism.go
SimonBaeumer, line 15: +1 for introducing an interface, this gives a better overview of the different implementations and gives more guidance for other engineers to add new auth mechanisms. I like it.

controllers/operator/authentication/ldap.go
MaciejKaras: TBH those variables are only used in the tests, so I can easily use factory method to retrieve them in tests and just delete those variables

controllers/operator/authentication/authentication_mechanism.go
MaciejKaras: I agree, connection could be set as field in each Mechanism struct. Adding it does not add any state to the struct itself. We can discuss this in followup call

controllers/operator/authentication/ldap.go
MaciejKaras: Done

controllers/operator/authentication/ldap.go
MaciejKaras: Let's talk about that on Monday, I'm curious what exactly you have in mind.

controllers/operator/authentication/scramsha.go
MaciejKaras: Done

controllers/operator/authentication/authentication_mechanism.go
MaciejKaras: The interface was already there, but previously it was not so easy to understand how to implement another Mechanism.

controllers/operator/authentication/authentication.go
```diff
@@ -254,76 +223,30 @@ func Disable(conn om.Connection, opts Options, deleteUsers bool, log *zap.Sugare
...
-
-		// if no authentication has been configured, the default value for "AutoAuthMechanism" is "MONGODB-CR"
-		// even if authentication is disabled, so we need to ensure that auth has been enabled.
-		if ac.Auth.AutoAuthMechanism == string(MongoDBCR) && ac.Auth.IsEnabled() {
-			return MongoDBCR
-		}
-		return ScramSha256
-	}
-	// this should never be reached as validation of this string happens at the CR level
-	panic(fmt.Sprintf("unknown mechanism name %s", mongodbResourceMode))
```
MaciejKaras, line 281: existing panic

controllers/operator/authentication/authentication_mechanism.go
MaciejKaras: I didn't add it, it was like that before -> https://github.com/mongodb/mongodb-kubernetes/pull/51/files#r2069371749

Actually I don't mind panic here. If we ever enter this line of code it will be due to programming issue, not based on user input. We already have validation in CR that safeguards us from it happening. So if it ever happens, it's programmer fault and a big one and we should know about sooner than later. 

On another note if we return error there is nothing we can do with it other than fail the reconciliation, because we cannot guess what user had on their mind and configure it for them. Panicking will be similar experience for the user as failing reconciliation I think

controllers/operator/authentication/ldap.go
MaciejKaras: Please take a look at this commit -> https://github.com/mongodb/mongodb-kubernetes/pull/51/commits/8f5ff0a72428e5fca31b81be58b6e9a020020405. Let me know if that is direction you were thinking about.

controllers/operator/authentication/ldap.go
```diff
@@ -122,9 +115,8 @@ func (l *ldapAuthMechanism) IsAgentAuthenticationConfigured() bool {
 	return true
 }
 
-func (l *ldapAuthMechanism) IsDeploymentAuthenticationConfigured() bool {
-	ac := l.AutomationConfig
-	return stringutil.Contains(ac.Auth.DeploymentAuthMechanisms, string(LDAPPlain)) && ldapObjectsEqual(ac.Ldap, l.Options.Ldap)
+func (l *ldapAuthMechanism) IsDeploymentAuthenticationConfigured(ac *om.AutomationConfig, opts Options) bool {
+	return stringutil.Contains(ac.Auth.DeploymentAuthMechanisms, string(LDAPPlain)) && ldapObjectsEqual(ac.Ldap, opts.Ldap)
```
lucian-tosa, line 119: Just spotted this bug with @anandsyncs 
We need 2 separate methods:
* IsDeploymentAuthMechanismConfigured - keep as it is
* IsDeploymentAuthMechanismEnabled - remove the right side of the &&

This is because this method is also reused when we try to remove an auth mechanism. When we remove LDAP (and OIDC) there is also a separate configuration (unlike SCRAM and x509). So when we try to disable it, the right side of the end will false (since there is an LDAP config upstream, but none locally, since we remove it). This means that when we try to disable LDAP, it will not be disabled since this method will always return false.

controllers/operator/authentication/ldap_test.go
```diff
@@ -4,61 +4,64 @@ import (
...
 	assert.Equal(t, "BindMethod", ac.Ldap.BindMethod)
 
-	if err := l.DisableDeploymentAuthentication(); err != nil {
-		t.Fatal(err)
-	}
+	err = LDAPPlainMechanism.DisableDeploymentAuthentication(conn, zap.S())
+	require.NoError(t, err)
+
+	ac, err = conn.ReadAutomationConfig()
+	require.NoError(t, err)
```
lucian-tosa, line 41: this test misses that bug I mentioned. It is fine to test DisableDeploymentAuthentication directly, but this is normally called in `removeUnsupportedMechanisms`

controllers/operator/authentication/authentication_mechanism.go
SimonBaeumer: > On another note if we return error there is nothing we can do with it other than fail the reconciliation, because we cannot guess what user had on their mind and configure it for them. Panicking will be similar experience for the user as failing reconciliation I think

Ah, overlooked the panic before. 
Imho this is fine. The difference in this case is, that a user might not have access to the operator pod's log but can access the status of a resource. But likely the same issue applies with a failing reconcile if not written to the status.

Resolving this, the logic stays the same.

controllers/operator/authentication/ldap.go
SimonBaeumer: This looks good. Except the panic 🙈
Maybe we need a "good-practice" or coding style when to use panics. I've left a comment at that line.

controllers/operator/authentication/ldap_test.go
MaciejKaras: I agree, testing only mechanism methods not the full `Configure()` function makes us miss such errors. I will think how to add testing for Configure method as well

controllers/operator/authentication/ldap.go
MaciejKaras: Thanks for for finding this issue, amazing!

We should definitely fix it, but I would not like to do it in this PR, because it is existing issue and not related to the refactor. It would be great to report it as a Jira ticket and fix it in separate PR.

If you fine with that can you create a ticket for this issue @lucian-tosa @anandsyncs and link it to OIDC epic?

controllers/operator/authentication/ldap.go
MaciejKaras: Based on our offline discussion we decided to go with renaming panicking function to `convertToMechanismOrPanic`

controllers/operator/authentication/ldap.go
anandsyncs: Let's add it in another ticket keeping in mind that this needs to be fixed for OIDC to go through.

lucian-tosa: Opened ticket: https://jira.mongodb.org/browse/CLOUDP-317713

controllers/operator/authentication/ldap_test.go
lucian-tosa: This can be done in the LDAP ticket but we should handle thsi for OIDC

PR #54, opened 2025-04-28
CLOUDP-314917 CLOUDP-314918 CLOUDP-314919 - Add ClusterMongoDBRole CRD

lucian-tosa:
# Summary

Introduced new CRD for MongoDB roles. The new CRD is cluster-scoped and can be reused in multiple MongoDB deployments even across-namespaces.

The CRD design can be found in the [TD](https://docs.google.com/document/d/11j_wg0-s3u8oBI6Ca12AZiHLlUv_ViilYwMvl2LgoiQ/edit?tab=t.bhdbox3ppbto#bookmark=id.jbs82dmgpgr6)

## Proof of Work

There are unit and E2E tests verifying that the custom roles are:
* Roles are added in automation config
* Removing a reference to the role removes the role from automation config
* Updating the role triggers a reconciliation, and role is updated in AC
* Deleting the role is blocked by the finalizer
* Deleting the role after it is removed from resources is not blocked anymore

## Checklist

  * **q**: Clarifying question.


public/tools/multicluster/pkg/common/common.go
```diff
@@ -319,9 +319,6 @@ func EnsureMultiClusterResources(ctx context.Context, flags Flags, clientMap map
 	}
 
 	centralClusterClient := clientMap[flags.CentralCluster]
-	if err != nil {
```
Julien-Ben, line 322: Why is it removed ?

controllers/operator/common_controller_test.go
```diff
@@ -265,9 +265,124 @@ func TestReadSubjectNoCertificate(t *testing.T) {
...
+	assert.Nil(t, customRole.Privileges)
+	rs := mdbv1.NewDefaultReplicaSetBuilder().SetRoles([]mdbv1.MongoDBRole{customRole}).SetRoleRefs([]mdbv1.MongoDBRoleRef{roleRef}).Build()
+
+	kubeClient, omConnectionFactory := mock.NewDefaultFakeClient()
+	controller := NewReconcileCommonController(ctx, kubeClient)
+	mockOm, _ := prepareConnection(ctx, controller, omConnectionFactory.GetConnectionFunc, t)
+
+	result := controller.ensureRoles(ctx, rs.Spec.Security.Roles, rs.Spec.Security.RoleRefs, mockOm, kube.ObjectKeyFromApiObject(rs), &zap.SugaredLogger{})
+	assert.False(t, result.IsOK())
+	assert.Equal(t, status.PhaseFailed, result.Phase())
```
Julien-Ben, line 294: Nit: Are we expecting failure here because of validations error ?
If yes we may want to make it explicit, e.g testing the err message

docker/mongodb-kubernetes-tests/kubetester/mongodb_role.py
```diff
@@ -0,0 +1,60 @@
...
+        with_defaults = {
+            "plural": "clustermongodbroles",
+            "kind": "ClusterMongoDBRole",
+            "group": "mongodb.com",
+            "version": "v1",
+        }
+        with_defaults.update(kwargs)
+        super(ClusterMongoDBRole, self).__init__(*args, **with_defaults)
+
+    def assert_reaches_phase(self, phase: Phase, msg_regexp=None, timeout=None, ignore_errors=False):
```
Julien-Ben: nit: is it worth adding a trace here ? or logging the reconciliation time
As we do e.g here: https://github.com/mongodb/mongodb-kubernetes/blob/6a1393da1c85ce7f02cc31986bc547bfd7adf03f/docker/mongodb-kubernetes-tests/kubetester/mongodb.py#L144

docker/mongodb-kubernetes-tests/tests/authentication/mongodb_custom_roles.py
```diff
@@ -0,0 +1,250 @@
...
+    replica_set: MongoDB, sharded_cluster: MongoDB, mc_replica_set: MongoDBMulti, mongodb_role: ClusterMongoDBRole
+):
+    rs_version = replica_set.get_automation_config_tester().automation_config["version"]
+    sc_version = sharded_cluster.get_automation_config_tester().automation_config["version"]
+    mcrs_version = mc_replica_set.get_automation_config_tester().automation_config["version"]
+
+    mongodb_role["spec"]["roles"][0]["role"] = "readWrite"
+    mongodb_role.update()
+
+    wait_until(lambda: replica_set.get_automation_config_tester().reached_version(rs_version + 1))
```
Julien-Ben: nit: should we have explicit timeouts for wait_until calls in this test ?

Julien-Ben (approved):
Great PR 👏 Very clean code
I just added a few nit, but overall LGTM

I approve the changes, but let's wait for a consensus on the finalizers before merging.
But on the implementation side, it follows what's described in the TD.

public/tools/multicluster/pkg/common/common.go
lucian-tosa: This err is already checked a few lines above. Getting the central cluster client does not generate any error

controllers/operator/common_controller_test.go
lucian-tosa: Yes. That is a good point, which I have tried to implement. But the error message is in a private field in the workflow struct. I don't know if I should do a refactor of the workflow struct in this PR.
I might be missing something, but I haven't found any other unit tests that assert the error message.

docker/mongodb-kubernetes-tests/tests/authentication/mongodb_custom_roles.py
lucian-tosa: Good point, I thought there was a sensible default, but it looks like the default is no timeout.

RELEASE_NOTES.md
```diff
@@ -1,6 +1,21 @@
 [//]: # (Consider renaming or removing the header for next release, otherwise it appears as duplicate in the published release, e.g: https://github.com/mongodb/mongodb-enterprise-kubernetes/releases/tag/1.22.0 )
 <!-- Next Release -->
 
+# MCK 1.2.0 Release Notes
+
+## New Features
+
+* Added new **ClusterMongoDBRole** CRD to support reusable roles across multiple MongoDB clusters.
+  * This allows users to define roles once and reuse them in multiple **MongoDB** or **MongoDBMultiCluster** resources. The role can be referenced through the `.spec.security.roleRefs` field. Note that only one of `.spec.security.roles` and `.spec.security.roleRefs` can be used at a time.
+  * The new resource is watched by default by the operator. This means that the operator will require a new **ClusterRole** and **ClusterRoleBinding** to be created in the cluster. These resources are created by default with the helm chart or the kubectl mongodb plugin. To disable this behavior, set the `operator.enableClusterMongoDBRoles` value to `false` in the helm chart values.
```
lsierant: Should we add a note to not watch the CRD if enableClusterMongoDBRoles is set to false?

lsierant:
```suggestion
  * The new resource is watched by default by the operator. This means that the operator will require a new **ClusterRole** and **ClusterRoleBinding** to be created in the cluster. **ClusterRole** and **ClusterRoleBinding** resources are created by default with the helm chart or the kubectl mongodb plugin. To disable this behavior, set the `operator.enableClusterMongoDBRoles` value to `false` in the helm chart values.
```

lucian-tosa: Setting that to false will disable watching and not install RBAC. The customer is not required to do anything else

RELEASE_NOTES.md
```diff
@@ -1,6 +1,21 @@
 [//]: # (Consider renaming or removing the header for next release, otherwise it appears as duplicate in the published release, e.g: https://github.com/mongodb/mongodb-enterprise-kubernetes/releases/tag/1.22.0 )
 <!-- Next Release -->
 
+# MCK 1.2.0 Release Notes
+
+## New Features
+
+* Added new **ClusterMongoDBRole** CRD to support reusable roles across multiple MongoDB clusters.
+  * This allows users to define roles once and reuse them in multiple **MongoDB** or **MongoDBMultiCluster** resources. The role can be referenced through the `.spec.security.roleRefs` field. Note that only one of `.spec.security.roles` and `.spec.security.roleRefs` can be used at a time.
```
lsierant, line 9:
```suggestion
  * This allows users to define roles once and reuse them in multiple **MongoDB** or **MongoDBMultiCluster** resources. The role can be referenced through the `.spec.security.roleRefs` field. Note that only one of `.spec.security.roles` and `.spec.security.roleRefs` can be used at a time. 
  * **ClusterMongoDBRole** resources are treated by the operator as a custom role templates that are only used when referenced by the database resources.
```

config/rbac/operator-roles.yaml
```diff
@@ -1,5 +1,18 @@
 ---
 # Source: mongodb-kubernetes/templates/operator-roles.yaml
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: mongodb-kubernetes-operator-mongodb-cluster-mongodb-role
```
lsierant, line 6: is it deliberate to not have the operator name in the ClusterRole? 
We will have helm install errors if someone installs more than one operator in the cluster.

lucian-tosa: The name of the clusterrole contains the operator name and namespace
```
  name: {{ .Values.operator.name }}-{{ include "mongodb-kubernetes-operator.namespace" . }}-cluster-mongodb-role
```

controllers/operator/mongodbuser_controller.go
```diff
@@ -174,12 +174,12 @@ func (r *MongoDBUserReconciler) Reconcile(ctx context.Context, request reconcile
...
 
-			if controllerutil.ContainsFinalizer(user, util.Finalizer) {
-				controllerutil.RemoveFinalizer(user, util.Finalizer)
+			if controllerutil.ContainsFinalizer(user, util.UserFinalizer) {
+				controllerutil.RemoveFinalizer(user, util.UserFinalizer)
 				if err := r.client.Update(ctx, user); err != nil {
 					return r.updateStatus(ctx, user, workflow.Failed(xerrors.Errorf("Failed to update the user with the removed finalizer: %w", err)), log)
 				}
-				return r.updateStatus(ctx, user, workflow.Pending("Finalizer will be removed. MongoDB resource not found"), log)
+				return r.updateStatus(ctx, user, workflow.Pending("UserFinalizer will be removed. MongoDB resource not found"), log)
```
lsierant: Should the message also be changed? The finalizer name wasn't changed.

helm_chart/values.yaml
```diff
@@ -66,6 +66,9 @@ operator:
   # kubectl mongodb plugin is used to configure multi-cluster resources
   createResourcesServiceAccountsAndRoles: true
 
+  # Set to false to not create the RBAC for enabling access to the ClusterMongoDBRole resources for the operator
```
nammn: let's play it safer and clearly note that this will create a create the ClusterRole and ClusterRoleBinding

main.go
```diff
@@ -123,6 +131,8 @@ func main() {
 	// Namespace where the operator is installed
 	currentNamespace := env.ReadOrPanic(util.CurrentNamespace)
 
+	watchClusterMongoDBRoles := slices.Contains(crds, clusterMongoDBRoleCRDPlural)
```
nammn: q: wdyt about `has/containsClusterMongoDBRoleCRD` this better fits a bool

docker/mongodb-kubernetes-tests/tests/authentication/mongodb_custom_roles.py
```diff
@@ -0,0 +1,260 @@
...
+@fixture(scope="module")
+def project_name_prefix(namespace: str) -> str:
+    return random_k8s_name(f"{namespace}-project-")
+
+
+@fixture(scope="module")
+def first_project(namespace: str, project_name_prefix: str) -> str:
+    cm = read_configmap(namespace=namespace, name="my-project")
+    project_name = f"{project_name_prefix}-first"
+    return create_or_update_configmap(
```
lsierant, line 25: I did a similar thing but wrapped in .configure function in this (not yet merged PR): https://github.com/mongodb/mongodb-kubernetes/pull/173/files#diff-9bd4ee69baff1e418bdf57281d036ca298a0b4aa5628cfc457b41e6b0fe565e7R111

I've left a comment there to change it after this PR is merged, so nothing to be done here for you.

RELEASE_NOTES.md
```diff
@@ -1,6 +1,21 @@
...
+  * The new resource is watched by default by the operator. This means that the operator will require a new **ClusterRole** and **ClusterRoleBinding** to be created in the cluster. **ClusterRole** and **ClusterRoleBinding** resources are created by default with the helm chart or the kubectl mongodb plugin. To disable this behavior, set the `operator.enableClusterMongoDBRoles` value to `false` in the helm chart values.
```
nammn: blocking: for conformity we should also shortly describe how to disable it via mongodb plugin or rephrase the following sentence.

docker/mongodb-kubernetes-tests/tests/authentication/mongodb_custom_roles.py
lucian-tosa: That would be a helpful addition, thanks

config/rbac/operator-roles.yaml
lsierant: yes, my bad, I thought it's a template!

RELEASE_NOTES.md
```diff
@@ -1,6 +1,21 @@
...
+  * The new **ClusterMongoDBRole** resource is designed to be read-only, meaning it can be used by MongoDB deployments managed by different operators.
+  * The **ClusterMongoDBRole** resource can be deleted at any time, but the operator will not delete any roles that were created using this resource. To properly remove access, you must remove the reference to the **ClusterMongoDBRole** in the **MongoDB** or **MongoDBMultiCluster** resources.
```
nammn:
```suggestion
  * The **ClusterMongoDBRole** resource can be deleted at any time, but the operator will not delete any roles that were created using this resource. To properly remove access, you must **manually** remove the reference to the **ClusterMongoDBRole** in the **MongoDB** or **MongoDBMultiCluster** resources.
```

helm_chart/templates/operator.yaml
```diff
@@ -53,6 +53,9 @@ spec:
             {{- if .Values.multiCluster.clusters }}
             - -watch-resource=mongodbmulticluster
             {{- end }}
+            {{- if .Values.operator.enableClusterMongoDBRoles }}
```
lsierant, line 56: blocking: I think we should have more granular control here. What does enabling cluster roles mean? Currently it means watching it by the operator *and* installing the custom role in one switch. But it will fail if someone install another instance of the operator (with the same/default name) in another namespace and also wants to use cluster roles. The cluster role will be already installed. So the only thing we need is to watch the resource, but not install. 

I'd suggest separating this setting to two: first to enable watching, second to install cluster role. Please see .Values.operator.telemetry.installClusterRole and consider following the same naming convention

controllers/operator/common_controller.go
```diff
@@ -127,6 +154,45 @@ func ensureRoles(roles []mdbv1.MongoDbRole, conn om.Connection, log *zap.Sugared
...
+		var role mdbv1.MongoDBRole
+		switch ref.Kind {
+
+		case util.ClusterMongoDBRoleKind:
+			customRole := &rolev1.ClusterMongoDBRole{}
+
+			err := r.client.Get(ctx, types.NamespacedName{Name: ref.Name}, customRole)
+			if err != nil {
+				if apiErrors.IsNotFound(err) {
+					return nil, xerrors.Errorf("ClusterMongoDBRole '%s' not found. If the resource was deleted, the role is still present in MongoDB. The correctly remove a role from MongoDB, please remove the reference from spec.security.roleRefs.", ref.Name)
```
nammn:
```suggestion
					return nil, xerrors.Errorf("ClusterMongoDBRole '%s' not found. If the resource was deleted, the role is still present in MongoDB. To correctly remove a role from MongoDB, please remove the reference from spec.security.roleRefs.", ref.Name)
```

docker/mongodb-kubernetes-tests/kubeobject/customobject.py
```diff
@@ -145,7 +145,7 @@ def _reload_if_needed(self):
             self.reload()
 
     @classmethod
-    def from_yaml(cls, yaml_file, name=None, namespace=None):
+    def from_yaml(cls, yaml_file, name=None, namespace=None, cluster_scoped=False):
```
nammn, line 148: can we update the method doc quickly for this?

helm_chart/templates/operator.yaml
lucian-tosa: It will not fail with more than one operator because the name of the clusterrole contains the namespace. Multiple installations will result in multiple clusterroles, but I think that makes the UX cleaner.

Do you think we shouldn't have multiple clusterroles? Doing it like this makes each operator self-contained and not depending on RBAC from other helm installations. If they want to watch the resource but not install RBAC, users can set the global `createOperatorServiceAccount`

docker/mongodb-kubernetes-tests/tests/authentication/mongodb_custom_roles.py
```diff
@@ -0,0 +1,260 @@
...
+    replica_set: MongoDB, sharded_cluster: MongoDB, mc_replica_set: MongoDBMulti, mongodb_role: ClusterMongoDBRole
+):
+    rs_version = replica_set.get_automation_config_tester().automation_config["version"]
+    sc_version = sharded_cluster.get_automation_config_tester().automation_config["version"]
+    mcrs_version = mc_replica_set.get_automation_config_tester().automation_config["version"]
+
+    mongodb_role["spec"]["roles"][0]["role"] = "readWrite"
+    mongodb_role.update()
+
+    wait_until(lambda: replica_set.get_automation_config_tester().reached_version(rs_version + 1), timeout=120)
```
nammn, line 179: can this be for any random reason be flaky? Example the operator updates the AC before we call this function and thus `rs_version` is already + 2 at this point of time

docker/mongodb-kubernetes-tests/tests/conftest.py
```diff
@@ -726,6 +726,37 @@ def multi_cluster_operator_manual_remediation(
     )
 
 
+@fixture(scope="module")
+def multi_cluster_operator_no_clustermongodbroles(
```
nammn: nit: should it be cluster_mongodb_roles?

nammn (changes requested):
Awesome changes! But some clarifications, suggestions required

helm_chart/templates/operator.yaml
lsierant: You've got me thinking... You're right - namespace in the role name solves this. And the UX is super clean with only one switch. Resolving for now!

controllers/operator/common_controller.go
```diff
@@ -97,7 +98,31 @@ func NewReconcileCommonController(ctx context.Context, client client.Client) *Re
...
+	roleRefs := db.GetSecurity().RoleRefs
+
+	if len(localRoles) > 0 && len(roleRefs) > 0 {
+		return workflow.Failed(xerrors.Errorf("At most one one of roles or roleRefs can be non-empty."))
+	}
+
+	var roles []mdbv1.MongoDBRole
+	if len(roleRefs) > 0 {
+		if !watchClusterMongoDBRoles {
+			return workflow.Failed(xerrors.Errorf("RoleRefs are not supported when ClusterMongoDBRoles are disabled. Please enable ClusterMongoDBRoles in the operator configuration."))
```
lsierant: **blocking:** We should tell in the message exactly what parameter to set or point to the documentation. In this case it will be helm's values.yaml parameter to enable cluster roles. 

But for someone who installs the operator manually using yaml and kubectl it will be confusing. As what is actually required here to pass this check is to add watch-resource parameter. 

What's worse - even without watching the resource, references might work because the only thing the operator needs is to have the clusterrole bound to the operator's service account.  

Consider rephrasing it this or even remove this if entirely and try to actually get the cluster role. Perhaps in the absence of necessary permissions we can recognize that and only then fail with some informative error?

go.mod
```diff
@@ -13,6 +13,8 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/hashicorp/vault/api v1.16.0
 	github.com/imdario/mergo v0.3.15
+	github.com/onsi/ginkgo/v2 v2.17.1
```
lsierant: why this is here?

main.go
lucian-tosa: Hmm, not sure about "CRD". What about `enableClusterMongoDBRoles`? That would also align with the helm value

RELEASE_NOTES.md
lucian-tosa: good point, haven't made a flag for the plugin yet. Doing that now

lsierant (changes requested):
Overall the PR is awesome! Well done! 👏 
I've left few minor and one blocking comment regarding checking in the operator if the cluster roles are enabled.

docker/mongodb-kubernetes-tests/tests/authentication/mongodb_custom_roles.py
lucian-tosa: I somewhat agree, but at the same time, the AC versions are increased only when it is changed, right?
If the rs_version would reach +2, then I think it should fail, because we should update the roles in one go.

I could also change the assertion to check that it reaches any version larger than `rs_version`, wdyt?

main.go
nammn: yes!

docker/mongodb-kubernetes-tests/kubeobject/customobject.py
lucian-tosa: done

docker/mongodb-kubernetes-tests/tests/conftest.py
lucian-tosa: I skimped on underscores, fixed it

go.mod
lucian-tosa: same reason as in #195 , fixed it now

main.go
lucian-tosa: done

controllers/operator/mongodbuser_controller.go
lucian-tosa: done

docker/mongodb-kubernetes-tests/tests/authentication/mongodb_custom_roles.py
nammn: I will leave that judgement to you as your point make sense. It shouldn't bump more than once but at the same time the operator might see something unrelated and bump the ac first because something outside of this code path has caused a change. 

I am fine with leaving this or your suggested change

RELEASE_NOTES.md
lucian-tosa: done

docker/mongodb-kubernetes-tests/tests/authentication/mongodb_custom_roles.py
lucian-tosa: I would leave it as is for now then. It really shouldn't bump the version more than once. If it will do it at some point, we can revisit this. Resolving for now

helm_chart/values.yaml
lucian-tosa: done

controllers/operator/common_controller.go
lucian-tosa: > What's worse - even without watching the resource, references might work because the only thing the operator needs is to have the clusterrole bound to the operator's service account.

Technically it could work, but we need that flag to have the operator watch the role resource for changes. Otherwise we don't set up the resource watcher for it. That is why I created this if statement to prevent this possibility.

> Consider rephrasing it this or even remove this if entirely and try to actually get the cluster role

I don't think the operator has permissions to see RBAC resources, right?

> Perhaps in the absence of necessary permissions we can recognize that and only then fail with some informative error?

This is another reason for which I had to create the if statement. Client-go has some weird behavior where it will retry continuously even if it doesn't have permissions. So a client.get will not return a 403, it will simply hang. I think it can be done using some custom error handlers, but that's not worth it.

Would it be best if I write something like this?

RoleRefs are not supported when ClusterMongoDBRoles are disabled. Please enable ClusterMongoDBRoles in the operator configuration. This can be done by setting the `operator.enableClusterMongoDBRoles` to true in the helm values file, which will automatically installed the necessary RBAC. Alternatively, it can be enabled by adding `-watch-resource=clustermongodbroles` flag to the operator deployment, and manually creating the necessary RBAC.

RELEASE_NOTES.md
```diff
@@ -7,7 +7,7 @@
...
+  * The new resource is watched by default by the operator. This means that the operator will require a new **ClusterRole** and **ClusterRoleBinding** to be created in the cluster. **ClusterRole** and **ClusterRoleBinding** resources are created by default with the helm chart or the kubectl mongodb plugin. To disable this behavior, set the `operator.enableClusterMongoDBRoles` value to `false` in the helm chart values.
   * The new **ClusterMongoDBRole** resource is designed to be read-only, meaning it can be used by MongoDB deployments managed by different operators.
   * The **ClusterMongoDBRole** resource can be deleted at any time, but the operator will not delete any roles that were created using this resource. To properly remove access, you must remove the reference to the **ClusterMongoDBRole** in the **MongoDB** or **MongoDBMultiCluster** resources.
   * The reference for this resource can be found here: **TODO** (link to documentation)
```
vinilage: Just a minor suggestion:
The reference **documentation** for this resource can be found here: TODO (link to documentation)

api/v1/mdb/mongodb_roles_validation.go
```diff
@@ -270,7 +270,7 @@ func isValidCIDR(cidr string) bool {
 	return err == nil
 }
 
-func roleIsCorrectlyConfigured(role MongoDbRole, mdbVersion string) v1.ValidationResult {
+func RoleIsCorrectlyConfigured(role MongoDBRole, mdbVersion string) v1.ValidationResult {
```
nammn, line 273: q: i didn't check out the PR, but did we expose this function? If yes please resolve this thread

lucian-tosa: Yes, we did because we need to run that validation on the role resource at binding time

controllers/operator/common_controller.go
lsierant: I find the message much, much better and we can have it like this! 👍

lsierant (approved):
LGTM! 
Please only update the error message.

PR #60, opened 2025-04-29
CLOUDP-314903 [OIDC] CRD Config Propagation to Automation Config

MaciejKaras:
# Summary

### Core Functionality Enhancements:
* Added a new authentication mechanism, `MongoDB-OIDC`, to the list of supported mechanisms in the `authentication_mechanism.go` file. 
* Introduced the `OIDCProviderConfigs` field in the `AutomationConfig` struct and implemented logic to merge and apply OIDC configurations into the deployment in the `automation_config.go` file.
* Removed default value for `groupClaim` because the value `groups` can result in hard to debug misconfiguration.

### API and Configuration Updates:
* Added the `IsOIDCEnabled()` method in the `Security` struct and `AuthResource` interface to check if OIDC is enabled. 
* Updated the `Options` struct in the `authentication.go` file to include `OIDCProviderConfigs`.

### Test Coverage:
* Added comprehensive test cases for OIDC provider configurations in `automation_config_test.go`, including scenarios for merging, clearing, and modifying configurations.
* Updated the `TestAutomationConfigEquality` test to include OIDC provider configurations. 

### JSON Configuration Example:
* Updated the `automation_config.json` test data file to include sample OIDC provider configurations for testing purposes.

## Checklist

  * **q**: Clarifying question.


controllers/operator/authentication/oidc.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func sortOIDCPProviderConfigs(configs []oidc.ProviderConfig) []oidc.ProviderConfig {
+	configsSeq := slices.Values(configs)
+	return slices.SortedFunc(configsSeq, func(l, r oidc.ProviderConfig) int {
+		return strings.Compare(l.AuthNamePrefix, r.AuthNamePrefix)
+	})
+}
+
+func oidcProviderConfigEqual(l oidc.ProviderConfig, r oidc.ProviderConfig) bool {
```
anandsyncs: nit:

```suggestion
func oidcProviderConfigEqual(l, r oidc.ProviderConfig) bool {
	return l.AuthNamePrefix == r.AuthNamePrefix &&
		l.Audience == r.Audience &&
		l.IssuerUri == r.IssuerUri &&
		slices.Equal(l.RequestedScopes, r.RequestedScopes) &&
		l.UserClaim == r.UserClaim &&
		l.GroupsClaim == r.GroupsClaim &&
		l.SupportsHumanFlows == r.SupportsHumanFlows &&
		l.UseAuthorizationClaim == r.UseAuthorizationClaim
}
```

helm_chart/crds/mongodb.com_opsmanagers.yaml
```diff
@@ -843,8 +843,84 @@ spec:
...
+                                  description: |-
+                                    Unique label that identifies this configuration. This label is visible to your Ops Manager users and is used when
+                                    creating users and roles for authorization. It is case-sensitive and can only contain the following characters:
+                                     - alphanumeric characters (combination of a to z and 0 to 9)
+                                     - hyphens (-)
+                                     - underscores (_)
+                                  pattern: ^[a-zA-Z0-9-_]+$
+                                  type: string
+                                groupsClaim:
+                                  default: groups
```
anandsyncs: Why do we give a default value here when we are throwing error if the `groupsClaim` is absent?

api/v1/mdb/mongodb_types.go
```diff
@@ -1089,7 +1090,6 @@ type OIDCProviderConfig struct {
 	// The identifier of the claim that includes the principal's IdP user group membership information.
 	// Accept the default value unless your IdP uses a different claim, or you need a custom claim.
 	// Required when selected GroupMembership as the authorization type, ignored otherwise
-	// +kubebuilder:default=groups
```
mircea-cosbuc, line 1100: Just to clarify, why was the default value removed?

controllers/om/automation_config.go
```diff
@@ -58,9 +60,67 @@ func applyInto(a AutomationConfig, into *Deployment) error {
...
+		if configs, ok := a.Deployment["oidcProviderConfigs"]; ok {
+			configsSlice := cast.ToSlice(configs)
+			for _, config := range configsSlice {
+				deploymentConfigs = append(deploymentConfigs, config.(map[string]any))
+			}
+		}
+
+		result := make([]map[string]any, 0)
+		for _, config := range a.OIDCProviderConfigs {
+			deploymentConfig := findOrCreateEmptyDeploymentConfig(deploymentConfigs, config.AuthNamePrefix)
```
mircea-cosbuc: Does this imply that if you configure the same OIDC provider multiple times, it gets overridden with the latest configuration? Why not just enforce uniqueness?

controllers/om/automation_config_test.go
```diff
@@ -627,6 +628,54 @@ func TestAutomationConfigEquality(t *testing.T) {
 	}
 	ldapConfig2 := ldapConfig
 
+	oidcProviderConfigs := []oidc.ProviderConfig{
+		{
+			AuthNamePrefix:        "provider-1",
```
mircea-cosbuc, line 634: Since the implementation implies that we support repeating the `AuthNamePrefix`, we should have a test displaying the expected behavior.

controllers/om/automation_config.go
mircea-cosbuc: I see that the validation webhook rejects resources which repeat the `AuthNamePrefix`.

mircea-cosbuc (approved):
LGTM, I think my comments just require clarification rather than changes.

api/v1/mdb/mongodb_types.go
anandsyncs: 1. To make sure that a value is set here explicitly, because this was not the default in any provider that I experimented with.
2. Setting a default here populated the value in places where it wasn't necessary, like in UserID authentication.

controllers/om/automation_config_test.go
MaciejKaras: What do you mean by repeating `AuthNamePrefix`? You mean overriding it with new values?

We don't allow multiple OIDC provider configs with the same configuration name aka `AuthNamePrefix`. We have validation rule to enforce uniqueness

anandsyncs: `AuthNamePrefix` needs to be unique amongst `oidcProviderConfigs`, we don't support repeating this.
We validated this behaviour here: https://github.com/mongodb/mongodb-kubernetes/blob/f0f8eaa987740fc8e99068cccb43ebdf97a3cf60/api/v1/mdb/mongodb_validation_test.go#L251

api/v1/mdb/mongodb_types.go
mircea-cosbuc: Thanks!

controllers/om/automation_config.go
mircea-cosbuc: Why do we use `findOrCreate` since it's invalid to have the same `AuthNamePrefix` ?

controllers/om/automation_config_test.go
mircea-cosbuc: I followed-up in the previous comment, but the code is implying that we're "finding" an already configured provider or creating it if doesn't exist already. If it's unique it should never exist already.

controllers/om/automation_config.go
anandsyncs: This is for update. If a config with the given name exists then we need to update it with latest changes.

anandsyncs: Although we don't need to do this, we can completely overwrite the existing oidcProviderConfig in the automation config with the new one's. We don't need to consider what already exists.
@MaciejKaras please correct me if I am wrong?

MaciejKaras: We should not replace existing oidcProviderConfig in AC. There are some other fields and readOnly values that Ops Manager populates. The current behaviour is that we just update the fields that we care about in Operator and I think it is safer approach.

controllers/om/automation_config_test.go
lsierant: Where is the validation (not the assertion) performed? I don't see any instance of "OIDC provider config name provider is not unique" string in the codebase.

controllers/om/automation_config.go
```diff
@@ -58,9 +60,67 @@ func applyInto(a AutomationConfig, into *Deployment) error {
...
+
+			deploymentConfig["authNamePrefix"] = config.AuthNamePrefix
+			deploymentConfig["audience"] = config.Audience
+			deploymentConfig["issuerUri"] = config.IssuerUri
+			deploymentConfig["userClaim"] = config.UserClaim
+			deploymentConfig["supportsHumanFlows"] = config.SupportsHumanFlows
+			deploymentConfig["useAuthorizationClaim"] = config.UseAuthorizationClaim
+
+			if config.ClientId == util.MergoDelete {
```
lsierant: how does it work? Where this will be set in real-life scenario (not in unit tests where we set this manually)?

controllers/om/automation_config.go
```diff
@@ -58,9 +60,67 @@ func applyInto(a AutomationConfig, into *Deployment) error {
 		}
 		(*into)["ldap"] = mergedLdap
 	}
+
+	if len(a.OIDCProviderConfigs) > 0 {
+		deploymentConfigs := make([]map[string]any, 0)
```
lsierant: nit: can we extract body of this if to a function?

controllers/om/automation_config.go
lsierant: I also think we should not overwrite what's set by the OM. We always try to do the merging and overwrite only the parts which the operator is owning. In this case we should (and I think we're doing that treating authNamePrefix as a key overwrite only the configs that are defined with particular authNamePrefix (which acts as a key IIUC).

controllers/om/automation_config.go
lsierant: blocking: I'm not sure using MergoDelete is the way to go here. If we expect something to be possible to reset to an empty value (so a default value at the same time), then this field should not have omitempty tag in the API. Consider changing all optional/resettable values to *string in `OIDCProviderConfig` and remove omitempty tags for them. 

I know we're using this MergoDelete hack in our codebase but it's a code smell for me. We shouldn't have any specific logic like that in our core business logic that results from using a particular merging library. If something is resettable, then if it's null or empty then we should treat it like that. 

I'm putting this a blocking comment for now.

docker/mongodb-kubernetes-tests/tests/webhooks/fixtures/invalid_oidc_mongodb_community.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: mongodb.com/v1
+kind: MongoDB
```
lsierant: This is not MongoDBCommunity

docker/mongodb-kubernetes-tests/tests/webhooks/e2e_mongodb_validation_webhook.py
```diff
@@ -79,6 +79,71 @@ def test_replicaset_members_is_specified(self):
             exception_reason="'spec.members' must be specified if type of MongoDB is ReplicaSet",
         )
 
+    def test_oidc_auth_with_mongodb_community(self):
```
lsierant: we should not put "community" into MongoDB validation test. Also the fixture is not using kind: MongoDBCommunity.

controllers/operator/authentication/oidc.go
```diff
@@ -0,0 +1,148 @@
...
+func (o *oidcAuthMechanism) DisableAgentAuthentication(_ om.Connection, _ *zap.SugaredLogger) error {
+	return xerrors.Errorf("OIDC agent authentication is not supported")
+}
+
+func (o *oidcAuthMechanism) EnableDeploymentAuthentication(conn om.Connection, opts Options, log *zap.SugaredLogger) error {
+	return conn.ReadUpdateAutomationConfig(func(ac *om.AutomationConfig) error {
+		if !stringutil.Contains(ac.Auth.DeploymentAuthMechanisms, string(MongoDBOIDC)) {
+			ac.Auth.DeploymentAuthMechanisms = append(ac.Auth.DeploymentAuthMechanisms, string(MongoDBOIDC))
+		}
+		ac.OIDCProviderConfigs = opts.OIDCProviderConfigs
```
lsierant, line 36: shouldn't we perform a proper merge here? It looks like we're just overwriting it. Or is it because we're enabling it, so there shouldn't be anything there in the first place?

controllers/operator/authentication/oidc.go
```diff
@@ -0,0 +1,148 @@
...
+func MapOIDCProviderConfigs(oidcProviderConfigs []mdbv1.OIDCProviderConfig) []oidc.ProviderConfig {
+	if len(oidcProviderConfigs) == 0 {
+		return nil
+	}
+
+	result := make([]oidc.ProviderConfig, len(oidcProviderConfigs))
+	for i, providerConfig := range oidcProviderConfigs {
+		clientId := providerConfig.ClientId
+		if clientId == "" {
+			clientId = util.MergoDelete
```
lsierant: also here: let's make the clientId and groupsClaim *string without omitempty and not use MergoDelete hack

controllers/operator/authentication/oidc.go
```diff
@@ -0,0 +1,148 @@
...
+
+func mapToSupportHumanFlows(authMethod mdbv1.OIDCAuthorizationMethod) bool {
+	switch authMethod {
+	case mdbv1.OIDCAuthorizationMethodWorkforceIdentityFederation:
+		return true
+	case mdbv1.OIDCAuthorizationMethodWorkloadIdentityFederation:
+		return false
+	}
+
+	panic(fmt.Sprintf("unsupported OIDC authorization method: %s", authMethod))
```
lsierant, line 123: this comes from the API, do we have validation (not webhook only) in place that will guarantee we won't have it here specified by the user accidentally? We don't want to crash the operator in case the user submit invalid stuff into the CR

controllers/operator/authentication/oidc.go
```diff
@@ -0,0 +1,148 @@
...
+
+func mapToUseAuthorizationClaim(authType mdbv1.OIDCAuthorizationType) bool {
+	switch authType {
+	case mdbv1.OIDCAuthorizationTypeGroupMembership:
+		return true
+	case mdbv1.OIDCAuthorizationTypeUserID:
+		return false
+	}
+
+	panic(fmt.Sprintf("unsupported OIDC authorization type: %s", authType))
```
lsierant, line 134: here the same thing as above

controllers/om/automation_config.go
MaciejKaras: Yes, exactly. We only overwrite configs when we find existing one under `authNamePrefix` which has to be unique

controllers/operator/authentication/oidc_test.go
```diff
@@ -0,0 +1,93 @@
...
+			UseAuthorizationClaim: true,
+		},
+	}
+
+	opts := Options{
+		Mechanisms:          []string{string(MongoDBOIDC)},
+		OIDCProviderConfigs: providerConfigs,
+	}
+
+	configured := MongoDBOIDCMechanism.IsDeploymentAuthenticationConfigured(ac, opts)
```
lsierant: this `MongoDBOIDCMechanism.IsDeploymentAuthenticationConfigured` is very suprising to read in go code. Looks like a class, type or weirdly named package, but it's a local global var. Perhaps it's a Java background, but it looks like a static method at first glance! 
Please either make it lowercase, rename or remove it completely as it's not really used outside the tests, nor its mutated to justify it being a global var. It's more confusing than necessary.

controllers/om/automation_config_test.go
MaciejKaras: Probably you couldn't find it because it uses placeholder in the string -> https://github.com/mongodb/mongodb-kubernetes/blob/f0f8eaa987740fc8e99068cccb43ebdf97a3cf60/api/v1/mdb/mongodb_validation.go#L203-L216

controllers/operator/authentication/oidc_test.go
lsierant: if necessary I'd suggest converting it to a function defaultOIDCMechanism() or sth like that to avoid the confusion

controllers/operator/authentication/oidc_test.go
```diff
@@ -0,0 +1,93 @@
...
+func TestOIDC_EnableAgentAuthentication(t *testing.T) {
+	conn := om.NewMockedOmConnection(om.NewDeployment())
+	opts := Options{
+		Mechanisms: []string{string(MongoDBOIDC)},
+	}
+
+	ac, err := conn.ReadAutomationConfig()
+	require.NoError(t, err)
+
+	configured := MongoDBOIDCMechanism.IsAgentAuthenticationConfigured(ac, opts)
```
lsierant: it's not obvious why the following assertion should return error. From the outside how it's configured it looks completely fine, so why error? The easiest would be to add comment or change the function name to indicate why it should fail.

lsierant (changes requested):
I've left few comments: most are minor, but I've put a blocking one regarding MergoDelete. 
I also miss a bit a proper e2e test verifying the automation config changes in a real world scenario.

controllers/operator/authentication/oidc_test.go
MaciejKaras: For other auth mechanism we set the variable in the test context only -> https://github.com/mongodb/mongodb-kubernetes/blob/8f5ff0a72428e5fca31b81be58b6e9a020020405/controllers/operator/authentication/ldap_test.go#L13-L14 This is definitely better than exposing it from the package.

As for the syntax, I don't view it like a bad thing (but I also have Java background!). I saw it couple of times in Go std lib i.e. `io.EOF.Error()`. We can change it to function, but easiest is to just use lowercase - it is part of the tests only.

controllers/operator/authentication/oidc.go
MaciejKaras: We have this in the webhook validation only. Is there an issue with that approach? Can webhook validations be ignored somehow?

MaciejKaras: This is only what we have:
https://github.com/mongodb/mongodb-kubernetes/blob/26e24a0cb1169cc22483469b591be45c58a0036c/api/v1/mdb/mongodb_types.go#L1114-L1118

controllers/operator/authentication/oidc_test.go
MaciejKaras: The following assertion does not check for error, only if the method returned false. Can you check if it is what you have meant?

docker/mongodb-kubernetes-tests/tests/webhooks/e2e_mongodb_validation_webhook.py
MaciejKaras: I think the tests only asserts if we block enabling OIDC for non-enterprise MongoDB image. Is that correct Anand? Maybe we need to change it from community to non-enterprise for better readability?

controllers/operator/authentication/oidc.go
MaciejKaras: I agree, from first look it seems that we are overriding all configs. But the replacing the values in `ac AutomationConfig` struct will not directly mean the update will happen in the `Deployment` map. `ac AutomationConfig` is only operator representation, mapping of actual deployment config in Ops Manager. The "magic" happens in `conn.ReadUpdateAutomationConfig()` which is calling `conn.UpdateAutomationConfig()` and then `ac.Apply()`-> https://github.com/mongodb/mongodb-kubernetes/blob/0ce0874677c3aaf51078a5a046cf2aaf1b01bc6c/controllers/om/automation_config.go#L64-L109

I really hate auth mechanisms and `conn.ReadUpdateAutomationConfig()` method. I tried to cleanup this a little bit, but it is still onion mess.

controllers/om/automation_config.go
MaciejKaras: >how does it work? Where this will be set in real-life scenario (not in unit tests where we set this manually)?

We set it [here](https://github.com/mongodb/mongodb-kubernetes/blob/05903d599aae744801825d266845d08eec13b7c8/controllers/operator/authentication/oidc.go#L104), but I think you found it already. 

I agree, MergoDelete is a code smell, but I didn't want to add additional logic then that we use in other places. Here we are actually merging without using Mergo (which I like more) so we don't have to depend on the MergoDelete at all.

controllers/operator/authentication/oidc_test.go
MaciejKaras: @lsierant Please check changes in https://github.com/mongodb/mongodb-kubernetes/pull/60/commits/13e76a37e8a5fa8137fc6fb1becd7cf20bcddc86

controllers/om/automation_config.go
MaciejKaras: addressed in https://github.com/mongodb/mongodb-kubernetes/pull/60/commits/646bec6295bec975f0ee46fa8cded62230e04b42. Additionally removed `findOrCreateEmptyDeploymentConfig` function and used map for checking if config under `authNamePrefix` exists.

controllers/om/automation_config.go
MaciejKaras: addressed in https://github.com/mongodb/mongodb-kubernetes/pull/60/commits/646bec6295bec975f0ee46fa8cded62230e04b42

controllers/operator/authentication/oidc.go
MaciejKaras: addressed in https://github.com/mongodb/mongodb-kubernetes/pull/60/commits/646bec6295bec975f0ee46fa8cded62230e04b42

controllers/om/automation_config.go
MaciejKaras: Removed `findOrCreateEmptyDeploymentConfig` function and used map for checking if config under `authNamePrefix` exists in https://github.com/mongodb/mongodb-kubernetes/pull/60/commits/646bec6295bec975f0ee46fa8cded62230e04b42

mircea-cosbuc: Got it, so this is handling the case where OM replies with additional fields in the OIDC config.

controllers/om/automation_config.go
lsierant: this is much better! Thanks!

docker/mongodb-kubernetes-tests/tests/webhooks/e2e_mongodb_validation_webhook.py
lsierant: ah, yes - please use 'non-enterprise' to avoid the confusion 🙏

controllers/om/automation_config_test.go
MaciejKaras: Is it ok to resolve the comment? @mircea-cosbuc @lsierant

controllers/operator/authentication/oidc.go
lsierant: Yes, webhooks are sometimes (deliberately) disabled. That's why we always reuse the validation logic in both webhooks and the first thing the reconciler does. We don't execute those kubebuidler validations in reconciles though (I'd wish we did that), so for the cases there is no webhook we don't want to crash the operator due to a typo in the CR. That should never be the case that the user submits invalid yaml and it's crashing the operator.

controllers/operator/authentication/oidc_test.go
lsierant: much better!

controllers/operator/authentication/oidc.go
MaciejKaras: Based on our discussion today about kubebuilder annotations and that they are also run in apiserver and cannot be disabled, are you fine with resolving?

controllers/om/automation_config_test.go
mircea-cosbuc: Ok from my side, I didn't resolve it because of @lsierant 's comment.

docker/mongodb-kubernetes-tests/tests/webhooks/fixtures/invalid_oidc_mongodb_community.yaml
anandsyncs: Fixed here: https://github.com/mongodb/mongodb-kubernetes/pull/60/commits/15a1cbc4aff121a9acbd1489b0a6513fc5f161d9

docker/mongodb-kubernetes-tests/tests/webhooks/e2e_mongodb_validation_webhook.py
anandsyncs: Fixed here: https://github.com/mongodb/mongodb-kubernetes/pull/60/commits/15a1cbc4aff121a9acbd1489b0a6513fc5f161d9

anandsyncs:
evergreen retry

anandsyncs:
evergreen retry

lsierant (approved):
LGTM! Great work 👏

PR #62, opened 2025-04-30
CLOUDP-314921 Telemetry for authentication modes

MaciejKaras:
# Summary

Adds Telemetry support for Authentication mode and Agent Authentication Mode. New fields `authenticationModes` and `authenticationAgentModes` will take any value passed from CRD. 

Because of the limitations of Telemetry subsystem used (Segment) we need to send flat `map[string]any` without any nested fields. For multiple Authentication modes we will flatten the values to `authenticationMode***` where `***` is one of the possible modes specified in the CRD. For example when `SCRAM` and `OIDC` are enabled we will send telemetry output:
```go
{
        [...]
	"authenticationModeOIDC":   true,
	"authenticationModeSCRAM":  true,
	"authenticationAgentMode":  util.SCRAM,
}
```

## Proof of Work

Passing unit tests.

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/pkg/telemetry/types.go
+++ b/pkg/telemetry/types.go
@@ -62,10 +62,8 @@ func (u DeploymentUsageSnapshotProperties) ConvertToFlatMap() (map[string]any, e
 		return nil, fmt.Errorf("failed to parse properties: %w", err)
 	}
 
-	if len(u.AuthenticationModes) > 0 {
-		for _, value := range u.AuthenticationModes {
-			properties["authenticationMode"+value] = true
-		}
+	for _, value := range u.AuthenticationModes {
+		properties["authenticationMode"+value] = true
 	}
 
 	return properties, nil


controllers/operator/mongodbstandalone_controller.go
```diff
@@ -177,7 +177,7 @@ func (r *ReconcileMongoDbStandalone) Reconcile(ctx context.Context, request reco
 	// cannot have a non-tls deployment in an x509 environment
 	// TODO move to webhook validations
 	security := s.Spec.Security
-	if security.Authentication != nil && security.Authentication.Enabled && security.Authentication.IsX509Enabled() && !s.Spec.GetSecurity().IsTLSEnabled() {
+	if security.Authentication != nil && security.Authentication.IsX509Enabled() && !s.Spec.GetSecurity().IsTLSEnabled() {
```
anandsyncs: Why did we remove the `security.Authentication.Enabled` check, was this duplicate?

MaciejKaras: The check for `security.Authentication.Enabled` is already in `Authentication.IsX509Enabled()` method.

anandsyncs: I don't see an explicit check in the function:
`func (a *Authentication) IsX509Enabled() bool {
	return stringutil.Contains(a.GetModes(), util.X509)
}`

Is there something I missed?

**EDIT: Saw it in you PR now.**

controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -1014,7 +1014,7 @@ func (r *ShardedClusterReconcileHelper) doShardedClusterProcessing(ctx context.C
 	}
 
 	security := sc.Spec.Security
-	if security.Authentication != nil && security.Authentication.Enabled && security.Authentication.IsX509Enabled() && !sc.Spec.GetSecurity().IsTLSEnabled() {
+	if security.Authentication != nil && security.Authentication.IsX509Enabled() && !sc.Spec.GetSecurity().IsTLSEnabled() {
```
anandsyncs: nit: These long chain calls are difficult to follow, we can probably make more readable with less nested method call when there is a lot of them.

Not really related to this PR.

controllers/operator/mongodbstandalone_controller.go
MaciejKaras: Great :) can you resolve the comment?

controllers/operator/mongodbshardedcluster_controller.go
MaciejKaras: actually I think the check `security.Authentication != nil` is not needed, because we already do this in `security.Authentication.IsX509Enabled()`. Let me fix that

MaciejKaras: Done, plz take a look

pkg/telemetry/types.go
```diff
@@ -14,25 +22,53 @@ type OperatorUsageSnapshotProperties struct {
...
+type FakeDeploymentUsageSnapshotProperties DeploymentUsageSnapshotProperties
+
+func (u DeploymentUsageSnapshotProperties) ConvertToFlatMap() (map[string]any, error) {
+	// FakeDeploymentUsageSnapshotProperties is used to avoid infinite recursion - maputil.StructToMap will call MarshalJSON
+	properties, err := maputil.StructToMap(FakeDeploymentUsageSnapshotProperties(u))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse properties: %w", err)
+	}
+
+	if len(u.AuthenticationModes) > 0 {
```
anandsyncs: This check is not necessary

MaciejKaras: True!

SimonBaeumer (approved):

PR #66, opened 2025-04-30
CLOUDP-315085: Rename member list cm

Julien-Ben:
# Summary

## Why

The multi cluster member-list configmap is created by both the helm chart and kubectl tool.
It should not be created by helm. E.g one drawback is that the configmap can be removed upon release uninstallation, which leads to a panic in the currently running operator.
This is also confusing and it should be the responsibility of a single tool to create that map.

The name of that configmap is 
mongodb-enterprise-operator-member-list and we should take the opportunity to rename it to mongodb-kubernetes-operator-member-list, to stay consistent with our new naming conventions.

Customers deploying multi cluster resources must run the kubectl plugin when upgrading, so we can safely rename it.

## How

The config map name was mostly defined by constants so it's a small renaming.

## Proof of Work

E2E tests pass (including upgrade tests)

### DIFF
--- a/docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_sharded_cluster.py
+++ b/docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_sharded_cluster.py
@@ -180,6 +180,7 @@ class TestOperatorDowngrade:
             member_cluster_clients=None,
             member_cluster_names=None,
             custom_operator_version=LEGACY_DEPLOYMENT_STATE_VERSION,
+            operator_name=LEGACY_OPERATOR_NAME,
         )
         operator.assert_is_running()
         # Dumping deployments in logs ensures we are using the correct operator version


public/tools/multicluster/cmd/setup.go
```diff
@@ -21,6 +21,7 @@ func init() {
 	setupCmd.Flags().StringVar(&setupFlags.CentralCluster, "central-cluster", "", "The central cluster the operator will be deployed in. [required]")
 	setupCmd.Flags().StringVar(&setupFlags.MemberClusterNamespace, "member-cluster-namespace", "", "The namespace the member cluster resources will be deployed to. [required]")
 	setupCmd.Flags().StringVar(&setupFlags.CentralClusterNamespace, "central-cluster-namespace", "", "The namespace the Operator will be deployed to. [required]")
+	setupCmd.Flags().StringVar(&setupFlags.OperatorName, "operator-name", common.DefaultOperatorName, "Name used to identify the deployment of the operator. [optional, default: mongodb-enterprise-operator]")
```
Julien-Ben, line 24: Should the default be `mongodb-enterprise-operator` ?
If a user downloads the new CLI tool and wants to install MCK, I think it should work without providing any operator name.
But if a user downloads the new CLI tool, and wants to make it work with MEKO, then they should provide the old operator name to create the old configmap

nammn: yeah lets set the default to mck

mircea-cosbuc: The default is `mck` it's just the comment that got missed 🤦

Julien-Ben: I fixed it but I refrained from pushing to wait for CI to pass

PR #93, opened 2025-05-07
[observability] most basic OpenTelemetry integration into MCK

nammn:
# Summary
This pull request introduces OpenTelemetry tracing support to the MongoDB Kubernetes Operator and its related components. Key changes include the integration of OpenTelemetry libraries, the addition of tracing configuration, and updates to ensure trace propagation across the application. These changes enhance observability and debugging capabilities.

In our CI suite this means we will have the following kind of traces:
```
trace_id: abc123

                         ┌────────────────────┐
                         │     Evergreen      │
                         │  span_id: ROOT     │
                         │  parent_id: none   │
                         └─────────┬──────────┘
                                   │
        ┌──────────────────────────┼─────────────────────────┐
        ▼                          ▼                         ▼
┌──────────────┐         ┌────────────────┐         ┌────────────────────┐
│   E2E Test   │         │   Operator     │         │     (Other…)       │
│ span_id: A1  │         │ span_id: B1    │         │                    │
│ parent: ROOT │         │ parent: ROOT   │         │                    │
└──────┬───────┘         └──────┬─────────┘         └────────────────────┘
       │                        │
       ▼                        ▼
┌──────────────┐         ┌────────────────────┐
│ E2E Function │         │   Reconcile Loop   │
│ span_id: A2  │         │   span_id: B2      │
│ parent: A1   │         │   parent: B1       │
└──────────────┘         └────────────────────┘

```

### OpenTelemetry Integration:

* **Tracing in `main.go`:**
  - Added OpenTelemetry setup in the `main` function, including trace and span ID extraction from environment variables and the creation of a root span for the operator. Tracing context is propagated across controllers and shutdown processes are handled gracefully. 
* **Telemetry in `pkg/telemetry/client.go`:** <--- this is good to know if we happen to make a change and happen to send to prod atlas
  - Added a span to the `SendEventWithRetry` function to capture telemetry events and include the Atlas base URL as a span attribute. 

### Helm Chart Updates:

* **Operator configuration:**
  - Added OpenTelemetry-specific environment variables (`OTEL_TRACE_ID`, `OTEL_PARENT_ID`, `OTEL_EXPORTER_OTLP_ENDPOINT`) to the operator's deployment template. (`[helm_chart/templates/operator.yamlR83-R90](diffhunk://#diff-5d2e377a6806023ca9eff60be4d7e5cd879803de2bd3800b630f479f8728f322R83-R90)`)
  - Introduced OpenTelemetry configuration options (`enabled`, `traceID`, `parentID`, `collectorEndpoint`) in the Helm chart's `values.yaml`.

### Dependency Updates:

* **Go module dependencies:**
  - Added OpenTelemetry-related libraries (`otel`, `otel/sdk`, `otel/trace`, etc.) to `go.mod`.
## Proof of Work
- e.g. [patch](https://spruce.mongodb.com/task/mongodb_kubernetes_e2e_mdb_kind_ubi_cloudqa_e2e_replica_set_pv_patch_943128faa1f738781f2b0e7442c8d63077c9ecd5_682dce94ed55bd000781c215_25_05_21_13_01_10/logs?execution=0&sortBy=STATUS&sortDir=ASC)

- generated traces in our ci: [Link](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/evergreen-agent/trace/uNv82G92XFD?fields[]=s_name&fields[]=s_serviceName&span=418471a7ba33179f)
Screenshot 2025-05-21 at 15 19 20

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/LICENSE-THIRD-PARTY
+++ b/LICENSE-THIRD-PARTY
@@ -10,7 +10,7 @@ github.com/fsnotify/fsnotify,v1.7.0,https://github.com/fsnotify/fsnotify/blob/v1
 github.com/ghodss/yaml,v1.0.0,https://github.com/ghodss/yaml/blob/v1.0.0/LICENSE,MIT
 github.com/go-jose/go-jose/v4,v4.0.5,https://github.com/go-jose/go-jose/blob/v4.0.5/LICENSE,Apache-2.0
 github.com/go-jose/go-jose/v4/json,v4.0.5,https://github.com/go-jose/go-jose/blob/v4.0.5/json/LICENSE,BSD-3-Clause
-github.com/go-logr/logr,v1.4.2,https://github.com/go-logr/logr/blob/v1.4.2/LICENSE,Apache-2.0
+github.com/go-logr/logr,v1.4.3,https://github.com/go-logr/logr/blob/v1.4.3/LICENSE,Apache-2.0
 github.com/go-logr/stdr,v1.2.2,https://github.com/go-logr/stdr/blob/v1.2.2/LICENSE,Apache-2.0
 github.com/go-logr/zapr,v1.3.0,https://github.com/go-logr/zapr/blob/v1.3.0/LICENSE,Apache-2.0
 github.com/go-openapi/jsonpointer,v0.19.6,https://github.com/go-openapi/jsonpointer/blob/v0.19.6/LICENSE,Apache-2.0
@@ -28,13 +28,13 @@ github.com/grpc-ecosystem/grpc-gateway/v2,v2.16.0,https://github.com/grpc-ecosys
 github.com/hashicorp/errwrap,v1.1.0,https://github.com/hashicorp/errwrap/blob/v1.1.0/LICENSE,MPL-2.0
 github.com/hashicorp/go-cleanhttp,v0.5.2,https://github.com/hashicorp/go-cleanhttp/blob/v0.5.2/LICENSE,MPL-2.0
 github.com/hashicorp/go-multierror,v1.1.1,https://github.com/hashicorp/go-multierror/blob/v1.1.1/LICENSE,MPL-2.0
-github.com/hashicorp/go-retryablehttp,v0.7.7,https://github.com/hashicorp/go-retryablehttp/blob/v0.7.7/LICENSE,MPL-2.0
+github.com/hashicorp/go-retryablehttp,v0.7.8,https://github.com/hashicorp/go-retryablehttp/blob/v0.7.8/LICENSE,MPL-2.0
 github.com/hashicorp/go-rootcerts,v1.0.2,https://github.com/hashicorp/go-rootcerts/blob/v1.0.2/LICENSE,MPL-2.0
 github.com/hashicorp/go-secure-stdlib/parseutil,v0.1.6,https://github.com/hashicorp/go-secure-stdlib/blob/parseutil/v0.1.6/parseutil/LICENSE,MPL-2.0
 github.com/hashicorp/go-secure-stdlib/strutil,v0.1.2,https://github.com/hashicorp/go-secure-stdlib/blob/strutil/v0.1.2/strutil/LICENSE,MPL-2.0
 github.com/hashicorp/go-sockaddr,v1.0.2,https://github.com/hashicorp/go-sockaddr/blob/v1.0.2/LICENSE,MPL-2.0
-github.com/hashicorp/hcl,v1.0.0,https://github.com/hashicorp/hcl/blob/v1.0.0/LICENSE,MPL-2.0
-github.com/hashicorp/vault/api,v1.16.0,https://github.com/hashicorp/vault/blob/api/v1.16.0/api/LICENSE,MPL-2.0
+github.com/hashicorp/hcl,v1.0.1-vault-7,https://github.com/hashicorp/hcl/blob/v1.0.1-vault-7/LICENSE,MPL-2.0
+github.com/hashicorp/vault/api,v1.20.0,https://github.com/hashicorp/vault/blob/api/v1.20.0/api/LICENSE,MPL-2.0
 github.com/imdario/mergo,v0.3.15,https://github.com/imdario/mergo/blob/v0.3.15/LICENSE,BSD-3-Clause
 github.com/josharian/intern,v1.0.0,https://github.com/josharian/intern/blob/v1.0.0/license.md,MIT
 github.com/json-iterator/go,v1.1.12,https://github.com/json-iterator/go/blob/v1.1.12/LICENSE,MIT
@@ -52,7 +52,7 @@ github.com/prometheus/common,v0.62.0,https://github.com/prometheus/common/blob/v
 github.com/prometheus/procfs,v0.15.1,https://github.com/prometheus/procfs/blob/v0.15.1/LICENSE,Apache-2.0
 github.com/r3labs/diff/v3,v3.0.1,https://github.com/r3labs/diff/blob/v3.0.1/LICENSE,MPL-2.0
 github.com/ryanuber/go-glob,v1.0.0,https://github.com/ryanuber/go-glob/blob/v1.0.0/LICENSE,MIT
-github.com/spf13/cast,v1.8.0,https://github.com/spf13/cast/blob/v1.8.0/LICENSE,MIT
+github.com/spf13/cast,v1.9.2,https://github.com/spf13/cast/blob/v1.9.2/LICENSE,MIT
 github.com/spf13/pflag,v1.0.5,https://github.com/spf13/pflag/blob/v1.0.5/LICENSE,BSD-3-Clause
 github.com/stretchr/objx,v0.5.2,https://github.com/stretchr/objx/blob/v0.5.2/LICENSE,MIT
 github.com/stretchr/testify/assert,v1.10.0,https://github.com/stretchr/testify/blob/v1.10.0/LICENSE,MIT
@@ -69,6 +69,7 @@ go.opentelemetry.io/otel/trace,v1.35.0,https://github.com/open-telemetry/opentel
 go.opentelemetry.io/proto/otlp,v1.0.0,https://github.com/open-telemetry/opentelemetry-proto-go/blob/otlp/v1.0.0/otlp/LICENSE,Apache-2.0
 go.uber.org/multierr,v1.11.0,https://github.com/uber-go/multierr/blob/v1.11.0/LICENSE.txt,MIT
 go.uber.org/zap,v1.27.0,https://github.com/uber-go/zap/blob/v1.27.0/LICENSE,MIT
+go.yaml.in/yaml/v2,v2.4.2,https://github.com/yaml/go-yaml/blob/v2.4.2/LICENSE,Apache-2.0
 gomodules.xyz/jsonpatch/v2,v2.4.0,https://github.com/gomodules/jsonpatch/blob/v2.4.0/v2/LICENSE,Apache-2.0
 google.golang.org/genproto/googleapis/api/httpbody,v0.0.0-20230726155614-23370e0ffb3e,https://github.com/googleapis/go-genproto/blob/23370e0ffb3e/googleapis/api/LICENSE,Apache-2.0
 google.golang.org/genproto/googleapis/rpc,v0.0.0-20230822172742-b8732ec3820d,https://github.com/googleapis/go-genproto/blob/b8732ec3820d/googleapis/rpc/LICENSE,Apache-2.0
@@ -92,5 +93,4 @@ k8s.io/utils/internal/third_party/forked/golang/net,v0.0.0-20240502163921-fe8a2d
 sigs.k8s.io/controller-runtime,v0.18.7,https://github.com/kubernetes-sigs/controller-runtime/blob/v0.18.7/LICENSE,Apache-2.0
 sigs.k8s.io/json,v0.0.0-20221116044647-bc3834ca7abd,https://github.com/kubernetes-sigs/json/blob/bc3834ca7abd/LICENSE,Apache-2.0
 sigs.k8s.io/structured-merge-diff/v4,v4.4.1,https://github.com/kubernetes-sigs/structured-merge-diff/blob/v4.4.1/LICENSE,Apache-2.0
-sigs.k8s.io/yaml,v1.4.0,https://github.com/kubernetes-sigs/yaml/blob/v1.4.0/LICENSE,Apache-2.0
-sigs.k8s.io/yaml/goyaml.v2,v1.4.0,https://github.com/kubernetes-sigs/yaml/blob/v1.4.0/goyaml.v2/LICENSE,Apache-2.0
+sigs.k8s.io/yaml,v1.5.0,https://github.com/kubernetes-sigs/yaml/blob/v1.5.0/LICENSE,Apache-2.0


scripts/evergreen/deployments/test-app/templates/mongodb-enterprise-tests.yaml
```diff
@@ -80,8 +80,9 @@ spec:
     - name: OTEL_RESOURCE_ATTRIBUTES
       value: {{ .Values.otel_resource_attributes }}
     {{ end }}
+    # Set service name to identify this component in the trace
     - name: OTEL_SERVICE_NAME
-      value: evergreen-agent
+      value: mongodb-e2e-tests
```
nammn, line 85: updated the service-name for better querying

scripts/evergreen/e2e/e2e.sh
```diff
@@ -102,48 +146,21 @@ else
...
-      for member_cluster in ${MEMBER_CLUSTERS}; do
-        echo "Tearing down cluster ${member_cluster}"
-        reset_namespace "${member_cluster}" "${NAMESPACE}" || true
-      done
-    else
-      # If the test pass, then the namespace is removed
-      reset_namespace "$(kubectl config current-context)" "${NAMESPACE}" || true
-    fi
+    # If the test passed, then the namespace is removed
+    delete_operator "${NAMESPACE}"
```
nammn: we should remove the operator to test teardown at the end of the test-run + it enables us to ensure traces are exported in time

main.go
```diff
@@ -123,11 +127,89 @@ func main() {
 	// Namespace where the operator is installed
 	currentNamespace := env.ReadOrPanic(util.CurrentNamespace)
 
+	// Get trace and span IDs from environment variables
+	traceIDHex := os.Getenv("OTEL_TRACE_ID")
```
lsierant, line 142: why Hex in the name?

helm_chart/values.yaml
```diff
@@ -32,6 +32,13 @@ operator:
   - mongodbcommunity
   - mongodbsearch
 
+  # OpenTelemetry configuration
+  opentelemetry:
```
lsierant: Is this supposed to be ever configured by the users? I wonder if we actually need that exposed. I'm a bit worried that it will be confusing why we have `operator.telemetry` and `operator.opentelemetry` in values.yaml. How this is related to each other, is there a possibility for the users to confuse the two and expect no telemetry at all when opentelemetry=false? 

I think we're lacking docs or explanations here. If we don't have good docs/story for it, I'd prefer to not expose the settings here at all.

main.go
```diff
@@ -123,11 +127,89 @@ func main() {
...
+	spanIDHex := os.Getenv("OTEL_PARENT_ID")
+	endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+
 	// Get a config to talk to the apiserver
 	cfg := ctrl.GetConfigOrDie()
 
+	log.Infof("Setting up tracing with ID: %s, Parent ID: %s, Endpoint: %s", traceIDHex, spanIDHex, endpoint)
+	traceCtx, tp, err := telemetry.SetupTracing(signalCtx, traceIDHex, spanIDHex, endpoint)
+	if err != nil {
+		log.Errorf("Failed to setup tracing: %v", err)
```
lsierant, line 152: can we change every message where we mention "tracing" to have it as "opentelemetry tracing"?

helm_chart/values.yaml
lsierant: I'm thinking that what we use otel here is just **tracing**. What if we don't mention any "telemetry" but always refer to it as tracing to not confuse with our telemetry gathering?

main.go
```diff
@@ -123,11 +127,89 @@ func main() {
...
+		// Continue without tracing instead of failing
+		traceCtx = signalCtx
+	} else if tp != nil {
```
lsierant: can we extract  that tracing code to a separate file/function? It's a quite a chunk of pretty verbose code.

main.go
```diff
@@ -123,11 +127,89 @@ func main() {
...
+	// We need to manually flush the span when the program ends
+	defer func() {
+		// Add more context to the root span before ending it
+		operatorSpan.SetAttributes(
+			attribute.String("operator.shutdown", "true"),
+			attribute.String("operator.span.status", "complete"),
+			attribute.Bool("root", true),
+		)
+
+		log.Info("Ending root operator span and flushing it to exporter")
```
lsierant: The messages on Info level will appear in normal, regular logs for every customer. And it looks like this is not guarded by having otel enabled or not. 

Please review every log message and consider if the customer seeing it in the logs will know what is this about and not be confusing.

pkg/telemetry/collector.go
```diff
@@ -82,8 +82,13 @@ func (l *LeaderRunnable) Start(ctx context.Context) error {
 type snapshotCollector func(ctx context.Context, memberClusterMap map[string]ConfigClient, operatorClusterMgr manager.Manager, operatorUUID, mongodbImage, databaseNonStaticImage string) []Event
 
 // RunTelemetry lists the specified CRDs and sends them as events to Segment
-func RunTelemetry(ctx context.Context, mongodbImage, databaseNonStaticImage, namespace string, operatorClusterMgr manager.Manager, clusterMap map[string]cluster.Cluster, atlasClient *Client, configuredOperatorEnv util.OperatorEnvironment) {
-	Logger.Debug("sending telemetry!")
+func RunTelemetry(leaderTrace context.Context, mongodbImage, databaseNonStaticImage, namespace string, operatorClusterMgr manager.Manager, clusterMap map[string]cluster.Cluster, atlasClient *Client, configuredOperatorEnv util.OperatorEnvironment) {
+	Logger.Debug("Collecting telemetry!")
```
lsierant, line 85: uppercased Logger is quite surprising to me. I know it's been there before this change, but can you explain why it's a different convention than other log *zap.SugaredLog conventions we use?

lsierant (changes requested):
This is a very, very nice change and I would like to have it included. But as of right now I think we should discuss this PR a bit more before deciding to merge. There are too many end-user implications to just LGTM it now.

helm_chart/values.yaml
nammn: right now its just tracing - but we might more in the future. But for now its tightly coupled. We could not expose this and just use it for us

main.go
nammn: because the api follows hex: https://www.w3.org/TR/trace-context/#parent-id

pkg/telemetry/collector.go
nammn: The reason was to expose a global package level Logger which is always ready and correctly initialized for the independent telemetry module. This saves us the need for creating a new sugaredLogger everytime + the potential of configuring it wrongly

main.go
nammn: done!

main.go
nammn: all reworked to debug, thanks for the catch!

helm_chart/values.yaml
nammn: i don't necessarily want to expose this; but how can we use it while not exposing it in our current e2e test infrastructure?

main.go
```diff
@@ -123,11 +127,39 @@ func main() {
...
+	// Get trace and span IDs from environment variables
+	traceIDHex := os.Getenv("OTEL_TRACE_ID")
+	spanIDHex := os.Getenv("OTEL_PARENT_ID")
+	endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+
 	// Get a config to talk to the apiserver
 	cfg := ctrl.GetConfigOrDie()
 
+	log.Debugf("Setting up tracing with ID: %s, Parent ID: %s, Endpoint: %s", traceIDHex, spanIDHex, endpoint)
+	traceCtxWithParentSpan, tp, err := telemetry.SetupTracingFromParent(signalCtx, traceIDHex, spanIDHex, endpoint)
```
lsierant: can we simplify this? 
We don't need to create new var names for each span - we can reuse ctx. Methods inside are returning ctx if err, so we don't lose anything if error happens.

main.go
```diff
@@ -123,11 +127,39 @@ func main() {
...
+	if err != nil {
+		log.Warnf("Failed to setup opentelemetry tracing: %v", err)
+		// Continue without tracing instead of failing
+		traceCtxWithParentSpan = signalCtx
+	} else {
+		defer func() {
+			if tp != nil {
+				shutdownTracerProvider(signalCtx, tp)
+			} else {
+				log.Warn("No tracer provider created, opentelemetry tracing will be disabled")
```
lsierant: this log could be in the same case as if err != nil above

main.go
```diff
@@ -123,11 +127,39 @@ func main() {
...
+			}
+		}()
+	}
+
+	ctxWithSpan, operatorSpan := startRootSpan(currentNamespace, spanIDHex, traceCtxWithParentSpan)
```
lsierant: another ctx var name - can we just shadow ctx?

main.go
```diff
@@ -123,11 +127,39 @@ func main() {
...
+	defer operatorSpan.End()
+
 	managerOptions := ctrl.Options{
 		Scheme: scheme,
+		BaseContext: func() context.Context {
+			// Ensures every controller gets the trace- and signal-aware context
```
lsierant:
```suggestion
			// Ensures every controller gets the trace and signal-aware context
```

main.go
```diff
@@ -153,7 +185,7 @@ func main() {
 		managerOptions.HealthProbeBindAddress = "127.0.0.1:8181"
 	}
 
-	webhookOptions := setupWebhook(ctx, cfg, log, slices.Contains(crds, mongoDBMultiClusterCRDPlural), currentNamespace)
+	webhookOptions := setupWebhook(ctxWithSpan, cfg, log, slices.Contains(crds, mongoDBMultiClusterCRDPlural), currentNamespace)
```
lsierant: we don't need ctxWithSpan, IUUC ctxWithSpan is the same type as `ctx` but with some additional values (what I mean is that we don't need to add new var names for spanned contexts)

helm_chart/values.yaml
lsierant: can we remove it from the helm's values.yaml and treat it as optional in the template?

main.go
```diff
@@ -123,11 +127,39 @@ func main() {
 	// Namespace where the operator is installed
 	currentNamespace := env.ReadOrPanic(util.CurrentNamespace)
 
+	// Get trace and span IDs from environment variables
+	traceIDHex := os.Getenv("OTEL_TRACE_ID")
```
lsierant, line 142: pls add OTEL_ to `printableEnvPrefixes`

scripts/funcs/kubernetes
```diff
@@ -142,9 +141,8 @@ reset_namespace() {
 
   set +e
 
-  helm uninstall --kube-context="${context}" mongodb-enterprise-operator || true &
-  helm uninstall --kube-context="${context}" mongodb-community-operator || true &
-  helm uninstall --kube-context="${context}" mongodb-enterprise-operator-multi-cluster || true &
+  helm uninstall --kube-context="${context}" mongodb-kubernetes-operator || true &
```
nammn, line 144: we have one operator now - this is a cleanup

scripts/funcs/kubernetes
```diff
@@ -162,7 +160,7 @@ reset_namespace() {
   # a while to delete it.
   should_wait="false"
   # shellcheck disable=SC2153
-  if [[ ${CURRENT_VARIANT_CONTEXT} == e2e_mdb_openshift_ubi_cloudqa || ${CURRENT_VARIANT_CONTEXT} == e2e_openshift_static_mdb_ubi_cloudqa ]]; then
+  if [[ ${KUBE_ENVIRONMENT_NAME} == "openshift" ]]; then
```
nammn, line 163: cleanup

scripts/funcs/kubernetes
```diff
@@ -35,11 +35,10 @@ EOF
 
 delete_operator() {
     local ns="$1"
-    local name=${OPERATOR_NAME:=mongodb-enterprise-operator}
+    local name=${OPERATOR_NAME:=mongodb-kubernetes-operator}
```
nammn, line 38: this was not running before and therefore we had no flushing

scripts/funcs/kubernetes
```diff
@@ -35,11 +35,10 @@ EOF
...
 
     title "Removing the Operator deployment ${name}"
-    ! kubectl --namespace "${ns}" get deployments | grep -q "${name}" \
-        || kubectl delete deployment "${name}" -n "${ns}" || true
+    kubectl delete deployment "${name}" -n "${ns}" --wait=true --timeout=10s|| true
```
nammn, line 41: let's wait until all things are flushed before we stop. PoW is in the PR description - you can see the trace of the oeprator

helm_chart/templates/operator.yaml
```diff
@@ -83,6 +83,16 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+    {{- $opentelemetry := default dict .Values.operator.opentelemetry }}
```
nammn, line 86: pow:

```
(venv) ~/projects/ops-manager-kubernetes git:[traces-operator]
helm template helm_chart | rg OTEL
(venv) ~/projects/ops-manager-kubernetes git:[traces-operator]
helm template --set operator.opentelemetry.tracing.enabled=true --set operator.opentelemetry.tracing.traceID=your-trace-id --set operator.opentelemetry.tracing.parentID=your-parent-id --set operator.opentelemetry.tracing.collectorEndpoint=http://jaeger:14268/api/traces helm_chart | rg OTEL
            - name: OTEL_TRACE_ID
            - name: OTEL_PARENT_ID
            - name: OTEL_EXPORTER_OTLP_ENDPOINT
(venv) ~/projects/ops-manager-kubernetes git:[traces-operator]
helm template --set operator.opentelemetry.tracing.enabled=true --set operator.opentelemetry.tracing.traceID=your-trace-id --set operator.opentelemetry.tracing.parentID=your-parent-id --set operator.opentelemetry.tracing.collectorEndpoint=http://jaeger:14268/api/traces helm_chart | rg OTEL -C 2
                fieldRef:
                  fieldPath: metadata.namespace
            - name: OTEL_TRACE_ID
              value: "your-trace-id"
            - name: OTEL_PARENT_ID
              value: "your-parent-id"
            - name: OTEL_EXPORTER_OTLP_ENDPOINT
              value: "http://jaeger:14268/api/traces"
            - name: WATCH_NAMESPACE
```

scripts/evergreen/e2e/e2e.sh
```diff
@@ -149,7 +149,7 @@ fi
 dump_cluster_information
 
 # We only have static clusters in OpenShift; otherwise, there's no need to mark and clean them up here.
-if [[ "${CLUSTER_TYPE}" == "openshift" ]]; then
+if [[ "${KUBE_ENVIRONMENT_NAME}" == *openshift* ]]; then
```
nammn, line 152:
```
KUBE_ENVIRONMENT_NAME="dev-openshift-cluster"

if [[ "${KUBE_ENVIRONMENT_NAME}" == *openshift* ]]; then
  echo "Contains openshift"
else
  echo "Does NOT contain openshift"
fi
```

-> Contains openshift

scripts/funcs/operator_deployment
```diff
@@ -69,9 +69,6 @@ get_operator_helm_values() {
     comma_separated_list="$(echo "${MEMBER_CLUSTERS}" | tr ' ' ',')"
     # shellcheck disable=SC2154
     config+=("multiCluster.clusters={${comma_separated_list}}")
-  fi
-
-  if [[ "${KUBE_ENVIRONMENT_NAME:-}" == "multi" ]]; then
```
nammn, line 57: for some reason it was duplicated

main.go
```diff
@@ -301,12 +326,38 @@ func main() {
...
 		log.Fatal(err)
 	}
 }
 
+func startRootSpan(currentNamespace string, spanIDHex string, traceCtx context.Context) (context.Context, trace.Span) {
+	opts := []trace.SpanStartOption{
+		trace.WithAttributes(
+			attribute.String("component", "operator"),
+			attribute.String("namespace", currentNamespace),
+			attribute.String("service.name", "mongodb-kubernetes-operator"),
```
lucian-tosa, line 339: Should this be using the actual operator name instead of being hardcoded?

lucian-tosa: Maybe it doesn't matter because we configure the namespace

lucian-tosa (approved):

main.go
nammn: do you mean service.name? In OTEL its used to identify a "group" of items. Usually its a microservice, in our case its our operator e2e tests. We wouldn't want a too high cardinality. We could make it configurable - but I was afraid that customers might deploy multiple operators and each operator would be its own group 🤔

lucian-tosa: Ah got it, wasn't aware of the use case

PR #97, opened 2025-05-07
Update MEKO to MCK OLM tests

m1kola:
# Summary

Instead of relying on MEKO -> MCK upgrade path in a catalog metadata, we now use uninstall-then-install migration approach.

While MEKO -> MCK upgrade path is technically possible, it is not allowed in operatorhub.io and Red Hat certified operators catalogs as it is prevented by validation.

## Proof of Work

CI must be green.

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
+++ b/docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
@@ -16,6 +16,7 @@ from kubetester.mongodb import Phase
 from kubetester.mongodb_user import MongoDBUser
 from kubetester.opsmanager import MongoDBOpsManager
 from pytest import fixture
+from tests import test_logger
 from tests.conftest import LEGACY_OPERATOR_NAME, OPERATOR_NAME
 from tests.olm.olm_test_commons import (
     get_catalog_image,
@@ -28,6 +29,8 @@ from tests.olm.olm_test_commons import (
     wait_for_operator_ready,
 )
 from tests.opsmanager.om_ops_manager_backup import create_aws_secret, create_s3_bucket
+
+logger = test_logger.get_test_logger(__name__)
 
 # See docs how to run this locally: https://wiki.corp.mongodb.com/display/MMS/E2E+Tests+Notes#E2ETestsNotes-OLMtests
 
@@ -395,15 +398,79 @@ def uninstall_meko_operator(namespace: str, meko_subscription: CustomObject):
 
 
 @pytest.mark.e2e_olm_meko_operator_upgrade_with_resources
-def test_migrate_meko_to_mck_operator(
+def test_uninstall_meko_operator(
     namespace: str,
-    version_id: str,
-    catalog_source: CustomObject,
     meko_subscription: CustomObject,
 ):
     # Uninstall the MEKO operator
     uninstall_meko_operator(namespace, meko_subscription)
 
+    # Get a list of all statefulsets
+    api_instance = kubernetes.client.AppsV1Api()
+    statefulsets = api_instance.list_namespaced_stateful_set(namespace)
+
+    # Kill one pod from each statefulset to simulate reschedule
+    for sts in statefulsets.items:
+        sts_name = sts.metadata.name
+        logger.info(f"Processing StatefulSet {sts_name}")
+
+        # Get pods for this statefulset
+        if sts.spec.selector and sts.spec.selector.match_labels:
+            # Build label selector string from match_labels dictionary
+            selector_parts = []
+            for key, value in sts.spec.selector.match_labels.items():
+                selector_parts.append(f"{key}={value}")
+            label_selector = ",".join(selector_parts)
+
+            pods = kubernetes.client.CoreV1Api().list_namespaced_pod(namespace, label_selector=label_selector)
+
+            if pods.items:
+                # Delete the first pod
+                pod_name = pods.items[0].metadata.name
+                logger.info(f"Deleting pod {pod_name} from StatefulSet {sts_name}")
+                kubernetes.client.CoreV1Api().delete_namespaced_pod(
+                    name=pod_name, namespace=namespace, body=kubernetes.client.V1DeleteOptions()
+                )
+
+    # Wait for all statefulsets to be ready again
+    def all_statefulsets_ready():
+        for sts in api_instance.list_namespaced_stateful_set(namespace).items:
+            if sts.status.ready_replicas != sts.status.replicas:
+                return (
+                    False,
+                    f"StatefulSet {sts.metadata.name} has {sts.status.ready_replicas}/{sts.status.replicas} ready replicas",
+                )
+        return True, "All StatefulSets are ready"
+
+    run_periodically(
+        all_statefulsets_ready, timeout=600, msg=f"Waiting for all StatefulSets to be ready after pod deletion"
+    )
+
+
+@pytest.mark.e2e_olm_meko_operator_upgrade_with_resources
+def test_connectivity_after_meko_uninstall(
+    ca_path: str,
+    ops_manager: MongoDBOpsManager,
+    oplog_replica_set: MongoDB,
+    blockstore_replica_set: MongoDB,
+    s3_replica_set: MongoDB,
+    mdb_sharded: MongoDB,
+):
+    """Verify resources are still connectable after MEKO operator uninstall but before MCK operator install"""
+    wait_for_om_healthy_response(ops_manager)
+
+    oplog_replica_set.assert_connectivity()
+    blockstore_replica_set.assert_connectivity()
+    s3_replica_set.assert_connectivity()
+    mdb_sharded.assert_connectivity(ca_path=ca_path)
+
+
+@pytest.mark.e2e_olm_meko_operator_upgrade_with_resources
+def test_install_mck_operator(
+    namespace: str,
+    version_id: str,
+    catalog_source: CustomObject,
+):
     current_operator_version = get_current_operator_version()
     incremented_operator_version = increment_patch_version(current_operator_version)
 


docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
```diff
@@ -344,59 +369,60 @@ def test_resources_in_running_state_before_upgrade(
...
+    # Load the subscription from API server
+    # so we can get CSV name from status
+    meko_subscription.load()
+    csv_name = meko_subscription["status"]["installedCSV"]
+
+    # Delete the subscription
+    meko_subscription.delete()
+
+    # Delete ClusterServiceVersion
+    api_instance = kubernetes.client.CustomObjectsApi()
```
nammn, line 390: does deleting the CSV cause anything to be deleted which is "related" to the csv? E.g. deployments/operator etc?

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
```diff
@@ -344,59 +369,60 @@ def test_resources_in_running_state_before_upgrade(
...
-            subscription.update()
-            return True
-        except kubernetes.client.ApiException as e:
-            if e.status == 409:
-                return False
-        else:
-            raise e
-
-    run_periodically(update_subscription, timeout=100, msg="Subscription to be updated")
+    # Create MCK subscription
```
nammn, line 477: q: for my understanding - this means we will not have 2 different subscriptions running at the same time; but rather delete `MEKO` subscription first and then install `MCK` one?

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
```diff
@@ -344,59 +369,60 @@ def test_resources_in_running_state_before_upgrade(
...
-    # Wait for the first resource to become reconciling after operator upgrade.
-    # Only then wait for all to not get a false positive when all resources are ready,
-    # because the upgraded operator haven't started reconciling
+def test_one_resource_not_in_running_state(ops_manager: MongoDBOpsManager):
+    # Wait for the first resource to become reconciling after operator replacement.
+    # This confirms the MCK operator has started reconciling the resources
     ops_manager.om_status().assert_reaches_phase(Phase.Pending, timeout=600)
 
 
 @pytest.mark.e2e_olm_meko_operator_upgrade_with_resources
```
nammn, line 491: nit: in all of those OLM we should have always tested general connectivity during the upgrade. Here we seem to only test this at the end 😅

scripts/evergreen/operator-sdk/prepare-openshift-bundles-for-e2e.sh
```diff
@@ -4,17 +4,20 @@ set -Eeou pipefail
 
 # This script prepares and publishes OpenShift bundles and catalog source for operator upgrade tests using OLM.
 # It enables testing two upgrade scenarios:
-#   1. Upgrade from the latest release of MEKO (MongoDB Enterprise Kubernetes Operator)
-#      to the current version of MCK (MongoDB Controllers for Kubernetes).
+#   1. Migration from the latest release of MEKO (MongoDB Enterprise Kubernetes Operator)
+#      to the current version of MCK (MongoDB Controllers for Kubernetes)
```
nammn, line 8: is it "current" or version 1.0.0?

config/manifests/bases/mongodb-kubernetes.clusterserviceversion.yaml
```diff
@@ -443,5 +443,4 @@ spec:
   maturity: stable
   provider:
     name: MongoDB, Inc
-  replaces: mongodb-enterprise.v1.33.0
```
nammn, line 446: q: but this will come back as soon as we have MCK:1.1.0 release, right?

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
```diff
@@ -89,6 +89,33 @@ def subscription(namespace: str, catalog_source: CustomObject):
...
+    Create a subscription object for the MCK operator.
+    This is a separate function (not a fixture) so it can be called after uninstalling MEKO.
+    """
+    static_value = get_default_architecture()
+    return get_subscription_custom_object(
+        OPERATOR_NAME,
+        namespace,
+        {
+            "channel": "migration",
+            "name": "mongodb-kubernetes",
```
nammn, line 106: do we have a variable for this?

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
Julien-Ben: q: I have the same question, what happens to CRDs when we uninstall MEKO through OLM ?
Since the test pass and the resources eventually reach Running Phase, I assume there's no issue. But I'm curious about how OLM handles the CRDs.

config/manifests/bases/mongodb-kubernetes.clusterserviceversion.yaml
m1kola: Yes, we need to add it back to any version after 1.0.0 (be it 1.1.0 or 1.0.1).

[This piece of code](https://github.com/mongodb/mongodb-kubernetes/pull/97/files#diff-362db66ba4c033cb3e6d12456bb89ac5782882ffc6f393e8f7d8d36b69aee0ffL110-L115) should add it back.

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
m1kola: We have a constant for operator name `LEGACY_OPERATOR_NAME` and `OPERATOR_NAME`, but not for OLM packages.

nammn: got it

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
m1kola: OLM never deletes CRDs installed with the bundle to prevent data loss.

Deleting subscription does not delete resources, but deleting CSV does. It deletes MEKO deployment and also supposed to delete RBAC accordingly [to the docs](https://olm.operatorframework.io/docs/tasks/uninstall-operator/#step-3-delete-the-operators-clusterserviceversion-csv):
> By deleting a `CSV`, OLM will delete the resources it created for the operator such as the `deployment`, `RBAC`, and any corresponding `CSVs` that OLM “Copied” into other namespaces watched by the operator.

However, our testing shows that OLM in fact does not delete RBAC. This means that either docs are incorrect/outdated or it is a regression in OLM.

We do not want OLM to delete RBAC and how it acts now is what we want. We initially didn't want to rely on the behaviour which is documented this way in OLM. This is one of the reasons why we went with MEKO -> MCK upgrade path with replaces (which in turn also undocumented, but seemed like a lesser of two evils). But now that we don't have a choice we are going with this option.

More info on two options can be found in [this doc](https://docs.google.com/document/d/12xTauemWsoEm9FG2cT_l-VW5KbUUYkKc9RN6YzEzCZc/edit?tab=t.oeh7pxtqthyl#heading=h.4fhdf31pqefl): original plan was [Option 2](https://docs.google.com/document/d/12xTauemWsoEm9FG2cT_l-VW5KbUUYkKc9RN6YzEzCZc/edit?tab=t.oeh7pxtqthyl#heading=h.cqtk4dewiafs) and now we are going for [Option 1](https://docs.google.com/document/d/12xTauemWsoEm9FG2cT_l-VW5KbUUYkKc9RN6YzEzCZc/edit?tab=t.oeh7pxtqthyl#heading=h.tsd6h96m0okp).

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
m1kola: Yes, correct.

scripts/evergreen/operator-sdk/prepare-openshift-bundles-for-e2e.sh
m1kola: It is current (build from the current checkout). Pinning it to 1.0.0 will add additional complexity. We can cherry-pick this PR to `release-1.0.0` branch to improve confidence.

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
m1kola: Yes, it is a good idea. I thought about it this morning as well. It is less of a concern in normal OLM upgrade tests, but I think it is important for uninstall-install migration.

I'll look into it now.

nammn: you could rely on something like this: https://github.com/mongodb/mongodb-kubernetes/blob/12d1b9082990481b7c32325d8d48589b24938b2b/docker/mongodb-kubernetes-tests/tests/shardedcluster/sharded_cluster_upgrade_downgrade.py#L69-L70

scripts/evergreen/operator-sdk/prepare-openshift-bundles-for-e2e.sh
nammn: release-1.0.0 branch has been merged back to master. We still have the branch but nothing long-running supported yet OOTB.
We might want to discuss how we want to handle this in the future via discuss/ticket (not here)

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
nammn: we just discussed in a call to add pod bouncing to tests to verify things still reconcile/start

m1kola: Added pod killing in 1e49e9ff443f260fab7ddb548e9341d3207b3d10

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
m1kola: Added connectivity checker in 1e49e9ff443f260fab7ddb548e9341d3207b3d10. 

I just use `assert_connectivity` as in other tests. This happens after we check that STSes are healthy again. I think this is enough, but I can look into adding `mdb_health_checker` if we think this is necessary.

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
```diff
@@ -395,15 +398,79 @@ def uninstall_meko_operator(namespace: str, meko_subscription: CustomObject):
...
+def test_uninstall_meko_operator(
     namespace: str,
-    version_id: str,
-    catalog_source: CustomObject,
     meko_subscription: CustomObject,
 ):
     # Uninstall the MEKO operator
     uninstall_meko_operator(namespace, meko_subscription)
 
+    # Get a list of all statefulsets
```
nammn, line 408: do you think it makes sense to do one quick check whether the appdb sa is still around? Its the most likely to cause problems if gone

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
Julien-Ben: Awesome, I think this gives enough confidence that deleting MEKO keeps the deployment stable.

docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py
m1kola: I think we should be good as is:
1. We kill app db pod as well here and check that all stateful sets recovered eventually
2. With `mongodb-enterprise-database-pods` role (shipped in the bundle) gone none of the DB resources would be re-created.

nammn: yes - thats true!

PR #129, opened 2025-05-15
Add test cases for test_calculate_images_to_build with real examples

SimonBaeumer:
# Summary

Add test cases for test_calculate_images_to_build with real examples

## Proof of Work

Unit tests passing

### DIFF
--- a/pipeline_test.py
+++ b/pipeline_test.py
@@ -75,6 +75,21 @@ def test_operator_build_configuration_defaults():
         (["a", "b", "c"], None, ["d"], ValueError),
         ([], ["a"], ["b"], ValueError),
         (["a", "b", "c"], None, None, {"a", "b", "c"}),
+
+        # Given an include, it should only return include images
+        (["cli", "ops-manager", "appdb-daily", "init-appdb"], ["cli"], [], {"cli"}),
+
+        # Given no include nor excludes it should return all images
+        (["cli", "ops-manager", "appdb-daily", "init-appdb"], [], [], {'init-appdb', 'appdb-daily', 'ops-manager', 'cli'}),
+
+        # Given an exclude, it should return all images except the excluded ones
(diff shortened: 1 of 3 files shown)


pipeline_test.py
```diff
@@ -75,6 +75,21 @@ def test_operator_build_configuration_defaults():
...
+        (["cli", "ops-manager", "appdb-daily", "init-appdb"], [], [], {'init-appdb', 'appdb-daily', 'ops-manager', 'cli'}),
+
+        # Given an exclude, it should return all images except the excluded ones
+        (["cli", "ops-manager", "appdb-daily", "init-appdb"], [], ['init-appdb', 'appdb-daily'], {'ops-manager', 'cli'}),
+
+        # Given an include and a different exclude, it should return all images except the exclusions
+        (["cli", "ops-manager", "appdb-daily", "init-appdb"], ['appdb-daily'], ['init-appdb'], {'appdb-daily', 'cli', 'ops-manager'}),
+
+        # Given multiple includes and a different exclude, it should return all images except the exclusions
+        (["cli", "ops-manager", "appdb-daily", "init-appdb"], ['cli', 'appdb-daily'], ['init-appdb'], {'appdb-daily', 'cli', 'ops-manager'}),
```
SimonBaeumer: These test cases look like bugs. All images are included, except the exclusions. Effectively `--include` loses its uses case.
In our code base we never use `--exclude`. Only `--include` is used.
Also we never run `pipeline.py` without `--include`.

MaciejKaras: I think `pipeline.py` argument syntax is overblown - we only use `--include` and pass single image. I think we should just remove all unnecessary features and stick with what is actually used.

SimonBaeumer: Agree, I've send PRs out to remove the `--include` already and renaming it to `--image`.

anandsyncs (approved):

PR #143, opened 2025-05-21
CLOUDP-317713: Separate authentication mechanism checks

anandsyncs:
## Summary of Changes

- Introduced a new `IsDeploymentAuthenticationEnabled` method to the `Mechanism` interface.
- Implemented `IsDeploymentAuthenticationEnabled` for all authentication mechanism types to check only if a mechanism is enabled (present in the config), without comparing configuration objects.
- Updated the removal/disable logic to use this new method, ensuring mechanisms like LDAP and OIDC can be properly disabled even when configuration objects differ or are missing.
- Retained the original `IsDeploymentAuthenticationConfigured` method for cases where strict configuration matching is required.

Previously, the `IsDeploymentAuthenticationConfigured` method was used for both modifying and removing authentication mechanisms. However, its implementation checked both presence and exact configuration, which caused issues when removing mechanisms like LDAP or OIDC—these have separate configuration objects that may not exist locally after removal, causing the method to return `false` and preventing proper disabling.

By separating this logic:
- `IsDeploymentAuthenticationEnabled` is used for removal/disabling and checks only for presence.
- `IsDeploymentAuthenticationConfigured` remains for full config matching.

## Proof of Work

Tests pass, added an Ldap removal test to verify that mechanism removal works.

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/controllers/operator/authentication/ldap.go
+++ b/controllers/operator/authentication/ldap.go
@@ -116,7 +116,7 @@ func (l *ldapAuthMechanism) IsAgentAuthenticationConfigured(ac *om.AutomationCon
 }
 
 func (l *ldapAuthMechanism) IsDeploymentAuthenticationConfigured(ac *om.AutomationConfig, opts Options) bool {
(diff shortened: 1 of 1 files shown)


lucian-tosa (approved):
LGTM if 🟢

nammn (approved):

PR #157, opened 2025-05-23
Batch dependabot PRs into single PR + bump python kubernetes dependency

MaciejKaras:
# Summary

Based on [GitHub Dependabot documentation](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--) we can define groups under each `package-ecosystem`. This allows to bundle multiple version updated to single PR. This will reduce the noise generated by dependabot.

Additionally bumped python `kubernetes` version to match what we use in `Go` and that is `v1.30.x`. Same as for kube client dependencies in Go added ignore rule for python `kubernetes` dependency so we have to manually update it.

## Proof of Work

There is no way to check `dependabot.yml` configuration other than simple validation [check](https://github.com/mongodb/mongodb-kubernetes/pull/157/checks?check_run_id=42781220754).

## Checklist

  * **q**: Clarifying question.


m1kola (approved):
I'm ok to try this, but want to share some previous experience.

Last time I tried grouping dependencies in a similar way it resulted in more manual work. When dependabot bumps multiple dependencies at the same time and one of them breaks something it is harder to troubleshoot which one caused the issue. I essentially end up doing bumps one-by-one manually.

It did however work very well with grouping k8s dependencies [like this](https://github.com/operator-framework/operator-controller/blob/71108b2e10c8fb60c0242a2d1e0f48a96f6689ae/.github/dependabot.yml#L15-L19), but it is not very relevant to us since we do not manage these with dependabot.

MaciejKaras:
@m1kola thanks Mikalai for sharing your experience. I agree that it can make fixing PRs harder, let's check if that will be an issue in our codebase. It is fairly easy to revert.

requirements.txt
```diff
@@ -8,7 +8,7 @@ MarkupSafe==2.0.1
 semver==2.13.0
 chardet==3.0.4
 jsonpatch==1.33
-kubernetes==17.17.0
+kubernetes==30.1.0
```
lsierant, line 11: It's awesome you've bumped this one. I also noticed it but with some e2e failures due to api changes I abandoned it  to have it as part of other work.

lsierant (approved):
LGTM!

PR #184, opened 2025-06-06
Split mongodb_types into smaller files

Julien-Ben:
# Summary

With the `mongodb_types.go` having over 1700 lines of code, I propose we split it into smaller files.
I identified 3 "groups": security, agent config and backup.

## Proof of Work

No functional changes, e2e must pass.


Julien-Ben:
Won't do, outdated

PR #193, opened 2025-06-16
CLOUDP-295785 - Calculate next version and release notes script

MaciejKaras:
# Summary

Initial work for calculating next operator version and release notes. This is based on https://docs.google.com/document/d/1eJ8iKsI0libbpcJakGjxcPfbrTn8lmcZDbQH1UqMR_g/edit?tab=t.aov8hxoyxtos#heading=h.6q8mwv95pdu6.

**This is not yet integrated with CI nor forces engineers to create change log files. There will be next smaller PRs that will address missing integration. Functionalities added by this PR:**

1. `scripts.release.calculate_next_version.py` -> this will calculate next version based on the changelog files provided.
```sh
➜  mongodb-kubernetes ✗ python3 -m scripts.release.calculate_next_version --help
usage: calculate_next_version.py [-h] [-p ] [-c ] [-s ] [-v ]

Calculate the next version based on the changes since the previous version tag.

options:
  -h, --help            show this help message and exit
  -p, --path            Path to the Git repository. Default is the current directory '.'
  -c, --changelog-path
                        Path to the changelog directory relative to the repository root. Default is 'changelog/'
  -s, --initial-commit-sha
                        SHA of the initial commit to start from if no previous version tag is found.
  -v, --initial-version
                        Version to use if no previous version tag is found. Default is '1.0.0'
```

2. `scripts.release.create_changelog.py` -> this is a tool for creating changelog files. It makes life easier for engineers and reduces a chance for errors in filename and contents syntax
```sh
➜  mongodb-kubernetes ✗python3 -m scripts.release.create_changelog --help
usage: create_changelog.py [-h] [-c ] [-d ] [-e] -k  title

Utility to easily create a new changelog entry file.

positional arguments:
  title                 Title for the changelog entry

options:
  -h, --help            show this help message and exit
  -c, --changelog-path
                        Path to the changelog directory relative to a current working directory. Default is 'changelog/'
  -d, --date            Date in 'YYYY-MM-DD' format to use for the changelog entry. Default is today's date
  -e, --editor          Open the created changelog entry in the default editor (if set, otherwise uses 'vi'). Default is True
  -k, --kind            Kind of the changelog entry:
                          - 'prelude' for prelude entries
                          - 'breaking' for breaking change entries
                          - 'feature' for feature entries
                          - 'fix' for bugfix entries
                          - 'other' for other entries
```

3. `scripts.release.release_notes.py` -> create release notes based on the changes since last release tag
```sh
➜  mongodb-kubernetes ✗ python3 -m scripts.release.release_notes --help
usage: release_notes.py [-h] [-p ] [-c ] [-s ] [-v ] [--output ]

Generate release notes based on the changes since the previous version tag.

options:
  -h, --help            show this help message and exit
  -p, --path            Path to the Git repository. Default is the current directory '.'
  -c, --changelog-path
                        Path to the changelog directory relative to the repository root. Default is 'changelog/'
  -s, --initial-commit-sha
                        SHA of the initial commit to start from if no previous version tag is found.
  -v, --initial-version
                        Version to use if no previous version tag is found. Default is '1.0.0'
  --output, -o          Path to save the release notes. If not provided, prints to stdout.
```

## Proof of Work

Passing unit tests for now is enough.

## Checklist

  * **q**: Clarifying question.


.gitignore
```diff
@@ -94,3 +94,4 @@ logs-debug/
 
 docs/**/log/*
 docs/**/test.sh.run.log
+/.venv/
```
m1kola: Nit: we already have `**/venv*/`. I normally just create `venv` in the root dir.

scripts/release/release_notes_tpl.md
```diff
@@ -0,0 +1,34 @@
...
+{%- endfor -%}
+{%- endif -%}
+{% if fixes %}
+## Bug Fixes
+
+{% for fix in fixes -%}
+{{- fix -}}
+{%- endfor -%}
+{%- endif -%}
+{% if other %}
```
m1kola:
```suggestion
{% if others %}
```

Otherwise this section will not be rendered.

m1kola (commented):
I haven't reviewed everything, but I tried the workflow with creating release notes and rendering it. Works really well. The only thing I noticed is that "Other Changes" section doesn't render (see a comment below with the reason).

scripts/release/changelog.py
```diff
@@ -0,0 +1,160 @@
...
+    change_entry = extract_changelog_entry_from_contents(file_content)
+
+    if change_entry.date != date:
+        raise Exception(
+            f"{file_name} - date in front matter {change_entry.date} does not match date extracted from file name {date}"
+        )
+
+    if change_entry.kind != kind:
+        raise Exception(
+            f"{file_name} - kind in front matter {change_entry.kind} does not match kind extracted from file name {kind}"
```
m1kola: This was mildly annoying when I tried. I first created a file with `fix` in the kind, but later I realised that `other` is more appropriate for my case (my changes are more like a change notice rather than a bug fix). I updated front matter metadata in the file, but not the filename itself and it failed. So basically I had to change in two places.

Not a big deal, IMO. I see benefits of having kind in the name.

One suggestion though - quote values. Because it reads a bit unnatural:

> kind in front matter fix does not match kind extracted from file name other

MaciejKaras: Having kind of change in filename makes it easier when skimming through the files. The downside is the one you've mentioned, but we have validation for this and mistakes shouldn't go through.

Good point with the quotes!

MaciejKaras: Done

.gitignore
MaciejKaras: But the one `**/venv*/` does not work and IDE marks it as committable for some reason

scripts/release/changelog.py
m1kola: > but we have validation for this and mistakes shouldn't go through.

Yes, I think this is good that we error explicitly on this. Resolving this thread.

.gitignore
m1kola: @MaciejKaras  I don't use UI for git, but I just checked in VS Code and it works fine. Perhaps worth submitting a bug report to JetBrains?

Resolving this thread. Not a big deal.

MaciejKaras: Ok, I'll remove it from the code, thx for checking

Julien-Ben (approved):
Awesome changes, and tests 👏

scripts/release/testdata/changelog/20250609_chore_not_oidc.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Some chore type of change
+kind: chore
+date: 2025-06-09
+---
+
+* Fixing CI/CD pipeline issues.
```
m1kola: I know it is test data, but I don't think we need change logs for this. In other words - `chore` kind seems unnecessary. If we do some clean up and it is user facing but not breaking we can use `other`.

scripts/release/create_changelog.py
```diff
@@ -0,0 +1,88 @@
...
+        formatter_class=argparse.RawTextHelpFormatter,
+    )
+    parser.add_argument(
+        "-c",
+        "--changelog-path",
+        default=DEFAULT_CHANGELOG_PATH,
+        metavar="",
+        action="store",
+        type=str,
+        help=f"Path to the changelog directory relative to the repository root. Default is {DEFAULT_CHANGELOG_PATH}",
```
m1kola: Nit: this is not factually correct. Repository root is not involved here as far as see.

Also formatting is not consistent with `scripts/release/release_notes.py`.

```suggestion
        help=f"Path to the changelog directory relative to a current working directory. Default is '{DEFAULT_CHANGELOG_PATH}'",
```

scripts/release/changelog.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # Compare base commit with current working tree
+    diff_index = base_commit.diff(other=repo.head.commit, paths=changelog_sub_path)
+
+    # No changes since the previous version
+    if not diff_index:
+        return changelog
+
+    # Traverse added Diff objects only (change type 'A' for added files)
+    for diff_item in diff_index.iter_change_type("A"):
```
m1kola: Could you please add a more detailed comment here or a docstring to the function explaining how this works? E.g. why we only getting added files? How do we account for modified changelog entries (e.g. added in one commit, modified in another)? Deleted?

This will be extremity helpful for the future us who will be maintaining the tool/making changes here.

scripts/release/changelog.py
```diff
@@ -0,0 +1,160 @@
...
+    if kind_str.lower() in PRELUDE_ENTRIES:
+        return ChangeKind.PRELUDE
+    if kind_str.lower() in BREAKING_CHANGE_ENTRIES:
+        return ChangeKind.BREAKING
+    elif kind_str.lower() in FEATURE_ENTRIES:
+        return ChangeKind.FEATURE
+    elif kind_str.lower() in BUGFIX_ENTRIES:
+        return ChangeKind.FIX
+    else:
+        return ChangeKind.OTHER
```
m1kola: Nit:

```suggestion
    return ChangeKind.OTHER
```

scripts/release/create_changelog.py
```diff
@@ -0,0 +1,88 @@
...
+        action="store",
+        metavar="",
+        required=True,
+        type=str,
+        help=f"""Kind of the changelog entry:
+  - '{", ".join(PRELUDE_ENTRIES)}' for prelude entries
+  - '{", ".join(BREAKING_CHANGE_ENTRIES)}' for breaking change entries
+  - '{", ".join(FEATURE_ENTRIES)}' for feature entries
+  - '{", ".join(BUGFIX_ENTRIES)}' for bugfix entries
+  - everything else will be treated as other entries""",
```
m1kola: I'm looking at the tests: e.g. `20250520_docs_update_readme` and `20250610_refactor_codebase`.

And I think we should make `other` to be explicit and reject anything else.

scripts/release/release_notes.py
```diff
@@ -0,0 +1,132 @@
...
+        help="Path to the Git repository. Default is the current directory '.'",
+    )
+    parser.add_argument(
+        "-c",
+        "--changelog-path",
+        default=DEFAULT_CHANGELOG_PATH,
+        metavar="",
+        action="store",
+        type=str,
+        help=f"Path to the changelog directory relative to the repository root. Default is '{DEFAULT_CHANGELOG_PATH}'",
```
m1kola: Nit: this is not factually correct. Repository root is not involved here as far as see.

```suggestion
        help=f"Path to the changelog directory relative to a current working directory. Default is '{DEFAULT_CHANGELOG_PATH}'",
```

scripts/release/release_notes.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Parameters:
+        repository_path: Path to the Git repository. Default is the current directory '.'.
+        changelog_sub_path: Path to the changelog directory relative to the repository root. Default is '{DEFAULT_CHANGELOG_PATH}.
+        initial_commit_sha: SHA of the initial commit to start from if no previous version tag is found.
+        initial_version: Version to use if no previous version tag is found. Default is '{DEFAULT_INITIAL_GIT_TAG_VERSION}'.
+
+    Returns:
+        Formatted release notes as a string.
+    """
```
m1kola: Let's get rid of default values and always assume we receive values here. We already have defaulting in the arguments: let's not maintain two places where we default.

scripts/release/version.py
```diff
@@ -0,0 +1,50 @@
+import semver
+from git import Commit, Repo, TagReference
+
+from scripts.release.changelog import ChangeKind
+
+
+def find_previous_version(repo: Repo, initial_commit_sha: str = None) -> (TagReference | None, Commit):
+    """Find the most recent version that is an ancestor of the current HEAD commit."""
```
m1kola, line 8: q: I believe you said that the tool supports multiple projects in the same repo. How is it going to work? Do they all need to share the versioning?

Or have I misunderstood something?

m1kola (commented):
I've got a few questions, nits and suggestions. But overall I think it looks great.

scripts/release/testdata/changelog/20250609_chore_not_oidc.md
MaciejKaras: yes, this is bad example, I agree :D 

There is no chore kind fortunately. Every unknown kind is just mapped to the `other` kind. This test case just proves that some arbitrary and unspecified kind is properly mapped as well.

scripts/release/create_changelog.py
MaciejKaras: Good catch!

scripts/release/version.py
MaciejKaras: I did not focus in this PR to make it work for multiple projects out of the box. We still don't know how this should be setup and implementing such use case would be just my understanding, which could be incorrect.

Choosing your own tool rather then using external one gives us possibility to just easily implement or change bits of the code, when the multiple projects in the repo design will be ready.

scripts/release/release_notes.py
MaciejKaras: For this particular script `release_notes.py` it makes sense. The `calculate_next_version.py` script will be used by other python scripts to retrieve version, so I will keep the defaults there.

scripts/release/create_changelog.py
MaciejKaras: I deliberately wanted to have some freedom in naming `other` kind of changes. The name `other` does not tell much, so if somebody wants to use `docs` or `config` they can and it will just land under `Other changes`.

Do you feel there is something wrong with this idea?

scripts/release/version.py
m1kola: Got it! Now I remember that you were saying that with our own tool we will have more flexibility to extend when the time comes and not that it will support it. Apologies. Resolving this.

scripts/release/create_changelog.py
m1kola: With having random strings treated as `other` it is hard to tell if it is a real kind or not. Unless you remember implementation, it is hard to say in which section it will be rendered. E.g. will `20250520_docs_update_readme` be rendered in features or in other? Or perhaps it is a prelude?

I think kind in file names and kind in front matter must always be consistent and explicit. This means that:
1. Let's not allow random strings in file names and in `kind` in front matter.
2. If one kind has multiple aliases (e.g. `major` and `breaking`) - let's choose one thing (e.g. `major`) as a canonical name and use it as a kind in both filenames and front matter.

So consistency in the data layer (files in our case), but flexibility on the client (CLI tool).

scripts/release/testdata/changelog/20250609_chore_not_oidc.md
m1kola: Resolving this thread in favour of https://github.com/mongodb/mongodb-kubernetes/pull/193#discussion_r2230547020

scripts/release/release_notes.py
m1kola: I'm not sure that I follow. Perhaps you did not explain it very clearly. I'll try again.

I'm suggesting:
1. Keep default values in `ArgumentParser` (e.g. lines 80-121 right now where we call ` parser.add_argument`).
2. Remove default values from `generate_release_notes` function because it duplicates defaulting from `ArgumentParser` and we always call it so we pass values for all the arguments. So defaults here are not really required.

I don't see a similar issue with `calculate_next_version.py`.

MaciejKaras: My bad, I had to change this some time ago.

Your point is valid and I will change this. I just wanted to mention that in case we are going to use some script not with CLI but from another python file, default values might be convienient.

scripts/release/create_changelog.py
MaciejKaras: I agree that the kind in filename and in the file contents itself has to be identical. We are right now only checking if they map to the same ChangeKind value. I can easily change this and compare actual string representations.

>Unless you remember implementation, it is hard to say in which section it will be rendered. E.g. will 20250520_docs_update_readme be rendered in features or in other? Or perhaps it is a prelude?
I don't find this true. The information about the mapping is presented in the CLI `help` of the method that creates a changelog file. I agree if we have less values to choose from it will be easier to memorize them and also see where the change will land in Release Notes just by scanning the filenames.

So you're proposing naming the same as the sections in the Release Notes template:
- breaking
- feature
- fix
- other
?

m1kola: > I don't find this true. The information about the mapping is presented in the CLI `help` of the method that creates a changelog file.

Ok, I can say "unless you remember the implementation or look it up in the `--help` command". And still you need to read a multiline help string and catch "everything else will be treated as other entries".

> So you're proposing naming the same as the sections in the Release Notes template:

I don't think it necessary has to be exact match with section names, but I would really want to have 1:1 mapping between sections and `kind`s in files. And I want the number of `kind` options to be deliberately very limited. Like `prelude`, `major`, `feature`, `fix`, `other` and that's it.

If we can make it dead simple we won't even need CLI to create change notes or read about how it works – just copy-paste existing one and modify it. Only read help if it fails to render (e.g. in CI). With limited number of options we can make it like so.

scripts/release/release_notes.py
```diff
@@ -0,0 +1,132 @@
...
+    initial_commit_sha: str | None,
+    initial_version: str,
+) -> str:
+    f"""Generate a release notes based on the changes since the previous version tag.
+
+    Parameters:
+        repository_path: Path to the Git repository. Default is the current directory '.'.
+        changelog_sub_path: Path to the changelog directory relative to the repository root. Default is '{DEFAULT_CHANGELOG_PATH}.
+        initial_commit_sha: SHA of the initial commit to start from if no previous version tag is found.
+        initial_version: Version to use if no previous version tag is found. Default is '{DEFAULT_INITIAL_GIT_TAG_VERSION}'.
```
m1kola: I see you updated args in the function and removed defaults, but defaults are still mentioned here.

scripts/release/create_changelog.py
MaciejKaras: I think you're right. Let's start simple and if there is requirement for more options we can add it later.

scripts/release/changelog.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # No changes since the previous version
+    if not diff_index:
+        return changelog
+
+    # Traverse added Diff objects only (change type 'A' for added files)
+    # Because we are traversing back to the most recent version, we only care about files that were added since then
+    # If a file was added in one commit and then modified in another, we only care about the final version of the file
+    # Same for deletions - if file was added in one commit but then deleted in another, we don't want to include it
+    # in the Release Notes
```
m1kola, line 65: q: don't we need to also look for deletions and remove them from `changelog` list? If not - how are the deletions tracked?

scripts/release/changelog.py
```diff
@@ -0,0 +1,164 @@
...
+        if kind_str_lower == str(ChangeKind.PRELUDE):
+            return ChangeKind.PRELUDE
+        if kind_str_lower in str(ChangeKind.BREAKING):
+            return ChangeKind.BREAKING
+        elif kind_str_lower in str(ChangeKind.FEATURE):
+            return ChangeKind.FEATURE
+        elif kind_str_lower in str(ChangeKind.FIX):
+            return ChangeKind.FIX
+        elif kind_str_lower in str(ChangeKind.OTHER):
+            return ChangeKind.OTHER
```
m1kola: Some of the conditions are `==` and some are `in`. Seems like a leftover from previous iterations?

This is an issue because when `kind_str` is `"e"`, `kind_str_lower in str(ChangeKind.BREAKING)` will evaluate to `True`.



```suggestion
        if kind_str_lower == str(ChangeKind.PRELUDE):
            return ChangeKind.PRELUDE
        if kind_str_lower == str(ChangeKind.BREAKING):
            return ChangeKind.BREAKING
        elif kind_str_lower == str(ChangeKind.FEATURE):
            return ChangeKind.FEATURE
        elif kind_str_lower == str(ChangeKind.FIX):
            return ChangeKind.FIX
        elif kind_str_lower == str(ChangeKind.OTHER):
            return ChangeKind.OTHER
```

scripts/release/changelog_test.py
```diff
@@ -0,0 +1,138 @@
...
+        self.assertEqual(date, datetime.date(2025, 1, 1))
+        self.assertEqual(kind, ChangeKind.BREAKING)
+
+        date, kind = extract_date_and_kind_from_file_name("20250508_breaking_remove_deprecated.md")
+        self.assertEqual(date, datetime.date(2025, 5, 8))
+        self.assertEqual(kind, ChangeKind.BREAKING)
+
+        date, kind = extract_date_and_kind_from_file_name("20250509_breaking_schema_change.md")
+        self.assertEqual(date, datetime.date(2025, 5, 9))
+        self.assertEqual(kind, ChangeKind.BREAKING)
```
m1kola: Q: Not sure that I'm following - why we need multiple test cases here and for other kinds?

scripts/release/create_changelog.py
```diff
@@ -0,0 +1,84 @@
...
+        metavar="",
+        required=True,
+        type=str,
+        help=f"""Kind of the changelog entry:
+  - '{str(ChangeKind.PRELUDE)}' for prelude entries
+  - '{str(ChangeKind.BREAKING)}' for breaking change entries
+  - '{str(ChangeKind.FEATURE)}' for feature entries
+  - '{str(ChangeKind.FIX)}' for bugfix entries
+  - '{str(ChangeKind.OTHER)}' for other entries""",
+    )
```
m1kola, line 55: I'm ok with it as it currently is. Just want to to clarify that in [this thread](https://github.com/mongodb/mongodb-kubernetes/pull/193#discussion_r2231266401) I meant that it is ok to have aliases like `major` for breaking changes on the CLI. I just wanted the to be created with a canonical kind in filename and in front matter.

Like if there we decide that `breaking` is a canonical kind for `breaking` and `major` - even if the user says `--kind major` we still write a file as `breaking` kind.

But as [you said](https://github.com/mongodb/mongodb-kubernetes/pull/193#discussion_r2235873771): we can keep it simple and add CLI level aliases later if we see the need.

m1kola (commented):
Left a few questions and one clarification for my previous comments.

I think overall it looks great - the only things which might need a bit of attention is `ChangeKind.from_str`.

scripts/release/create_changelog.py
MaciejKaras: If in the end the filename and contents will have `breaking` change and not `major` alias that was passed to CLI, I don't see much value in it. It will be only used through CLI. Let's keep it simple for now and less mapping code ;)

scripts/release/changelog_test.py
MaciejKaras: No we don't. Those are redundant, I can clean them

scripts/release/changelog.py
MaciejKaras: Nice catch!

scripts/release/changelog.py
MaciejKaras: I don't follow. You create a diff between commit range. You don't care about the deleted changelog files, because they were before the first commit from the range. You only care about the additions. Example:

```mermaid
%%{
  init: {
    'logLevel': 'debug',
    'gitGraph': {
      'showBranches': true,
      'mainBranchName': 'master',
      'parallelCommits': 'true'
    }
  }
}%%
gitGraph
    commit id: "20230505_fix_some_fix.md"
    commit id: "20230506_feature_some_feature.md" tag: "1.1.0"
    commit id: "20230507_feature_some_feature_2.md"
    commit id: "delete: 20230505_fix_some_fix.md"
    commit id: "20230508_fix_some_fix_2.md" tag: "1.2.0"
```

Commit `delete: 20230505_fix_some_fix.md` shouldn't affect anyhow the Release Notes for "1.2.0" because it was already released in "1.1.0"

m1kola: Yes, in this case it doesn't matter. But what about this?

```mermaid
%%{
  init: {
    'logLevel': 'debug',
    'gitGraph': {
      'showBranches': true,
      'mainBranchName': 'master',
      'parallelCommits': 'true'
    }
  }
}%%
gitGraph
    commit id: "20230506_feature_some_feature.md" tag: "1.1.0"
    commit id: "20230505_fix_some_fix.md"
    commit id: "20230507_feature_some_feature_2.md"
    commit id: "delete: 20230505_fix_some_fix.md"
    commit id: "20230508_fix_some_fix_2.md" tag: "1.2.0"
```

`20230505_fix_some_fix.md` is added after 1.1.0 and deleted before the next release (1.2.0).

scripts/release/create_changelog.py
m1kola: Sounds good! Resolving.

scripts/release/changelog.py
MaciejKaras: Then the diff between `1.1.0...1.2.0` won't contain `20230505_fix_some_fix.md` file as well as Release Notes for `1.2.0` version. 

I you would modify the `20230505_fix_some_fix.md` rather then delete it, then the modifications will be part of Release Notes for `1.2.0` version

m1kola: Aha! That makes sense now. Resolving this.

m1kola (approved):
Looks great! :shipit:

PR #173, opened 2025-06-03
Allow for more than one deployment per test under cloud-qa

lsierant:
# Summary

TBD

## Checklist
- [ ] Have you linked a jira ticket and/or is the ticket in the title?
- [ ] Have you checked whether your jira ticket required DOCSP changes?
- [ ] Have you checked for release_note changes?

## Reminder (Please remove this when merging)
- Please try to Approve or Reject Changes the PR, keep PRs in review as short as possible
- Our Short Guide for PRs: [Link](https://docs.google.com/document/d/1T93KUtdvONq43vfTfUt8l92uo4e4SEEvFbIEKOxGr44/edit?tab=t.0)
- Remember the following Communication Standards - use comment prefixes for clarity:
  * **blocking**: Must be addressed before approval.
  * **follow-up**: Can be addressed in a later PR or ticket.
  * **q**: Clarifying question.
  * **nit**: Non-blocking suggestions.
  * **note**: Side-note, non-actionable. Example: Praise 
  * --> no prefix is considered a question

PR #206, opened 2025-06-20
CLOUDP-324440 Save rs member ids

lucian-tosa:
# Summary

## Checklist

  * **q**: Clarifying question.

### DIFF
--- a/controllers/om/mockedomclient.go
+++ b/controllers/om/mockedomclient.go
@@ -734,7 +734,7 @@ func (oc *MockedOmConnection) CheckResourcesAndBackupDeleted(t *testing.T, resou
 	// This can be improved for some more complicated scenarios when we have different resources in parallel - so far
 	// just checking if deployment
 	assert.Empty(t, oc.deployment.getProcesses())
-	assert.Empty(t, oc.deployment.getReplicaSets())
+	assert.Empty(t, oc.deployment.GetReplicaSets())
 	assert.Empty(t, oc.deployment.getShardedClusters())
 	assert.Empty(t, oc.deployment.getMonitoringVersions())
 	assert.Empty(t, oc.deployment.getBackupVersions())
--- a/controllers/om/mockedomclient.go
+++ b/controllers/om/mockedomclient.go
@@ -141,6 +141,22 @@ func (oc *MockedOmConnection) ConfigureProject(project *Project) {
 	oc.context.OrgID = project.OrgID
 }
 
+func (oc *MockedOmConnection) GetReplicaSetMemberIds() (map[string]map[string]int, error) {
+	oc.addToHistory(reflect.ValueOf(oc.GetReplicaSetMemberIds))
+	dep, err := oc.ReadDeployment()
+	if err != nil {
+		return nil, err
+	}
+
+	finalProcessIds := make(map[string]map[string]int)
+
+	for _, replicaSet := range dep.GetReplicaSets() {
+		finalProcessIds[replicaSet.Name()] = replicaSet.MemberIds()
+	}
+
+	return finalProcessIds, nil
+}
+
 var _ Connection = &MockedOmConnection{}
 
 // NewEmptyMockedOmConnection is the standard function for creating mocked connections that is usually used for testing
--- a/controllers/om/omclient.go
+++ b/controllers/om/omclient.go
@@ -64,6 +64,10 @@ type Connection interface {
 	GetPreferredHostnames(agentApiKey string) ([]PreferredHostname, error)
 	AddPreferredHostname(agentApiKey string, value string, isRegexp bool) error
 
+	// GetReplicaSetMemberIds returns a map with the replicaset name as the key.
+	// The value is another map where the key is the replicaset member name and the value is its member id.
+	GetReplicaSetMemberIds() (map[string]map[string]int, error)
+
 	backup.GroupConfigReader
 	backup.GroupConfigUpdater
 
@@ -271,6 +275,21 @@ func (oc *HTTPOmConnection) GetAgentAuthMode() (string, error) {
 		return "", nil
 	}
 	return ac.Auth.AutoAuthMechanism, nil
+}
+
+func (oc *HTTPOmConnection) GetReplicaSetMemberIds() (map[string]map[string]int, error) {
+	dep, err := oc.ReadDeployment()
+	if err != nil {
+		return nil, err
+	}
+
+	finalProcessIds := make(map[string]map[string]int)
+
+	for _, replicaSet := range dep.GetReplicaSets() {
+		finalProcessIds[replicaSet.Name()] = replicaSet.MemberIds()
+	}
+
+	return finalProcessIds, nil
 }
 
 var _ Connection = &HTTPOmConnection{}
--- a/controllers/operator/mongodbmultireplicaset_controller.go
+++ b/controllers/operator/mongodbmultireplicaset_controller.go
@@ -198,8 +198,14 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 		return r.updateStatus(ctx, &mrs, status, log)
 	}
 
+	// Save replicasets member ids in annotation
+	finalMemberIds, err := conn.GetReplicaSetMemberIds()
+	if err != nil {
+		return r.updateStatus(ctx, &mrs, workflow.Failed(err), log)
+	}
+
 	mrs.Status.FeatureCompatibilityVersion = mrs.CalculateFeatureCompatibilityVersion()
-	if err := r.saveLastAchievedSpec(ctx, mrs); err != nil {
+	if err := r.saveLastAchievedSpec(ctx, mrs, finalMemberIds); err != nil {
 		return r.updateStatus(ctx, &mrs, workflow.Failed(xerrors.Errorf("Failed to set annotation: %w", err)), log)
 	}
 
@@ -624,7 +630,7 @@ func getMembersForClusterSpecItemThisReconciliation(mrs *mdbmultiv1.MongoDBMulti
 }
 
 // saveLastAchievedSpec updates the MongoDBMultiCluster resource with the spec that was just achieved.
-func (r *ReconcileMongoDbMultiReplicaSet) saveLastAchievedSpec(ctx context.Context, mrs mdbmultiv1.MongoDBMultiCluster) error {
+func (r *ReconcileMongoDbMultiReplicaSet) saveLastAchievedSpec(ctx context.Context, mrs mdbmultiv1.MongoDBMultiCluster, rsMemberIds map[string]map[string]int) error {
 	clusterSpecs, err := mrs.GetClusterSpecItems()
 	if err != nil {
 		return err
@@ -652,6 +658,14 @@ func (r *ReconcileMongoDbMultiReplicaSet) saveLastAchievedSpec(ctx context.Conte
 	annotationsToAdd[util.LastAchievedSpec] = string(achievedSpecBytes)
 	if string(clusterNumBytes) != "null" {
 		annotationsToAdd[mdbmultiv1.LastClusterNumMapping] = string(clusterNumBytes)
+	}
+
+	rsMemberIdsBytes, err := json.Marshal(rsMemberIds)
+	if err != nil {
+		return err
+	}
+	if string(rsMemberIdsBytes) != "null" {
+		annotationsToAdd[util.LastAchievedRsMemberIds] = string(rsMemberIdsBytes)
 	}
 
 	return annotations.SetAnnotations(ctx, &mrs, annotationsToAdd, r.client)
@@ -696,6 +710,10 @@ func (r *ReconcileMongoDbMultiReplicaSet) updateOmDeploymentRs(ctx context.Conte
 	}
 
 	processIds := getReplicaSetProcessIdsFromReplicaSets(mrs.Name, existingDeployment)
+	// If there is no replicaset configuration saved in OM, it might be a new project, so we check the ids saved in annotation
+	if len(processIds) == 0 {
+		processIds = getReplicaSetProcessIdsFromAnnotation(mrs)
+	}
 	log.Debugf("Existing process Ids: %+v", processIds)
 
 	certificateFileName := ""
@@ -789,6 +807,16 @@ func getReplicaSetProcessIdsFromReplicaSets(replicaSetName string, deployment om
 	}
 
 	return processIds
+}
+
+func getReplicaSetProcessIdsFromAnnotation(mrs mdbmultiv1.MongoDBMultiCluster) map[string]int {
+	processIds := make(map[string]map[string]int)
+	if processIdsStr, ok := mrs.Annotations[util.LastAchievedRsMemberIds]; ok {
+		if err := json.Unmarshal([]byte(processIdsStr), &processIds); err != nil {
+			return map[string]int{}
+		}
+	}
+	return processIds[mrs.Name]
 }
 
 func getSRVService(mrs *mdbmultiv1.MongoDBMultiCluster) corev1.Service {
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -280,7 +280,10 @@ const (
 	// TODO: remove this from here and move it to the certs package
 	// This currently creates an import cycle
 	InternalCertAnnotationKey = "internalCertHash"
-	LastAchievedSpec          = "mongodb.com/v1.lastSuccessfulConfiguration"
+
+	// Annotation keys used by the operator
+	LastAchievedSpec        = "mongodb.com/v1.lastSuccessfulConfiguration"
+	LastAchievedRsMemberIds = "mongodb.com/v1.lastAchievedRsMemberIds"
 
 	// SecretVolumeName is the name of the volume resource.
 	SecretVolumeName = "secret-certs"
--- a/controllers/om/automation_config_test.go
+++ b/controllers/om/automation_config_test.go
@@ -1059,7 +1059,7 @@ func TestApplyInto(t *testing.T) {
 }
 
 func changeTypes(deployment Deployment) error {
-	rs := deployment.getReplicaSets()
+	rs := deployment.GetReplicaSets()
 	deployment.setReplicaSets(rs)
 	return nil
 }
--- a/controllers/om/deployment.go
+++ b/controllers/om/deployment.go
@@ -280,13 +280,13 @@ func (d Deployment) AddMonitoringAndBackup(log *zap.SugaredLogger, tls bool, caF
 	d.addBackup(log)
 }
 
-// DEPRECATED: this shouldn't be used as it may panic because of different underlying type; use getReplicaSets instead
+// DEPRECATED: this shouldn't be used as it may panic because of different underlying type; use GetReplicaSets instead
 func (d Deployment) ReplicaSets() []ReplicaSet {
 	return d["replicaSets"].([]ReplicaSet)
 }
 
 func (d Deployment) GetReplicaSetByName(name string) ReplicaSet {
-	for _, rs := range d.getReplicaSets() {
+	for _, rs := range d.GetReplicaSets() {
 		if rs.Name() == name {
 			return rs
 		}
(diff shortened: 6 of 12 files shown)


RELEASE_NOTES.md
```diff
@@ -25,6 +25,9 @@
     * [Manage Database Users using OIDC](https://www.mongodb.com/docs/kubernetes/upcoming/manage-users/) # TODO
     * [Authentication and Authorization with OIDC/OAuth 2.0](https://www.mongodb.com/docs/manual/core/oidc/security-oidc/)
 
+## Bug Fixes
```
vinilage, line 28: Can you please highlight better the introduction of OIDC? Checking the release notes, I see the new features but our main goal for this release is the OIDC, so I would suggest to add a section for OIDC. I will create a suggestion in GDocs and share with you.

vinilage: Just did a proposal:
https://docs.google.com/document/d/143v_VS4graZZgGp2r63WzLcPH1tcv3OvShHtM5HPiSY/edit?usp=sharing

lucian-tosa: I would like to keep this PR focused on the bug fix. I think we can do this refactoring when we release. Wdyt?

controllers/om/mockedomclient.go
```diff
@@ -141,6 +141,22 @@ func (oc *MockedOmConnection) ConfigureProject(project *Project) {
 	oc.context.OrgID = project.OrgID
 }
 
+func (oc *MockedOmConnection) GetReplicaSetMemberIds() (map[string]map[string]int, error) {
```
anandsyncs: **Unrelated to this PR:** Given that we are just copy pasting the the whole functionality, we can simply wrap the omclient with this mock client and only override what is necessary.

controllers/om/omclient.go
```diff
@@ -273,6 +277,21 @@ func (oc *HTTPOmConnection) GetAgentAuthMode() (string, error) {
 	return ac.Auth.AutoAuthMechanism, nil
 }
 
+func (oc *HTTPOmConnection) GetReplicaSetMemberIds() (map[string]map[string]int, error) {
```
lsierant: **blocking:** Do we need that separate method? Could we have an external function calling ReadDeployment and extracting the ids from the result? 

I'd prefer to limit the interface of OMConnection to the distinct API operations and have them combined for the logic like this in a different layer above it.

lsierant: With this we'd also avoic changing the mock's implementation

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -657,6 +663,14 @@ func (r *ReconcileMongoDbMultiReplicaSet) saveLastAchievedSpec(ctx context.Conte
 		annotationsToAdd[mdbmultiv1.LastClusterNumMapping] = string(clusterNumBytes)
 	}
 
+	rsMemberIdsBytes, err := json.Marshal(rsMemberIds)
+	if err != nil {
+		return err
+	}
+	if string(rsMemberIdsBytes) != "null" {
```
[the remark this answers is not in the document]
lsierant: Good suggestion.

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -794,6 +812,16 @@ func getReplicaSetProcessIdsFromReplicaSets(replicaSetName string, deployment om
 	return processIds
 }
 
+func getReplicaSetProcessIdsFromAnnotation(mrs mdbmultiv1.MongoDBMultiCluster) map[string]int {
+	processIds := make(map[string]map[string]int)
+	if processIdsStr, ok := mrs.Annotations[util.LastAchievedRsMemberIds]; ok {
+		if err := json.Unmarshal([]byte(processIdsStr), &processIds); err != nil {
+			return map[string]int{}
```
lsierant: Did you consider making this a fatal error? If there is some garbage in the annotation, isn't it too risky to ignore its content?

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -699,6 +713,10 @@ func (r *ReconcileMongoDbMultiReplicaSet) updateOmDeploymentRs(ctx context.Conte
 	}
 
 	processIds := getReplicaSetProcessIdsFromReplicaSets(mrs.Name, existingDeployment)
+	// If there is no replicaset configuration saved in OM, it might be a new project, so we check the ids saved in annotation
+	if len(processIds) == 0 {
+		processIds = getReplicaSetProcessIdsFromAnnotation(mrs)
```
nammn: why not use `conn.GetReplicaSetMemberIds()` here instead? Couldn't that save us the whole annotation work?

controllers/operator/mongodbmultireplicaset_controller_test.go
```diff
@@ -945,6 +945,90 @@ func TestScaling(t *testing.T) {
...
+		assertMemberNameAndId(t, members, fmt.Sprintf("%s-1-0", mrs.Name), 1)
+		assertMemberNameAndId(t, members, fmt.Sprintf("%s-2-0", mrs.Name), 2)
+
+		// Assert that the member ids are updated in the annotation
+		rsMemberIds[mrs.GetName()][fmt.Sprintf("%s-0-1", mrs.Name)] = 3
+		rsMemberIdsBytes, _ = json.Marshal(rsMemberIds)
+		assert.Equal(t, mrs.GetAnnotations()[util.LastAchievedRsMemberIds], string(rsMemberIdsBytes))
+
+		// We simulate a changing the project by recreating the omConnection. The resource will keep the annotation.
+		// This part would have failed before 1.2.0.
```
lsierant, line 1016: no need to have it written, but a good practice is to manually check if this actually is true by running this unit test against master and verifying it's indeed failing here

controllers/operator/mongodbmultireplicaset_controller.go
nammn: as discussed offline - its the migration case

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
```diff
@@ -17,6 +25,26 @@
...
+@pytest.fixture(scope="module")
+def project_name_prefix(namespace: str) -> str:
+    return random_k8s_name(f"{namespace}-project-")
+
+
+@pytest.fixture(scope="module")
+def new_project_configmap(namespace: str, project_name_prefix: str) -> str:
+    cm = read_configmap(namespace=namespace, name="my-project")
+    project_name = f"{project_name_prefix}-new-project"
+    return create_or_update_configmap(
```
lsierant, line 38: again, reconfiguring the project! We need to followup on this [PR](https://github.com/mongodb/mongodb-kubernetes/pull/173)... 
I'm leaving the comment there to adjust it in here as well.

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
```diff
@@ -97,7 +131,6 @@ def test_ops_manager_has_been_updated_correctly_before_scaling():
 
 @pytest.mark.e2e_multi_cluster_scale_up_cluster
 def test_scale_mongodb_multi(mongodb_multi: MongoDBMulti, member_cluster_clients: List[MultiClusterClient]):
-    mongodb_multi.load()
```
lsierant, line 100: to not need load() you must change mongodb_multi's fixture scope to "function" to load it every time it's injected into function

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
```diff
@@ -139,3 +159,59 @@ def test_ops_manager_has_been_updated_correctly_after_scaling():
 def test_replica_set_is_reachable(mongodb_multi: MongoDBMulti, ca_path: str):
     tester = mongodb_multi.tester()
     tester.assert_connectivity(opts=[with_tls(use_tls=True, ca_path=ca_path)])
+
+
+# From here on, the tests are for verifying that we can change the project of the MongoDBMulti resource even with
```
lsierant, line 157: nit: I find this comment a bit unnecessary, but you don't need to remove it. 
It would be better to wrap such a set of test functions into a class and name it properly, e.g. `TestNonSequentialMemberIdsInReplicaSet` or sth like that.

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
```diff
@@ -139,3 +159,59 @@ def test_ops_manager_has_been_updated_correctly_after_scaling():
...
+    mongodb_multi.assert_abandons_phase(phase=Phase.Running, timeout=300)
+    mongodb_multi.assert_reaches_phase(phase=Phase.Running, timeout=600)
+
+    newRsMembers = mongodb_multi.get_automation_config_tester().get_replica_set_members(mongodb_multi.name)
+
+    # Assert that the replica set member ids have not changed after changing the project.
+    assert oldRsMembers == newRsMembers
+
+
+def sts_are_ready(mdb: MongoDBMulti, member_cluster_clients: List[MultiClusterClient], members: List[int]):
```
lsierant: this looks to me like something we must already have somewhere in our tests. Or is it doing something different than just waiting for sts to be ready? If yes, then pls name it so we can see in the name why it's different.

pkg/util/constants.go
```diff
@@ -283,7 +283,10 @@ const (
 	// TODO: remove this from here and move it to the certs package
 	// This currently creates an import cycle
 	InternalCertAnnotationKey = "internalCertHash"
-	LastAchievedSpec          = "mongodb.com/v1.lastSuccessfulConfiguration"
+
+	// Annotation keys used by the operator
+	LastAchievedSpec        = "mongodb.com/v1.lastSuccessfulConfiguration"
+	LastAchievedRsMemberIds = "mongodb.com/v1.lastAchievedRsMemberIds"
```
lsierant, line 289: @nammn to your concern about not putting the state into annotations - I agree, we should not do that. But mdbmc is still not migrated to the deployments state config map, and doing that should not be part of this PR. IMO it's fine to handle it this way, and migrate those annotations to the state cm once we decide to do so.

lsierant (changes requested):
Generally - awesome fix! 👏 
But I've left one blocking comment regarding not adding another method to the OM's connection interface. Consider the rest optional.

controllers/operator/mongodbmultireplicaset_controller.go
anandsyncs: @lucian-tosa Can you please explain the "migration case" that requires use of annotation?

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -201,8 +201,14 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 		return r.updateStatus(ctx, &mrs, status, log)
 	}
 
+	// Save replicasets member ids in annotation
```
nammn: nit: we can remove this comment, we can find its usage via the ide

controllers/operator/mongodbmultireplicaset_controller.go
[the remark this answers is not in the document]
nammn: blocking: oh - yeah definitely please don't check for a null string. I think in other code places we have a better way of doing that

controllers/operator/mongodbmultireplicaset_controller.go
nammn: blocking: let's add a comment here on why this can happen (the migration case) for clarity

controllers/operator/mongodbmultireplicaset_controller.go
nammn: if not - we should at least log this very clearly

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
nammn: q: can you check whether those new projects are properly cleaned up in cloudmanager?

pkg/util/constants.go
nammn: thanks!

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
lsierant: added this to the PR

controllers/operator/mongodbmultireplicaset_controller_test.go
lucian-tosa: I already ran it manually and it failed. Are you suggesting to remove the entire comment, or just the second line?

controllers/operator/mongodbmultireplicaset_controller.go
lucian-tosa: The only case where a broken annotation would be fatal is if the deployment is being migrated to a new OM project, and the annotation was manually changed at the same time.
I can't think of any other reason for why the annotation can have garbage, I wouldn't make this fatal. If there is garbage in the annotation, the operator will rewrite anyway. 
Wdyt?

controllers/operator/mongodbmultireplicaset_controller_test.go
lsierant: nothing to do here, if you tested it it's fine!

controllers/operator/mongodbmultireplicaset_controller.go
lsierant: if there is garbage in annotation that means someone or something changed it and in this case we should probably error the reconciliation as the annotation is there, but we don't have correct information from it

lsierant: it's like, check spec.version and find garbage in it - should we ignore it and continue reconciling?

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
lucian-tosa: We don't already have it. But I will add it as part of the MongoDBMulti class so we can reuse it at least for that resource.

controllers/om/mockedomclient.go
lucian-tosa: I will remove this method entirely following Lukasz's suggestion

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
lucian-tosa: done

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
lucian-tosa: done

controllers/operator/mongodbmultireplicaset_controller.go
lucian-tosa: done

controllers/operator/mongodbmultireplicaset_controller.go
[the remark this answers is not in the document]
lucian-tosa: done

controllers/om/omclient.go
lucian-tosa: done

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_scale_up_cluster.py
lucian-tosa: done

controllers/operator/mongodbmultireplicaset_controller.go
lucian-tosa: makes sense, done

controllers/operator/mongodbmultireplicaset_controller.go
lucian-tosa: So the migration case is detailed in the PR description. The steps that lead to a broken deployment is:
1. scale the deployment in a way that results in non-sequential ids
2. move the deployment to a new project.

I also added some other considerations and edge cases to the PR description. 
@anandsyncs let me know if I should explain something further

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -663,12 +662,14 @@ func (r *ReconcileMongoDbMultiReplicaSet) saveLastAchievedSpec(ctx context.Conte
 		annotationsToAdd[mdbmultiv1.LastClusterNumMapping] = string(clusterNumBytes)
 	}
 
-	rsMemberIdsBytes, err := json.Marshal(rsMemberIds)
-	if err != nil {
-		return err
-	}
-	if string(rsMemberIdsBytes) != "null" {
-		annotationsToAdd[util.LastAchievedRsMemberIds] = string(rsMemberIdsBytes)
+	if len(rsMemberIds) > 0 {
```
nammn, line 665: nit: i think its redundant because - if you just marshal it and its empty you will get nil; marshalling nil return "null"

docker/mongodb-kubernetes-tests/kubetester/mongodb_multi.py
```diff
@@ -77,6 +78,21 @@ def service_names(self) -> List[str]:
...
 
+    def assert_statefulsets_are_ready(self, clients: List[MultiClusterClient], timeout: int = 600):
+        def fn():
+            statefulsets = self.read_statefulsets(clients)
+
+            assert len(statefulsets) == len(self["spec"]["clusterSpecList"])
+
+            for i, mcc in enumerate(clients):
+                cluster_sts = statefulsets[mcc.cluster_name]
+                if cluster_sts.status.ready_replicas != self.get_item_spec(mcc.cluster_name)["members"]:
```
nammn, line 89: q; for my understanding - shouldn't that suffice or is that due to how multicluster sts members are structured in the spec? https://github.com/mongodb/mongodb-kubernetes/blob/ca24d93d7a931f7853a679b4576674cace37bb16/docker/mongodb-kubernetes-tests/tests/olm/olm_meko_operator_upgrade_with_resources.py#L434

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -716,7 +716,10 @@ func (r *ReconcileMongoDbMultiReplicaSet) updateOmDeploymentRs(ctx context.Conte
 	processIds := getReplicaSetProcessIdsFromReplicaSets(mrs.Name, existingDeployment)
 	// If there is no replicaset configuration saved in OM, it might be a new project, so we check the ids saved in annotation
```
nammn, line 718: can you add a comment on how this can even happen (e.g. migration between om projects)

nammn (approved):

RELEASE_NOTES.md
vinilage: Fine for me!!!

vinilage (approved):

docker/mongodb-kubernetes-tests/kubetester/mongodb_multi.py
lucian-tosa: `sts.status.replicas` is not the same as the members field in the clusterspec item. The statefulset is scaled one by one.

controllers/operator/mongodbmultireplicaset_controller.go
lucian-tosa: If rsMemberIds is nil, indeed the string will end up being "null". But do we want to add this string to the annotation?
If instead rsMemberIds is an empty map, then we would end up with the string "{}". Same question in this case

controllers/operator/mongodbmultireplicaset_controller.go
lucian-tosa: done

lsierant (approved):
LGTM!

controllers/operator/mongodbmultireplicaset_controller.go
nammn: no - its fine!

PR #220, opened 2025-07-02
Remove set -x in install_csi_driver.sh

lsierant:
# Summary

This removes unnecessary bash debug switch in install_csi_driver.sh and aligns it to the standard `set -Eeou pipefail` preamble we use in all shell scripts.


lucian-tosa (approved):

nammn (approved):
finally

PR #224, opened 2025-07-03
CLOUDP-329300: Bump Ops Manager Container Image version to 8.0.10

mms-build-account:
_Opened by Private Cloud Tools (PCT)_.

# Ticket
[CLOUDP-329300](https://jira.mongodb.org/browse/CLOUDP-329300)

# Description
Bump Ops Manager container image version to 8.0.10.

# Reviewer Checklist

Before merging this PR, verify the following:
- [ ] the following tasks are passing in Evergreen:
  - `publish_ops_manager` task (variant: `publish_om80_images`)
- [ ] the `agent_version` was updated correctly
- [ ] the `tools_version` was updated correctly


viveksinghggits:
As we can see this PR is failing in CI and when we shared the CI failure with the OM team they recommended some app settings that we should use in OM in our tests to check if the test passes, but the test didn't pass even after using those app settings.

Then it was recommended by the OM team that we should wait for the next release (8.0.11) of the OM and the issue (that we are seeing in CI) will be fixed in that release. And were provided a custom build that we tested and the CI passed. Detailed discussion can be found in this slack thread https://mongodb.slack.com/archives/C06GQK609CJ/p1750440836507499 

Since the fix is going to be part of the next release, we should not release the versions where the issue exists. That's why I am closing this PR and the next PR that gets raised for version 8.0.11 should pass in CI.

PR #230, opened 2025-07-04
Handle new mongot config schema

lsierant:
# Summary

Implements minimum necessary changes to run against newest mongot container built from master: 

* Workaround for mongot's requirement of password file having owner-only permissions
* New config schema structure
* Switched default mongot image to ECR
* Simulated searchCoordinated role

### DIFF
--- a/docker/mongodb-kubernetes-tests/tests/common/search/movies_search_helper.py
+++ b/docker/mongodb-kubernetes-tests/tests/common/search/movies_search_helper.py
@@ -1,5 +1,7 @@
 import logging
 
+import pymongo.errors
+
 from kubetester import kubetester
 from tests import test_logger
 from tests.common.search.search_tester import SearchTester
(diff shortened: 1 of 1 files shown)


controllers/search_controller/mongodbsearch_reconcile_helper.go
```diff
@@ -207,19 +208,41 @@ func buildSearchHeadlessService(search *searchv1.MongoDBSearch) corev1.Service {
...
+				TLS: mongot.ConfigTLS{Mode: "disabled"},
+			},
+		},
+		Metrics: mongot.ConfigMetrics{
 			Enabled: true,
 			Address: fmt.Sprintf("localhost:%d", search.GetMongotMetricsPort()),
 		},
-		Logging: mongot.Logging{
+		HealthCheck: mongot.ConfigHealthCheck{
+			Address: "0.0.0.0:8080",
```
anandsyncs, line 240: Do we need to read from a config?

lsierant: I don't think so, we know in the operator what we're setting here, so the readiness probe can be configured accordingly.

controllers/search_controller/search_construction.go
```diff
@@ -180,7 +186,20 @@ func mongodbSearchContainer(mdbSearch *searchv1.MongoDBSearch, volumeMounts []co
...
+chown 2000:2000 /tmp/keyfile
+chmod 0600 /tmp/keyfile
+
+cp /mongot/sourceUserPassword/password /tmp/sourceUserPassword
+chown 2000:2000 /tmp/sourceUserPassword
+chmod 0600 /tmp/sourceUserPassword
+
+if ! /mongot-community/mongot --config /mongot/config/config.yml; then
+	echo "mongot error code, sleeping to allow to ssh"
+	sleep 600
```
lsierant: this is temporary, to be removed before merging

PR #231, opened 2025-07-04
Fix flakiness in the E2E test `e2e_multi_cluster_replica_set_scale_up`

viveksinghggits:
# Summary

The E2E test `e2e_multi_cluster_replica_set_scale_up` has been flaky and @lucian-tosa suggested that we should fix it. It has been failing while waiting for statefulsets (STSs) to have correct number of in case multi cluster mongoDB deployment. And the problem was sometimes after the `MongoDBMultiCluster (mdbmc)` resource got into running phase (that would mean all the STSs are ready), some of the STSs got into not ready state.
When we see that `mdbmc` resource is Running we try to make sure that STSs have correct number of replicas but because of above problem (STSs transitioning into not ready state from ready), STS didn't have the correct number of replicas and tests failed.
The reason why STS was transitioning into not ready state from ready is, the pod that it was maintaining did the same, i.e., it transitioned from ready state to not ready state. After looking into it further we got to know that the pod is behaving like this because sometimes, it's readiness probe fails momentarily. And because of that the pod gets to ready and then transitioned to not ready (readiness probe failed) and then eventually becomes ready. This is documented in much more detail in the document [here](https://jira.mongodb.org/browse/CLOUDP-329231).  

The ideal fix of the problem would be to figure out why the readiness probe fails and then fix that. But this PR has the workaround that changes the test slightly to wait for STSs to get correct number of replicas.

Jira ticket: https://jira.mongodb.org/browse/CLOUDP-329422

## Proof of Work

Ran the test `e2e_multi_cluster_replica_set_scale_up` manually locally to make sure that it's passing consistently. I am not able to reproduce the flakiness now.

## Checklist

### DIFF
--- a/docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_replica_set_scale_up.py
+++ b/docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_replica_set_scale_up.py
@@ -81,9 +81,10 @@ def test_statefulsets_have_been_created_correctly(
     mongodb_multi: MongoDBMulti,
     member_cluster_clients: List[MultiClusterClient],
 ):
-    # even though we already verified, in previous test, that the MongoDBMultiCluster resource's phase is running (that would mean all STSs are ready);
-    # checking the expected number of replicas for STS makes the test flaky. That's why we are waiting for STS to have expected number of replicas.
-    # Details are documented here https://jira.mongodb.org/browse/CLOUDP-329231
+    # Even though we already verified, in previous test, that the MongoDBMultiCluster resource's phase is running (that would mean all STSs are ready);
+    # checking the expected number of replicas for STS makes the test flaky because of an issue mentioned in detail in this ticket https://jira.mongodb.org/browse/CLOUDP-329231.
+    # That's why we are waiting for STS to have expected number of replicas. This change can be reverted when we make the proper fix as
+    # mentioned in the above ticket.
     def fn():
         cluster_one_client = member_cluster_clients[0]
         cluster_one_statefulsets = mongodb_multi.read_statefulsets([cluster_one_client])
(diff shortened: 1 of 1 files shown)


docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_replica_set_scale_up.py
```diff
@@ -80,18 +81,29 @@ def test_statefulsets_have_been_created_correctly(
     mongodb_multi: MongoDBMulti,
     member_cluster_clients: List[MultiClusterClient],
 ):
-    statefulsets = mongodb_multi.read_statefulsets(member_cluster_clients)
-    cluster_one_client = member_cluster_clients[0]
-    cluster_one_sts = statefulsets[cluster_one_client.cluster_name]
-    assert cluster_one_sts.status.ready_replicas == 1
+    # even though we already verified, in previous test, that the MongoDBMultiCluster resource's phase is running (that would mean all STSs are ready);
+    # checking the expected number of replicas for STS makes the test flaky. That's why we are waiting for STS to have expected number of replicas.
+    # Details are documented here https://jira.mongodb.org/browse/CLOUDP-329231
```
MaciejKaras: I understand that this is a temporary fix for this particular test flakiness. I would write down comment that after fixing the issue we should revert those changes.

We should also add this as a Jira subtask to https://jira.mongodb.org/browse/CLOUDP-329231 and link the PR to revert.

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_replica_set_scale_up.py
```diff
@@ -116,18 +128,35 @@ def test_statefulsets_have_been_scaled_up_correctly(
...
-    cluster_two_client = member_cluster_clients[1]
-    cluster_two_sts = statefulsets[cluster_two_client.cluster_name]
-    assert cluster_two_sts.status.ready_replicas == 1
-
-    cluster_three_client = member_cluster_clients[2]
-    cluster_three_sts = statefulsets[cluster_three_client.cluster_name]
-    assert cluster_three_sts.status.ready_replicas == 2
+    # even though we already verified, in previous test, that the MongoDBMultiCluster resource's phase is running (that would mean all STSs are ready);
+    # checking the expected number of replicas for STS makes the test flaky. That's why we are waiting for STS to have expected number of replicas.
+    # Details are documented here https://jira.mongodb.org/browse/CLOUDP-329231
```
MaciejKaras: same here

MaciejKaras (approved):
Great work Vivek! Nice investigation and I like the description of the issues + the Jira ticket for root cause issue 🥇

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_replica_set_scale_up.py
viveksinghggits: I have created a new ticket and linked it to https://jira.mongodb.org/browse/CLOUDP-329231 like suggested.

lucian-tosa (approved):
Nice work, LGTM

anandsyncs (approved):

PR #244, opened 2025-07-09
CLOUDP-321075: fail to reconcile mongo db search for version 1.47.0

anandsyncs:
# Summary

This pull request introduces validation for unsupported MongoDBSearch image versions and adds a corresponding test case. The key changes include implementing a new validation method, updating the reconciliation logic, and adding constants for unsupported versions.

### Validation for unsupported MongoDBSearch image versions:

* **New validation method**: Added `ValidateSearchImageVersion` to `MongoDBSearchReconcileHelper` to check if the specified or container image version matches an unsupported version (`1.47.0`). If so, it returns an error. 

* **Integration into reconciliation workflow**: Updated the `reconcile` method to invoke `ValidateSearchImageVersion` before proceeding with other validations. 

* **Constants for unsupported version**: Introduced `unsupportedSearchVersion` and `unsupportedSearchVersionErrorFmt` constants to centralize the unsupported version logic and error formatting. 
### Test case for validation:

* **New test**: Added `TestMongoDBSearchReconcile_InvalidSearchImageVersion` to validate the error handling for unsupported MongoDBSearch versions. This ensures the reconciliation fails with the appropriate error message. 

## Proof of Work

Test pass

## Checklist

  * **q**: Clarifying question.


controllers/search_controller/mongodbsearch_reconcile_helper.go
```diff
@@ -264,7 +270,32 @@ func (r *MongoDBSearchReconcileHelper) ValidateSingleMongoDBSearchForSearchSourc
...
+		}
+		return nil
+	}
+
+	if r.mdbSearch.Spec.StatefulSetConfiguration == nil {
+		return nil
+	}
+
+	for _, container := range r.mdbSearch.Spec.StatefulSetConfiguration.SpecWrapper.Spec.Template.Spec.Containers {
+		if strings.Contains(container.Image, unsupportedSearchVersion) {
```
fealebenpae: This could trigger on any sidecar containers users add. Perhaps the if check should also test whether the container name equals the `MongotContainerName` constant in `search_construction.go`?

controllers/search_controller/mongodbsearch_reconcile_helper.go
```diff
@@ -264,7 +270,32 @@ func (r *MongoDBSearchReconcileHelper) ValidateSingleMongoDBSearchForSearchSourc
...
+			strings.Join(resourceNames, ", "),
+		)
+	}
+
+	return nil
+}
+
+func (r *MongoDBSearchReconcileHelper) ValidateSearchImageVersion() error {
+	version := strings.TrimSpace(r.mdbSearch.Spec.Version)
+	if version != "" {
```
fealebenpae, line 435: If the version field in the spec is empty, you should check `r.operatorSearchConfig.SearchVersion` (see the `createOrUpdateStatefulSet()` method on `MongoDBSearchReconcileHelper` and perhaps extract a helper to calculate the version).

fealebenpae (changes requested):

controllers/search_controller/mongodbsearch_reconcile_helper.go
```diff
@@ -264,8 +270,44 @@ func (r *MongoDBSearchReconcileHelper) ValidateSingleMongoDBSearchForSearchSourc
...
+	if r.operatorSearchConfig.SearchVersion != "" {
+		return r.operatorSearchConfig.SearchVersion
+	}
+
+	if r.mdbSearch.Spec.StatefulSetConfiguration == nil {
+		return ""
+	}
+
+	for _, container := range r.mdbSearch.Spec.StatefulSetConfiguration.SpecWrapper.Spec.Template.Spec.Containers {
+		if strings.Contains(container.Name, MongotContainerName) {
```
fealebenpae: I think this should check for equality instead. We want to check for the exact container name we construct, otherwise it could still match other sidecars inserted in the statefulset overrides if its name just contains `mongot`

fealebenpae (approved):

controllers/search_controller/mongodbsearch_reconcile_helper.go
```diff
@@ -264,8 +270,44 @@ func (r *MongoDBSearchReconcileHelper) ValidateSingleMongoDBSearchForSearchSourc
...
+		return xerrors.Errorf(unsupportedSearchVersionErrorFmt, unsupportedSearchVersion)
 	}
 
 	return nil
 }
+
+func (r *MongoDBSearchReconcileHelper) getMongotImage() string {
+	version := strings.TrimSpace(r.mdbSearch.Spec.Version)
+	if version != "" {
+		return version
```
lsierant, line 436: I find the method a bit confusing what is really returned. In this codepath we'll get a version, but below we'll return a full image (which is also consistent with the method name). 

I'd recommend renaming the method to getMongotVersion() and return an actual version tag from the container.Image. Be also mindful that this container.Image might contain digest pinned image instead of a semver tag.

controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -203,3 +203,13 @@ func TestMongoDBSearchReconcile_MultipleSearchResources(t *testing.T) {
 
 	checkSearchReconcileFailed(ctx, t, reconciler, c, search1, "multiple MongoDBSearch")
 }
+
+func TestMongoDBSearchReconcile_InvalidSearchImageVersion(t *testing.T) {
```
lsierant, line 242: can we add a testcase for the statefulset codepath? I'd do it by doing the reconcile and then updating search.Spec.Version to "" (not sure if that will work)

controllers/search_controller/mongodbsearch_reconcile_helper.go
```diff
@@ -30,7 +30,9 @@ import (
 )
 
 const (
-	MongoDBSearchIndexFieldName = "mdbsearch-for-mongodbresourceref-index"
+	MongoDBSearchIndexFieldName      = "mdbsearch-for-mongodbresourceref-index"
+	unsupportedSearchVersion         = "1.47.0"
+	unsupportedSearchVersionErrorFmt = "MongoDBSearch version %s is not supported"
```
lsierant: Can we make this error message more informative for the user? Let's mention why it's not supported, how the operator is treating it (ignoring it essentially), and perhaps that the existing workload will still work but cannot be reconfigured by the operator and they need to start from scratch.

controllers/search_controller/mongodbsearch_reconcile_helper.go
anandsyncs: I would say that I just change the function name to `getMongotImageOrVersion` because at the end we just need to do a string contains operation and we can live without over-engineering this temporary code.
Wdyt?

controllers/search_controller/mongodbsearch_reconcile_helper.go
anandsyncs: This sounds like something that needs to be in the docs. 
Would a doc link be more helpful?

PR #272, opened 2025-07-16
Remove undocumented `operator.enablePVCResize` Helm value

m1kola:
# Summary

In https://github.com/mongodb/mongodb-kubernetes/pull/260 I moved `PersistentVolumeClaim` permissions into a separate role, enabling modularity because these permissions are controlled by `operator.enablePVCResize` Helm value and appeared to be optional.

Later I noticed that `operator.enablePVCResize` is undocumented.

I propose that we revert changes done in https://github.com/mongodb/mongodb-kubernetes/pull/260 and remove the undocumented setting. Permissions are enabled by default already and I do not think there is a rationale for this to be configurable.

Furthermore - we do not test this with `operator.enablePVCResize` set to `false`. We don't even know if the operator will still function correctly without the permissions.

## Proof of Work

N/A - CI must be green.

## Checklist

  * **q**: Clarifying question.


RELEASE_NOTES.md
```diff
@@ -4,7 +4,7 @@
 # MCK 1.3.0 Release Notes
 
 ## Other Changes
-* Optional permissions for `PersistentVolumeClaim` moved to a separate role. When managing the operator with Helm it is possible to disable permissions for `PersistentVolumeClaim` resources by setting `operator.enablePVCResize` value to `false` (`true` by default). When enabled, previously these permissions were part of the primary operator role. With this change, permissions have a separate role.
+* The undocumented `operator.enablePVCResize` Helm value has been removed. If you previously set this value to `false`, please note that the operator roles will now include permissions for `PersistentVolumeClaim` resources by default.
```
lsierant, line 7: This should be considered a breaking change, so requiring bumping major version...

lsierant (changes requested):
Removing this switch should be considered as a breaking change.

RELEASE_NOTES.md
m1kola: Arguably it is not a breaking change since we are adding new RBAC by default. Yes, we are removing the setting which only appears in `values.yaml`, but at runtime it won't negatively affect the operator or workloads.

PR #278, opened 2025-07-21
CLOUDP-323995 CLOUDP-321068 Ingress and Egress TLS in mongot

fealebenpae:
# Summary

## Checklist

  * **q**: Clarifying question.


api/v1/search/mongodbsearch_types.go
```diff
@@ -47,6 +49,22 @@ type MongoDBSource struct {
...
+	// +optional
+	TLS TLS `json:"tls"`
+}
+
+type TLS struct {
+	Enabled bool `json:"enabled"`
+	// CertificateKeySecret is a reference to a Secret containing a private key and certificate to use for TLS.
+	// The key and cert are expected to be PEM encoded and available at "tls.key" and "tls.crt".
+	// This is the same format used for the standard "kubernetes.io/tls" Secret type, but no specific type is required.
+	// Alternatively, an entry tls.pem, containing the concatenation of cert and key, can be provided.
```
lsierant, line 62: do we normally support tls.pem AND tls.crt in secrets? I thought we're only supporting standard tls secret types so tls.key and tls.crt only. We might have some leftovers in the codebase handling tls.pem-type secrets but we shouldn't support them anymore for new things. Could you verify it's the case here?

controllers/operator/mongodbsearch_controller.go
```diff
@@ -89,5 +90,6 @@ func AddMongoDBSearchController(ctx context.Context, mgr manager.Manager, operat
 		For(&searchv1.MongoDBSearch{}).
 		Watches(&mdbcv1.MongoDBCommunity{}, r.mdbcWatcher).
 		Owns(&appsv1.StatefulSet{}).
+		Owns(&corev1.Secret{}).
```
lsierant, line 93: just to ensure - we won't take any ownership of tls certificates referenced in the CR? I.e. those won't be deleted with deleting the CR?

controllers/search_controller/mongodbsearch_reconcile_helper.go
```diff
@@ -85,13 +89,28 @@ func (r *MongoDBSearchReconcileHelper) reconcile(ctx context.Context, log *zap.S
 		return workflow.Failed(err)
 	}
 
-	mongotConfig := createMongotConfig(r.mdbSearch, r.db)
-	configHash, err := r.ensureMongotConfig(ctx, mongotConfig)
+	ingressTlsMongotModification, ingressTlsStsModification, err := r.ensureIngressTlsConfig(ctx)
```
lsierant, line 92: nit: can we have Tls as uppercased TLS in those names?

api/v1/search/mongodbsearch_types.go
fealebenpae: We’re using the community operator’s tls secret handling behind the scenes which does support the concatenated tls.pem field. I copied this comment directly from the Community CRD. I am happy to drop the tls.pem thing altogether, though.

controllers/search_controller/mongodbsearch_reconcile_helper.go
lsierant: While I agree it's quite nice naming convention specifying mongot's server as ingress and mongot->mongod as egress, I'm a bit worried it's initially misleading/confusing as ingress in k8s is generally considered as a mechanism for exposing workloads externally.

controllers/operator/mongodbsearch_controller.go
fealebenpae: Correct. This is for the secret the operator creates which contains the concatenated certificate and key, just like the mongod secret.

controllers/search_controller/mongodbsearch_reconcile_helper.go
fealebenpae: How about if we call them “query” and “replication” as the search team does?

api/v1/search/mongodbsearch_types.go
lsierant: Let's drop it and keep it simple, exactly the same as tls-type secrets are defined 👍

controllers/search_controller/mongodbsearch_reconcile_helper.go
lsierant: replication is regarded also as syncSource, at least in the config schema. Query seems OK, let's see how those work out in the codebase, but it's not super important

fealebenpae: That sounds good to me. I’m not attached to the names I used - I just wanted to see if things worked first.

docker/mongodb-kubernetes-tests/tests/search/search_community_tls.py
```diff
@@ -0,0 +1,324 @@
...
+
+TLS_SECRET_NAME = "tls-secret"
+TLS_CA_CONFIGMAP_NAME = "tls-ca-configmap"
+
+# MongoDBSearch TLS configuration
+MDBS_TLS_SECRET_NAME = "mdbs-tls-secret"
+
+
+@fixture(scope="module")
+def ca_certificate(tmp_path_factory: TempPathFactory):
```
lsierant: I think we're relying on already generated ca.crt ca.key in the test resource files (look for ca.crt). So if we don't do anything more here than just ca file, we might not need this code

docker/mongodb-kubernetes-tests/tests/search/search_community_tls.py
```diff
@@ -0,0 +1,324 @@
...
+        .sign(ca_private_key, hashes.SHA256())
+    )
+
+    cert_file_path = tmp_path_factory.mktemp("certs").joinpath("ca.crt")
+    cert_file_path.write_bytes(ca_cert.public_bytes(serialization.Encoding.PEM))
+
+    return ca_cert, ca_private_key, cert_file_path
+
+
+def generate_server_certificate(
```
lsierant: also for this consider using existing code: look for `generate_cert`, `create_tls_certs` functions and usages. We use cert-manager in e2e tests, so we don't need this custom RSA code around.

docker/mongodb-kubernetes-tests/tests/search/search_community_tls.py
```diff
@@ -0,0 +1,324 @@
...
+
+@fixture(scope="module")
+def mdbc(namespace: str) -> MongoDBCommunity:
+    resource = MongoDBCommunity.from_yaml(
+        yaml_fixture("community-replicaset-sample-mflix.yaml"),
+        name=MDBC_RESOURCE_NAME,
+        namespace=namespace,
+    )
+
+    # Add TLS configuration
```
lsierant, line 45: please add:
if tryLoad(resource): return resource 

and change CR fixture scope to function. It's a pattern allowing loading (and thus refreshing) the CR with each test's invocation if it's already created.

docker/mongodb-kubernetes-tests/tests/search/search_community_tls.py
```diff
@@ -0,0 +1,324 @@
...
+@mark.e2e_search_community_tls
+def test_wait_for_community_resource_ready(mdbc: MongoDBCommunity):
+    mdbc.assert_reaches_phase(Phase.Running, timeout=1800)
+
+
+@mark.e2e_search_community_tls
+def test_validate_tls_connections(mdbc: MongoDBCommunity, mdbs: MongoDBSearch, namespace: str, ca_certificate):
+    ca_file_path = str(ca_certificate[2])
+
+    with pymongo.MongoClient(
```
lsierant, line 127: can we extend SearchTester to handle TLS? it's already using MongoTester which does those operations already. SearchTester inherits from MongoTester which has `assert_connectivity` method and I believe MongoTester handles TLS already

lsierant: movies_search_helper.assert_connectivity will be available just like that

mongodb-community-operator/pkg/tls/tls.go
```diff
@@ -0,0 +1,113 @@
...
+		Build()
+
+	return fileName, secret.CreateOrUpdate(ctx, getUpdateCreator, operatorSecret)
+}
+
+// getCertAndKey will fetch the certificate and key from the user-provided Secret.
+func getCertAndKey(ctx context.Context, getter secret.Getter, secretName types.NamespacedName) string {
+	cert, err := secret.ReadKey(ctx, getter, tlsSecretCertName, secretName)
+	if err != nil {
+		return ""
```
lsierant, line 56: is it correct to not error in case referenced secret is not there? If we look for something and it's not there, isn't it better to requeue the reconcile and try again?

mongodb-community-operator/pkg/tls/tls.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	return combineCertificateAndKey(cert, key)
+}
+
+// getPem will fetch the pem from the user-provided secret
+func getPem(ctx context.Context, getter secret.Getter, secretName types.NamespacedName) string {
+	pem, err := secret.ReadKey(ctx, getter, tlsSecretPemName, secretName)
+	if err != nil {
+		return ""
```
lsierant, line 71: here as well, it doesn't look good to ignore errors like that

mongodb-community-operator/pkg/tls/tls.go
```diff
@@ -0,0 +1,113 @@
...
+	trimmedCert := strings.TrimRight(cert, "\n")
+	trimmedKey := strings.TrimRight(key, "\n")
+	return fmt.Sprintf("%s\n%s", trimmedCert, trimmedKey)
+}
+
+// getPemOrConcatenatedCrtAndKey will get the final PEM to write to the secret.
+// This is either the tls.pem entry in the given secret, or the concatenation
+// of tls.crt and tls.key
+// It performs a basic validation on the entries.
+func getPemOrConcatenatedCrtAndKey(ctx context.Context, getter secret.Getter, secretName types.NamespacedName) (string, error) {
```
lsierant, line 86: complexity of this function is too much at first glance. Let's see how it will be if we not handle pem keys in the secret

mongodb-community-operator/pkg/tls/tls.go
fealebenpae: This file is a copy of [mongodb_tls.go](https://github.com/mongodb/mongodb-kubernetes/blob/master/mongodb-community-operator/controllers/mongodb_tls.go). I don't believe it should exist as it does now in this PR, but it was the quickest way I could get TLS up and running for testing purposes.

mongodb-community-operator/pkg/tls/tls.go
```diff
@@ -0,0 +1,113 @@
...
+// ensureTLSSecret will create or update the operator-managed Secret containing
+// the concatenated certificate and key from the user-provided Secret.
+// Returns the file name of the concatenated certificate and key
+func EnsureTLSSecret(ctx context.Context, getUpdateCreator secret.GetUpdateCreator, resource TLSConfigurableResource) (string, error) {
+	certKey, err := getPemOrConcatenatedCrtAndKey(ctx, getUpdateCreator, resource.TLSSecretNamespacedName())
+	if err != nil {
+		return "", err
+	}
+	// Calculate file name from certificate and key
+	fileName := OperatorSecretFileName(certKey)
```
lsierant, line 40: do we support or do we plan to support the old + current file names in the secret like in MongoDB now?

fealebenpae: The code for creating the operator TLS secret does what the Community reconciler does because that's where I copied it from - it does support multiple pem entries in the secret, but it doesn't track which one is the current one like the MongoDB reconciler does. That said, tracking the previous certificate is not necessary for mongot to handle certificate rotations.

docker/mongodb-kubernetes-tests/tests/search/search_community_tls.py
fealebenpae: I agree with what you're saying, but I thought using SearchTester for that might be confusing - we create it with the mongod connection string so we can use its assert_connectivity method here as it's already supporting tls, but we need the additional connection string for testing mongot connectivity. Could you elaborate a bit on the changes you envision for SearchTester?

lsierant (approved):
Let's merge and cleanup/refactor this later on.

docker/mongodb-kubernetes-tests/tests/search/search_community_tls.py
lsierant: oh, wait - you're connecting also here directly to mongot. I missed that part. For direct mongot connection it's OK to connect like that here - it will be one shot probably. 
For the connection to the replicaset we can use SearchTester. 

Anyway it's not important and we can leave it like that.

anandsyncs (approved):

PR #283, opened 2025-07-22
CLOUDP-332914: Cleanup: Remove `MDB_ASSUME_ENTERPRISE_IMAGE`

m1kola:
# Summary

We read this env var, but:
1. We provide no way to specify it in the Helm chart
2. It is undocumented: no mention in the docs or even release notes.

## Proof of Work

CI must pass

## Checklist

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@
 ## Other Changes
 * Optional permissions for `PersistentVolumeClaim` moved to a separate role. When managing the operator with Helm it is possible to disable permissions for `PersistentVolumeClaim` resources by setting `operator.enablePVCResize` value to `false` (`true` by default). When enabled, previously these permissions were part of the primary operator role. With this change, permissions have a separate role.
 * `subresourceEnabled` Helm value was removed. This setting used to be `true` by default and made it possible to exclude subresource permissions from the operator role by specifying `false` as the value. We are removing this configuration option, making the operator roles always have subresource permissions. This setting was introduced as a temporary solution for [this](https://bugzilla.redhat.com/show_bug.cgi?id=1803171) OpenShift issue. The issue has since been resolved and the setting is no longer needed.
-* The `MDB_ASSUME_ENTERPRISE_IMAGE` environment variable has been removed. This undocumented environment variable, when set to `true`, forced the `-ent` suffix for the database image version in static architecture.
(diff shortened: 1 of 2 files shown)


changelog/20250722_other_MDB_ASSUME_ENTERPRISE_IMAGE_env_var_removed.md
```diff
@@ -0,0 +1,7 @@
+---
+title: MDB_ASSUME_ENTERPRISE_IMAGE environment variable has been removed
+kind: other
+date: 2025-07-22
+---
+
+* The `MDB_ASSUME_ENTERPRISE_IMAGE` environment variable has been removed. This undocumented environment variable, when set to `true`, forced the `-ent` suffix for the database image version in static architecture.
```
nammn: just to be clear - if the operator does not add -ent to the enterprise image - it will not work. This can only happen if someone mirrors the image with a different name than: `mongodb-enterprise-server`. We might want to add this into the release note

m1kola: Updated in 6fc194d.

RELEASE_NOTES.md
```diff
@@ -6,6 +6,7 @@
...
 * `subresourceEnabled` Helm value was removed. This setting used to be `true` by default and made it possible to exclude subresource permissions from the operator role by specifying `false` as the value. We are removing this configuration option, making the operator roles always have subresource permissions. This setting was introduced as a temporary solution for [this](https://bugzilla.redhat.com/show_bug.cgi?id=1803171) OpenShift issue. The issue has since been resolved and the setting is no longer needed.
+* The `MDB_ASSUME_ENTERPRISE_IMAGE` environment variable has been removed. This undocumented environment variable, when set to `true`, forced the `-ent` suffix for the database image version in static architecture. If you are mirroring images and were using this variable, ensure that you do not rename the server image. The name must contain `mongodb-enterprise-server`; otherwise, the operator will not function correctly.
```
viveksinghggits, line 9: > This undocumented environment variable

nit: should we skip mentioning undocumented here? But I am fine with merging even without removing that. I am not really sure how useful this would be to the users and we are able to perfectly convey the change without using it.

m1kola: I would like to leave it as is. This is a soft justification for the change not being considered breaking.

viveksinghggits: Sounds good.

PR #271, opened 2025-07-16
[CLOUDP-332196] Move `kubectl-mongodb` plugin to `cmd` dir for atomic releases

viveksinghggits:
# Summary

In the efforts towards atomic releases this PR tries to move the `kubectl-mongodb` plugin code from the directory `public/tools/multicluster` to the root of the project `cmd/`. This is to align the project with how other Go based projects are structured. 
Apart from that this PR also moves the `.goreleaser.yaml` for the `kubectl-mongodb` plugin from `public/tools/multicluster/.goreleaser.yaml` directory to the root of the repo. This is also done in order to align how `.goreleaser.yaml` is kept in other Go based opensource projects.
This PR also moves the evergreen file that releases `kubectl-mongodb` plugin from the location `public/.evergreen.yml` to the root of the repo and is renamed to `.evergreen-kubectlplugin.yml`. This is to make sure the evergreen files are kept at the same location, i.e. the root of the repo.

### Note on LICENSE-THIRD-PARTY

The licenses that are used by the third party go libraries that we use in our kubectl plugin, are documented in public docs at [this link](https://www.mongodb.com/docs/mongocli/current/third-party-licenses/).
When we generated the file (`LICENSE-THIRD-PARTY`) after moving the kubectl-mongodb code to `cmd/` directory, we can only see  a new package added, that is `sigs.k8s.io/yaml/goyaml` this seems to be part of the module `sigs.k8s.io/yaml` which was already part of the `LICENSE-THIRD-PARTY`, that's why we don't need to re-publish the docs for third party licenses.

To generate the `LICENSE-THIRD-PARTY`, `make precommit` can be run after setting env var `export MDB_UPDATE_LICENSES=true`.

This PR also fixes https://jira.mongodb.org/browse/CLOUDP-332234

## Proof of Work

Run goreleaser command locally and verify the binaries that are generated in the `dist` directory that they output the correct subcommands.

```
~/work/opensource/mongodb-kubernetes/dist/kubectl-mongodb_darwin_arm64_v8.0 (mongodb-kubectlplugin-refactor*) » ./kubectl-mongodb                             
This application is a tool to simplify maintenance tasks
of MongoDB resources in your kubernetes cluster.

Build: 1cd782d97de76ed25d6170a1d0aaa50382022c08, 2025-07-17T12:10:32Z

Usage:
  kubectl-mongodb [command]

Available Commands:
  completion   Generate the autocompletion script for the specified shell
  debug        Downloads all resources required for debugging and stores them into the disk
[...]

PR #288, opened 2025-07-23
CLOUDP-332943: Refactor unit test execution

lsierant:
# Summary

This PR simplifies unit tests execution to allow a simple `go test ./...` to always correctly execute *all* golang unit tests without running community e2e tests.
Additionally we use gotestsum wrapper by default to better present test output in the command line. 

## Changes
- There is only one, unified go.mod in the repo now, so running `go test ./...` correctly identifies all go files. The last one from init-ops-manager was removed in this PR, and another one from kubectl plugin was removed recently in #271.
- All community e2e tests were marked with `community_e2e` tag, so we don't need to run tests with a filtered package list anymore. 
- [gotestsum](https://github.com/gotestyourself/gotestsum) test runner is used to execute unit tests giving more readable output
- gotestsum has been added as a tool in go.mod by running `go get -tool gotest.tools/gotestsum@latest`. 
  - While using [go.mod's tool](https://go.dev/doc/modules/gomod-ref#tool) directive allows for easy tool access (by `go tool <tool>`) and is not resulting in adding dependencies into the operators binary, it is still cluttering dependency list (in `require` section). We might consider removing it from go mod and installing it as a separate step. 
- All test (python and golang) are reporting results as xml and are all gathered by evg. Test output is gathered in the teardown_task section now. 
- Enabled back golang race tests - those were mistakenly disabled during operator merge to MCK [(diff unit-tests.sh)](https://github.com/mongodb/mongodb-kubernetes/pull/17/files#diff-2a75349bb340f98985096d14d89d5543d43abe5a7662454278e74a59d68a71d9R14)

## How to run tests now
Essentially as before:
```bash
$ make golang-tests
```
or 
```bash
$ make test
```

**Difference**: we don't run with coverage enabled by default, thus allowing to use test caching for faster repeated runs from cli. 

In order to run all unit tests without cache and with coverage run: 
```bash
$ make golang-tests-race
```

We is use gotestsum as:
```bash
go tool gotestsum --junitfile golang-unit-result.xml --format-icons hivis -- -v ./...
```

But thanks to the overall go.mod/unitest simplifications you can use whatever suits you:
1. Use gotestsum
```bash
$ go tool gotestsum
```
2. Use plain go test
```bash
$ go test ./...
```

3. Use other test runner/viewers. I recommend to look into gotestsum options as well.. 

4. Also there is `tparse` which also has very nice summary table:
```bash
go get -tool github.com/mfridman/tparse@latest
go test -v ./... -json | go tool tparse
```
tparse is not supporting xunit output thus it's not used here as a default. 

## Proof of Work
Golang unit tests with remove locks to trigger race test error: [link](https://spruce.mongodb.com/task/mongodb_kubernetes_unit_tests_unit_tests_golang_patch_d841a8e222405ea73cd012d66ab62fae2fc14824_688094bed734fc00078b45fe_25_07_23_07_52_33/tests?execution=0&sorts=STATUS:ASC)
Python unit test with injected error: [evg link](https://spruce.mongodb.com/task/mongodb_kubernetes_unit_tests_unit_tests_python_patch_d841a8e222405ea73cd012d66ab62fae2fc14824_688094bed734fc00078b45fe_25_07_23_07_52_33/tests?execution=0&sorts=STATUS:ASC)

Passing tests: [evg link](https://spruce.mongodb.com/version/688095f702299d0007c1959e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,7 +8,7 @@
 vendor/
 pkg/client/
 docker/mongodb-kubernetes-tests
-# scripts/
+scripts/
 samples/
 .multi_cluster_local_test_files/
 


.dockerignore
```diff
@@ -8,7 +8,7 @@
 vendor/
 pkg/client/
 docker/mongodb-kubernetes-tests
-scripts/
+# scripts/
```
nammn: q: on purpose?

pkg/telemetry/init_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"os"
+
+	"go.uber.org/zap"
+)
+
+func init() {
+	logger, _ := zap.NewDevelopment()
+	zap.ReplaceGlobals(logger)
+	os.Clearenv() // nolint:forbidigo
```
lsierant, line 12: this (and similar additions in other pkgs) is allowing for consistent exec of tests that runs reconcile code across all contexts. The biggest offender was static architecture env which caused different reconcile behavior if tests were executed with different contexts. 
Unit tests should never rely on env vars set outside. If those are needed then should be specified in the test itself.

docker/mongodb-kubernetes-init-ops-manager/scripts/backup-daemon-liveness-probe.sh
```diff
@@ -6,4 +6,4 @@ check_backup_daemon_alive () {
     pgrep --exact 'mms-app' || pgrep -f '/mongodb-ops-manager/jdk/bin/mms-app'
 }
 
-check_backup_daemon_alive
+check_backup_daemon_alive
```
viveksinghggits, line 9:
```suggestion
check_backup_daemon_alive

```

.dockerignore
viveksinghggits: should we just remove this, or commenting it ok?

pkg/telemetry/init_test.go
nammn: but in theory - the proper way is to fix all those tests which are using global env to t.setEnv() instead, right? 
Have you considered doing this instead? Or is that too much of a change?

lsierant: It's just too much of a change for this PR. This hack just works and if we encounter problems with it we should handle it properly.

.dockerignore
lsierant: reverting

lsierant: reverting

PR #289, opened 2025-07-23
CLOUDP-333181: Combined dockerfiles

Julien-Ben:
# Summary

Since we will stop re-building images daily, we won't have `-context` images anymore. We then don't need Dockerfiles templating.
This PR creates self-sufficient Dockerfiles that build our images without templating. They only need variables passed with ARG.
For images which already had a plain Dockerfile, it was renamed `Dockerfile.old`, and inventories modified accordingly.

For now, we keep templates and Sonar, until the Atomic Release epic is completed and we fully get rid of all of it.

## Proof of Work

The proof of work will come when actually using them in a build pipeline. In a future PR.
This one is a noop, Dockerfiles are created but not used yet.

## Methodology

The Dockerfiles were automatically generated, using Sonar templating, by adding special `final_dockerfile` tag to stages necessary to generate one. Using code like:

```
def test_build_init_database_dockerfile():  
    process_image(  
        image_name="init-database",  
        skip_tags=["release"],  
        include_tags=["final_dockerfile"],  
        build_args={  
            "registry": "localhost:5000",  
            "version": "1.1.0",  
            "is_appdb": False,  
            "mongodb_tools_url_ubi": "https://downloads.mongodb.org/tools/db/mongodb-database-tools-rhel93-x86_64-100.12.0.tgz",  
        },  
        build_options={},  
        inventory="inventories/init_database.yaml",  
    )  
```

## Checklist
- [X] Have you checked for release_note changes?


docker/mongodb-agent-non-matrix/Dockerfile
MaciejKaras, line 1: b: for every Dockerfile that is modified (not new file) we need to make sure those changes are not breaking anything. If we want to be more defensive we can copy existng Dockerfile with different extension that sonar will use during build process.

docker/mongodb-agent/Dockerfile
MaciejKaras, line 1: b: same here

docker/mongodb-enterprise-ops-manager/Dockerfile.dcar
MaciejKaras, line 1: I think this file is not used anywhere (this why I have removed it), but please double check.

docker/mongodb-enterprise-ops-manager/LICENSE
MaciejKaras, line 1: same here, the LICENSE file was unused I think

docker/mongodb-kubernetes-readinessprobe/Dockerfile
MaciejKaras, line 1: b: same here

docker/mongodb-kubernetes-upgrade-hook/Dockerfile
MaciejKaras, line 1: b: same here

inventories/database.yaml
MaciejKaras, line 1: b: we should not change anything in inventories .yaml files. Those changes were introduced by me to be able to generate final Dockerfiles. After this they are not needed

lib/sonar/sonar.py
MaciejKaras, line 1: file permission were changed so I can run the script manually, but this was a mistake. It should be reverted

lib/sonar/test/test_final_dockerfiles.py
MaciejKaras, line 1: This script should be removed and only added in the PR description as information how we generated the Dockerfiles

inventories/init_appdb.yaml
```diff
@@ -6,13 +6,13 @@ images:
 - name: init-appdb
   vars:
     context: .
-    template_context: docker/mongodb-kubernetes-init-database
```
Julien-Ben: @MaciejKaras Should we keep this renaming ?

docker/mongodb-agent-non-matrix/Dockerfile.old
```diff
@@ -0,0 +1,60 @@
+ARG imagebase
```
Julien-Ben, line 1: For review: Dockerfile.old files are the copy of the initial Dockerfile

inventories/init_appdb.yaml
```diff
@@ -22,7 +22,6 @@ images:
   - name: init-appdb-template-ubi
     task_type: dockerfile_template
     template_file_extension: ubi_minimal
-    tags: ["ubi"]
```
MaciejKaras: lets keep this tag

inventories/init_appdb.yaml
MaciejKaras: I would rather keep the same name here to prevent any changes. `mongodb-kubernetes-init-appdb` can unused for now

Julien-Ben:
I see that `init_release_agents_on_ecr` ran (and failed) on some commits, I'm confused why

docker/mongodb-kubernetes-init-appdb/Dockerfile
lucian-tosa, line 1: Keep in mind that we should get rid of this image entirely [CLOUDP-329374](https://jira.mongodb.org/browse/CLOUDP-329374).
But we can merge it for now and make that change before 1.3

lucian-tosa (approved):

docker/mongodb-agent/Dockerfile
```diff
@@ -1,5 +1,40 @@
...
+
+FROM public.ecr.aws/docker/library/golang:1.24 AS readiness_builder
+
+WORKDIR /go/src/github.com/mongodb/mongodb-kubernetes/
+
+COPY --from=dependency_downloader /go/pkg /go/pkg
+COPY . /go/src/github.com/mongodb/mongodb-kubernetes
+
+RUN CGO_ENABLED=0 GOFLAGS=-buildvcs=false go build -o /readinessprobe ./mongodb-community-operator/cmd/readiness/main.go
+RUN CGO_ENABLED=0 GOFLAGS=-buildvcs=false go build -o /version-upgrade-hook ./mongodb-community-operator/cmd/versionhook/main.go
```
viveksinghggits, line 22: nit: should we have these two `go build` commands in just one `RUN` statement? That would result into less layers and potentially smaller image.

viveksinghggits (commented):
LGTM

docker/mongodb-kubernetes-init-appdb/Dockerfile
MaciejKaras: There are the same apart from the:
 - upgrade hook 
 - different docker labels

If we want this to be the same file, with the same contents and labels, I'm fine. I don't know if we should do it as part of atomic, or do it a little later.

docker/mongodb-agent/Dockerfile
MaciejKaras: I agree, but in this PR we are not optimising the Dockerfiles. We want to make sure we don't break anything and have exactly the same Dockerfiles as we used before. We should optimise them definitely, but a little bit later

viveksinghggits: Sounds good to me. I will resolve this conversation.

PR #267, opened 2025-07-15
CLOUDP-331841 - remove the agent matrix work

nammn:
# Summary

- [TD](https://docs.google.com/document/d/1zu95dakDbIXCncakW2FbPSayKqPYPc-ga6YSEMMzB1M/edit?tab=t.0)
## Database:
from:

```
+------------------------------------------------------------+
|                        Pod: rs1                            |
+------------------------------------------------------------+
| Container 1: mongodb-agent (operator suffixed)             |
|   - entry: agent-launcher.sh                               |
|   - livenessProbe                                          |
|   - readinessProbe                                         |
+------------------------------------------------------------+
| Container 2: mongodb-enterprise-database                   |
|   - mongod                                                 |
|   - mongosh                                                |
+------------------------------------------------------------+
```

to:
```
+------------------------------------------------------------+
|                        Pod: rs1                            |
+------------------------------------------------------------+
| Container 1: mongodb-agent (not operator suffixed)         |
|   - entry: agent-launcher-shim.sh                          |
|   - dummy-livenessProbe                                    |
|   - dummy-readinessProbe                                   |
+------------------------------------------------------------+
| Container 2: utilities-holder (name tbd)                   |
|   - readinessProbe                                         |
|   - livenessProbe                                          |
|   - agent-launcher.sh                                      |
+------------------------------------------------------------+
| Container 3: mongodb-enterprise-database                   |
|   - mongod                                                 |
|   - mongosh                                                |
+------------------------------------------------------------+
```

AppDB:

from:

```
+------------------------------------------------------------+
|                        Pod: rs1                            |
+------------------------------------------------------------+
| Container 1: mongodb-agent (operator suffixed)             |
|   - entry: agent/mongodb-agent                             |
|   - agent-launcher.sh                                      |
|   - livenessProbe                                          |
|   - readinessProbe                                         |
+------------------------------------------------------------+
| Container 2: mongodb-enterprise-database                   |
|   - mongod                                                 |
|   - mongosh                                                |
+------------------------------------------------------------+
```

to:

```
+------------------------------------------------------------+
|                        Pod: rs1                            |
+------------------------------------------------------------+
| Container 1: mongodb-agent (not operator suffixed)         |
|   - entry: /usr/local/bin/setup-agent-files.sh &&          |
|            agent/mongodb-agent                             |
|   - agent-launcher-shim.sh (unused)                        |
|   - dummy-livenessProbe                                    |
|   - dummy-readinessProbe                                   |
+------------------------------------------------------------+
[...]

PR #292, opened 2025-07-24
CLOUDP-331886 - add new agent binaries as base

nammn:
# Summary

- those scripts are required for agent matrix removal. They are placed here - since we overwrite the tag and this would race with the images created in other e2e tests

### Dockerfile updates:
* Updated `docker/mongodb-agent-non-matrix/Dockerfile` to copy new scripts (`agent-launcher-shim.sh`, `setup-agent-files.sh`, `dummy-probe.sh`, `dummy-readinessprobe.sh`) into the container at build time. These scripts are placed in `/usr/local/bin` for execution.
* Updated `docker/mongodb-agent-non-matrix/Dockerfile.builder` to include the same script files in the build process, ensuring they are available in the image.

### New scripts added:
* [`agent-launcher-shim.sh`](diffhunk://#diff-7050c43142cef16fa7268fd7d1356e96bdf5de2c8bc0d050f277babee01a397cR1-R31): A script to initialize the agent launcher by setting up agent files and starting the launcher if the required script is found.
* [`setup-agent-files.sh`](diffhunk://#diff-fbe2cd71b8fc58242e04e18562c58b1085f6b059dce42d94a2f4e5cbdf7a3469R1-R95): A script to set up dummy probe scripts initially, wait for an init container to provide real probe and launcher scripts, and replace the dummy scripts with the real ones.
* [`dummy-probe.sh`](diffhunk://#diff-ff81b0610fd08dd2d32bcbe72c0d78fb402d5f6dbe8a388bad94c7cdf9b95e3bR1-R5): A dummy liveness probe script that always returns success to keep the container alive during the setup process.
* [`dummy-readinessprobe.sh`](diffhunk://#diff-4378f36df04937eb7473771a47dc6ac5a424ec8ffdb171e201b9fc8052a54c26R1-R5): A dummy readiness probe script that always returns a "not ready" status until the real probe script is available.

## Proof of Work

- not used anywhere, ci needs to pass
- used by this pr (that contains more information on how its used): https://github.com/mongodb/mongodb-kubernetes/pull/267

## Checklist

### DIFF
--- a/docker/mongodb-agent-non-matrix/dummy-probe.sh
+++ b/docker/mongodb-agent-non-matrix/dummy-probe.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# Dummy liveness probe that returns success to keep container alive during script copying
+# This prevents container from being killed while waiting for real probe scripts
+echo "Using dummy liveness probe - keeping container alive until real probe script is copied"
+exit 0
--- a/docker/mongodb-agent-non-matrix/Dockerfile
+++ b/docker/mongodb-agent-non-matrix/Dockerfile
@@ -46,6 +46,12 @@ COPY --from=base /data/mongodb-agent.tar.gz /agent
 COPY --from=base /data/mongodb-tools.tgz /agent
 COPY --from=base /data/LICENSE /licenses/LICENSE
 
+# Copy scripts to a safe location that won't be overwritten by volume mount
+COPY --from=base /opt/scripts/agent-launcher-shim.sh /usr/local/bin/agent-launcher-shim.sh
+COPY --from=base /opt/scripts/setup-agent-files.sh /usr/local/bin/setup-agent-files.sh
+COPY --from=base /opt/scripts/dummy-probe.sh /usr/local/bin/dummy-probe.sh
+COPY --from=base /opt/scripts/dummy-readinessprobe.sh /usr/local/bin/dummy-readinessprobe
+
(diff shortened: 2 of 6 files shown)


lucian-tosa (approved):

docker/mongodb-agent-non-matrix/dummy-probe.sh
viveksinghggits, line 1: Just for my understanding can you please briefly explain why do we need dummy probes?

nammn: when configuring the sts template you need to configure readiness and liveness probe. Because we don't have an init container nor those probes as part of the image, we will need to get them from somewhere. The dummy probes are the ones which are getting moved over by the launcher-shim

viveksinghggits: > Because we don't have an init container

Does this mean we don't have init container initially, or we don't have initcontainer in STS at all? If we have initcontianer in the sts, can actual readiness/liveness probe not be prepared in that initContainer? I am assuming we have to prepare these probes and can not just use it directly in STS.

nammn: we don't have an initContainer in static at all.

viveksinghggits (commented):
LGTM

MaciejKaras (approved):
LGTM!

nammn:
only tests failing are the broken ones (opsmanager upgrade)

PR #303, opened 2025-07-29
CLOUDP-333692: Re-design images building

Julien-Ben:
# Re-design images building

# Changes

## Usage of standalone Dockerfiles
Added in a previous PR, they eliminate the need for templating, and make it possible to **retire Sonar** once the Epic is completed.

## Building with docker buildx, multi-platform builds
In `build_images.py` we use docker buildx through a python API. It eliminates the need for building images separately for each platform, and manually bundling them in a manifest.

## Handle build environments explicitly,
Each build scenario, such as merge to master or a local patch, is handled and defined explicitly, at top level, in the main.py CLI. All options are inferred from the environment as before, but can be overridden through flags.
They define parameters such as registry, image tag, signing enabled ...

# Proof of work

CI

# Note

For the duration of the Atomic Releases epic, both pipelines will be in the repository, until we are done with the staging and promotion process. For now, this new system only work for patches (local, PR, merge to master).
Releases and daily builds still happen with the legacy `pipeline.py`, the function ` legacy_pipeline` was created in `.evergreen-functions.yml` for this.

Once the Epic is complete, we'll be able to remove:

- Sonar
- Inventories
- Periodic builds
- `pipeline.py`

## Checklist


scripts/release/main.py
MaciejKaras, line 1: The file is named `main.py`, but there are other files in the `scripts/release` folder already that are not related to the container images. Either move the to subfolder or change it's name

.evergreen-functions.yml
```diff
@@ -538,7 +538,43 @@ functions:
         shell: bash
         <<: *e2e_include_expansions_in_env
         working_dir: src/github.com/mongodb/mongodb-kubernetes
-        binary: scripts/evergreen/run_python.sh pipeline.py --include ${image_name} --parallel --sign
+        binary: scripts/evergreen/run_python.sh scripts/release/main.py --parallel ${image_name}
+
+  legacy_pipeline:
```
MaciejKaras, line 544: please mark it as TODO and to be removed after Atomic Releases

scripts/release/main.py
```diff
@@ -0,0 +1,205 @@
...
+"""
+
+SUPPORTED_PLATFORMS = ["linux/amd64", "linux/arm64"]
+
+
+def get_builder_function_for_image_name() -> Dict[str, Callable]:
+    """Returns a dictionary of image names that can be built."""
+
+    image_builders = {
+        "cli": build_CLI_SBOM,
```
MaciejKaras: this can be removed right now

scripts/release/main.py
```diff
@@ -0,0 +1,205 @@
...
+    provider = TracerProvider(resource=resource, active_span_processor=sp)
+    trace.set_tracer_provider(provider)
+
+
+def main():
+
+    _setup_tracing()
+    parser = argparse.ArgumentParser(description="Build container images.")
+    parser.add_argument("image", help="Image to build.")  # Required
+    parser.add_argument("--parallel", action="store_true", help="Build images in parallel.")
```
MaciejKaras: I don't think we need parallel anymore. We don't build so many agents anymore after agent matrix is removed.

Plus it's confusing the logs

scripts/release/main.py
```diff
@@ -0,0 +1,205 @@
...
+    )
+
+    # Agent specific arguments
+    parser.add_argument(
+        "--all-agents",
+        action="store_true",
+        help="Build all agent variants instead of only the latest.",
+    )
+    parser.add_argument(
+        "--parallel-factor",
```
MaciejKaras: same here

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,794 @@
...
+    build_image_generic(
+        image_name=image_name,
+        dockerfile_path="docker/mongodb-kubernetes-operator/Dockerfile",
+        build_configuration=build_configuration,
+        extra_args=args,
+    )
+
+
+def build_operator_image_patch(build_configuration: BuildConfiguration):
+    if not build_operator_image_fast(build_configuration):
```
MaciejKaras: why do we have two different operator build functions?

scripts/release/build_images.py
```diff
@@ -0,0 +1,159 @@
...
+    dockerfile_path: str,
+    dockerfile_args: Dict[str, str],
+    base_registry: str,
+    platforms: list[str] = None,
+    sign: bool = False,
+    build_path: str = ".",
+    push: bool = True,
+):
+    # Login to ECR
+    ecr_login_boto3(region="us-east-1", account_id="268558157000")
```
MaciejKaras: we only need to login when using ECR registry

MaciejKaras: BTW we don't need to login to quay?

scripts/release/build_images.py
```diff
@@ -0,0 +1,159 @@
...
+from botocore.exceptions import BotoCoreError, ClientError
+from python_on_whales.exceptions import DockerException
+
+import docker
+from lib.base_logger import logger
+from scripts.evergreen.release.images_signing import sign_image, verify_signature
+
+
+def ecr_login_boto3(region: str, account_id: str):
+    """
```
MaciejKaras, line 17: maybe check if already logged in before making another login step?

scripts/release/build_context.py
```diff
@@ -0,0 +1,82 @@
...
+            scenario = BuildScenario.RELEASE
+            logger.info(f"Build scenario: {scenario} (git_tag: {git_tag})")
+        elif is_patch:
+            scenario = BuildScenario.PATCH
+            logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
+        elif is_evg:
+            scenario = BuildScenario.STAGING
+            logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
+        else:
+            scenario = BuildScenario.DEVELOPMENT
```
MaciejKaras, line 36: do we need this scenario?

scripts/release/build_context.py
```diff
@@ -0,0 +1,82 @@
...
+        if git_tag:
+            # Release scenario and the git tag will be used for promotion process only
+            scenario = BuildScenario.RELEASE
+            logger.info(f"Build scenario: {scenario} (git_tag: {git_tag})")
+        elif is_patch:
+            scenario = BuildScenario.PATCH
+            logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
+        elif is_evg:
+            scenario = BuildScenario.STAGING
+            logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
```
MaciejKaras, line 34: here we should use `commit_sha` instead of `patch_id`

scripts/release/build_context.py
```diff
@@ -0,0 +1,82 @@
...
+    signing_enabled: bool = False
+    multi_arch: bool = True
+    version: Optional[str] = None
+
+    @classmethod
+    def from_scenario(cls, scenario: BuildScenario) -> "BuildContext":
+        """Create build context from a given scenario."""
+        git_tag = os.getenv("triggered_by_git_tag")
+        patch_id = os.getenv("version_id")
+        signing_enabled = scenario == BuildScenario.RELEASE
```
MaciejKaras, line 58: we should sign for staging as well

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,794 @@
...
+    with ProcessPoolExecutor(max_workers=max_workers) as executor:
+        logger.info(f"running with factor of {max_workers}")
+
+        # We need to regularly push legacy agents, otherwise ecr lifecycle policy will expire them.
+        # We only need to push them once in a while to ecr, so no quay required
+        if not is_release:
+            for legacy_agent in legacy_agent_versions_to_build:
+                tasks_queue.put(
+                    executor.submit(
+                        build_multi_arch_agent_in_sonar,
```
MaciejKaras: b: why cannot we use new process to build those images?

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,794 @@
...
+        sign=build_configuration.sign,
+        build_path=build_path,
+    )
+
+    if with_sbom:
+        produce_sbom(dockerfile_args)
+
+
+@TRACER.start_as_current_span("produce_sbom")
+def produce_sbom(args):
```
MaciejKaras: this we don't want

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,794 @@
...
+
+
+@TRACER.start_as_current_span("sonar_build_image")
+def pipeline_process_image(
+    image_name: str,
+    dockerfile_path: str,
+    build_configuration: BuildConfiguration,
+    dockerfile_args: Dict[str, str] = None,
+    build_path: str = ".",
+    with_sbom: bool = True,
```
MaciejKaras: always without SBOM

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,794 @@
...
+        platforms = ["linux/amd64"]
+        logger.info("Skipping ARM64 builds for community image as this is running in EVG pipeline as a patch")
+    else:
+        platforms = build_configuration.platforms or ["linux/amd64", "linux/arm64"]
+
+    # Extract architectures from platforms for build args
+    architectures = [platform.split("/")[-1] for platform in platforms]
+    multi_arch_args_list = []
+
+    for arch in architectures:
```
MaciejKaras: why are we iterating over architectures for community images? cannot we use the multiarch build?

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,794 @@
...
+    version = args_list[0].get("version", "")
+    platforms = [args.get("architecture") for args in args_list]
+
+    for base_args in args_list:
+        # merge in the registry without mutating caller’s dict
+        build_args = {**base_args, "quay_registry": registry}
+        logger.debug(f"Build args: {build_args}")
+
+        for arch in platforms:
+            logger.debug(f"Building {image_name} for arch={arch}")
```
MaciejKaras: b: every usage of `build_image_generic` (there are 9) is building the image iterating over platforms - not using multi arch approach. I thought only the agent was limited and could not be run using `buildx --platforms` arg. Is this not the case?

.evergreen-functions.yml
Julien-Ben: I renamed it, as a follow up of this PR I think we should reorganize the release folder as we now have 15+ files
We can have for example at least the two following folders:
- Changelog/release notes
- Image builds

scripts/release/main.py
Julien-Ben: Done

scripts/release/main.py
Julien-Ben: I replied above

scripts/release/atomic_pipeline.py
Julien-Ben: This is an optimization we have in the pipeline, that I preserved, I don't know if it is still relevant. I can measure performance of both as a follow up

scripts/release/build_images.py
Julien-Ben: We have a Quay login function as a prerequisite in release tasks in evergreen

```
  quay_login:
    - command: subprocess.exec
      type: setup
      params:
        command: "docker login quay.io -u ${quay_prod_username} -p ${quay_prod_robot_token}"
```

scripts/release/build_images.py
Julien-Ben: From what I've read online, there's no optimized way to easily check if we're already logged in:
We can check if we have the credentials in docker config, but they could be expired
Or we can try to send a request to a registry and see if we get a `200` or `401`

If we think it is important to optimize this, I can look into it further

scripts/release/build_context.py
Julien-Ben: I think it's important to preserve the development scenario, which is when we build locally on our machine.
While it may have the same default configuration as others, it is semantically different and we should be able to evolve the default settings (parallel, platforms, signing, etc.) independently of the others.

scripts/release/atomic_pipeline.py
Julien-Ben: I removed the agent functions that were confusing, they will be added back (and probably heavily modified) after the completion of matrix removal epic

scripts/release/atomic_pipeline.py
Julien-Ben: That's a good point, it is confusing so I removed it

I initially kept it in case we would realize it is not possible to have unified args between platform, but I made the assumption that folks working on the multi platforms epic will manage to do it.

Now we shouldn't have any iteration over platforms and only a list passed through the `BuildConfiguration`, all the way down to `docker buildx`

scripts/release/atomic_pipeline.py
Julien-Ben: Good point, same as my comment above, I made assumptions and removed it.

scripts/release/build_context.py
MaciejKaras: We talked about this yesterday and I now see better the point of having this. Resolving!

scripts/release/atomic_pipeline.py
Julien-Ben: I removed all SBOM related stuff, again I made the assumption we won't need it in the pipeline anytime soon, and will instead use the manual script that Vivek is working on. We can add it again from the old pipeline if really needed

scripts/release/atomic_pipeline.py
Julien-Ben: Answered about SBOMs in another comment

scripts/release/main.py
Julien-Ben: I renamed it, as a follow up of this PR I think we should reorganize the release folder as we now have 15+ files
We can have for example at least the two following folders:

    Changelog/release notes
    Image builds

scripts/release/main.py
Julien-Ben: For now we still have 5+ agent builds per patch, and as agent builds is the longest task in init_test_run variant I'd rather keep it for now

MaciejKaras: Agree!

scripts/release/atomic_pipeline.py
MaciejKaras: @nammn @Julien-Ben do we need to keep this? How much is this useful right now?

scripts/release/build_images.py
MaciejKaras: Thank you

.evergreen-functions.yml
Julien-Ben: Done https://github.com/mongodb/mongodb-kubernetes/pull/303/commits/972b23c9aeacc347d6d0f5b309858a563253bf7e

scripts/release/build_context.py
```diff
@@ -0,0 +1,82 @@
...
+
+    RELEASE = "release"  # Official release triggered by a git tag
+    PATCH = "patch"  # CI build for a patch/pull request
+    STAGING = "staging"  # CI build from a merge to the master
+    DEVELOPMENT = "development"  # Local build on a developer machine
+
+    @classmethod
+    def infer_scenario_from_environment(cls) -> "BuildScenario":
+        """Infer the build scenario from environment variables."""
+        git_tag = os.getenv("triggered_by_git_tag")
```
nammn, line 20: guiding question - should we handle this via env vars or should we have all of that as args instead?

scripts/release/build_images.py
```diff
@@ -0,0 +1,159 @@
+# This file is the new Sonar
+import base64
+from typing import Dict
+
+import boto3
+import python_on_whales
```
nammn, line 6: do we describe/compare python_on_whales in the td compared to other wrappers?

scripts/release/build_images.py
```diff
@@ -0,0 +1,159 @@
...
+            tags=[tag],
+            platforms=platforms,
+            builder=builder_name,
+            build_args=build_args,
+            push=push,
+            provenance=False,  # To not get an untagged image for single platform builds
+            pull=False,  # Don't always pull base images
+        )
+
+        logger.info(f"Successfully built {'and pushed' if push else ''} {tag}")
```
nammn, line 131: should we wrap this important function into a trace?

scripts/release/build_images.py
```diff
@@ -0,0 +1,159 @@
...
+    build_image(
+        tag=image_full_uri,
+        dockerfile=dockerfile_path,
+        path=build_path,
+        args=dockerfile_args,
+        push=push,
+        platforms=platforms,
+    )
+
+    if sign:
```
nammn: same here and all related functions

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,552 @@
...
+def is_running_in_evg_pipeline():
+    return os.getenv("RUNNING_IN_EVG", "") == "true"
+
+
+def load_release_file() -> Dict:
+    with open("release.json") as release:
+        return json.load(release)
+
+
+@TRACER.start_as_current_span("sonar_build_image")
```
nammn: can we rename the span names to match the function name?

scripts/release/atomic_pipeline.py
nammn: i don't remember using this

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,552 @@
+#!/usr/bin/env python3
+
```
nammn, line 2: how do we test this? Do we have a related unit test file integrated?

scripts/release/build_context.py
MaciejKaras: We can always override the scenario by passing the `--scenario` arg. I'm fine with env vars if they are at least documented in the command help

scripts/release/atomic_pipeline.py
MaciejKaras: will this not break existing dashboards (if we have them?)

nammn: nope we don't have them

scripts/release/build_context.py
nammn: thats what I was worrying, right now pipeline is a mess partly due to the fact that we have multiple entrypoints. Which are env vars, arguments and files. This adds complexity as we need to codify argument hierachy.

scripts/release/atomic_pipeline.py
MaciejKaras: The previous tests we had in `pipeline_test.py` were mostly testing some internals that are not used anymore i.e. `run_command_with_retries`, `is_version_in_range` or `test_is_release_step_executed` (sonar stuff) 

Apart from e2e tests that verify image creation I would also ask for:
- moving some tests that still qualify for `atomic_pipeline.py`:
  - `test_build_latest_agent_versions`
  - `test_get_versions_to_rebuild_same_version`
  - possibly others
- add tests for `process_image` and verify what commands are actually invoked using mocks?

Most complex building process is for `agent` image, but this will greatly change after non-matrix is merged, so I would wait with adding tests for it.

scripts/release/build_context.py
MaciejKaras: I agree, but `triggered_by_git_tag`, `version_id`, `RUNNING_IN_EVG` and `is_patch` are internals of `pipeline.py` and I don't like the option to pass those values as cmd args. For example it will be hard to motivate why developer need to provide them when running locally.

Another way is to have a separate cmd tool i.e. `calculate_build_scenario` that calculates the build scenario for given args `triggered_by_git_tag`, `version_id`, `RUNNING_IN_EVG` and `is_patch`. The result will be `scenario` output that can be provided to `atomic_pipeline.py` as arg. The only issue is that evg functions don't support sharing outputs and we would need to somehow embed it in env vars 😭

MaciejKaras: Another improvement is to move all env var definitions to separate python file and hide the values behind python functions. This way at least we will know all env var usages and it will be easier to refactor later

MaciejKaras: I think calculating the scenario and passing it to `atomic_pipeline.py` is the way to go

scripts/release/build_images.py
Julien-Ben: I'll add a section, but TL;DR it is lower level and have 100% parity with docker cli. Especially for using buildx

scripts/release/atomic_pipeline.py
Julien-Ben: I'll remove it then !

scripts/release/atomic_pipeline.py
Julien-Ben: I agree with what Maciej said, all unit tests were for low level internals. I don't think the pipeline is very relevant to test with unit tests.

scripts/release/build_images.py
Julien-Ben: Can we make observability related stuff part of the follow up ticket ?
https://jira.mongodb.org/browse/CLOUDP-335471

scripts/release/build_context.py
Julien-Ben: I think this should be discussed in the TD. The point of the `build_context` and the `pipeline_main` files is to give full flexibility and a central place to decide how we pass arguments.

Once we decide the source of truth (e.g env vars vs JSON config files) we can modify it. 
In the meantime I kept the behavior of the `pipeline.py`, but made it easy to change the configurations.

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,552 @@
...
+    build_image_generic(
+        image_name="mongodb-kubernetes-database",
+        dockerfile_path="docker/mongodb-kubernetes-database/Dockerfile",
+        build_configuration=build_configuration,
+        extra_args=args,
+    )
+
+
+@TRACER.start_as_current_span("sign_image_in_repositories")
+def sign_image_in_repositories(args: Dict[str, str], arch: str = None):
```
lucian-tosa: is this used anywhere?

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,552 @@
...
+    args = {
+        "version": build_configuration.version,
+        "log_automation_config_diff": log_automation_config_diff,
+        "test_suffix": test_suffix,
+    }
+
+    logger.info(f"Building Operator args: {args}")
+
+    image_name = "mongodb-kubernetes"
+    build_image_generic(
```
lucian-tosa: There are a bit too many methods that could easily be merged. There is `pipeline_process_image`, `build_image_generic`, `process_image`. They should be only one function.
I even notice that both `build_image_generic` and `process_image` sign the image

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,552 @@
...
+    :return: the latest agent for each major version
+    """
+    agent_versions_to_build = list()
+    agent_versions_to_build.append(
+        (
+            release["supportedImages"]["mongodb-agent"]["opsManagerMapping"]["cloud_manager"],
+            release["supportedImages"]["mongodb-agent"]["opsManagerMapping"]["cloud_manager_tools"],
+        )
+    )
+
```
lucian-tosa, line 482: Can we reuse the method above for this?

scripts/release/atomic_pipeline.py
Julien-Ben: Renamed ✅

lucian-tosa (commented):
How will this be merged with Maciej's changes (the release_info file)?

scripts/release/build_images.py
nammn: works for me

scripts/release/build_context.py
nammn: we can discuss this in the td, but i rather give up flexibility in favour of clarity

scripts/release/atomic_pipeline.py
Julien-Ben: Good catch, not anymore

scripts/release/atomic_pipeline.py
Julien-Ben: I haven't really touched agent related stuff, once both Nam's PR (matrix removal) and mine are merged, we will make the necessary changes to the atomic pipeline.

scripts/release/atomic_pipeline.py
Julien-Ben: Very good point, with all the simplifications there's no need for so many abstraction layers anymore. I merged and renamed methods.
https://github.com/mongodb/mongodb-kubernetes/pull/303/commits/493d4d67075ebb494be82dc0e09adffa41c2556a
https://github.com/mongodb/mongodb-kubernetes/pull/303/commits/a21b254ba8cd9b7dcb9e34de63dd14a3ea11129d

scripts/release/atomic_pipeline.py
Julien-Ben: https://github.com/mongodb/mongodb-kubernetes/pull/303/commits/6dd208f9ebc35788ae6990197ea6b4c3d081ec26

scripts/release/build_context.py
```diff
@@ -0,0 +1,83 @@
...
+    RELEASE = "release"  # Official release triggered by a git tag
+    PATCH = "patch"  # CI build for a patch/pull request
+    STAGING = "staging"  # CI build from a merge to the master branch
+    DEVELOPMENT = "development"  # Local build on a developer machine
+
+    @classmethod
+    def infer_scenario_from_environment(cls) -> "BuildScenario":
+        """Infer the build scenario from environment variables."""
+        git_tag = os.getenv("triggered_by_git_tag")
+        is_patch = os.getenv("is_patch", "false").lower() == "true"
```
lucian-tosa, line 21: So this is true only when triggering manual patches or PR commits? Will it be false on merges to master?

Julien-Ben:
> How will this be merged with Maciej's changes (the release_info file)?

Maciej answered above, but I also plan to sync with him as soon as this is merged, to properly rebase and wire everything togetger

scripts/release/build_context.py
Julien-Ben: I made it more explicit, and we'll make proper use of staging registry in a follow up:
https://github.com/mongodb/mongodb-kubernetes/pull/303/commits/2ec75879a950aa6ea579792aaa088ba65e48b7fc

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,512 @@
...
+def get_tools_distro(tools_version: str) -> Dict[str, str]:
+    new_rhel_tool_version = "100.10.0"
+    default_distro = {"arm": "rhel90-aarch64", "amd": "rhel90-x86_64"}
+    if Version(tools_version) >= Version(new_rhel_tool_version):
+        return {"arm": "rhel93-aarch64", "amd": "rhel93-x86_64"}
+    return default_distro
+
+
+def is_running_in_evg_pipeline():
+    return os.getenv("RUNNING_IN_EVG", "") == "true"
```
nammn: following your definition of build_context, could we remove all of env access into there as well?

Julien-Ben: Sure, it was indeed unused

https://github.com/mongodb/mongodb-kubernetes/pull/303/commits/e17b32356b779ce9c75cc962f27b296be9bbf7ef

nammn: i meant generally removal of all os.getEnvs into build_context. i really want things to be streamlined

scripts/release/build_images.py
```diff
@@ -0,0 +1,149 @@
...
+        docker.buildx.create(
+            name=builder_name,
+            driver="docker-container",
+            use=True,
+            bootstrap=True,
+        )
+        logger.info(f"Created new buildx builder: {builder_name}")
+    except DockerException as e:
+        # Check if this is a race condition (another process created the builder)
+        if hasattr(e, 'stderr') and 'existing instance for' in str(e.stderr):
```
Julien-Ben: @nammn @lucian-tosa @MaciejKaras 

I'd like your opinion on two solutions here
This function is sometimes failing when building multiple agents concurrently.
I tried two alternatives:
- Catching the exception
- Use ensure_buildx_builder at top leven, in the pipeline_main

None of them are ideal. Using exception is not the right way of handling concurrency. Making the main aware of buildx increase coupling.
My knowledge of python multiprocessing is limited. I know that for multi threading it is easy to make a lock with a global variable and `threading.Lock()` . From what I've seen online, for multi processing we would need to pass the lock from the agent building method all the way down to here. 
I feel like it adds a lot of complexity and optional variables.

What solutions do you prefer ?
Do you see an alternative ?

Here is the alternative solution with ensuring builder in main:
https://github.com/mongodb/mongodb-kubernetes/pull/303/commits/1badff0183576a697aaab94ea1530951b7c52914

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,511 @@
...
+
+    build_image(
+        image_name="mongodb-enterprise-ops-manager-ubi",
+        dockerfile_path="docker/mongodb-enterprise-ops-manager/Dockerfile",
+        build_configuration=build_configuration,
+        extra_args=args,
+    )
+
+
+@TRACER.start_as_current_span("build_image_generic")
```
nammn: the name doesn't match the function

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,511 @@
...
+    image_name: str,
+    dockerfile_path: str,
+    build_configuration: BuildConfiguration,
+    extra_args: dict | None = None,
+    build_path: str = ".",
+):
+    """
+    Build an image then (optionally) sign the result.
+    """
+    # Tracing setup
```
nammn: lets remove that comment

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,511 @@
...
+
+    # merge in the registry without mutating caller's dict
+    build_args = {**args_list, "quay_registry": registry}
+
+    if build_args:
+        span.set_attribute("mck.build_args", str(build_args))
+
+    logger.info(f"Building {image_name}, dockerfile args: {build_args}")
+    logger.debug(f"Build args: {build_args}")
+    logger.debug(f"Building {image_name} for platforms={build_configuration.platforms}")
```
nammn, line 240: since we have the span already setup here - can we save the span attributes as we already log them anyway?

scripts/release/build_images.py
```diff
@@ -0,0 +1,149 @@
...
+            return builder_name
+        
+        # Otherwise, it's a real error
+        logger.error(f"Failed to create buildx builder: {e}")
+        raise
+
+    return builder_name
+
+
+def execute_docker_build(
```
nammn, line 75: or here - i think adding tracing costs 1 minute and saves us all the troubles later.

nammn: especially since you are logging the exact information we need as metrics anyway

scripts/release/atomic_pipeline.py
```diff
@@ -0,0 +1,511 @@
...
+
+
+def _build_agent_operator(
+    agent_version: Tuple[str, str],
+    build_configuration: BuildConfiguration,
+    executor: ProcessPoolExecutor,
+    operator_version: str,
+    tasks_queue: Queue,
+):
+    agent_distro = "rhel9_x86_64"
```
nammn, line 429: you removed the the multiarch agent support with atomic

scripts/release/atomic_pipeline.py
Julien-Ben: Yes that's something we're aiming for
Maciej is already doing it in his PR: github.com/mongodb/mongodb-kubernetes/pull/317/files

scripts/release/atomic_pipeline.py
Julien-Ben: Yes that's on purpose, Lucian is aware of that and these args computation should be done inside Dockerfiles.
I think the diagram in the TD is the best way to understand it

scripts/release/atomic_pipeline.py
Julien-Ben: I think it should really be done as a follow up, it might look like one minute but I have no idea how to properly setup spans. 
The scope of this PR is getting quite large

scripts/release/build_images.py
Julien-Ben: I think it should really be done as a follow up, it might look like one minute but I have no idea how to properly setup spans. 
The scope of this PR is getting quite large

scripts/release/atomic_pipeline.py
Julien-Ben: It's listed in the follow up ticket
https://jira.mongodb.org/browse/CLOUDP-335471

scripts/release/atomic_pipeline.py
nammn: nice!

scripts/release/atomic_pipeline.py
Julien-Ben: Done

scripts/release/atomic_pipeline.py
Julien-Ben: Nice catch thx
https://github.com/mongodb/mongodb-kubernetes/pull/303/commits/715c4adb81c0e87d7346fafc64f0ac3b374ccc16

scripts/release/atomic_pipeline.py
nammn: you don't have to setup anything. All you need to do is already there.
its
span.set_attribute("mck.platforms", build_configuration.platforms)
span.set_attribute("mck.registry", registry)

and the same everywhere...

nammn (approved):

scripts/release/build_context.py
lucian-tosa: I was referring to the `is_patch` variable from evergreen. This is what I found from the devprod docs

> ${is_patch} is "true" if the running task is in a patch build and undefined if it is not.

Does this mean that `is_patch` will be false in a merge to master?

PR #306, opened 2025-07-30
Add diagnostic tool selection with optional DTOOL/DTOOLARGS flags in …

ansh921992:
# Summary

This change updates the diagnostic script to support selecting a tool (oc or kubectl) using the DTOOL environment variable and passing additional flags using DTOOLARGS. The script detects which tools are installed and chooses the preferred tool if specified.

## Proof of Work

The script was tested in an environment where both oc and kubectl were installed. It was verified that when no preference was set, the script correctly defaulted to oc. It was also tested with DTOOL set to either oc or kubectl, and in both cases the script correctly used the specified tool.

## Checklist

### DIFF
--- a/changelog/20250730_feature_add_diagnostic_tool_selection_with_dtool_and.md
+++ b/changelog/20250730_feature_add_diagnostic_tool_selection_with_dtool_and.md
@@ -1,0 +1,6 @@
+---
+title: Add diagnostic tool selection with DTOOL and DTOOLARGS support
+kind: feature
+date: 2025-07-30
+---
+


changelog/20250730_feature_add_diagnostic_tool_selection_with_dtool_and.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Add diagnostic tool selection with DTOOL and DTOOLARGS support
```
viveksinghggits, line 2:
```suggestion
title: Add support to select diagnostic tool with DTOOL and DTOOLARGS 
```

public/support/mdb_operator_diagnostic_data.sh
```diff
@@ -2,6 +2,45 @@
 
 set -Eeou pipefail
 
+# Check if oc and kubectl exist
+command -v oc >/dev/null 2>&1
+ocExists=$?
+
+command -v kubectl >/dev/null 2>&1
+kcExists=$?
```
viveksinghggits, line 10:
```suggestion
kubectlExists=$?
```

viveksinghggits: nit: I think it's better to be explicit about what exactly we are talking about. wdyt?

public/support/mdb_operator_diagnostic_data.sh
```diff
@@ -2,6 +2,45 @@
...
+
+# Let the user override via environment variable
+dToolUser=${DTOOL:-""}
```
viveksinghggits, line 13: `kubectl` and `oc` are not really the diagnostic tools right? they are just tools to interact with the api server. So I would recommend naming this to something else. Maybe `clientTool`, `kubeClientTool` or something like that.

public/support/mdb_operator_diagnostic_data.sh
```diff
@@ -49,7 +89,7 @@ usage() {
 contains() {
   local e match=$1
   shift
-  for e; do [[ "${e}" == "${match}" ]] && return 0; done
+  for e; do [[ "$e" == "$match" ]] && return 0; done
```
viveksinghggits, line 92: is this change necessary? this might upset the shellcheck. or the changed version is better suited for shellcheck? 😆

public/support/mdb_operator_diagnostic_data.sh
```diff
@@ -2,6 +2,45 @@
...
+# Check if oc and kubectl exist
+command -v oc >/dev/null 2>&1
+ocExists=$?
+
+command -v kubectl >/dev/null 2>&1
+kcExists=$?
+
+# Let the user override via environment variable
+dToolUser=${DTOOL:-""}
+dToolUserArgs=${DTOOLARGS:-""}
```
viveksinghggits, line 14: what exactly is the motivation to introduce this? do we want users to be able to configure the kubeconfig?

PR #309, opened 2025-08-01
CLOUDP-342325 Basic enterprise search support

fealebenpae:
# Summary

## Checklist

### DIFF
--- a/controllers/operator/mongodbreplicaset_controller.go
+++ b/controllers/operator/mongodbreplicaset_controller.go
@@ -239,15 +239,15 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
 		}
 	}
 
-	agentCertSecretName := rs.GetSecurity().AgentClientCertificateSecretName(rs.Name).Name
-	agentCertSecretName += certs.OperatorGeneratedCertSuffix
+	agentCertSecretSelector := rs.GetSecurity().AgentClientCertificateSecretName(rs.Name)
+	agentCertSecretSelector.Name += certs.OperatorGeneratedCertSuffix
 
 	// Recovery prevents some deadlocks that can occur during reconciliation, e.g. the setting of an incorrect automation
 	// configuration and a subsequent attempt to overwrite it later, the operator would be stuck in Pending phase.
 	// See CLOUDP-189433 and CLOUDP-229222 for more details.
 	if recovery.ShouldTriggerRecovery(rs.Status.Phase != mdbstatus.PhaseRunning, rs.Status.LastTransition) {
 		log.Warnf("Triggering Automatic Recovery. The MongoDB resource %s/%s is in %s state since %s", rs.Namespace, rs.Name, rs.Status.Phase, rs.Status.LastTransition)
-		automationConfigStatus := r.updateOmDeploymentRs(ctx, conn, rs.Status.Members, rs, sts, log, caFilePath, agentCertSecretName, prometheusCertHash, true, shouldMirrorKeyfile).OnErrorPrepend("Failed to create/update (Ops Manager reconciliation phase):")
+		automationConfigStatus := r.updateOmDeploymentRs(ctx, conn, rs.Status.Members, rs, sts, log, caFilePath, agentCertSecretSelector, prometheusCertHash, true, shouldMirrorKeyfile).OnErrorPrepend("Failed to create/update (Ops Manager reconciliation phase):")
 		deploymentError := create.DatabaseInKubernetes(ctx, r.client, *rs, sts, rsConfig, log)
 		if deploymentError != nil {
 			log.Errorf("Recovery failed because of deployment errors, %w", deploymentError)
@@ -263,7 +263,7 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
 	}
 	status = workflow.RunInGivenOrder(publishAutomationConfigFirst(ctx, r.client, *rs, lastSpec, rsConfig, log),
 		func() workflow.Status {
-			return r.updateOmDeploymentRs(ctx, conn, rs.Status.Members, rs, sts, log, caFilePath, agentCertSecretName, prometheusCertHash, false, shouldMirrorKeyfile).OnErrorPrepend("Failed to create/update (Ops Manager reconciliation phase):")
+			return r.updateOmDeploymentRs(ctx, conn, rs.Status.Members, rs, sts, log, caFilePath, agentCertSecretSelector, prometheusCertHash, false, shouldMirrorKeyfile).OnErrorPrepend("Failed to create/update (Ops Manager reconciliation phase):")
 		},
 		func() workflow.Status {
 			workflowStatus := create.HandlePVCResize(ctx, r.client, &sts, log)
@@ -437,7 +437,7 @@ func AddReplicaSetController(ctx context.Context, mgr manager.Manager, imageUrls
 
 // updateOmDeploymentRs performs OM registration operation for the replicaset. So the changes will be finally propagated
 // to automation agents in containers
-func (r *ReconcileMongoDbReplicaSet) updateOmDeploymentRs(ctx context.Context, conn om.Connection, membersNumberBefore int, rs *mdbv1.MongoDB, set appsv1.StatefulSet, log *zap.SugaredLogger, caFilePath string, agentCertSecretName string, prometheusCertHash string, isRecovering bool, shouldMirrorKeyfileForMongot bool) workflow.Status {
+func (r *ReconcileMongoDbReplicaSet) updateOmDeploymentRs(ctx context.Context, conn om.Connection, membersNumberBefore int, rs *mdbv1.MongoDB, set appsv1.StatefulSet, log *zap.SugaredLogger, caFilePath string, agentCertSecretSelector corev1.SecretKeySelector, prometheusCertHash string, isRecovering bool, shouldMirrorKeyfileForMongot bool) workflow.Status {
 	log.Debug("Entering UpdateOMDeployments")
 	// Only "concrete" RS members should be observed
 	// - if scaling down, let's observe only members that will remain after scale-down operation
@@ -471,7 +471,7 @@ func (r *ReconcileMongoDbReplicaSet) updateOmDeploymentRs(ctx context.Context, c
 		internalClusterPath = fmt.Sprintf("%s%s", util.InternalClusterAuthMountPath, hash)
 	}
 
-	status, additionalReconciliationRequired := r.updateOmAuthentication(ctx, conn, processNames, rs, agentCertSecretName, caFilePath, internalClusterPath, isRecovering, log)
+	status, additionalReconciliationRequired := r.updateOmAuthentication(ctx, conn, processNames, rs, agentCertSecretSelector, caFilePath, internalClusterPath, isRecovering, log)
 	if !status.IsOK() && !isRecovering {
 		return status
 	}
--- a/.evergreen-functions.yml
+++ b/.evergreen-functions.yml
@@ -456,6 +456,7 @@ functions:
         content_type: text/plain
     - command: attach.xunit_results
       params:
+        continue_on_err: true
         file: "src/github.com/mongodb/mongodb-kubernetes/logs/myreport.xml"
 
   upload_e2e_logs_gotest:
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -1480,7 +1480,7 @@ buildvariants:
     run_on:
       - rhel9-power-small
       - rhel9-power-large
-    allowed_requesters: [ "patch", "github_tag" ]
+    allowed_requesters: [ "patch", "github_tag" , "commit"]
     depends_on:
       - name: build_operator_ubi
         variant: init_test_run
@@ -1503,7 +1503,7 @@ buildvariants:
     run_on:
       - rhel9-zseries-small
       - rhel9-zseries-large
-    allowed_requesters: [ "patch", "github_tag" ]
+    allowed_requesters: [ "patch", "github_tag", "commit"]
     depends_on:
       - name: build_operator_ubi
         variant: init_test_run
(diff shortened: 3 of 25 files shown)


controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -609,3 +633,70 @@ func getAllHostsRs(set appsv1.StatefulSet, clusterName string, membersCount int,
...
+
+	if rs.Spec.AdditionalMongodConfig == nil {
+		rs.Spec.AdditionalMongodConfig = mdbv1.NewEmptyAdditionalMongodConfig()
+	}
+	searchMongodConfig := search_controller.GetMongodConfigParameters(search)
+	rs.Spec.AdditionalMongodConfig.AddOption("setParameter", searchMongodConfig["setParameter"])
+
+	mdbVersion, err := semver.ParseTolerant(rs.Spec.Version)
+	if err != nil {
+		log.Warnf("Failed to parse MongoDB version %q: %w. Proceeding without the automatic creation of the searchCoordinator role that's necessary for MongoDB <8.2", rs.Spec.Version, err)
```
lsierant, line 657: shouldn't we return false here?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -609,3 +633,70 @@ func getAllHostsRs(set appsv1.StatefulSet, clusterName string, membersCount int,
...
+		if rs.Spec.Security == nil {
+			rs.Spec.Security = &mdbv1.Security{}
+		}
+		rs.Spec.Security.Roles = append(rs.Spec.Security.Roles, search_controller.SearchCoordinatorRole())
+	}
+
+	return true
+}
+
+func (r *ReconcileMongoDbReplicaSet) mirrorKeyfileIntoSecretForMongot(ctx context.Context, d om.Deployment, rs *mdbv1.MongoDB, log *zap.SugaredLogger) error {
```
lsierant, line 670: I wonder if we could make it part of the search reconciler. It feels a bit awkward to have this special code for making the search reconciler work properly. I feel we should limit changes only to the setParameters. We know the secret name and we can look and mirror it when we need it while reconciling MongoDBSearch resource.

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -609,3 +633,70 @@ func getAllHostsRs(set appsv1.StatefulSet, clusterName string, membersCount int,
...
+	} else {
+		return nil
+	}
+}
+
+func (r *ReconcileMongoDbReplicaSet) lookupCorrespondingSearchResource(ctx context.Context, rs *mdbv1.MongoDB, log *zap.SugaredLogger) *searchv1.MongoDBSearch {
+	var search *searchv1.MongoDBSearch
+	searchList := &searchv1.MongoDBSearchList{}
+	if err := r.client.List(ctx, searchList, &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(search_controller.MongoDBSearchIndexFieldName, rs.GetNamespace()+"/"+rs.GetName()),
```
lsierant, line 691: Do we the assumption that MongoDBSearch must be in the same namespace as MongoDB resource?

controllers/operator/mongodbsearch_controller.go
```diff
@@ -14,29 +14,35 @@ import (
...
+	secretWatcher        watch.ResourceWatcher
 	operatorSearchConfig search_controller.OperatorSearchConfig
 }
 
 func newMongoDBSearchReconciler(client client.Client, operatorSearchConfig search_controller.OperatorSearchConfig) *MongoDBSearchReconciler {
-	mdbcWatcher := watch.New()
 	return &MongoDBSearchReconciler{
 		kubeClient:           kubernetesClient.NewClient(client),
-		mdbcWatcher:          &mdbcWatcher,
+		mdbcWatcher:          watch.New(),
```
lsierant: **blocking:** we cannot have these watchers in the controller instance. It must be placed in MongoDBSearchReconcilerHelper as the controller is a singleton executed concurrently by the runtime in case we have more than one reconcile at a time configured.

Also the resource watcher here is not made to be thread safe and we should migrate to the one from MEKO packages, which have locks (controllers/operator/watch/resource_watcher.go). 

Also `#2`, I thought we will migrate away from using those resource watchers in favor of using indexes?

controllers/operator/mongodbreplicaset_controller.go
lsierant: What I mean is that we don't copy/mirror it for community, so we don't need it for enterprise. We can just mount it directly. But this will only work if we limit it for the same namespace as we cannot ref secrets across namespaces.

controllers/search_controller/enterprise_search_source_test.go
```diff
@@ -0,0 +1,290 @@
...
+			},
+		},
+	}
+
+	// Set topology directly since it's inlined from DbCommonSpec
+	src.Spec.Topology = topology
+	return src
+}
+
+func TestEnterpriseResourceSearchSource_Validate(t *testing.T) {
```
lsierant, line 52: I like how exhaustive the cases are! 👏

docker/mongodb-kubernetes-tests/tests/common/search/search_tester.py
```diff
@@ -32,7 +32,18 @@ def mongorestore_from_url(self, archive_url: str, ns_include: str, mongodb_tools
                 mongorestore_cmd += " --ssl"
             if ca_path := self.default_opts.get("tlsCAFile"):
                 mongorestore_cmd += " --sslCAFile=" + ca_path
-            process_run_and_check(mongorestore_cmd.split())
+            process_run_and_check(mongorestore_cmd.split(), capture_output=True)
+
+    def assert_search_enabled(self):
+        try:
+            result = self.client.admin.command("getCmdLineOpts")
```
lsierant: is this a reliable test if we connect to the primary in the replicaset mode? We won't check other instances. 

Do you think it would be better to ssh directly into all pods and check autogenerated mongod config there if the parameters are set in each pod (we can use KubernetesTester.run_command_in_pod_container for cat-ing the config probably)

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_basic.py
```diff
@@ -0,0 +1,171 @@
...
+def test_wait_for_database_resource_ready(mdb: MongoDB):
+    mdb.assert_abandons_phase(Phase.Running, timeout=1800)
+    mdb.assert_reaches_phase(Phase.Running, timeout=1800)
+    SearchTester(get_connection_string(mdb, ADMIN_USER_NAME, ADMIN_USER_PASSWORD)).assert_search_enabled()
+
+
+@fixture(scope="function")
+def sample_movies_helper(request, mdb: MongoDB) -> SampleMoviesSearchHelper:
+    credentials = (USER_NAME, USER_PASSWORD)
+    if request.node.get_closest_marker("use_admin_user"):
```
lsierant: what does it do?

scripts/dev/contexts/e2e_static_mdb_kind_ubi_cloudqa
```diff
@@ -17,3 +17,7 @@ export CUSTOM_OM_VERSION
 
 export CUSTOM_MDB_PREV_VERSION=6.0.16
 export CUSTOM_MDB_VERSION=7.0.5
+
+export MDB_SEARCH_COMMUNITY_VERSION="fbd60fb055dd500058edcb45677ea85d19421f47" # Nolan's fixes
```
lsierant: let's extract this to one file similarly like we have 

`source "${script_dir}/variables/mongodb_latest"` in `e2e_mdb_community`
as we have at least three copies of it and it will be a mess soon, especially that we have also this image details specified in values.yaml and release.json. 

Feel free to create a ticket for it if you feel it doesn't belong in this PR.

controllers/operator/mongodbreplicaset_controller.go
lsierant: After reviewing the whole PR I think it's not even necessary as we already have mirroring on the search side in controllers/search_controller/mongodbsearch_reconcile_helper.go:142

controllers/operator/mongodbreplicaset_controller.go
fealebenpae: I'm not sure. If we return false then we won't mirror the keyfile from Ops Manager into the secret that mongot will be waiting for. If customers are using a customized MongoDB version it might either not need the `searchCoordinator` role polyfill, or they might apply it themselves in the `MongoDB` spec, so not returning false if we can't parse the MongoDB version doesn't prevent that very convoluted scenario from actually working with Search.

On the other hand, I can't realistically imagine when we might fail parsing the MongoDB version, so I can't decide if this case is actually worth guarding against.

controllers/operator/mongodbreplicaset_controller.go
fealebenpae: We need to read the keyfile contents from Ops Manager. Unfortunately in Enterprise the only copy of the keyfile is in the automation config in Ops Manager, unlike in Community where the Operator generates the keyfile and stores it in a secret we can just mount in mongot. I considered mirroring the keyfile into a secret inside the search reconciler, but then we'd need to establish the Ops Manager connection there, so I felt the replica set reconciler would be a smaller pain.

FWIW we don't mirror the keyfile on the search side - the `ensureSourceKeyfile` method on the search reconcile helper fails the search resource reconciliation until the replica set controller actually creates the keyfile secret, and applies a hash annotation to mongot's pod template so that if the keyfile gets rotated mongot will be restarted.

controllers/operator/mongodbreplicaset_controller.go
fealebenpae: We currently assume that the two resources are in the same namespace in Community as well, this is mostly a copy of the same code in the Community replicaset reconciler. I believe all we need to support different namespaces is to start mirroring the keyfile and CA secrets in the Search namespace in Community.

controllers/operator/mongodbsearch_controller.go
fealebenpae: So, I tried repurposing field indexes for watches as described in [Kubebuilder's last-version docs](https://book-v3.book.kubebuilder.io/reference/watching-resources/externally-managed) and it works pretty neatly for resource we have a direct relationship against. We can generate reconcile requests for changes in `MongoDBCommunity` resources by checking them against the index.

However, where this breaks down for me is watching related resources like the CA or keyfile secrets since we don't build indexes against them. I was looking into transitive events, as in "if the CA or keyfile secrets change, shouldn't that cause at least a temporary change in the MongoDBCommunity status that will trigger the search controller's mdbc field index watch?" but I didn't have the time to verify that assumption. That's why I fell back to using watches.

Field indexes aside, I'll switch to the thread-safe watcher used in `ReconcileCommonController`, thanks for pointing out the issue.

docker/mongodb-kubernetes-tests/tests/common/search/search_tester.py
fealebenpae: That's a good point, thanks! I'll check `run_command_in_pod_container` out.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_basic.py
fealebenpae: The `test_search_restore_sample_database` step is decorated with the `used_admin_user` marker which causes the `sample_movies_helper` fixture to return a helper configured with the admin user credentials instead of the regular user.

I found out that the `restore` role can be granted on the admin database only (the automation config failed to apply when I requested a user with the `restore` role on the `sample_mflix` database) so I opted to use the admin user instead to restore the database, that's why the restore step needs the admin user credentials instead of the regular user.

lsierant: > I found out that the restore role can be granted on the admin database only (the automation config failed to apply when I requested a user with the restore role on the sample_mflix database) 

This is worrying. Does it mean that Ops Manager has some more AC validations than the agents doing exactly the same thing in a headless mode in community?

lsierant: Also usage of 'get_closest_marker("use_admin_user")' is making my head hurt. I'd like to avoid using such pytest features and try to organize the code in a more explicit way. Can we do it differently? We don't need to use fixtures at all for sample_movies_helper. In any call site we can just execute it with get_sample_movies_helper function directly. 

Also the way we're using marks is to execute whole e2e tests, so adding @mark.use_admin_user might suggest we're adding a whole new e2e test (evergreen task) here.

controllers/operator/mongodbreplicaset_controller.go
lsierant: I believe this behavior should be consistent with community in this regard, i.e. we should never store any secret in AC without having it stored as a secret as well. I've asked this on slack and we have other instances of such behavior. 

Now the question is: 
 - should we add a simple logic of creating a secret when the keyfile is generated (and creating the secret if there is key in AC, but not in k8s) and keep the logic consistent for enterprise and community
 - ignore it for now and fix all those instances of keeping secrets in AC only in on go as part of that effort.

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -408,14 +417,27 @@ func AddReplicaSetController(ctx context.Context, mgr manager.Manager, imageUrls
 			zap.S().Errorf("Failed to watch for vault secret changes: %w", err)
 		}
 	}
+
+	err = c.Watch(source.Kind(mgr.GetCache(), &searchv1.MongoDBSearch{},
+		handler.TypedEnqueueRequestsFromMapFunc(func(ctx context.Context, search *searchv1.MongoDBSearch) []reconcile.Request {
+			source := search.GetMongoDBResourceRef()
+			if source == nil {
+				return []reconcile.Request{}
```
lsierant, line 425: what does it mean? 
I'm asking because ref is optional and we can work on a naming convention as well. Would it work in this case?

lsierant (approved):

lsierant:
evergreen refresh

PR #321, opened 2025-08-07
Fix auth transition on edge-cases

nammn:
# Summary
**Add Not-Ready Handling for Ongoing Auth Transitions**:
This patch refines our readiness logic to correctly reflect the state of authentication transitions. Previously, we treated LastGoalVersionAchieved == GoalVersion as a signal that the cluster was "Running", but this assumption breaks down when auth transitions are still in progress.
This happened because we returned "ready" during a wait step (WaitAuthCanUpdate) — and [we generally return ready for all wait steps](https://github.com/mongodb/mongodb-kubernetes/blob/f0050b8942545701e8cb9e42d54d14f0cb58ee6a/mongodb-community-operator/cmd/readiness/main.go#L139), regardless of whether auth is fully transitioned. Example status:
```
{
  "step": "WaitAuthUpdate",
  "stepDoc": "Wait to update Auth",
  "isWaitStep": true,
  "started": "2025-08-07T14:59:40.213178437Z",
  "attempts": 512,
  "latestAttempt": "2025-08-07T15:09:20.966699961Z",
  "completed": null,
  "result": "wait"
}
```

**Why implemented in the operator and not readinessProbe**:
I didn't fix the readinessProbe but rather  the operator
* if the readinessProbe blocks new nodes are not coming up
* we want new nodes coming up 
* but we also want to block new configurations being applied, which the automation_status check in the
  operator does

**The core idea:**
* Configuration applied ≠ transition fully complete.

**What happened in our tests**:
* we update auth via CR x509 -> scram
* `node-0` completed its auth transition (now uses scram, instead of x509)
* `Config server` hasn't finished its auth transition yet
* We hit a race condition where clusters were marked as "Running" too early and thus continued the rolling restart of `nod e-0` 
* `node-0` restarted with the old X509 config (see below comment from the agent code)
* The X509 process couldn’t access the SCRAM automation user
* Leads to Error: "process...doesn't have the automation user"

- in the mms-automation there is also a comment; that indicates thats they are handling the edge-case if an auth transition was not successful, they start the process with old auth to "finish" it. But this is exactly what causes our race condition
```
	// If a process went down unexpectedly in the middle of an auth transition,
	// we want to restart it with the old auth args.
	// Otherwise, it could be upgraded to the new auth state too soon,
	// and not be able to communicate with other shard members.
```

tl;dr: first `node-0` moved to new auth, `config` not yet, `node-0` restarted and during the restart `config` transitioned to the new auth while `node-0` is again running old auth

## Proof of Work

- auth change tests are passing multiple times in a row: [Link](http://spruce.mongodb.com/version/6894b98218a2e90007437e99/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) - the most flaky auth tests + [Link2](https://spruce.mongodb.com/task/mongodb_kubernetes_e2e_static_mdb_kind_ubi_cloudqa_e2e_sharded_cluster_x509_to_scram_transition_patch_b29fb4ace63eec7102f8f034fd6c553b5d75c1a1_6894c0785c119f0007a58f3c_25_08_07_15_04_26/logs?execution=0) - from the patch

## Checklist


controllers/om/automation_status.go
```diff
@@ -113,17 +138,29 @@ func checkAutomationStatusIsGoal(as *AutomationStatus, relevantProcesses []strin
 	}
 }
 
+// isAuthenticationTransitionMove returns true if the given move is related to authentication transitions
+func isAuthenticationTransitionMove(move string) bool {
```
mircea-cosbuc, line 142: I really like the approach of moving those phase specific checks outside of the readiness probe where we can only afford to treat them with a very wide brush.

controllers/om/automation_status.go
```diff
@@ -113,17 +138,29 @@ func checkAutomationStatusIsGoal(as *AutomationStatus, relevantProcesses []strin
...
+	return ok
+}
+
 func areAnyAgentsInKubeUpgradeMode(as *AutomationStatus, relevantProcesses []string, log *zap.SugaredLogger) bool {
 	for _, p := range as.Processes {
 		if !stringutil.Contains(relevantProcesses, p.Name) {
 			continue
 		}
-		for _, plan := range p.Plan {
+		for _, planStep := range p.Plan {
```
mircea-cosbuc: nit: In the code you introduced, you're ranging over `p.Plan` with `move` and this is refactored to use `planStep` instead. Let's stay consistent (I think we can use `move` since automation uses the same terminology).

mircea-cosbuc (changes requested):
LGTM with a nit

changelog/20250808_fix_fixing_auth_transition_edge_cases.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Fixing auth transition edge-cases
+kind: fix
+date: 2025-08-08
+---
+
+* The agent returns ready if the cluster is ready to accept requests. The operator uses this information to continue operational actions like restarts.
```
viveksinghggits: > The agent returns ready if the cluster is ready to accept requests.

nit: Are there are chances that user might get confused if it's K8s cluster or MongoDB cluster that we are talking about? Should we explicit and mention which cluster this is?

changelog/20250808_fix_fixing_auth_transition_edge_cases.md
```diff
@@ -0,0 +1,8 @@
...
+* This can be problematic during auth transitions. We can have a period where we invalidate one auth while the other is not activated yet and we try to use the not supported one
```
viveksinghggits: > This can be problematic during auth transitions.

By this, are we talking about the thing that is discussed in the first point? If yes, should be just have one point to explain it well, seems both the points are related closely to each other.

viveksinghggits:
```suggestion
* This can be problematic during auth transitions. We can have a period where we invalidate one auth while the other is not activated yet and we try to use the not supported one.
```
?

controllers/om/automation_status.go
```diff
@@ -85,12 +85,25 @@ func checkAutomationStatusIsGoal(as *AutomationStatus, relevantProcesses []strin
 
 	goalsNotAchievedMap := map[string]int{}
 	goalsAchievedMap := map[string]int{}
+	authTransitionInProgress := map[string]string{}
```
viveksinghggits: nit:
```suggestion
	authTransitionsInProgress := map[string]string{}
```
?

controllers/om/automation_status.go
```diff
@@ -85,12 +85,25 @@ func checkAutomationStatusIsGoal(as *AutomationStatus, relevantProcesses []strin
...
 		if !stringutil.Contains(relevantProcesses, p.Name) {
 			continue
 		}
 		if p.LastGoalVersionAchieved == as.GoalVersion {
 			goalsAchievedMap[p.Name] = p.LastGoalVersionAchieved
+
+			// Check if authentication transitions are in the current plan
+			// If a process has reached goal version but still has auth-related moves in plan,
+			// it means authentication transition is likely in progress
+			// The plan contains non-completed move names from the API
```
viveksinghggits:
```suggestion
			// Check if authentication transitions are in the current plan.
			// If a process has reached goal version but still has auth-related moves in plan,
			// it means authentication transition is likely in progress.
			// The plan contains non-completed move names from the API.
```

controllers/om/automation_status.go
```diff
@@ -113,17 +138,29 @@ func checkAutomationStatusIsGoal(as *AutomationStatus, relevantProcesses []strin
 	}
 }
 
+// isAuthenticationTransitionMove returns true if the given move is related to authentication transitions
+func isAuthenticationTransitionMove(move string) bool {
+	authMoves := map[string]struct{}{
+		"UpdateAuth":     {},
+		"WaitAuthUpdate": {},
+	}
```
viveksinghggits, line 146: nit: should this map be defined outside of the function?

controllers/om/automation_status.go
viveksinghggits: sorry, when I add `?`, I mean what do you think about this suggestion.

controllers/om/automation_status.go
nammn: hmmm usually yes, but since its tightly coupled to auth we can leave it there. If we find other waitMoves which we should return unready -> lets move it then, ok?

controllers/om/automation_status.go
nammn: added!

changelog/20250808_fix_fixing_auth_transition_edge_cases.md
nammn: fully reworded! PTAL

controllers/om/automation_status.go
nammn: renamed!

controllers/om/automation_status.go
viveksinghggits: sure, thank you.

changelog/20250808_fix_fixing_auth_transition_edge_cases.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Fixing auth transition edge-cases
+kind: fix
+date: 2025-08-08
+---
+
+* The readinessProbe always returns ready if the agent is in a wait step. This can be problematic during auth transitions as we can have a period where we invalidate one auth while the other is not activated yet and we try to use the not supported one.
```
viveksinghggits: > a period where we invalidate one auth while the other is not activated yet and we try to use the not supported one.

We seem to have auth in two states, invalidated and inactive. But at the end the note says, we try to use the not supported one? What is not supported one? Is it the invalidated one or the inactive one? 
Sorry, it might be just me how is not able to understand this.

The other point that I have is, usually we try to convey what exactly what improved/fixed/changed using the release note. But after reading this note, I am not able to understand what exactly was changed. Was an issue fixed, or a feature added? Or something else?

nammn: as discussed offline updated

mircea-cosbuc (approved):

PR #317, opened 2025-08-06
CLOUDP-295785 - using build_info.json in new atomic pipeline

MaciejKaras:
# Summary

This PR focuses on using `build_info.json` information in `atomic_pipeline.py` code. The logic is mostly hidden in `image_build_config_from_args` function. It is binding the parts from:
- https://github.com/mongodb/mongodb-kubernetes/pull/303 - atomic pipeline for building images
- https://github.com/mongodb/mongodb-kubernetes/pull/307 - `build_info.json` with information for building images

[Section in Atomic Releases TD](https://docs.google.com/document/d/1eJ8iKsI0libbpcJakGjxcPfbrTn8lmcZDbQH1UqMR_g/edit?tab=t.p76ry15gwmkk#heading=h.23xqxc1qypvb) that describes the `build_info.json` and how it is used in the image building process.

- Added signing configuration in `build_info.json`
- Added missing `meko-tests` and `mco-tests` images in `build_info.json`
- Added configuration for `agent` and `ops-manager` images in `build_info.json`
- Moved `images_signing.py` from `scripts/evergreen/release`
- Simplifications in the existing code

## Proof of Work

CI tests passing + updated unit tests

## Checklist

- [ ] Have you linked a jira ticket and/or is the ticket in the title?
- [ ] Have you checked whether your jira ticket required DOCSP changes?
- [ ] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

PR #331, opened 2025-08-11
Fix master builds

Julien-Ben:
# Summary

Master patches are broken because they build with `latest` tag only, while the patch still need the `patch_id` tagged images.
This PR fixes this issue to get master green again, but remove builds to `latest`. They will be added again in https://github.com/mongodb/mongodb-kubernetes/pull/317, using build_info.

## Proof of Work
Unfortunately, we cannot master patches before merging.

### DIFF
--- a/scripts/release/build_context.py
+++ b/scripts/release/build_context.py
@@ -71,7 +71,7 @@ class BuildContext:
             return self.git_tag
         if self.scenario == BuildScenario.STAGING:
             # On master merges, always use "latest" (preserving legacy behavior)
-            return "latest"
+            return self.patch_id
         if self.patch_id:
             return self.patch_id
         # Alternatively, we can fail here if no ID is explicitly defined


MaciejKaras (approved):
LGTM

scripts/release/build_context.py
```diff
@@ -71,10 +86,12 @@ def get_version(self) -> str:
             return self.git_tag
         if self.scenario == BuildScenario.STAGING:
             # On master merges, always use "latest" (preserving legacy behavior)
-            return "latest"
+            return self.patch_id
```
nammn, line 91: The comment is not valid anymore, right?

Julien-Ben: That's correct, good catch
Fixed

MaciejKaras: The whole version calculation based on variables will be soon extracted outside of `build_context` and `pipeline` in general. It is not a proper place and it hides a lot of details

PR #324, opened 2025-08-08
CLOUDP-336989: Bump Ops Manager Container Image version to 7.0.17

mms-build-account:
_Opened by Private Cloud Tools (PCT)_.

# Ticket
[CLOUDP-336989](https://jira.mongodb.org/browse/CLOUDP-336989)

# Description
Bump Ops Manager container image version to 7.0.17.

# Reviewer Checklist

Before merging this PR, verify the following:
- [ ] the following tasks are passing in Evergreen:
  - `publish_ops_manager` task (variant: `publish_om70_images`)
- [ ] the `agent_version` was updated correctly
- [ ] the `tools_version` was updated correctly

PR #332, opened 2025-08-11
Add change log for unavailability of OM container images

viveksinghggits:
# Summary

Some of the versions of ops manager container images are not built/published by us, because of an issue that we have been facing while updating the OM version in our project.
Those issues are fixed and like suggested [here](https://github.com/mongodb/mongodb-kubernetes/pull/324#pullrequestreview-3105253213) this PR adds change log to inform users that some version of this image are unavailable.

## Proof of Work

NA


changelog/20250811_other_container_images_for_opsmanager_versions_7016_800.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Some versions of container images for OpsManager are not published
+kind: other
+date: 2025-08-11
+---
+
+* Container images for OpsManager versions `7.0.16`, `8.0.0`, `8.0.9` and `8.0.10` are not published because of a technical issue in encountered in our build process.
```
MaciejKaras: typo:

```
technical issue in encountered in our build process
```

viveksinghggits: aah, sorry.

MaciejKaras (approved):
LGTM

changelog/20250811_other_container_images_for_opsmanager_versions_7016_800.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Some versions of container images for OpsManager are not published
+kind: other
+date: 2025-08-11
+---
+
+* We have deliberately not published the container images for OpsManager versions `7.0.16`, `8.0.0`, `8.0.9` and `8.0.10` due to technical issues with our build process and because of this customers will not be able to update their OpsManager installations to these versions.
```
mircea-cosbuc: The reason is not our build process, but a bug in OM that would block customers from upgrading to those versions.
It's also `8.0.8` not `8.0.0`.

mircea-cosbuc (changes requested):

changelog/20250811_other_container_images_for_opsmanager_versions_7016_800.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Some versions of container images for OpsManager are not published
```
anandsyncs:
```suggestion
title: Ops Manager Container Images Not Published Due to Product Bug
```

viveksinghggits: It was not really a product bug @anandsyncs. It was an issue in the OM. But I don't think we have to into that much details. I can change it to title case though but I don't see title case being used in other changelogs.

changelog/20250811_other_container_images_for_opsmanager_versions_7016_800.md
viveksinghggits: > The reason is not our build process, but a bug in OM that would block customers from upgrading to those versions.

Right, I remember this. But I knowingly just mentioned `technical issue in our build process`, because I thought it's not really necessary to do into that much details. But I can add that if that's a good idea.

Sorry about the typo. I will push the fix.

mircea-cosbuc: Unfortunately we need to raise this and document this, because customers could explicitly ask why we failed to publish when those releases for OM are published on other platforms. Those versions are incompatible with the operator, this is why we knowingly did not publish them.

changelog/20250811_other_container_images_for_opsmanager_versions_7016_800.md
```diff
@@ -0,0 +1,7 @@
+---
+title: OpsManager container images not published due to a product bug
+kind: other
+date: 2025-08-11
+---
+
+* We have deliberately not published the container images for OpsManager versions `7.0.16`, `8.0.8`, `8.0.9` and `8.0.10` due to a bug in the OpsManager and because of that customers will not be able to update their OpsManager installations to these versions.
```
mircea-cosbuc:
```suggestion
* We have deliberately not published the container images for OpsManager versions `7.0.16`, `8.0.8`, `8.0.9` and `8.0.10` due to a bug in the OpsManager which prevents MCK customers to upgrade their OpsManager deployments to those versions.
```

PR #336, opened 2025-08-11
CLOUDP-295785 - staging support for image building

MaciejKaras:
# Summary

PR integrates `staging` build scenario with `atomic_pipeline.py` and the e2e tests. `staging` repositories are created in AWS ECR under `/staging` dir. List of new `staging/` repositories:

Screenshot 2025-08-13 at 14 39 55

Example AWS page with details:

Screenshot 2025-08-11 at 21 26 51

Previously all of the evergreen patch builds were targeting `268558157000.dkr.ecr.us-east-1.amazonaws.com/dev` repository. Now, depending on the `BUILD_SCENARIO` environment variable images will be pushed to and pulled from
different registry:
 - PATCH_REGISTRY="268558157000.dkr.ecr.us-east-1.amazonaws.com/dev"
 - STAGING_REGISTRY="268558157000.dkr.ecr.us-east-1.amazonaws.com/staging"
 - RELEASE_REGISTRY="quay.io/mongodb"

Additional changes:
 - re-added missing `latest` tag for staging builds. This is used for local testing
 - moved all registry variables declarations to `root-context`. Previously they were duplicated in both `scripts/dev/contexts/evg-private-context` and `scripts/dev/contexts/local-defaults-context`
 - fixes multi arch smoke tests configuration. We can now run them also for released images (tbd)
 - started to use `OPERATOR_VERSION` instead of `VERSION_ID`. `VERSION_ID` was only related to patch builds, which did not apply for staging and release builds
 - BASE_REPO_URL_SHARED is now `268558157000.dkr.ecr.us-east-1.amazonaws.com/staging` because all latest tags are pushed on master builds
 - `build_scenario` now needs to be passed explicitly to `pipeline.py` as well as `version` (apart from `agent` image)

## Proof of Work

Building all images is passing for staging scenario -> https://spruce.mongodb.com/version/68c7ce53672666000716401a/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Building all images is passing for release scenario (apart from kubectl which requires tag) -> https://spruce.mongodb.com/version/68c911ef8d38f100074fa2db/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Example image stored together with signature -> https://us-east-1.console.aws.amazon.com/ecr/repositories/private/268558157000/staging/mongodb-kubernetes-init-database?region=us-east-1

## Remaining work

- [ ] Add guide on how to run `development` scenario for building the images using evergreen and directly `pipeline.py`
- [ ] Fix `dummy` version in the `release_info.json` - this is already planned in [here](https://docs.google.com/document/d/1eJ8iKsI0libbpcJakGjxcPfbrTn8lmcZDbQH1UqMR_g/edit?tab=t.p76ry15gwmkk#bookmark=id.egtg1p95d1)

## Checklist


scripts/release/build/build_scenario.py
```diff
@@ -28,7 +28,7 @@ def infer_scenario_from_environment(cls) -> "BuildScenario":
             scenario = BuildScenario.RELEASE
             logger.info(f"Build scenario: {scenario} (git_tag: {git_tag})")
         elif is_patch or is_evg:
-            scenario = BuildScenario.PATCH
+            scenario = BuildScenario.STAGING
```
MaciejKaras: this is just for testing purpose, remove when finished

scripts/dev/contexts/evg-private-context
```diff
@@ -33,7 +33,16 @@ else
 fi
 export NAMESPACE
 
-export BASE_REPO_URL="268558157000.dkr.ecr.us-east-1.amazonaws.com/dev"
+DEV_REPO_URL="268558157000.dkr.ecr.us-east-1.amazonaws.com/dev"
+STAGING_REPO_URL="268558157000.dkr.ecr.us-east-1.amazonaws.com/staging"
+
+if [ "${is_patch:-false}" = "true" ]; then
+  echo "is_patch is set, setting BASE_REPO_URL=${STAGING_REPO_URL}"
+  export BASE_REPO_URL="${STAGING_REPO_URL}"
```
MaciejKaras: change to DEV_REPO_URL before merging

MaciejKaras: and remove override_version_id

scripts/dev/contexts/variables/om60
```diff
@@ -13,10 +13,10 @@ export CUSTOM_OM_VERSION
 export CUSTOM_MDB_VERSION=6.0.21
 export CUSTOM_MDB_PREV_VERSION=5.0.7
 
-export AGENT_VERSION=12.0.33.7866-1
+export AGENT_VERSION=12.0.34.7888-1
```
MaciejKaras: this had to change, because the previously used agent version does not appear in `release.json` and there is no way to rebuild it.

scripts/dev/contexts/local-defaults-context
MaciejKaras, line 1: we should later change `BASE_REPO_URL_SHARED` to use `268558157000.dkr.ecr.us-east-1.amazonaws.com/staging` when we have staging images already

scripts/release/release_info_test.py
```diff
@@ -15,37 +15,37 @@ def test_create_release_info_json(
         "images": {
             "operator": {
                 "repository": "quay.io/mongodb/mongodb-kubernetes",
-                "platforms": ["linux/arm64", "linux/amd64"],
```
nammn: why this change?

scripts/release/build/build_scenario.py
nammn: when building something new we should keep testability in mind - I think if we need to hardcode this - this means we forgot something as part of the planning

scripts/dev/contexts/evg-private-context
```diff
@@ -33,7 +33,19 @@ else
 fi
 export NAMESPACE
 
-export BASE_REPO_URL="268558157000.dkr.ecr.us-east-1.amazonaws.com/dev"
+DEV_REPO_URL="268558157000.dkr.ecr.us-east-1.amazonaws.com/dev"
+STAGING_REPO_URL="268558157000.dkr.ecr.us-east-1.amazonaws.com/staging"
+COMMIT_SHA_SHORT=$(git rev-parse --short=8 HEAD)
+
+if [ "${is_patch:-false}" = "true" ]; then
```
nammn: we should have a way to test this change without doing these hacks to verify it. Otherwise, we will run into the same problem like the last pipeline

build_info.json
```diff
@@ -12,15 +12,13 @@
         "sign": true,
         "repository": "268558157000.dkr.ecr.us-east-1.amazonaws.com/staging/mongodb-kubernetes",
         "platforms": [
-          "linux/arm64",
```
mircea-cosbuc, line 15: q: Why no arm64 images on staging? Doesn't this conflict with multi-arch tests? cc @nammn

scripts/dev/contexts/local-defaults-context
mircea-cosbuc: q: When would this be and how would we trigger the change then?

scripts/release/build/build_scenario.py
mircea-cosbuc: q: The PR is ready for review. Are those comments meant to be addressed before merging?

build_info.json
```diff
@@ -12,15 +12,13 @@
         "sign": true,
         "repository": "268558157000.dkr.ecr.us-east-1.amazonaws.com/staging/mongodb-kubernetes",
```
mircea-cosbuc: q: It seems like this was already configured. What specifically in this PR enables staging support?

scripts/dev/contexts/local-defaults-context
Julien-Ben: q: Are we going to use staging registry the same way we were using ECR `/dev` before ? 
Does it mean we will push to staging with both tags: the commit sha and `latest` ?

(just for my understanding, but I think it makes sense)

scripts/release/atomic_pipeline.py
```diff
@@ -49,14 +49,12 @@ def build_image(
     span.set_attribute("mck.platforms", build_configuration.platforms)
 
     # Build docker registry URI and call build_image
-    image_full_uri = f"{build_configuration.registry}:{build_configuration.version}"
-
-    logger.info(
-        f"Building {image_full_uri} for platforms={build_configuration.platforms}, dockerfile args: {build_args}"
-    )
+    tags = [f"{build_configuration.registry}:{build_configuration.version}"]
+    if build_configuration.latest_tag:
```
Julien-Ben: Is it possible to handle tags as a list in the build configuration instead of treating `latest` as a special case ?

MaciejKaras: for latests yes, but not for the `olm` tag where we will have a dynamic timestamp suffix. Because of this I've chose the bool rather than array

scripts/dev/contexts/local-defaults-context
MaciejKaras: As I understand, please correct me if I'm wrong, `latest` images from `/dev` were used for local testing, so we can reuse most recent images. Previously the `latest` tag was added when the image was built for master commits. To preserve existing behaviour we are going to add `latest` tag also when we merge to master (staging build), but those images will land now in `/staging` repositories.

Because of this we will need to change our `scripts/dev/contexts/local-defaults-context` to use `\staging` base repository instead.

scripts/release/build/build_scenario.py
```diff
@@ -27,13 +27,12 @@ def infer_scenario_from_environment(cls) -> "BuildScenario":
...
             logger.info(f"Build scenario: {scenario} (git_tag: {git_tag})")
-        elif is_patch or is_evg:
-            scenario = BuildScenario.PATCH
+        elif is_patch and is_evg:
+            scenario = BuildScenario.STAGING
+            logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
+        elif is_evg:
+            scenario = BuildScenario.STAGING
             logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
-        # TODO: Uncomment the following lines when starting to work on staging builds
```
MaciejKaras, line 36: uncommenting this code makes `BuildScenario.STAGING` being actually used

build_info.json
MaciejKaras: It was configured but unused. Please check -> https://github.com/mongodb/mongodb-kubernetes/pull/336/files#r2273696424

scripts/release/build/build_scenario.py
MaciejKaras: The change was only made to test the PR. We need to revet this before merging

build_info.json
MaciejKaras: I don't know if "linux/arm64" will work for staging. I will leave this to @nammn to add when he will provide his PR with multi-arch support.

scripts/dev/contexts/local-defaults-context
Julien-Ben: Sounds good to me ! Thanks for clarifying :)

Julien-Ben: (I'll let @mircea-cosbuc resolve the thread as he commented as well)

build_info.json
nammn: why shouldn't it work? atomic_pipeline just reads platform and passes it as it is to docker whale?

scripts/release/release_info_test.py
MaciejKaras: as discussed in the meeting I didn't want to test the multi-arch in this PR

scripts/release/build/build_scenario.py
MaciejKaras: agree! Reverted the change and we can think how to test this properly

scripts/dev/contexts/evg-private-context
MaciejKaras: agree! Reverted the change and we can think how to test this properly

build_info.json
MaciejKaras: as discussed in the meeting I didn't want to face any possible issues for the multi-arch in this PR, that's why I have removed it

.evergreen-release.yml
```diff
@@ -13,7 +13,8 @@ tasks:
       - func: pipeline
         vars:
           image_name: meko-tests
-          build_scenario: --build-scenario patch
+          BUILD_SCENARIO: patch #TODO: we should use standard release scenario here and just push to staging or dev repo bases on build_info.json
```
MaciejKaras: this will be solved in future PR

scripts/dev/contexts/variables/e2e_smoke_base
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -Eeou pipefail
+
+if [[ "${BUILD_SCENARIO}" == "release" ]]; then
+  export MEKO_TESTS_REGISTRY="268558157000.dkr.ecr.us-east-1.amazonaws.com/dev"
```
nammn, line 6: shouldn't it always just use this registry?

scripts/release/pipeline_ops_manager.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
nammn, line 1: does it make sense to consider merging those bash scripts and/or give them the option to use params? Something like script as pipelien_auxilliary_images or so ?

scripts/release/release_info.py
```diff
@@ -30,17 +30,12 @@
     UPGRADE_HOOK_IMAGE,
 ]
 
+# TODO: this is dummy version, to be replaced with actual versioning logic
```
nammn: when? is this tracked?

scripts/release/retag_images_to_staging.sh
```diff
@@ -0,0 +1,97 @@
...
+    local manifest_json
+    manifest_json=$(docker buildx imagetools inspect --raw "${image}" 2>/dev/null || echo '{}')
+
+    local media_type
+    media_type=$(echo "${manifest_json}" | jq -r '.mediaType // empty')
+
+    if [[ "${media_type}" == *"manifest.list"* ]]; then
+        # this is a multi-arch manifest
+        local arch_digest
+        arch_digest=$(echo "${manifest_json}" | jq -r ".manifests[] | select(.platform.architecture == \"${arch}\" and .platform.os == \"linux\") | .digest")
```
nammn, line 31: do we need the linux and check?

scripts/release/retag_images_to_staging.sh
```diff
@@ -0,0 +1,97 @@
...
+
+        if [[ -n "${arch_digest}" && "${arch_digest}" != "null" ]]; then
+            echo "${arch_digest}"
+            return 0
+        fi
+    elif [[ "${arch}" == "amd64" ]]; then
```
nammn, line 37: if its not a list; just return the digest, why check for arch here?

scripts/release/retag_images_to_staging.sh
```diff
@@ -0,0 +1,97 @@
...
+
+process_image() {
+    local source_image="$1"
+    local target_image="$2"
+
+    echo "  Processing ${source_image}..."
+
+    local digest_arm64
+    local digest_amd64
+    digest_arm64=$(get_arch_digest "${source_image}" arm64)
```
nammn, line 56: should we also consider other archs? Maybe its worth to just iterate through the list and retag that list?

scripts/release/retag_images_to_staging.sh
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env bash
+
+set -Eeou pipefail
+
+# Utility used to retag and push container images from one repo to another
+# Useful for migrating images between different repositories (e.g. dev -> staging)
```
nammn, line 7: is this only manually used? Why not rely on atomic_pipeline to have that functionality?

EVERGREEN.md
```diff
@@ -0,0 +1,43 @@
+# Evergreen CI/CD Configuration Guide
```
nammn, line 1: I believe we will have all the documantion on our ci/cd in a google doc (?) if yes - we should link to that one instead. Otherwise - we will scatter things around a lot of places and things will go even more stale

scripts/dev/contexts/variables/e2e_smoke_base
MaciejKaras: The logic of building and pushing the `meko-tests` image is hidden in `build_info.json` file. It is a little bit different than for other images. When running smoke tests on released artifacts we need to build `meko-tests` image. I didn't want to push it to quay, because it serves no use for our customers, so I decided to push it to `268558157000.dkr.ecr.us-east-1.amazonaws.com/dev` on release builds. This is configured over here -> https://github.com/mongodb/mongodb-kubernetes/blob/dda39623beb942cf590913554d7983e69e235f9a/build_info.json#L182-L187

So for running e2e smoke tests in release scenario we need to pull from ecr dev instead of quay.io. Maybe I should add a comment about that in this file.

scripts/release/release_info.py
MaciejKaras: Yes it is -> https://docs.google.com/document/d/1eJ8iKsI0libbpcJakGjxcPfbrTn8lmcZDbQH1UqMR_g/edit?tab=t.p76ry15gwmkk#bookmark=id.egtg1p95d1. Right now the `release_info.json` file is not used nor generated.

scripts/release/retag_images_to_staging.sh
MaciejKaras: TBH I have copied this logic from @lsierant script and used it to upload ops-manager images to staging. It worked, but I haven't get into the details.

scripts/release/retag_images_to_staging.sh
MaciejKaras: I didn't want to create some automation for single shot script. This won't be needed in the near future.

I can also delete the file.

EVERGREEN.md
MaciejKaras: I thought about having all CI/CD documentation in the codebase tbh. That is why I have started this doc.

scripts/release/build/build_scenario.py
```diff
@@ -1,69 +1,9 @@
...
 
 class BuildScenario(StrEnum):
     RELEASE = "release"  # Official release triggered by a git tag
     MANUAL_RELEASE = "manual_release"  # Manual release, not part of operator release cycle
     PATCH = "patch"  # CI build for a patch/pull request
     STAGING = "staging"  # CI build from a merge to the master
     DEVELOPMENT = "development"  # Local build on a developer machine
-
-    @classmethod
-    def infer_scenario_from_environment(cls) -> "BuildScenario":
```
Julien-Ben, line 21: Have we moved this somewhere ? Or do we drop scenario inference ?

Julien-Ben: After reading further, it was moved to `scripts/dev/contexts/evg-private-context`, is that correct ?
Is the development scenario chosen by default locally ?

MaciejKaras: `BuildScenario` inference from env vars was a code smell, because the env vars were not visible in the script inputs. Right now you have to specify the `BuildScenario` explicitly whenever you want to build the image locally. You can use `development` of course.

If you rather want evergreen to build the images for you, then the `BuildScenario` is calculated based on evergreen env variables in `scripts/dev/contexts/evg-private-context` as you noticed.

scripts/release/retag_images_to_staging.sh
nammn: ahhhh single shot script - maybe mark it accordingly. I am fine with keeping it, but we should clearly note that it is not used in any of our ci automation. Then good to resolve from my side

EVERGREEN.md
nammn: problem is that ci/cd (evg) is only viewable and available internally. Not much external contributors can rely on + we will have release related things in google docs (right?) and this here - I think this will make things more difficult. I rather have that md file really short and just link to the google doc you are writing/have prepared

.evergreen.yml
```diff
@@ -190,20 +192,6 @@ parameters:
     value: "false"
     description: set this to true if you would like the pipeline to automatically push a branch with updated snippets outputs
 
-# Each variant needs to be tagged with one or more tags referencing related build scenario:
```
mircea-cosbuc, line 198: Is this comment not relevant anymore?

scripts/dev/contexts/e2e_smoke_ibm_power
```diff
@@ -6,9 +6,8 @@ script_name=$(readlink -f "${BASH_SOURCE[0]}")
 script_dir=$(dirname "${script_name}")
 
 source "${script_dir}/root-context"
+source "${script_dir}/variables/e2e_smoke_base"
 source "${script_dir}/variables/e2e_ibm_smoke_base"
 
-# This is required to be able to rebuild the om image and use that image which has been rebuilt
```
mircea-cosbuc, line 11: Is this not required anymore?

scripts/dev/contexts/root-context
```diff
@@ -39,27 +39,57 @@ export OPERATOR_ENV=${OPERATOR_ENV:-"dev"}
...
-export INIT_APPDB_VERSION="${VERSION_ID}"
-export INIT_DATABASE_VERSION="${VERSION_ID}"
-export INIT_OPS_MANAGER_VERSION="${VERSION_ID}"
-export DATABASE_VERSION="${VERSION_ID}"
+export OPERATOR_VERSION="${OPERATOR_VERSION}"
+export INIT_APPDB_VERSION="${INIT_APPDB_VERSION:-${OPERATOR_VERSION}}"
+export INIT_DATABASE_VERSION="${INIT_DATABASE_VERSION:-${OPERATOR_VERSION}}"
+export INIT_OPS_MANAGER_VERSION="${INIT_OPS_MANAGER_VERSION:-${OPERATOR_VERSION}}"
+export DATABASE_VERSION="${DATABASE_VERSION:-${OPERATOR_VERSION}}"
+export READINESS_PROBE_VERSION="${READINESS_PROBE_VERSION:-${READINESS_PROBE_VERSION}}"
```
mircea-cosbuc: Is this intentional? We shouldn't have a default here, right?

scripts/release/atomic_pipeline.py
```diff
@@ -236,13 +235,7 @@ def build_init_om_image(build_configuration: ImageBuildConfiguration):
 
 
 def build_om_image(build_configuration: ImageBuildConfiguration):
-    # Make this a parameter for the Evergreen build
```
mircea-cosbuc, line 239: Just to understand, this PR also addresses making OM_VERSION a parameter for the build.

scripts/release/release_info.py
mircea-cosbuc: Why do we need it in master now?

scripts/release/retag_images_to_staging.sh
```diff
@@ -0,0 +1,97 @@
...
+
+    if [[ -n "${digest_amd64}" ]]; then
+      docker pull "${source_image}@${digest_amd64}"
+      docker tag "${source_image}@${digest_amd64}" "${target_image}-amd64"
+      docker push "${target_image}-amd64"
+    fi
+
+    if [[ -n "${digest_arm64}" ]]; then
+      docker pull "${source_image}@${digest_arm64}"
+      docker tag "${source_image}@${digest_arm64}" "${target_image}-arm64"
```
mircea-cosbuc, line 67: Why not `skopeo copy` ?

scripts/dev/contexts/variables/e2e_smoke_base
MaciejKaras: @nammn let me know if that answers your question

scripts/release/pipeline_ops_manager.sh
MaciejKaras: I will try to unify bash scripts, but in next PR. Is that acceptable? 

I've added this to project tasks -> https://docs.google.com/document/d/1eJ8iKsI0libbpcJakGjxcPfbrTn8lmcZDbQH1UqMR_g/edit?tab=t.45ig7xr3e3w4#bookmark=id.cn8t47bpc4c8

scripts/release/release_info.py
MaciejKaras: @nammn It was added to master months ago when I was still working on the promotion part of the TD. I had to switch my priorities and help with the image building process, that is why it is still not used and on master.

EVERGREEN.md
MaciejKaras: The release guide already [exists](https://docs.google.com/document/d/1QWpYu-YOrxwLgyHhSKQCAf6byJp2M6cq7C1FBxr3rbE/edit?tab=t.jrkme3tf79aa#heading=h.78zktspfbmi7), but is has less to do with evergreen CI/CD documentation. This `EVERGREEN.md` files describes the structure, aliases and usage of evergreen in our project. I really think it is beneficial to keep it close to code, so important changes in the Evergreen configuration/usage can be easily documented.

scripts/release/build/build_scenario.py
MaciejKaras: We have agreed to make `Development` scenario a default one. I will also document how to use new `pipeline.py` locally to build container images. Added this as a task in https://docs.google.com/document/d/1eJ8iKsI0libbpcJakGjxcPfbrTn8lmcZDbQH1UqMR_g/edit?tab=t.45ig7xr3e3w4#bookmark=id.7li492lbc7ux

Let me know if that is enough to resolve the thread.

.evergreen.yml
MaciejKaras: This was moved to `EVERGREEN.md` file

scripts/dev/contexts/e2e_smoke_ibm_power
MaciejKaras: no, because the `OPS_MANAGER_REGISTRY` is properly specified in `root-context` now

scripts/dev/contexts/root-context
MaciejKaras: Good catch, this is unnecessary. 

I've fixed it in the next PR as well -> https://github.com/mongodb/mongodb-kubernetes/blob/15d63da5c59a54eb379eee83de4e8233a5656a91/scripts/dev/contexts/root-context#L61-L62

scripts/release/atomic_pipeline.py
MaciejKaras: yes, exactly

scripts/release/retag_images_to_staging.sh
MaciejKaras: Good question. I didn't, but Łukasz did and he experienced issues trying skopeo and crane as well.

I've tested now that using `skopeo copy --preserve-digests --all ` works. I will try to use it in the promotion pipeline later. After that this script will not be needed anymore

scripts/release/retag_images_to_staging.sh
MaciejKaras: Added comment mentioning that it is not used in CI/CD

scripts/release/retag_images_to_staging.sh
MaciejKaras: Added comment mentioning that it is not used in CI/CD

scripts/release/retag_images_to_staging.sh
MaciejKaras: Added comment mentioning that it is not used in CI/CD

scripts/release/retag_images_to_staging.sh
MaciejKaras: Added comment mentioning that it is not used in CI/CD

scripts/release/release_info.py
nammn: oh - i meant, if we keep this todo - let's link to the ticket/google doc so we know when this is done and we can remove it.
So other folks outside of the initial writer knows when to remove it

EVERGREEN.md
nammn: I think there was a discussion with @viveksinghggits on where to put things. The content is great, but I want us to be consistent with what everyone decided

MaciejKaras: I think that we agreed to link the `EVERGREEN.md` in the WiKi ToC and in related documents.

scripts/release/build/build_scenario.py
Julien-Ben: I'm okay to resolve. Maybe we should also track these follow ups in a Jira ticket ?

MaciejKaras: For bugs and help tickets it makes sense, but for projects it's additional burden to duplicate what we have in the doc. I wouldn't go that way, it's unnecessary bureaucracy

EVERGREEN.md
viveksinghggits, line 1: Hi @MaciejKaras ,
Are you doing to add the section that we talked about, about how to load build info for a particular scenario? For example if I want to do something similar to 

```python
build_scenario = BuildScenario.infer_scenario_from_environment()
kubectl_plugin_build_info = load_build_info(build_scenario).binaries[KUBECTL_PLUGIN_BINARY_NAME]
```
How would I do that now?

MaciejKaras: You should expose the `build_scenario` from your script and user should provide it. In `evg-private-context` there is `BUILD_SCENARIO` env variable set that you can use when calling the script in evg task. For local runs you can make it a default `Development` value

EVERGREEN.md
viveksinghggits, line 1: This file name seems too generic to me. It looks like this file has complete information about evergreen but that's not the case right? If I understood things clearly, we are just explaining how to add a new variant in the evergreen file.
If that is the case should we change the title of the file (and maybe file name as well), to `Adding a new variant in evergreen file`?

EVERGREEN.md
viveksinghggits: Ok, makes sense. So we will have to two evg variants one will be run in PR patches and one will be run in staging. If I am using the same script in these variants, I can make my script accept the build_scenario. Which can even be hardcoded in the variants. Please correct me if I am wrong.
Also, are we going to document this? Or you documenting it doesn't make much sense.

scripts/dev/contexts/variables/e2e_smoke_base
nammn: it answers, i wonder whether its not "cleaner" to just push to quay, it might not be useful for customers but it will reduce the complexity a bit. WDYT?

EVERGREEN.md
nammn: we could consider to use a script that just returns the env vars for build scenarios. `evg-private-context` can just and any other user the same way. It follow my suggestion in the call to extract that logic out to be re-usable

scripts/dev/contexts/variables/e2e_smoke_base
MaciejKaras: I think so as well, good point.

@mircea-cosbuc what do you think about creating quay.io repository for the `mongodb-kubernetes-tests` images that is needed during release process (smoke and olm tests)?

EVERGREEN.md
MaciejKaras: >So we will have to two evg variants one will be run in PR patches and one will be run in staging. If I am using the same script in these variants, I can make my script accept the build_scenario. Which can even be hardcoded in the variants. Please correct me if I am wrong.

@viveksinghggits You don't need separate variants for PR patches and staging builds. Just call the `switch_context` evg function and it will create `.generated` context files with the `BUILD_SCENARIO` environment variable available for you. This is what we do in all other tasks, including e2e tests or image building tasks. I believe `clone` task already sets up context files so you just need to read it in your script.

MaciejKaras: >we could consider to use a script that just returns the env vars for build scenarios. evg-private-context can just and any other user the same way. It follow my suggestion in the call to extract that logic out to be re-usable

@nammn We could extract that logic, I agree. But still we need `BUILD_SCENARIO` to be provided and it's currently calculated in `evg-private-context` and it only makes sense to calculate it there. The evg variables that are used to calculate `BUILD_SCENARIO` are not present locally.

EVERGREEN.md
MaciejKaras: It's more about our Evergreen setup, usage etc. Adding new variant is only single topic of this guide. I also discuss running patches representing specific build scenarios and there will be more things added in the future.

scripts/dev/contexts/variables/e2e_smoke_base
mircea-cosbuc: I think there's no problem in creating a separate quay.io repository for this. 👍

scripts/release/retag_images_to_staging.sh
mircea-cosbuc: Got it, this should be mandatory for the promotion process (or an equivalent that preserves digests).

mircea-cosbuc: Do we still need this script as an intermediary step to the promotion pipeline? I assume yes, just making sure.

MaciejKaras: This script is not used in promoting the images, just for initial supply of the OM images in staging repo. I've added a comment on the top of the script:

```bash
# Utility used to retag and push container images from one repo to another
# Useful for migrating images between different repositories (e.g. dev -> staging)
```

scripts/dev/contexts/variables/e2e_smoke_base
MaciejKaras: Done! @nammn plz take a look

scripts/dev/contexts/evg-private-context
```diff
@@ -121,3 +98,51 @@ export cognito_workload_url="${cognito_workload_url}"
...
+else
+  BUILD_SCENARIO=$BUILD_SCENARIO_STAGING
+fi
+echo "Setting BUILD_SCENARIO=${BUILD_SCENARIO}"
+export BUILD_SCENARIO="${BUILD_SCENARIO}"
+
+# Set REGISTRY and OPERATOR_VERSION based on build scenario
+# REGISTRY and OPERATOR_VERSION can be overridden externally if needed
+PATCH_REGISTRY="268558157000.dkr.ecr.us-east-1.amazonaws.com/dev"
+STAGING_REGISTRY="268558157000.dkr.ecr.us-east-1.amazonaws.com/staging"
```
viveksinghggits, line 122: nit: this seems to be part of our `build_info.json` file as well. does it make sense to have these registries defined at just one place?
would the python scripts not figure out the registries by looking at the build_info.json based on the build scenario?

scripts/dev/contexts/evg-private-context
```diff
@@ -121,3 +98,51 @@ export cognito_workload_url="${cognito_workload_url}"
...
+  "${BUILD_SCENARIO_STAGING}")
+    COMMIT_SHA_SHORT=$(git rev-parse --short=8 HEAD)
+    REGISTRY="${REGISTRY:-$STAGING_REGISTRY}"
+    OPERATOR_VERSION="${OPERATOR_VERSION:-$COMMIT_SHA_SHORT}"
+    ;;
+
+  "${BUILD_SCENARIO_PATCH}")
+    REGISTRY="${REGISTRY:-$PATCH_REGISTRY}"
+    OPERATOR_VERSION="${OPERATOR_VERSION:-$VERSION_ID}"
+    ;;
```
viveksinghggits, line 139: q: would I be correct if I say, in case of PR patches the OPERATOR_VERSION would be evg patch id and in case of master merges it would be commit sha?
I am confirming this, because I noticed in case of master merges version_id is `mongodb_kubernetes_<commit_sha>`.

viveksinghggits (approved):

scripts/dev/contexts/evg-private-context
MaciejKaras: Yes, this would be good idea. Right now we specify the full repository for each image separately and this will be hard to it this way. I saw other comments in the TD that maybe we should move the base repository url from each image definition and use it as default build scenario value. This way we could easily replace those hardcoded values and read directly from `build_info.json`. I've added this as optional task in our milestones

scripts/dev/contexts/evg-private-context
MaciejKaras: yes, exactly. The `version_id` for master commits is `mongodb_kubernetes_<commit_sha>`, but we prefer our own logic that is easier to change

PR #341, opened 2025-08-12
CLOUDP-337727: Add operator with race checker image in init variant

mircea-cosbuc:
# Summary

This patch adds a separate task in the `init_test_run` variant for building the operator image with the race-checker enabled. 

This patch also moved the tests under `scripts/release` according to [standard pytest best practices](https://doc.pytest.org/en/latest/explanation/goodpractices.html#tests-as-part-of-application-code) as currently our python test targets were not capturing all the tests.

## Proof of Work

[EVG Job](https://spruce.mongodb.com/version/689b8a097e5e41000778b82e)

## Checklist


build_info.json
```diff
@@ -24,6 +24,14 @@
         ]
       }
     },
+    "operator-race": {
+      "patch": {
```
MaciejKaras, line 30: we will run the same set of tests in staging as in patch, so the staging part is also needed. Please use
```"repository": "268558157000.dkr.ecr.us-east-1.amazonaws.com/staging/mongodb-kubernetes"```

File: .evergreen.yml [hunk 2]
```diff
@@ -156,6 +156,25 @@ variables:
       - name: build_agent_images_ubi
         variant: init_test_run
 
+  - &base_om7_dependency_with_race
```
MaciejKaras, line 159: does it make sense to add new `base_om7_dependency_with_race` variable instead of updating existing one `base_om7_dependency`? Also if we change the `e2e_operator_race_ubi_with_telemetry` to use OM8 it might be inconsistent, because `operator-race` might not be pushed yet. It is still unlikely, because the image build is really fast. I'm just mentioning here that specifying new `base_om7_dependency_with_race` does not solve any problem while possibly making it harder to see the dependency issues.

I would rather see simplification for all `&base_om_dependency` like this. It will be much shorter and easier to read:
```yaml
- &base_om7_dependency
   depends_on:
      - name: build_om_images
        variant: build_om70_images
      - name: "*"
        variant: init_test_run
```

Also I don't know why `&base_no_om_image_dependency` is dependent on OM image ❓ 
```yaml
  - &base_no_om_image_dependency
    depends_on:
      - name: build_om_images
        variant: build_om70_images
      - name: build_operator_ubi
        variant: init_test_run
      - name: build_init_database_image_ubi
        variant: init_test_run
      - name: build_database_image_ubi
        variant: init_test_run
      - name: build_init_om_images_ubi
        variant: init_test_run
      - name: build_test_image
        variant: init_test_run
      - name: build_init_appdb_images_ubi
        variant: init_test_run
      - name: build_agent_images_ubi
        variant: init_test_run
```

scripts/release/build/build_info.py
```diff
@@ -3,11 +3,17 @@
 from typing import Dict
 
 from scripts.release.build.build_scenario import BuildScenario
-from scripts.release.constants import DEFAULT_REPOSITORY_PATH, DEFAULT_CHANGELOG_PATH, RELEASE_INITIAL_VERSION_ENV_VAR, \
```
Julien-Ben, line 6: nit: Do we have diverging linter configuration ? 
I assume this change is coming from `black`.

The change in this PR looks closer to the structure we usually have

scripts/release/pipeline_main.py
```diff
@@ -67,6 +69,7 @@ def get_builder_function_for_image_name() -> Dict[str, Callable]:
     image_builders = {
         MEKO_TESTS_IMAGE: build_tests_image,
         OPERATOR_IMAGE: build_operator_image,
+        OPERATOR_RACE_IMAGE: partial(build_operator_image, with_race_detection=True),
```
Julien-Ben, line 72: nit: Do you think this deserves a comment ? I think most people who will read this in the future won't know what `partial` is

scripts/release/build/build_info.py
mircea-cosbuc: I'm not sure, I'm using the linter configuration in the repo.

build_info.json
Julien-Ben: Have we already decided on a final name for the staging repo ? 
We might want to be more specific than just `staging` 
Like `mongodb-kubernetes-staging`

(this can be addressed in the TD/another PR, I'm just curious)

.evergreen.yml
mircea-cosbuc: In the context of how we're doing those dependencies it makes sense: we're optimizing the exact amount of images needed, not only for the general "run all tests" patches but for manual patches as well where we just want to build exactly the images we need. 
Using 
```
      - name: "*"
        variant: init_test_run
```
will include all the community images as well, for example.
I agree at the same time it can be annoying to keep so many different YAML anchors. I can refactor our om dependency anchors in this PR, though I don't think this is the point of the PR.

`base_no_om_image_dependency` : this is just poorly named, we need the OM image for the cloudqa variant of the test. There is one or two tests that spin up an OM. I can change this to `base_cloudqa_dependency`. Still, this is a different topic from this PR.

build_info.json
mircea-cosbuc: Thanks for the explanation, adding this.

scripts/release/pipeline_main.py
mircea-cosbuc: I'd rather not add a comment to document partial function application that's been in the python stdlib forever, but if we consider it necessary I will. Something like "Use the same image builder as the operator image with defaults set to enable race detection" ?

.evergreen.yml
mircea-cosbuc: I'm happy to have this in a different PR.

build_info.json
MaciejKaras: I wanted to keep it similar as we have for `dev`, so short name `staging` is ok. I will add it to TD.

.evergreen.yml
MaciejKaras: Thanks for explanation. You're right with the locally triggering evg patch, makes sense!

scripts/release/conftest.py
MaciejKaras, line 1: probably this should be part of scripts/release/tests

scripts/release/pipeline_main.py
Julien-Ben: Yeah that sounds good, but if you'd rather not comment I understand, don't have strong opinion

Julien-Ben (approved):
LGTM, thanks for fixing ! 🙏

PR #353, opened 2025-08-20
CLOUDP-327089: Populate .status.version in MongoDBSearch

anandsyncs:
# Summary

**Version Handling Improvements:**

* Refactored `ValidateSearchImageVersion` to accept a version string instead of extracting it internally, and renamed the internal method to `getMongotVersion` for clarity. The helper now sets `mdbSearch.Status.Version` to the determined version. 
* Updated the logic for extracting the image tag from a container image to use a new `extractImageTag` helper function, ensuring the version is parsed correctly from the image string.

**Testing Updates:**

* Modified the `TestMongoDBSearchReconcile_Success` test to use the new operator config and to assert that `Status.Version` is set to the expected value after reconciliation. 

## Proof of Work

Tests Pass

## Checklist

### DIFF
--- a/controllers/search_controller/mongodbsearch_reconcile_helper.go
+++ b/controllers/search_controller/mongodbsearch_reconcile_helper.go
@@ -86,9 +86,11 @@ func (r *MongoDBSearchReconcileHelper) reconcile(ctx context.Context, log *zap.S
 		return workflow.Failed(err)
 	}
 
-	if err := r.ValidateSearchImageVersion(); err != nil {
+	searchImageVersion := r.getMongotVersion()
+	if err := r.ValidateSearchImageVersion(searchImageVersion); err != nil {
 		return workflow.Failed(err)
 	}
+	r.mdbSearch.Status.Version = searchImageVersion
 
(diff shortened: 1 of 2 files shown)


controllers/search_controller/mongodbsearch_reconcile_helper.go
```diff
@@ -86,9 +86,11 @@ func (r *MongoDBSearchReconcileHelper) reconcile(ctx context.Context, log *zap.S
 		return workflow.Failed(err)
 	}
 
-	if err := r.ValidateSearchImageVersion(); err != nil {
+	searchImageVersion := r.getMongotVersion()
+	if err := r.ValidateSearchImageVersion(searchImageVersion); err != nil {
 		return workflow.Failed(err)
 	}
+	r.mdbSearch.Status.Version = searchImageVersion
```
lsierant, line 93: I think it's not a good way to update the status field at the beginning of the reconcile. What if there will be error? Will it set the version anyway? 
We should probably set this when we return workflow.OK() and we can use workflow.OK().WithAdditionalOptions() and pass there an option that will set the version.

PR #320, opened 2025-08-07
Add an evergreen function to build kubectl plugin and push to S3

viveksinghggits:
# Summary

## Checklist

- [ ] Have you linked a jira ticket and/or is the ticket in the title?
- [ ] Have you checked whether your jira ticket required DOCSP changes?
- [ ] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

PR #362, opened 2025-08-22
Enable `staging` build scenario that was commented out by default

viveksinghggits:
# Summary

The `staging` build scenario was commented out by default because we have not been using it in our workflows. But after [this PR](https://github.com/mongodb/mongodb-kubernetes/pull/320) gets merged we will start using the the staging build scenario for kubectl plugin to build and push it to the staging bucket. That's why this PR enables that build scenario.

## Proof of Work

Once the PR https://github.com/mongodb/mongodb-kubernetes/pull/320 related to kubectl plugin gets merged the goreleaser artifacts should be pushed to the `mongodb-kubernetes-staging/kubectl-mongodb/<commit-sha>` path.

## Checklist

### DIFF
--- a/scripts/release/build/build_scenario.py
+++ b/scripts/release/build/build_scenario.py
@@ -31,10 +31,9 @@ class BuildScenario(StrEnum):
         elif is_patch or is_evg:
             scenario = BuildScenario.PATCH
             logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
-        # TODO: Uncomment the following lines when starting to work on staging builds
-        # elif is_evg:
-        #     scenario = BuildScenario.STAGING
-        #     logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
+        elif is_evg:
+            scenario = BuildScenario.STAGING
+            logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
         else:
             scenario = BuildScenario.DEVELOPMENT
             logger.info(f"Build scenario: {scenario}")
--- a/scripts/release/build/build_scenario.py
+++ b/scripts/release/build/build_scenario.py
@@ -3,7 +3,8 @@ from enum import StrEnum
 from git import Repo
 
 from lib.base_logger import logger
-from scripts.release.constants import triggered_by_git_tag, is_evg_patch, is_running_in_evg, get_version_id
+from scripts.release.constants import triggered_by_git_tag, is_evg_patch, is_running_in_evg, get_version_id, \
+    DEFAULT_REPOSITORY_PATH
 from scripts.release.version import calculate_next_version
 
 COMMIT_SHA_LENGTH = 8
@@ -17,7 +18,7 @@ class BuildScenario(StrEnum):
     DEVELOPMENT = "development"  # Local build on a developer machine
 
     @classmethod
-    def infer_scenario_from_environment(cls) -> "BuildScenario":
+    def infer_scenario_from_environment(cls, repository_path: str = DEFAULT_REPOSITORY_PATH) -> "BuildScenario":
         """Infer the build scenario from environment variables."""
         git_tag = triggered_by_git_tag()
         is_patch = is_evg_patch()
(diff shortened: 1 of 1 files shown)


scripts/release/build/build_scenario.py
```diff
@@ -31,10 +31,9 @@ def infer_scenario_from_environment(cls) -> "BuildScenario":
         elif is_patch or is_evg:
             scenario = BuildScenario.PATCH
             logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
-        # TODO: Uncomment the following lines when starting to work on staging builds
-        # elif is_evg:
-        #     scenario = BuildScenario.STAGING
-        #     logger.info(f"Build scenario: {scenario} (patch_id: {patch_id})")
+        elif is_evg:
```
nammn: how can this work? Have we tested this? if `is_evg` is true, then its always patch already, no?

nammn (changes requested):
the code doesn't seem to work semantically, can we have a proof of work?

scripts/release/build/build_scenario.py
Julien-Ben: When I wrote the infer method I matched our previous code for the usage of env variables. If `is_patch`is false but `is_evg` is true, then we are on master.
I'm confused by the above `elif` tho.
`elif is_patch or is_evg` seems incorrect

nammn: we really need a way to test those things in the ci/e2e outside of requiring to merge to master

viveksinghggits: Hi @Julien-Ben,
I was looking at the docs here https://docs.devprod.prod.corp.mongodb.com/evergreen/Reference/Comparison#whats-a-version-or-patch and [here](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#default-expansions) and it is not very clear that `is_patch` is going to be false if the build is run from the master merge. Do you have reference to a documentation that we can use to confirm that `is_patch` is false in case of builds to master merges.

Julien-Ben: I tried to dig as much as I could in the documentation and commented everything in methods (they were moved to `constants.py`)

Do the following explanations help ? That's the best I could find

```
def triggered_by_git_tag() -> str | None:
    """
    "triggered_by_git_tag" is the name of the tag that triggered this version, if applicable. It will be None for
    all patches except when tagging a commit on GitHub
    :return: tag name if the build was triggered by a git tag, otherwise None.
    """
    return os.getenv("triggered_by_git_tag")


def is_evg_patch() -> bool:
    """
    A patch build is a version not triggered by a commit to a repository. "is_patch" is passed automatically by Evergreen.
    It either runs tasks on a base commit plus some diff if submitted by the CLI or on a git branch if created by
    a GitHub pull request.
    :return: "true" if the running task is in a patch build, otherwise "false".
    """
    return os.getenv("is_patch", "false").lower() == "true"


def is_running_in_evg() -> bool:
    """
    "RUNNING_IN_EVG" is set by us in evg-private-context
    :return: "true" if the script is running in Evergreen, otherwise "false".
    """
    return os.getenv("RUNNING_IN_EVG", "false").lower() == "true"


def get_version_id() -> str | None:
    """
    Get the version ID from the environment variable. This is typically used for patch builds in the Evergreen CI system.
    It is generated automatically for each task run. For example: `6899b7e35bfaee00077db986` for a manual/PR patch,
    or `mongodb_kubernetes_5c5a3accb47bb411682b8c67f225b61f7ad5a619` for a master merge
    :return: version_id (patch ID) or None if not set
    """
    return os.getenv("version_id")
```

MaciejKaras (changes requested):
This change will not work out of the box. Pull Request misses requires PoW that would prove this will make `master` branch e2e tests will start to fail. The reason is it will push the images with new version (commit_sha) but the e2e tests will still depend on the `build_id` as the image version - thus the image for the e2e tests will be missing.

This is also conflicting with another change https://github.com/mongodb/mongodb-kubernetes/pull/336, that was supposed to enable `staging` builds as well. It is closer to being approved, because it was tested - still needs some improvement though.

viveksinghggits:
> This change will not work out of the box. Pull Request misses requires PoW that would prove this will make `master` branch e2e tests will start to fail. The reason is it will push the images with new version (commit_sha) but the e2e tests will still depend on the `build_id` as the image version - thus the image for the e2e tests will be missing.
> 
> This is also conflicting with another change #336, that was supposed to enable `staging` builds as well. It is closer to being approved, because it was tested - still needs some improvement though.


Hi @MaciejKaras,
I was not really able to test this, because I didn't have a way to imitate the scenario where the patch is run after merge to master. Do you think it's a good idea merge this as part of the PR that you raised?

PR #365, opened 2025-08-23
Fix races in pre-commit checks and improve error handling

lsierant:
# Summary

Introduction of parallelism in pre-commit hook introduced races or even caused ignoring errors for some checks executed in background jobs. 

## Changes
- in pre_commit function each check is send to background with tracking its pid to. Previously we dispatched tasks into background and execute `wait` builtin without arguments. While it's OK for just waiting until completion it's ignoring any errors. Now, we're using `wait "${pid}"`, which fails if the background job return non zero code.
  - Note, that we cannot just use simple wait with $(jobs -p) because it's race'y (see the next point)
- in start_shellcheck we're using xargs that errors if any of the checks in background resulted in error. Previously there was a race condition because until we reached `jobs -p` some of the checks might be already failed and finished and it won't show on the job list. 
  

## Proof of Work
* [Clean lint_repo in evg](https://parsley.mongodb.com/evergreen/mongodb_kubernetes_unit_tests_lint_repo_patch_637923f515139843848eb9bd335bc6d66aab3f78_68aa3ac8e2d2f40007605d15_25_08_23_22_03_54/0/task?bookmarks=0%2C925&selectedLineRange=L916&shareLine=916)
```
[2025/08/24 00:11:23.223] pre-commit: All checks passed!
```

* [Deliberate error in shellcheck](https://parsley.mongodb.com/evergreen/mongodb_kubernetes_unit_tests_lint_repo_patch_637923f515139843848eb9bd335bc6d66aab3f78_68aa4cf0a398bc00070a8e1b_25_08_23_23_21_21/0/task?bookmarks=0%2C974&selectedLineRange=L963&shareLine=963): 
```
[2025/08/24 01:26:57.709] Some checks have failed:
[2025/08/24 01:26:57.709]     shellcheck (PID 4211)
```

## Checklist

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -163,18 +163,9 @@ lint_code() {
   scripts/evergreen/lint_code.sh
 }
 
-# bg_job_ vars are global; capture_bg_job function is appending to them on each call
+# bg_job_ vars are global; run_job_in_background function is appending to them on each call
 bg_job_pids=()
 bg_job_pids_with_names=()
-
-# Helper function to capture background job PID with job name
-capture_bg_job() {
-  local job_name="$1"
-  local job_pid=$!
-  bg_job_pids+=("${job_pid}")
-  bg_job_pids_with_names+=("${job_pid}:${job_name}")
-  echo "Started ${job_name} with PID: ${job_pid}"
-}
 
 get_job_name() {
   local search_pid="$1"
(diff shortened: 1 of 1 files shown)


.githooks/pre-commit
```diff
@@ -126,78 +126,133 @@ function regenerate_public_rbac_multi_cluster() {
...
+}
 
-  ( (time check_erroneous_kubebuilder_annotations) 2>&1 | prepend "check_erroneous_kubebuilder_annotations" ) &
+# Executes function given on the first argument as background job.
+# It's ensuring logs are properly prefixed by the name and
+# the job's pid is captured in bg_jobs array in order to wait for completion.
+run_job_in_background() {
+  job_name=$1
+  time ${job_name} 2>&1 | prepend "${job_name}" &
+  capture_bg_job "${job_name}"
```
nammn: why not inline capture_bg_job here?

nammn (approved):
lgtm, consider my comment

.githooks/pre-commit
lsierant: great catch! It was a leftover from my previous approach where I had capture_bg_job used for every check in pre-commit function. Removed.

Julien-Ben (approved):

PR #381, opened 2025-08-28
Improve the message in `check_precommit.sh` for go mod licenses case

viveksinghggits:
# Summary

If we (MCK) started consuming a new go mod deps and because of that we would need to regenerate (`update_licenses.sh`) the `LICENSE-THIRD-PARTY` file for the main project as well as `kubectl-mongodb` plugin. If we forget doing that, the CI would fail with below error

```
We have files that differ after running pre-commit, please run the pre-commit locally
```

and running `pre-commit` doesn't really update the licenses file, to do that, we will have to run `make precommit-with-licenses`. And that's why this PR updates above message.

## Proof of Work

NA

### DIFF
--- a/scripts/evergreen/check_precommit.sh
+++ b/scripts/evergreen/check_precommit.sh
@@ -23,7 +23,7 @@ if [ "${initial_index_state}" != "${final_index_state}" ]; then
   echo "Final index state:"
   echo "${final_index_state}"
 
-  echo "We have files that differ after running pre-commit, please run the pre-commit locally"
+  echo "We have files that differ after running pre-commit, please run the pre-commit and precommit-with-licenses locally"
   echo "Full diff: "
   git diff --cached --diff-filter=AM
   echo "The following files differ: "
--- a/scripts/evergreen/check_precommit.sh
+++ b/scripts/evergreen/check_precommit.sh
@@ -23,7 +23,7 @@ if [ "${initial_index_state}" != "${final_index_state}" ]; then
   echo "Final index state:"
   echo "${final_index_state}"
 
-  echo "We have files that differ after running pre-commit, please run the make target precommit-with-licenses locally"
+  echo "We have files that differ after running pre-commit, please run make precommit-with-licenses locally"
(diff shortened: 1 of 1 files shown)


anandsyncs (approved):

scripts/evergreen/check_precommit.sh
```diff
@@ -23,7 +23,7 @@ if [ "${initial_index_state}" != "${final_index_state}" ]; then
   echo "Final index state:"
   echo "${final_index_state}"
 
-  echo "We have files that differ after running pre-commit, please run the pre-commit locally"
+  echo "We have files that differ after running pre-commit, please run the pre-commit and precommit-with-licenses locally"
```
MaciejKaras: I don't think we need both, `precommit-with-licenses` will suffice

viveksinghggits: If a python/go file is not formatted well, we will have to run `precommit` right?

MaciejKaras: If you check the code for make `precommit-with-licenses` target it runs the same command, but with license option:

https://github.com/mongodb/mongodb-kubernetes/blob/1e901704a46e0544f3a72cf162f602f88b0f3f63/Makefile#L59-L63

viveksinghggits: Makes sense. I have changed the message.

scripts/evergreen/check_precommit.sh
```diff
@@ -23,7 +23,7 @@ if [ "${initial_index_state}" != "${final_index_state}" ]; then
   echo "Final index state:"
   echo "${final_index_state}"
 
-  echo "We have files that differ after running pre-commit, please run the pre-commit locally"
+  echo "We have files that differ after running pre-commit, please run the make target precommit-with-licenses locally"
```
viveksinghggits:
```suggestion
  echo "We have files that differ after running pre-commit, please run make precommit-with-licenses locally"
```

MaciejKaras (approved):

PR #383, opened 2025-08-28
WIP: Remove certs hash annotations

m1kola:
Includes #371 which needs to be merged first.

# Summary

## Checklist


controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -171,12 +171,32 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
...
 
+	// If tls is enabled we need to configure the "processes" array in opsManager/Cloud Manager with the
+	// correct tlsCertPath, with the new tls design, this path has the certHash in it(so that cert can be rotated
+	// without pod restart).
+	tlsCertPath := ""
+	internalClusterCertPath := ""
+	if mrs.Spec.Security.IsTLSEnabled() {
+		certSecretName := mrs.Spec.GetSecurity().MemberCertificateSecretName(mrs.Name)
+		internalClusterCertSecretName := mrs.Spec.GetSecurity().InternalClusterAuthSecretName(mrs.Name)
+		tlsCertHash := enterprisepem.ReadHashFromSecret(ctx, r.SecretClient, mrs.Namespace, certSecretName, "", log)
```
MaciejKaras, line 182: I see that `tlsCertHash` and `tlsCertPath` are calculated here and also in the other `controllers/operator/mongodbreplicaset_controller.go`. Can we move this logic to common controller?
I thought about method like:
```go
func (r *ReconcileCommonController) tlsCertHashAndPath(ctx, ...) (string, string) {
  tlsCertHash := enterprisepem.ReadHashFromSecret(ctx, r.SecretClient, mrs.Namespace, certSecretName, "", log)
  if tlsCertHash == "" {
    return "", ""
  }
  
  return tlsCertHash, fmt.Sprintf("%s/%s", util.TLSCertMountPath, tlsCertHash)  
}
```

controllers/operator/certs/certificates.go
```diff
@@ -31,7 +31,6 @@ type certDestination string
 
 const (
 	OperatorGeneratedCertSuffix = "-pem"
-	CertHashAnnotationKey       = "certHash"
```
MaciejKaras, line 34: q: was this annotation `CertHashAnnotationKey` and `InternalCertAnnotationKey` only used internally? User might use it for any reason? Even if not I see this as API change and we should add changelog mentioning that this has gone away + detail the reasoning behind the change.

m1kola: This is internal implementation detail. Users do not interact with this annotation: controller takes user proved secret with a certs, generates a concatenated cert and writes into a new secret. And from from that generated secret it calculates the hash which is being written into STS's annotation.

It is not an API change (since users do not interface with it). I do not think it is worth adding a change log about it.

controllers/operator/mongodbmultireplicaset_controller.go
m1kola: I would like to leave this out of the scope of this PR. Here I want to make a minimal change so that I can progress with my main goal (agent cert rotation).

There are more places where we read the secret and calculate the hash and I have a feeling that after adding agent cert rotation it will be a bit more clear if there is a pattern which we can abstract somehow. I'll look into this if I have time left after implementing agent cert rotation.

MaciejKaras: Understand. In that case can you add TODO item for improving this later?

controllers/operator/certs/certificates.go
MaciejKaras: How do we know users don't interact with this annotation? It's visible to them so they in theory can.

m1kola: @MaciejKaras this annotation (together with the rest of the stateful set) is reconciled by the operator. So even if they change this somehow it will be reconciled.

There is no practical reason for the users to read/write this annotation and use it for anything.

MaciejKaras: I totally understand your point and you're right in 99,9% that users are not using it for anything. 

Still this annotation is something that is visible through Kubernetes API and we should at least note down that is was removed and why.

m1kola: @MaciejKaras I don't think it is reasonable to add a release note every time we change a shape of underlying resources/primitives. It will just add noise to the release notes without adding any value whatsoever.

This specific release note might also sound like a breaking change (removing something) when it actually is not. These annotations also have never been documented in any way and are not useful to anyone apart from the operator itself.

I can add a release note if you still think that we should do that (please let me know), but my opinion is that it is not release not worthy change.

MaciejKaras: In this case where should we draw the line for need or no need of documenting the changes made in the API? This condition should be easy to follow by all engineers and properly documented. 

I come from a place where we should document every change visible to users, which is simple rule and easy to understand.

Adding @vinilage and @mircea-cosbuc as this impacts our release notes process.

m1kola: > In this case where should we draw the line for need or no need of documenting the changes made in the API?

@MaciejKaras this is not the user's API. StatefulSets are created and managed by the operator. This is a very clear line: there is no user API change here.

> come from a place where we should document every change visible to users, which is simple rule and easy to understand.

Users have great visibility and can inspect many things in Kubernetes. It doesn't necessary mean that we should be adding release notes for everything they can see. It will bloat the release notes with noise and will do more harm than good. Also I can give examples where we have made user visible changes and haven't added release notes. So perhaps it is more systematic issue?

---

As I said earlier - I'm happy to add a release note so we can make progress. Any ideas on how to word this to not be too alarming?

_The operator no longer adds `certHash` and `internalCertHash` annotations to database `StatefulSet`s. Existing annotations will be removed by the operator. These annotations were used by the operator in the process of certificate rotation and are no longer needed._

vinilage: I would add a note in the release notes only if the API is documented. If this is an internal implementation, we don't need to mention in the release notes, keeping it as clean as possible and focused in the changes that adds value to the customers.

I tried to find "certHash" in our public documentation and I couldn't find anything, so in this case I would not mention it in the release notes.

This may be a topic for the team discuss if you want to discuss deeply about what information we want to have in the release notes.

Obs.: since our code is now open-source, the commit history is also an "official registry" of our changes, and I would really use the release notes for customer facing changes.

MaciejKaras: Thanks Vini and Mikalai for this discussion. I'm fine with resolving this thread, but I'm still not certain if I can make correct decision in the future, what changes include in the release notes. My misunderstanding comes from the fact, that we lack documentation for:
 - what changes should we mention in the release notes and what not. This should cover changes to Kubernetes objects we create and manage (not only user created CRDs)
 - additionally how changes fall into different categories, fix, feature, breaking, other etc.

...or maybe this is documented and I'm just not aware of it?

MaciejKaras (approved):

anandsyncs (approved):

controllers/operator/mongodbmultireplicaset_controller.go
m1kola: I've added a to do in another branch (I'll eventually push the to do or a change to #389). Don't want to push to this branch and wait for CI again.

m1kola:
`e2e_om_ops_manager_queryable_backup` is red, but it is a known CI issue and is not related to this change. Merging.

controllers/operator/certs/certificates.go
m1kola: > or maybe this is documented and I'm just not aware of it?

@MaciejKaras I don't think we have any documentation like this and I don't know if it is going to be possible to clearly draw the line.

Would you like to bring this as a discuss topic?

MaciejKaras: Thanks. Good idea!

PR #378, opened 2025-08-27
CLOUDP-290847: Remove outdated code

m1kola:
# Summary

This piece of code is meant to preserve cert hash annotation in the pod template. We currently do not use cert hash annotation on the pod: we only use sts level annotations.

## Proof of Work

Tests must pass

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

PR #389, opened 2025-08-29
CLOUDP-290847: Improve agent certificate rotation

m1kola:
# Summary

Pods now mount a secret containing both old and new certificates, with file names being the hash of the certificate. When it is time to rotate the certificate, the operator updates the automation config with a new path (including the hash) to the certificate. This eliminates the need to manually restart Pods during certificate rotation.

Other related PRs:
* https://github.com/mongodb/mongodb-kubernetes/pull/383 - removes cert hash annotations, simplifies how we pass cert hashes a bit.
* https://github.com/mongodb/mongodb-kubernetes/pull/378 - clean up around Opaque cert secrets which are no longer supported.
* https://github.com/mongodb/mongodb-kubernetes/pull/371 - preparation refactoring which made implementation of this PR a bit easier).
* https://github.com/mongodb/mongodb-kubernetes/pull/444 - this last one is not directly related to agent cert rotation. It just unifies tests so we don't restart STS in all the tests.

---

Copilot summary:

This pull request implements improved agent certificate rotation for database Pods, allowing seamless certificate updates without requiring manual Pod restarts. The main changes revolve around passing the agent certificate file path (which now includes a hash for rotation) through the operator's reconciliation and authentication logic, updating method signatures and tests accordingly, and simplifying how agent certificate secret names are handled.

**Agent certificate rotation and configuration enhancements:**

* Database Pods now mount a secret containing both old and new agent certificates, with filenames based on the certificate hash. The operator updates the automation config with the new certificate path (including the hash) during rotation, eliminating the need for manual Pod restarts.
* The agent certificate file path (`AutoPEMKeyFilePath`) is now passed through the reconciliation and authentication logic, allowing dynamic updates to the path as certificates rotate. This includes changes to method signatures, struct fields, and function calls in the operator and authentication code.

**API and method signature updates:**

* The `AgentClientCertificateSecretName` method now returns a string (the secret name) instead of a `SecretKeySelector`, simplifying its usage throughout the codebase. Corresponding updates were made to all call sites and tests.
* The `EnableX509Authentication` methods for backup and monitoring agent configs now accept the agent PEM file path as a parameter, ensuring the correct path (including hash) is set during configuration.

**Test updates:**

* Tests have been updated to reflect the new method signatures and to check that the correct (possibly hashed) PEM file path is being set and passed through during agent authentication and backup/monitoring configuration.

These changes collectively improve the robustness and automation of agent certificate rotation, reducing operational overhead and risk during certificate updates.

## Proof of Work

Existing tests `e2e_tls_x509_configure_all_options_rs` and `e2e_tls_x509_configure_all_options_sc` now have extra certificate rotation tests `test_rotate_agent_certificate` (same name in both) which prove that we update the automation config on certificate rotation.

## Checklist

- [x] Create a DOCSP ticket/a docs PR to clean [this leftover](https://www.mongodb.com/docs/kubernetes/current/reference/k8s-operator-specification/#mongodb-setting-spec.security.authentication.agents.clientCertificateSecretRef.name) about supporting opaque Kubernetes secrets for agent certs (this is no longer supported since MongoDB Enterprise Kubernetes Operator 1.17). Created [DOCSP-53906](https://jira.mongodb.org/browse/DOCSP-53906).

### DIFF
--- a/changelog/20250909_feature_improved_agent_certificate_rotation.md
+++ b/changelog/20250909_feature_improved_agent_certificate_rotation.md
@@ -1,0 +1,7 @@
+---
+title: Improved agent certificate rotation
+kind: feature
+date: 2025-09-09
+---
+
+* Improve automation agent certificate rotation: the agent now restarts automatically when its certificate is renewed, ensuring smooth operation without manual intervention and allowing seamless certificate updates without requiring manual Pod restarts.
--- a/controllers/operator/authentication_test.go
+++ b/controllers/operator/authentication_test.go
@@ -91,8 +91,7 @@ func TestUpdateOmAuthentication_NoAuthenticationEnabled(t *testing.T) {
 
 	kubeClient, omConnectionFactory := mock.NewDefaultFakeClient(rs)
 	r := newReplicaSetReconciler(ctx, kubeClient, nil, "", "", false, false, omConnectionFactory.GetConnectionFunc)
-	agentCertSecretSelector := rs.GetSecurity().AgentClientCertificateSecretName(rs.Name)
-	r.updateOmAuthentication(ctx, conn, processNames, rs, agentCertSecretSelector, "", "", false, zap.S())
+	r.updateOmAuthentication(ctx, conn, processNames, rs, "", "", "", false, zap.S())
 
 	ac, _ := conn.ReadAutomationConfig()
 
@@ -113,8 +112,7 @@ func TestUpdateOmAuthentication_EnableX509_TlsNotEnabled(t *testing.T) {
 
 	kubeClient, omConnectionFactory := mock.NewDefaultFakeClient(rs)
 	r := newReplicaSetReconciler(ctx, kubeClient, nil, "", "", false, false, omConnectionFactory.GetConnectionFunc)
-	agentCertSecretSelector := rs.GetSecurity().AgentClientCertificateSecretName(rs.Name)
-	status, isMultiStageReconciliation := r.updateOmAuthentication(ctx, conn, []string{"my-rs-0", "my-rs-1", "my-rs-2"}, rs, agentCertSecretSelector, "", "", false, zap.S())
+	status, isMultiStageReconciliation := r.updateOmAuthentication(ctx, conn, []string{"my-rs-0", "my-rs-1", "my-rs-2"}, rs, "", "", "", false, zap.S())
 
 	assert.True(t, status.IsOK(), "configuring both options at once should not result in a failed status")
 	assert.True(t, isMultiStageReconciliation, "configuring both tls and x509 at once should result in a multi stage reconciliation")
@@ -126,8 +124,7 @@ func TestUpdateOmAuthentication_EnableX509_WithTlsAlreadyEnabled(t *testing.T) {
 	omConnectionFactory := om.NewCachedOMConnectionFactoryWithInitializedConnection(om.NewMockedOmConnection(deployment.CreateFromReplicaSet("fake-mongoDBImage", false, rs)))
 	kubeClient := mock.NewDefaultFakeClientWithOMConnectionFactory(omConnectionFactory, rs)
 	r := newReplicaSetReconciler(ctx, kubeClient, nil, "", "", false, false, omConnectionFactory.GetConnectionFunc)
-	agentCertSecretSelector := rs.GetSecurity().AgentClientCertificateSecretName(rs.Name)
-	status, isMultiStageReconciliation := r.updateOmAuthentication(ctx, omConnectionFactory.GetConnection(), []string{"my-rs-0", "my-rs-1", "my-rs-2"}, rs, agentCertSecretSelector, "", "", false, zap.S())
+	status, isMultiStageReconciliation := r.updateOmAuthentication(ctx, omConnectionFactory.GetConnection(), []string{"my-rs-0", "my-rs-1", "my-rs-2"}, rs, "", "", "", false, zap.S())
 
 	assert.True(t, status.IsOK(), "configuring x509 when tls has already been enabled should not result in a failed status")
 	assert.False(t, isMultiStageReconciliation, "if tls is already enabled, we should be able to configure x509 is a single reconciliation")
@@ -143,8 +140,7 @@ func TestUpdateOmAuthentication_AuthenticationIsNotConfigured_IfAuthIsNotSet(t *
 	kubeClient := mock.NewDefaultFakeClientWithOMConnectionFactory(omConnectionFactory, rs)
 	r := newReplicaSetReconciler(ctx, kubeClient, nil, "", "", false, false, omConnectionFactory.GetConnectionFunc)
 
-	agentCertSecretSelector := rs.GetSecurity().AgentClientCertificateSecretName(rs.Name)
-	status, _ := r.updateOmAuthentication(ctx, omConnectionFactory.GetConnection(), []string{"my-rs-0", "my-rs-1", "my-rs-2"}, rs, agentCertSecretSelector, "", "", false, zap.S())
+	status, _ := r.updateOmAuthentication(ctx, omConnectionFactory.GetConnection(), []string{"my-rs-0", "my-rs-1", "my-rs-2"}, rs, "", "", "", false, zap.S())
 	assert.True(t, status.IsOK(), "no authentication should have been configured")
 
 	ac, _ := omConnectionFactory.GetConnection().ReadAutomationConfig()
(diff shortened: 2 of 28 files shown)


controllers/operator/certs/certificates.go
```diff
@@ -327,25 +317,19 @@ func VerifyAndEnsureClientCertificatesForAgentsAndTLSType(ctx context.Context, s
 			return err
 		}
 
-		if s.Type == corev1.SecretTypeTLS {
-			needToCreatePEM = true
-			secretData = s.Data
```
m1kola, line 332: This is a leftover from times when MEKO had support for PEM certificates in Opaque Kubernetes secrets. This support was removed in MEKO 1.17.

controllers/operator/certs/certificates.go
```diff
@@ -271,16 +271,6 @@ func validatePemData(data []byte, additionalDomains []string) error {
...
-// validatePemSecret returns true if the given Secret contains a parsable certificate and contains all required domains.
-func validatePemSecret(secret corev1.Secret, key string, additionalDomains []string) error {
-	data, ok := secret.Data[key]
-	if !ok {
-		return xerrors.Errorf("the secret %s does not contain the expected key %s\n", secret.Name, key)
-	}
-
-	return validatePemData(data, additionalDomains)
-}
-
```
m1kola, line 283: This is a leftover from times when MEKO had support for PEM certificates in Opaque Kubernetes secrets. This support was removed in MEKO 1.17.

controllers/operator/common_controller.go
```diff
@@ -480,30 +482,23 @@ func (r *ReconcileCommonController) updateOmAuthentication(ctx context.Context,
...
-			// If the agent secret is of type TLS, we can find the certificate under the standard key,
-			// otherwise the concatenated PEM secret would contain the certificate and keys under the selector's
-			// Key identifying the agent.
-			// In single cluster workloads this path is working with the concatenated PEM secret,
-			//
-			// Important: In multi cluster it is working with the TLS secret in the central cluster, hence below selector update.
-			if agentSecret.Type == corev1.SecretTypeTLS {
-				agentCertSecretSelector.Key = corev1.TLSCertKey
-			}
-
```
m1kola, line 501: This change is important.

Previously this method had different sources of truth for auth configuration depending on a resource:
* For standalone and single-cluster replica set resources we were reading agent subjects from an Opaque secret **generated by the operator**. The secret contains key-value pairs where key is cert hash (calculated from original TLS secret provided by an user) and value is a certificate concatenated by the operator (from the original TLS secret) and converted into a PEM container.
* For multi-cluster and sharded resources we were reading from the original TLS secret **provided by a user**.

The logic was scattered across all the containers where in some cases we were adding `-pem` suffix and in some cases were not (depending on the resource as mentioned above) but also in this function where we had to read the original secret, check the type and override either a key name based on the secret type. It was extremely difficult to comprehend.

Now all the controllers provide the original TLS secret name to this function and the function uses only this to read subjects. I believe the end result is the same, but the implementation a bit easier to understand.

controllers/operator/common_controller.go
```diff
@@ -529,15 +524,6 @@ func (r *ReconcileCommonController) updateOmAuthentication(ctx context.Context,
...
 	} else {
-		agentSecret := &corev1.Secret{}
-		if err := r.client.Get(ctx, kube.ObjectKey(ar.GetNamespace(), agentCertSecretSelector.Name), agentSecret); client.IgnoreNotFound(err) != nil {
-			return workflow.Failed(err), false
-		}
-
-		if agentSecret.Type == corev1.SecretTypeTLS {
-			agentCertSecretSelector.Name = fmt.Sprintf("%s%s", agentCertSecretSelector.Name, certs.OperatorGeneratedCertSuffix)
-		}
-
```
m1kola, line 540: Same explanation [as above](https://github.com/mongodb/mongodb-kubernetes/pull/389/files#r2337173529).

controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -1169,6 +1176,9 @@ func getTLSSecretNames(sc *mdbv1.MongoDB) func() []string {
 		for i := 0; i < sc.Spec.ShardCount; i++ {
 			secretNames = append(secretNames, sc.GetSecurity().MemberCertificateSecretName(sc.ShardRsName(i)))
 		}
+		if sc.GetSecurity().ShouldUseX509("") {
+			secretNames = append(secretNames, sc.GetSecurity().AgentClientCertificateSecretName(sc.Name))
+		}
```
m1kola, line 1181: Right now Sharded Cluster Controller doesn't watch for changes in the agent cert secrets at all unlike other controllers.

This addition ensures that the resource gets reconciled when secret changes. Otherwise automation config won't be updated until something else triggers reconciliation.

controllers/operator/authentication_test.go
```diff
@@ -812,17 +808,14 @@ func Test_NoExternalDomainPresent(t *testing.T) {
 }
 
 // createAgentCSRs creates all the agent CSRs needed for x509 at the specified condition type
-func createAgentCSRs(t *testing.T, ctx context.Context, numAgents int, client kubernetesClient.Client, conditionType certsv1.RequestConditionType) {
-	if numAgents != 1 && numAgents != 3 {
-		return
-	}
```
m1kola, line 818: Just removed `numAgents` as it is not doing doing anything apart from this silly condition. Looks like a letfover.

changelog/20250909_feature_improved_agent_certificate_rotation.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Improved agent certificate rotation
+kind: feature
+date: 2025-09-09
+---
+
+* Database Pods now mount a secret containing both old and new certificates, with file names being the hash of the certificate. When it is time to rotate the certificate, the operator updates the automation config with a new path (including the hash) to the certificate. This eliminates the need to manually restart Pods during certificate rotation.
```
lsierant: Release notes are too technical imo. "How" - we have in PR's description. Here we should just say "what". 
For example: "Improved agent certificate rotation. Pods are automatically restarted now when the agent's certificate is rotated."

lsierant: Also I think the title is not visible in the release notes, so it might be confusing if this is about normal TLS certs or agent's

controllers/operator/mongodbshardedcluster_controller.go
lsierant: Nice catch!!!

controllers/operator/common_controller.go
m1kola: Adding a bit of extra info about differences between single-cluster replica set controller and others.

I think this came to be this way because in multi cluster setups we always have the original TLS secret in the central cluster and in the member clusters we have only operator generates certs with `certs.OperatorGeneratedCertSuffix` (a constant for `"-pem"` literal).

Here we can see that we were adding `certs.OperatorGeneratedCertSuffix` to the secret name. This effectively makes us refer to the Opaque secret generated by the operator from the original TLS secret provided by the user:

https://github.com/mongodb/mongodb-kubernetes/blob/cb20e63069f1bf87f969a17dee06f6314fad6b95/controllers/operator/mongodbreplicaset_controller.go#L244-L245

In other controllers where we were not adding `certs.OperatorGeneratedCertSuffix` to the secret name. Then in `controllers/operator/common_controller.go` we either add `certs.OperatorGeneratedCertSuffix` suffix to the secret `Name` or replace `Key` in `agentCertSecretSelector` with `corev1.TLSCertKey` because we are dealing with the original TLS cert provided by the user (which is always TLS secret).

After this change we always work with the original TLS secret in all controllers. This eliminates the need of checking for secret type and having custom code for overriding `Key` or `Name` on `agentCertSecretSelector`.

lsierant: Thanks for additional details! Really helpful. I think it's much better to align this and only read from TLS-type secret. Reading from -pem might be just a leftover from the past.

Please do not resolve this comment it's super useful.

lsierant (approved):
Great work! 👏 LGTM!

changelog/20250909_feature_improved_agent_certificate_rotation.md
m1kola: Great feedback, thanks! I'm waiting for more feedback to come in and then I'll update the release note together with other changes.

lucian-tosa (approved):

changelog/20250909_feature_improved_agent_certificate_rotation.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Improved agent certificate rotation
+kind: feature
+date: 2025-09-09
+---
+
+* Improved automation agent certificate rotation. The automation agent is automatically restarted when the agent's certificate is rotated.
```
m1kola: @lsierant updated the release note. What do you think?

Also adding @vinilage for a review of the release note.

vinilage: Minor change:

"Improve automation agent certificate rotation: the agent now restarts automatically when its certificate is renewed, ensuring smooth operation without manual intervention and allowing seamless certificate updates without requiring manual Pod restarts."

m1kola: @vinilage thanks! I updated the PR with this exact wording.

controllers/operator/database_statefulset_options.go
```diff
@@ -54,6 +54,13 @@ func CertificateHash(hash string) func(options *construct.DatabaseStatefulSetOpt
 	}
 }
 
+// AgentCertHash will assign the given AgentCertHash during StatefulSet construction.
```
viveksinghggits, line 57: nit: I think it would be useful to expand this comment a little.
The comments says, it will assign, but doesn't mention It will assign to what. Assigns to `statefulsetoptions.AgentCertHash`?

m1kola:
I rebased on top of master (without any changes/conflicts) to run with the recent tests.

PR #404, opened 2025-09-04
Fix master builds - it was using BuildScenario.DEVELOPMENT for master merges by mistake

MaciejKaras:
# Summary

## Checklist


scripts/release/build/build_scenario.py
```diff
@@ -30,7 +30,7 @@ def infer_scenario_from_environment(cls) -> "BuildScenario":
             # Release scenario and the git tag will be used for promotion process only
             scenario = BuildScenario.RELEASE
             logger.info(f"Build scenario: {scenario} (git_tag: {git_tag})")
-        elif is_patch and is_evg:
+        elif is_patch:
```
lucian-tosa: So the issue was that `is_patch` && `is_evg` was false, but we want it to be true?
I am not sure this change fixes is, because `is_evg` is always true, when running in CI

PR #409, opened 2025-09-04
always print version and path and error handling

nammn:
# Summary

## Checklist

### DIFF
--- a/scripts/dev/recreate_python_venv.sh
+++ b/scripts/dev/recreate_python_venv.sh
@@ -76,6 +76,9 @@ ensure_required_python() {
     if pyenv install --skip-existing "${required_version}"; then
         pyenv global "${required_version}"
         return 0
+    else
+        echo "Error: Failed to install Python ${required_version} via pyenv" >&2
+        return 1
     fi
 }
 
@@ -88,7 +91,7 @@ fi
 # Ensure required Python version is available
 ensure_required_python
 
-python3 -m venv venv
+python -m venv venv
 source venv/bin/activate
 pip install --upgrade pip
 
@@ -102,5 +105,4 @@ else
 fi
 
 echo "Python venv was recreated successfully."
-echo "Current python path: $(which python)"
-python --version
+echo "Using Python: $(which python) ($(python --version))" >&2


scripts/release/atomic_pipeline.py
```diff
@@ -126,7 +126,7 @@ def build_meko_tests_image(build_configuration: ImageBuildConfiguration):
     shutil.copyfile("release.json", "docker/mongodb-kubernetes-tests/release.json")
     shutil.copyfile("requirements.txt", requirements_dest)
 
-    python_version = os.getenv("PYTHON_VERSION", "3.13")
+    python_version = os.getenv("PYTHON_VERSION", "3.13.7")
```
MaciejKaras: I think it's better to fail here, if the PYTHON_VERSION is not set. Otherwise we would need to remember to update it in multiple places

nammn: yes! true, let me do that

scripts/dev/run_python.sh
```diff
@@ -16,8 +16,9 @@ fi
 
 export PYTHONPATH="${PROJECT_DIR}"
 
-current_python_version=$(python --version 2>&1 | awk '{split($2, a, "."); print a[1] "." a[2]}')
-if [[ "${current_python_version}" != "${PYTHON_VERSION}" ]]; then
+current_python_version=$(python --version 2>&1 | awk '{print $2}')
+required_version="${PYTHON_VERSION:-3.13.7}"
```
MaciejKaras: if `PYTHON_VERSION` is not set, we should exit the script, not using the default value.

nammn: yeah will be handled the same everywhere

scripts/dev/recreate_python_venv.sh
```diff
@@ -70,25 +70,38 @@ ensure_required_python() {
         return 1
     fi
 
-    # Install specific pinned version for consistency across runs
+    # Check if the required version is already installed
+    if pyenv versions --bare | grep -q "^${required_version}$"; then
+        echo "Python ${required_version} already installed via pyenv" >&2
+        pyenv global "${required_version}"
```
lucian-tosa, line 82: should we use local/shell instead of global?

nammn: i am not sure whether the next task is running under a different shell, global is the safer bet for me, albeit not ci friendly indeed

nammn: but yeah shell should work since we create the venv and later only rely on the venv 🤔

nammn: ive tried shell and it didn't for some reason, unfortunately I didn't have the time to figure out why and just used the one that works

PR #21, opened 2025-04-22
[migration] rename all paths to match repo path

nammn:
# Summary

- ensures we are removing 10gen leftovers and use the correct path
- its a search and replace for all `github.com/10gen` to `github.com/mongodb`
- its a search and replace for all `github.com/10gen/ops-manager-kubernetes` to `github.com/mongodb/mongodb-kubernetes`

## Proof of Work

- green ci

## Checklist
- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you checked for release_note changes?

## Reminder (Please remove this when merging)
- Please try to Approve or Reject Changes the PR, keep PRs in review as short as possible
- Our Short Guide for PRs: [Link](https://docs.google.com/document/d/1T93KUtdvONq43vfTfUt8l92uo4e4SEEvFbIEKOxGr44/edit?tab=t.0)
- Remember the following Communication Standards - use comment prefixes for clarity:
  * **blocking**: Must be addressed before approval.
  * **follow-up**: Can be addressed in a later PR or ticket.
  * **q**: Clarifying question.
  * **nit**: Non-blocking suggestions.
  * **note**: Side-note, non-actionable. Example: Praise 
  * --> no prefix is considered a question

PR #410, opened 2025-09-05
Fix `.gitignore` to not ignore `mongodb-kubernetes` dir/file

viveksinghggits:
# Summary

As part of this PR https://github.com/mongodb/mongodb-kubernetes/pull/21 we changed all the occurrences of old repo to the new repo and in doing so we ended up adding `mongodb-kubernetes` to the `.gitignore` file.

Becauase of this, if we try to add any file to our new repo in dir `public/dockerfiles/mongodb-kubernetes` it would be ignored by default. This PR fixes that. We don't really need that in .gitignore.

## Proof of Work

Create a file in the dir `public/dockerfiles/mongodb-kubernetes` and make sure that it shows up in `git status`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,7 +42,6 @@ public/architectures/**/certs/*
 public/architectures/**/secrets/*
 
 docker/mongodb-kubernetes-appdb/content/readinessprobe
-mongodb-kubernetes
 docker/mongodb-kubernetes-operator/content/mongodb-kubernetes-operator.tar
 docker/mongodb-kubernetes-tests/helm_chart/
 docker/mongodb-kubernetes-tests/public/


.gitignore
```diff
@@ -42,7 +42,6 @@ public/architectures/**/certs/*
 public/architectures/**/secrets/*
 
 docker/mongodb-kubernetes-appdb/content/readinessprobe
-mongodb-kubernetes
 docker/mongodb-kubernetes-operator/content/mongodb-kubernetes-operator.tar
```
viveksinghggits, line 45: or instead of removing this we can change it to 
`/mongodb-kubernetes`
if in case someone is used to build controller locally and don't want it to show up in `git status`.

nammn: q: why would you have a folder in project dir called /mongodb-kubernetes? Is that a common pattern?

viveksinghggits: sorry @nammn I didn't understand things here, I am not suggesting that people might have dir called `mongodb-kubernetes`. 
Is your comment on the change that is made in the PR? Or Is your comment on my reply? Because both of those say entirely different things.

anandsyncs: @viveksinghggits I think that ignoring  `/mongodb-kubernetes` is a better approach. 


> or instead of removing this we can change it to `/mongodb-kubernetes` if in case someone is used to build controller locally and don't want it to show up in `git status`.

m1kola: > sorry @nammn I didn't understand things here, I am not suggesting that people might have dir called `mongodb-kubernetes`. Is your comment on the change that is made in the PR? Or Is your comment on my reply? Because both of those say entirely different things.

I think the question is - why do we want to keep `/mongodb-kubernetes` in git ignore (instead of just removing from git ignore as in the current version)?

viveksinghggits: @m1kola if someone wants to build the project just using `go build`, we would have a binary named `mongodb-kubernetes` and that would not be ignored.
If we add `/mongodb-kubernetes` in `.gitignore` we would fix the problem that this PR tries to fix and it will also handle the `go build` case.

m1kola: Got it. That makes sense. In that case - I think ignoring `/mongodb-kubernetes` is better.

viveksinghggits: @m1kola I have made the change, can you look into again/approve the PR please.

m1kola (approved):

PR #446, opened 2025-09-12
Make Dependabot add `skip-changelog` label

m1kola:
# Summary

So we don't have to add it manually.

## Proof of Work

N/A

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,10 @@ updates:
       - dependency-name: k8s.io/client-go
       - dependency-name: k8s.io/code-generator
       - dependency-name: sigs.k8s.io/controller-runtime
+    labels:
+      - dependencies
+      - go
+      - skip-changelog
 
   - package-ecosystem: pip
     directory: "/"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -16,6 +16,10 @@ updates:
       - dependency-name: k8s.io/client-go
       - dependency-name: k8s.io/code-generator
       - dependency-name: sigs.k8s.io/controller-runtime
+    labels:
+      - dependencies
+      - go
```
m1kola, line 21: Default labels. Have to explicitly define them when providing custom labels.

Same below for python deps.

viveksinghggits: q: are these labels `dependencies1` and `go`, `python` used by any workflows?

m1kola: @viveksinghggits I don't think so. I added these because I wanted to preserve what we have already.

It might be handy to filter PRs by these labels (e.g. if you are looking for old go bumps). But I also don't mind removing them if we thimk that we don't need them.

viveksinghggits: I think it's ok.

PR #449, opened 2025-09-12
Use enterprise mongodb for enterprise search snippets

lsierant:
# Summary

In enterprise snippets we declare the mdb version as 8.0.10 without the `-ent` suffix. 
The agents then will download a community version of the mongodb, not the enterprise:

## Checklist


docs/search/04-search-external-mongod/env_variables.sh
```diff
@@ -2,7 +2,7 @@ export K8S_CTX="<your kubernetes context here>"
 
 export MDB_NS="mongodb"
 
-export MDB_VERSION="8.0.10"
+export MDB_VERSION="8.0.10-ent"
```
anandsyncs: external mongod snippets were setup for community, enterprise version would need extra steps.
Like extracting the keyfile separately.

fealebenpae: This will still use the MCO replica set controller so the keyfile and setParameter behavior will be the same, but it will force it to use the MongoDB Enterprise Server image instead of the Community one, so it should be fine.

PR #450, opened 2025-09-12
[CLOUDP-343558] Run helm chart-testing tool `ct` as part of CI (`lint_repo`)

viveksinghggits:
# Summary

Whenever we raised PR in to helm chart repo CI failed there because we had linting issues in our helm chart. A previous PR that [I raised](https://github.com/mongodb/mongodb-kubernetes/pull/445) fixed the linting problem. But we should have a linting mechanism in our CI that would consistently check if we have made any mistake in our charts. This PR does that.

The `chart_schema.yaml` and `lintconf.yaml` are the default yaml files using which `ct` should be run. And `ct` is the tool that is used for linting in our helm charts repo.

## Proof of Work

Passing `lint_repo` task here https://spruce.mongodb.com/version/68c43aae1d5d5d000770c624/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC and in this PR as well.

## Checklist

### DIFF
--- a/.evergreen-functions.yml
+++ b/.evergreen-functions.yml
@@ -265,7 +265,7 @@ functions:
           - ${workdir}/venv/bin
         working_dir: src/github.com/mongodb/mongodb-kubernetes
         binary: scripts/evergreen/check_precommit.sh
-    
+
   # Configures docker authentication to ECR and RH registries.
   setup_building_host:
     - *switch_context


MaciejKaras (changes requested):
Changes look good, still I have one general ask.

When we add or change logic in our linting requirements we should also consider adding them to `.githooks/pre-commit`. Otherwise there is no way locally to lint the helm charts.

Secondly we should make scripts that are run in EVG CI easy to run locally as well. I wouldn't put them in `scripts/evergreen/` directory, but in `scripts/helm` and make them available for developers. I think also `.githooks/pre-commit` takes care of installing all prerequisite applications.

Makefile
```diff
@@ -420,3 +420,8 @@ prepare-local-olm-e2e:
 
 prepare-operator-configmap: # prepares the local environment to run a local operator
 	source scripts/dev/set_env_context.sh && source scripts/funcs/printing && source scripts/funcs/operator_deployment && prepare_operator_config_map "$(kubectl config current-context)"
+
+# Lint the helm chart using helm chart-testing tool (`ct`) using the default configuration available at `helm_chart/tests/schemas/`
+lint-chart:
```
MaciejKaras, line 425: what about target for installing `chart-testing` locally?

scripts/dev/lint_helm_chart.sh
MaciejKaras, line 1: both scripts should not be in `scripts/evergreen/` dir, they can be used outside of evg

lsierant: we should probably migrate away from scripts/evergreen completely.

scripts/dev/setup_chart_testing_cli.sh
lsierant, line 1: Please check if you could hook up into [install.sh](https://github.com/mongodb/mongodb-kubernetes/blob/0946c60768d7e3f84ec68a64f5557ab25b6cc072/scripts/dev/install.sh#L1-L0) to setup this on a mac too.

.evergreen-functions.yml
```diff
@@ -258,6 +265,14 @@ functions:
           - ${workdir}/venv/bin
         working_dir: src/github.com/mongodb/mongodb-kubernetes
         binary: scripts/evergreen/check_precommit.sh
+    - command: subprocess.exec
```
lsierant: I would prefer this to be part of pre-commit script instead

scripts/dev/lint_helm_chart.sh
viveksinghggits: @MaciejKaras so what's our policy on where a script should be put? Can a script be used by evg if it's outside of evg dir but other workflows can't/shouldn't use it if it's in evg dir?
I am asking this so that I can decide where to move these scripts.

MaciejKaras: Basically we should not keep any script in evergreen dir. It is misleading, especially for scripts that can be run locally. The confusion is even greater because evergreen uses scripts from `scripts/evergreen/`, `scripts/` and even `scripts/dev/` dir.

Makefile
viveksinghggits: I have added it as part of `install.sh`, which gets run from `make prerequisites`.

.evergreen-functions.yml
viveksinghggits: Since we didn't discuss this in last discuss meeting @lsierant. Is it ok if we handle this comment later after the discussion, so that we can merge this PR? Right now I have added a way to run this script locally.

MaciejKaras (approved):

scripts/dev/setup_chart_testing_cli.sh
```diff
@@ -0,0 +1,23 @@
...
+  bindir="${PROJECT_DIR}/bin"
+  tmpdir="${PROJECT_DIR}/tmp"
+else
+  bindir="${GOPATH}/bin"
+  tmpdir="${GOPATH}/tmp"
+fi
+
+mkdir -p "${bindir}" "${tmpdir}"
+
+curl -s --retry 3 -L -o "${tmpdir}/chart-testing_3.13.0_linux_amd64.tar.gz" "https://github.com/helm/chart-testing/releases/download/v3.13.0/chart-testing_3.13.0_linux_amd64.tar.gz"
```
nammn, line 21: will this work on our multi-arch systems on master for smoke tests?

.evergreen-functions.yml
nammn: I also prefer it to be part of our pre-commit script, otherwise i need to chase-down scripts

scripts/dev/setup_chart_testing_cli.sh
viveksinghggits: do we run `lint_repo` on such systems @nammn? thanks for pointing it out, can you please tell me which variant runs this, I will go ahead and check if this needs modifications there.

nammn: We probably will be fine with the current code. But might be better to just rely on the arch tag. E.g if we run this locally. You can check all the other scripts how they do it

.evergreen-functions.yml
m1kola: I would prefer:
1. To be able easily to run this separately (`make lint-chart` is perfect). So I can run it separately from everything else.
2. Have a separate CI job for it (so not part of `lint_repo`) because when `lint_repo` fails it you need to go through other logs because it is not immediately clear it is an issue with Python linting or Go linting or Bash linting or maybe an issue with licenses?

I'm good for it to be included in a pre-commit if it

scripts/dev/setup_chart_testing_cli.sh
viveksinghggits: So if we run it locally, it will be called via [install.sh](https://github.com/mongodb/mongodb-kubernetes/pull/450/files#diff-50de9b4b83598458bdb8adf703dbb057d8fcb7fe96d4f12a9ce9850bf3a29380), that takes care of two archs, linux and mac.

nammn: How would that work? Wouldn't this script download the amd64 binary for our arm architecture?

viveksinghggits: no, in install.sh (run locally, not in evg), we check whether the arch is darwin or linux. If it's darwin we run `brew install`, if it's linux we run the `setup_chart_testing_cli.sh` to setup on linux machines.

.evergreen-functions.yml
viveksinghggits: I have made this part of the pre-commit and have kept `lint-chart` as well.

@m1kola for the second point that you mentioned, when we make this script part of pre-commit it would be automatically be part of the `lint_repo` task because it eventually just runs the pre-commit script. And I think it makes sense as well considering, we are actually linting the helm chart under `lint_repo`.

lsierant (approved):

.evergreen-functions.yml
m1kola: That is the issue - I think `lint_repo` is too loaded with checks so the output became overly difficult to comprehend for a linting task. Today when I see `lint_repo` failed it takes me more time than it should to find out what linting I messed up.

`lint_repo` checks (probably not the full list):
1. Release json/generated yamls
2. Licenses
3. Go linting
4. Python linting
5. Bash linting
6. Now helm chart as well.

I think each of these things should be a separate task so we can quickly glance at them and understand what has failed instead of going through logs searching for specific failures. It gets worse when you mess up multiple things at the same time (e.g. Python and bash linting).

That is why I was suggesting not to add more to this complexity. I was hoping that the new checks can go into separate jobs and we will eventually split existing linting into separate jobs.

I don't want to block this PR on this - feel free to merge. But I think this is the direction we should be going with our linting jobs.

MaciejKaras: I agree that it makes sense to separate those tasks, but setting new evergreen job per each linting tasks seems overkill. I would rather have them separated in bash scripts and then call them in separate evergreen commands/functions. This will help navigate in the task output and make it easier find the offending bash script/linting task.

scripts/dev/setup_chart_testing_cli.sh
viveksinghggits: @nammn is it ok we go ahead and merge this?

viveksinghggits: Other comments are almost resolved 😄.

PR #467, opened 2025-09-18
CLOUDP-342858: Running search test under OM 8.2

lsierant:
# Summary

This PR adds testing for running search with Ops Manager 8.0.14 and with mongod 8.2. 

Changes to search/search_enterprise_tls.py: 
* Added ability to run the test with either OM or cloud qa.
* Added mongod migration steps from 8.0.14 to 8.2.0. 
  * This adds automated verification that we can have search running and upgrade from mdb 8.0 (having polyfilled searchCoordinator role) to mdb 8.2 (with the role being a built-in one). 

## Proof of Work

EVG passing for search_enterprise_tls.py in both cloud-qa and om80 variants: [evg link](https://spruce.mongodb.com/version/68da6bb997a26c000793eff0/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&taskName=e2e_search_enterprise_tls)

Also the test is only run in om80 variant and not for earlier OM. 

## Checklist

### DIFF
--- a/docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
+++ b/docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
@@ -1,7 +1,6 @@
 import pymongo
 import yaml
-from kubernetes import client
-from kubetester import create_or_update_secret, run_periodically, try_load, wait_until
+from kubetester import create_or_update_secret, run_periodically, try_load
 from kubetester.certs import create_mongodb_tls_certs, create_tls_certs
 from kubetester.kubetester import KubernetesTester
 from kubetester.kubetester import fixture as yaml_fixture
@@ -212,11 +211,12 @@ def test_wait_for_mongod_parameters(mdb: MongoDB):
     run_periodically(check_mongod_parameters, timeout=200)
 
 
-# After picking up MongoDBSearch CR, MongoDB reconciler will add mongod parameters.
-# But it will not immediately mark the MongoDB CR as Pending
-# spinning
+# After picking up MongoDBSearch CR, MongoDB reconciler will add mongod parameters to each process.
+# Due to how MongoDB reconciler works (blocking on waiting for agents and not changing the status to pending)
+# the phase won't be updated to Pending and we need to wait by checking agents' status directly in OM.
 @mark.e2e_search_enterprise_tls
-def test_wait_for_database_resource_ready(mdb: MongoDB):
+def test_wait_for_agents_ready(mdb: MongoDB):
+    mdb.get_om_tester().wait_agents_ready()
     mdb.assert_reaches_phase(Phase.Running, timeout=300)
 
 
--- a/docker/mongodb-kubernetes-tests/kubetester/__init__.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/__init__.py
@@ -131,7 +131,6 @@ def create_or_update_configmap(
     data: Dict[str, str],
     api_client: Optional[kubernetes.client.ApiClient] = None,
 ) -> str:
-    print("Logging inside create_or_update configmap")
     try:
         create_configmap(namespace, name, data, api_client)
     except kubernetes.client.ApiException as e:
--- a/docker/mongodb-kubernetes-tests/kubetester/kubetester.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/kubetester.py
@@ -979,16 +979,6 @@ class KubernetesTester(object):
         return response.json()
 
     @staticmethod
-    def get_automation_status(group_id=None, group_name=None):
-        if group_id is None:
-            group_id = KubernetesTester.get_om_group_id(group_name=group_name)
-
-        url = build_automation_status_endpoint(KubernetesTester.get_om_base_url(), group_id)
-        response = KubernetesTester.om_request("get", url)
-
-        return response.json()
-
-    @staticmethod
     def get_monitoring_config(group_id=None):
         if group_id is None:
             group_id = KubernetesTester.get_om_group_id()
--- a/docker/mongodb-kubernetes-tests/kubetester/omtester.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/omtester.py
@@ -15,14 +15,22 @@ import pytest
 import requests
 import semver
 from kubetester.automation_config_tester import AutomationConfigTester
-from kubetester.kubetester import build_agent_auth, build_auth, run_periodically
+from kubetester.kubetester import (
+    KubernetesTester,
+    build_agent_auth,
+    build_auth,
+    run_periodically,
+)
 from kubetester.mongotester import BackgroundHealthChecker
 from kubetester.om_queryable_backups import OMQueryableBackup
 from opentelemetry import trace
 from requests.adapters import HTTPAdapter, Retry
+from tests import test_logger
 from tests.common.ops_manager.cloud_manager import is_cloud_qa
 
 skip_if_cloud_manager = pytest.mark.skipif(is_cloud_qa(), reason="Do not run in Cloud Manager")
+
+logger = test_logger.get_test_logger(__name__)
 
 
 class BackupStatus(str, Enum):
@@ -414,7 +422,7 @@ class OMTester(object):
                 span.set_attribute(key=f"mck.om.request.retries", value=retries - retry_count)
                 return resp
             except Exception as e:
-                print(f"Encountered exception: {e} on retry number {retries-retry_count}")
+                print(f"Encountered exception: {e} on retry number {retries - retry_count}")
                 span.set_attribute(key=f"mck.om.request.exception", value=str(e))
                 last_exception = e
                 time.sleep(1)
@@ -677,6 +685,42 @@ class OMTester(object):
     def api_update_version_manifest(self, major_version: str = "8.0"):
         body = requests.get(url=f"https://opsmanager.mongodb.com/static/version_manifest/{major_version}.json").json()
         self.om_request("put", "/versionManifest", json_object=body)
+
+    def api_get_automation_status(self) -> dict[str, str]:
+        return self.om_request("get", f"/groups/{self.context.project_id}/automationStatus").json()
+
+    def wait_agents_ready(self, timeout: Optional[int] = 600):
+        """Waits until all the agents reached the goal automation config version."""
+        log_prefix = f"[{self.context.group_name}/{self.context.project_id}] "
+
+        def agents_are_ready():
+            auto_status = self.api_get_automation_status()
+            goal_version = auto_status.get("goalVersion")
+
+            logger.info(f"{log_prefix}Checking if all agent processes have reached goal version: {goal_version}")
+            processes_not_ready = []
+            for process in auto_status.get("processes", []):
+                process_name = process.get("name", "unknown")
+                process_version = process.get("lastGoalVersionAchieved")
+                if process_version != goal_version:
+                    logger.info(
+                        f"{log_prefix}Process {process_name} at version {process_version}, expected {goal_version}"
+                    )
+                    processes_not_ready.append(process_name)
+
+            all_processes_ready = len(processes_not_ready) == 0
+            if all_processes_ready:
+                logger.info(f"{log_prefix}All agent processes have reached the goal version")
+            else:
+                logger.info(f"{log_prefix}{len(processes_not_ready)} processes have not yet reached the goal version")
+
+            return all_processes_ready
+
+        KubernetesTester.wait_until(
+            agents_are_ready,
+            timeout=timeout,
+            sleep_time=3,
+        )
 
 
 class OMBackgroundTester(BackgroundHealthChecker):


docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -115,6 +126,16 @@ def test_install_operator(namespace: str, operator_installation_config: dict[str
...
 
 
+@mark.e2e_search_enterprise_tls
+@skip_if_cloud_manager
+def test_create_ops_manager(namespace: str):
+    ops_manager = get_ops_manager(namespace)
+    if ops_manager is not None:
+        ops_manager.update()
+    ops_manager.om_status().assert_reaches_phase(Phase.Running, timeout=1200)
+    ops_manager.appdb_status().assert_reaches_phase(Phase.Running, timeout=600)
```
fealebenpae: Would `get_ops_manager()` ever return `None` provided this step is decorated with `skip_if_cloud_manager`? Or maybe I'm not understanding why we're guarding against `ops_manager` being `None` before calling `.update()` but not for the following method calls?

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -169,27 +190,117 @@ def test_create_search_resource(mdbs: MongoDBSearch):
...
+            )
+            pod_parameters.append(f"pod {idx} setParameter: {set_parameter}")
+
+        return parameters_are_set, f'Not all pods have mongot parameters set:\n{"\n".join(pod_parameters)}'
+
+    run_periodically(lambda: check_mongod_parameters(), timeout=200)
+
+
+@mark.e2e_search_enterprise_tls
+def test_wait_for_database_resource_ready2(mdb: MongoDB):
```
fealebenpae: Why do we wait again here?

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -1,19 +1,22 @@
...
+from kubetester import create_or_update_secret, run_periodically, try_load
 from kubetester.certs import create_mongodb_tls_certs, create_tls_certs
 from kubetester.kubetester import KubernetesTester
 from kubetester.kubetester import fixture as yaml_fixture
 from kubetester.mongodb import MongoDB
 from kubetester.mongodb_search import MongoDBSearch
 from kubetester.mongodb_user import MongoDBUser
+from kubetester.omtester import skip_if_cloud_manager
 from kubetester.phase import Phase
+from mypyc.irbuild.function import check_native_override
```
fealebenpae: This doesn't seem used.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -1,19 +1,22 @@
...
 from pytest import fixture, mark
 from tests import test_logger
 from tests.common.search import movies_search_helper
-from tests.common.search.movies_search_helper import SampleMoviesSearchHelper
 from tests.common.search.search_tester import SearchTester
-from tests.conftest import get_default_operator
+from tests.conftest import get_default_operator, get_issuer_ca_filepath
+from tests.opsmanager.conftest import custom_om_prev_version
```
fealebenpae: This doesn't seem used.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -169,27 +190,117 @@ def test_create_search_resource(mdbs: MongoDBSearch):
...
+            ADMIN_USER_NAME, ADMIN_USER_PASSWORD, "SCRAM-SHA-256", 1, ssl=True, tlsCAFile=get_issuer_ca_filepath()
+        )
+        # TODO check why assert version works without auth for 8.0 and not for 8.2
+        mdb_tester.assert_version(MDB_VERSION_WITH_BUILT_IN_ROLE)
+
+
+@mark.e2e_search_enterprise_tlssh
+def test_search_assert_search_query_2(mdb: MongoDB):
+    get_user_sample_movies_helper(mdb).assert_search_query(retry_timeout=60)
+
```
fealebenpae: I think we should move this as the last step in `TestUpgradeMongod` to make sure it doesn't get moved around.

docker/mongodb-kubernetes-tests/tests/search/om_deployment.py
```diff
@@ -0,0 +1,31 @@
...
+    resource: MongoDBOpsManager = MongoDBOpsManager.from_yaml(
+        yaml_fixture("om_ops_manager_basic.yaml"), namespace=namespace
+    )
+
+    if try_load(resource):
+        return resource
+
+    resource.set_version(get_custom_om_version())
+    resource.set_appdb_version(get_custom_appdb_version())
+    resource.allow_mdb_rc_versions()
```
mircea-cosbuc: Is `8.2` an RC version? Is it not available without enabling this?

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: you're right, removed the if. The annotation is sufficient.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: Yeah, it looks it's redundant. Removing it.

docker/mongodb-kubernetes-tests/tests/search/om_deployment.py
lsierant: Oh this is not necessary anymore. Thanks for flagging.

docker/mongodb-kubernetes-tests/kubetester/mongodb.py
```diff
@@ -247,6 +258,39 @@ def configure(
         self["spec"]["credentials"] = om.api_key_secret(self.namespace, api_client=api_client)
         return self
 
+    def configure_cloud_qa(
+        self,
+        project_name,
+        src_project_config_map_name: str = None,
```
m1kola: Nit: Doesn't look like we use `src_project_config_map_name` anywhere. Am I missing something?

docker/mongodb-kubernetes-tests/kubetester/mongodb.py
```diff
@@ -247,6 +258,39 @@ def configure(
...
+        api_client: Optional[client.ApiClient] = None,
+    ) -> MongoDB:
+        if "opsManager" in self["spec"]:
+            del self["spec"]["opsManager"]
+
+        if src_project_config_map_name is None and "cloudManager" in self["spec"]:
+            src_project_config_map_name = self["spec"]["cloudManager"]["configMapRef"]["name"]
+        else:
+            # my-project cm and my-credentials secret are created by scripts/evergreen/e2e/configure_operator.sh
+            src_project_config_map_name = "my-project"
```
m1kola: Nit: Seems like we can drop this branch since we are defaulting to `my-project` anyway in case of 404 below. Or alternatively drop the defaulting in 404 case?

docker/mongodb-kubernetes-tests/tests/search/fixtures/enterprise-replicaset-sample-mflix.yaml
```diff
@@ -19,16 +19,15 @@ spec:
...
-          containers:
-          - name: mongodb-enterprise-database
-            resources:
-              limits:
-                cpu: "2"
-                memory: 2Gi
-              requests:
-                cpu: "1"
-                memory: 1Gi
+  podSpec:
```
m1kola, line 22: Q: Could you please explain why change this fixture?

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -169,27 +187,112 @@ def test_create_search_resource(mdbs: MongoDBSearch):
 
 @mark.e2e_search_enterprise_tls
 def test_wait_for_database_resource_ready(mdb: MongoDB):
-    mdb.assert_abandons_phase(Phase.Running, timeout=300)
     mdb.assert_reaches_phase(Phase.Running, timeout=300)
```
m1kola: Nit: doesn't look like a very useful test. We do nothing here apart from the assertion. In my mind the test should perform some change and then assert some condition.

docker/mongodb-kubernetes-tests/kubetester/mongodb.py
```diff
@@ -230,7 +230,18 @@ def __repr__(self):
...
-        om: MongoDBOpsManager,
+        om: Optional[MongoDBOpsManager],
+        project_name: str,
+        api_client: Optional[client.ApiClient] = None,
+    ) -> MongoDB:
+        if om is not None:
+            return self.configure_ops_manager(om, project_name, api_client=api_client)
+        else:
+            return self.configure_cloud_qa(project_name, api_client=api_client)
+
```
lucian-tosa, line 246: nit: do we need this? the `opsManager` and `cloudManager` fields behave the same

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -169,27 +187,112 @@ def test_create_search_resource(mdbs: MongoDBSearch):
...
-        ), "mongot parameters not found in mongod config"
+            set_parameter = mongod_config.get("setParameter", {})
+            parameters_are_set = parameters_are_set and (
+                "mongotHost" in set_parameter and "searchIndexManagementHostAndPort" in set_parameter
+            )
+            pod_parameters.append(f"pod {idx} setParameter: {set_parameter}")
+
+        return parameters_are_set, f'Not all pods have mongot parameters set:\n{"\n".join(pod_parameters)}'
+
+    run_periodically(lambda: check_mongod_parameters(), timeout=200)
```
lucian-tosa: nit: why not use `wait_until`?

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -169,27 +187,112 @@ def test_create_search_resource(mdbs: MongoDBSearch):
...
+
+
+@mark.e2e_search_enterprise_tls
+def test_search_restore_sample_database(mdb: MongoDB):
+    get_admin_sample_movies_helper(mdb).restore_sample_database()
+
+
+@mark.e2e_search_enterprise_tls
+def test_search_create_search_index(mdb: MongoDB):
+    get_user_sample_movies_helper(mdb).create_search_index()
```
m1kola, line 235: I don't understand the correlation between test actions and the assertions here and in some other tests. Perhaps it is too granular split into separate tests?

I think it will be easier to follow if there is a test which performs an action and checks the outcome of the action.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lucian-tosa: I am confused how this works. The `()` are calling the function, so you are passing the result of that function, not the function itself

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -169,27 +187,112 @@ def test_create_search_resource(mdbs: MongoDBSearch):
...
+def test_search_assert_search_query(mdb: MongoDB):
+    get_user_sample_movies_helper(mdb).assert_search_query(retry_timeout=60)
+
+
+@mark.e2e_search_enterprise_tls
+class TestUpgradeMongod:
+    def test_check_polyfilled_role_in_ac(self, mdb: MongoDB):
+        custom_roles = mdb.get_automation_config_tester().automation_config.get("roles", [])
+        assert len(custom_roles) > 0
+        assert "searchCoordinator" in [role["role"] for role in custom_roles]
```
m1kola, line 262: Not clear to me - why we assert it here?

---

I had to look at the controller code to understand it better. I see that for DB versions prior to 8.2 we create a polyfill role. Would it then make more sense to assert this after creating the resource with `MDB_VERSION_WITHOUT_BUILT_IN_ROLE`? E.g. somewhere in `test_create_database_resource`?

And then in the upgrade test we can test that there is no polyfill role after the upgrade.

This way test actions are bundled together with a relevant assertions which makes it easier to understand. Also I think a comment with an explanation about the polyfill role wouldn't hurt somewhere around role assertions.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -169,27 +187,112 @@ def test_create_search_resource(mdbs: MongoDBSearch):
...
+
+    def test_mongod_version(self, mdb: MongoDB):
+        mdb.tester(ca_path=get_issuer_ca_filepath(), use_ssl=True).assert_version(MDB_VERSION_WITHOUT_BUILT_IN_ROLE)
```
m1kola, line 257: Not sure what is the value of this test as part of the upgrade test.

Perhaps it is also should be an assertion after initial creation  in `test_create_database_resource`? 

And then in the upgrade test we will be:
* Taking action: upgrading to a new version
* Making assertions: checking running state, custom roles, version, etc

m1kola (commented):
Looks good overall, but the tests are not easy to follow. Left some suggestions.

lucian-tosa (approved):

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
m1kola: > so you are passing the result of that function, not the function itself

There is a `lambda` function which calls `check_mongod_parameters`.

This can probably be simplified to be just `check_mongod_parameters` instead of `lambda: check_mongod_parameters()`

lucian-tosa: 🤦 didn't see the lambda, thank you
I agree having only `check_mongod_parameters` is simpler. I'll keep this thread open as a `nit`

docker/mongodb-kubernetes-tests/kubetester/mongodb.py
lsierant: Because cloud-qa and om have different pathways to provision the config map and credentials. For OM we use OM's CR. For cloud-qa we use setup_cloud_qa.py which is executed before the test pod is even deployed.

To unify both it's much more work, so it was easier to just put if in there.

docker/mongodb-kubernetes-tests/kubetester/mongodb.py
```diff
@@ -247,6 +258,39 @@ def configure(
...
+        else:
+            # my-project cm and my-credentials secret are created by scripts/evergreen/e2e/configure_operator.sh
+            src_project_config_map_name = "my-project"
+
+        try:
+            src_cm = read_configmap(self.namespace, src_project_config_map_name, api_client=api_client)
+        except client.ApiException as e:
+            if e.status == 404:
+                logger.debug("project config map is not specified, trying my-project as the source")
+                src_cm = read_configmap(self.namespace, "my-project", api_client=api_client)
```
anandsyncs: The whole `except` section is redundant due to default value assigned to `src_project_config_map_name`

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: no hard reason, KubernetesTester.wait_until suggests it's more about kubernetes, but here we wait on something different. 
And wait_until is a thin wrapper over run_periodically anyway.

docker/mongodb-kubernetes-tests/tests/search/fixtures/enterprise-replicaset-sample-mflix.yaml
lsierant: Because spec.statefulSet is not used in MongoDB, but in MongoDBMultiCluster only. It's available in MongoDB only by accident because it's part of DbCommon. 

We should unify this for sure, but it's much broader topic.

lsierant: We have this one: https://jira.mongodb.org/browse/CLOUDP-222185

lsierant: So essentially it was a bug, the resource requests were not applied and were ignored.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: it is useful, it's checking the state of the world. Not every test must change the world AND observe it as part of the same test case. Especially when you deal with multiple components deployed asynchronously. In this case we create users, search and mongodb which are reconciled by different controllers

Having smaller test cases makes it easier to develop, as you can retry smaller individual functions if needed.  Also we should consider all the functions top to bottom as part of one testing steps and story.

lsierant: In case of search, after it's deployed, the MongoDB reconciler will pick it up and perform changes to mongod - agents will perform rolling restart of mongods, even without changing to MongoDB CR at all.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: oh, the lambda is redundant, I'll remove it

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: I agree that this assertion alone is of little value as it stands as we deploy this version at the beginning and the assertion could be there. 

OTOH, I've prepared this *TestUpgradeMongod* to be a potentially reusable bit of upgrade test/steps. If you look at this assertion as part of the steps grouped under a single TestClass, then it's OK to test whether we run the test under specific circumstances.

Let me think about the overall structure, perhaps some comments would be needed here.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
m1kola: > it is useful, it's checking the state of the world. Not every test must change the world AND observe it as part of the same test case. Especially when you deal with multiple components deployed asynchronously. In this case we create users, search and mongodb which are reconciled by different controllers
> 
> Having smaller test cases makes it easier to develop, as you can retry smaller individual functions if needed. Also we should consider all the functions top to bottom as part of one testing steps and story.

This is an end to end test and we shouldn't be concerned about specific controllers or other implementation details. We should focus on how the system behaves as a whole instead.

Perhaps it is easier to develop/debug the tests like this, but tests structured this way leave the reader guessing. Even if you consider all the functions top to bottom it raises a question - why we do this check here? We already check ready state in `test_create_database_resource` - why check it again? You explained this later, but imagine not having this context.

> In case of search, after it's deployed, the MongoDB reconciler will pick it up and perform changes to mongod - agents will perform rolling restart of mongods, even without changing to MongoDB CR at all.

Given that explanation I would argue that in the end to end test we should be:
1. Deploying everything (the DB resource, user and the search)
2. Do the ready state waits and other checks after the waiting.

This is probably how the users will be using it (especially with GitOps). And it eliminates the questions about strange checks.

docker/mongodb-kubernetes-tests/kubetester/mongodb.py
lsierant: Good catch!

docker/mongodb-kubernetes-tests/kubetester/mongodb.py
lsierant: true, simplified it a bit

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: Your comments made me rethink this and I changed that function to `test_wait_for_agents_ready`. After search is deployed, mongod parameters are added to all processes and the agents are doing their work for at least few minutes. 
Due to how we report (or rather not doing properly) changes to Pending when reconciling, we cannot see the agents are doing anything by checking the spec.status alone. 

Therefore I've added in the test `mdb.get_om_tester().wait_agents_ready()` that checks if all the agents reached the goal state (it's exactly the same logic as the operator is waiting on in the reconcile loop). 

I hope now it's more clear what's happening.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: This is difficult to fix tbh and it would require a more broader alignment how we should write the test... 
I don't know how to fix this exact case, but it's just part of the testing story: in order to query the search we need the index in the first place.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: I've added some more comments around and reordered the steps a bit - PTAL.

docker/mongodb-kubernetes-tests/kubetester/mongodb.py
lsierant: I refactored this part a bit.

docker/mongodb-kubernetes-tests/kubetester/mongodb.py
lsierant: I've removed it.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -168,28 +187,128 @@ def test_create_search_resource(mdbs: MongoDBSearch):
...
+            )
+            set_parameter = mongod_config.get("setParameter", {})
+            parameters_are_set = parameters_are_set and (
+                "mongotHost" in set_parameter and "searchIndexManagementHostAndPort" in set_parameter
+            )
+            pod_parameters.append(f"pod {idx} setParameter: {set_parameter}")
+
+        return parameters_are_set, f'Not all pods have mongot parameters set:\n{"\n".join(pod_parameters)}'
+
+    run_periodically(check_mongod_parameters, timeout=200)
```
m1kola, line 211: I think both this and `test_wait_for_agents_ready` are testing implementation details. I understand that `test_wait_for_agents_ready` is necessary (at least for now, while we are not reporting `Pending`). But do we need to do `test_wait_for_mongod_parameters`?

I'm trying to look at the flow from the user perspective. For me as a user it is important that I can:

1. Deploy `MongoDB`
2. Deploy `MongoDBSearch`
3. Get some indication that everything is ready (we don't have that at the moment unfortunately as you explained in the comment for `test_wait_for_agents_ready`).
4. Create search index
5. And start searching.

Can we just drop `test_wait_for_mongod_parameters` and verify "I can search" functionality (you already do that later in the test)?

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
m1kola: I'm not going to insist on changing this. But IMO something like this makes more sense:

```python
def test_can_search(mdb: MongoDB):
    get_admin_sample_movies_helper(mdb).restore_sample_database()
    get_user_sample_movies_helper(mdb).create_search_index()
    get_user_sample_movies_helper(mdb).assert_search_query(retry_timeout=60)
```

You can immediately see that the user should be able to search after creating a search index.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
m1kola: After looking at this again I now feel like ideally we should not be checking for implementation details in an E2E test: we should not be concerned about polyfill role. It might be an useful unit/integration test, but on the E2E level we shouldn't wory about it. We care about an answer to the question - "does search work after upgrade?".

Not requesting you to change this (but I'll welcome it) as there is no harm in the additional checks apart from them adding a bit of extra complexity to the test.

lsierant: we don't have integration tests, and our e2e serves that exact purpose. We could consider having snippets tests as the real end-to-end tests where the outcome only matters. In our python tests we should check some implementation details, especially if they are relevant only when working with third party components.

m1kola: Just to reiterate: I'm not saying we should change this.

I still believe end to end tests should simulate end user experiences and implementation details like this are not relevant to the end user experience. But if you want to keep it - it is not a big deal.

> we don't have integration tests, and our e2e serves that exact purpose.

(Off topic) Arguably we do have integration tests. We do not distinguish them from unit tests, but there are definitively tests that cover multiple functions/packages at once (multiple parts of the system) as opposed to being isolated to a specific unit (as unit tests should).

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: This probably begs some longer discussion. I generally believe we should test it, even if it's testing some internal implementation assumptions.

I think we need a place to test our assumptions how some components few layers down the chain (operator -> OM -> agent -> mongod) are behaving as we exactly know what is there to be configured in order to enable the feature correctly. So for us it's more about testing the exact outcome of enabling search, rather than the search queries alone as those are tested by the search team mostly and we do the smoke tests here. 

Where could we test if we perform search-related configuration properly if not here?

PR #473, opened 2025-09-24
CLOUDP-347194 - enable Pod Security Admission at `restricted` level

MaciejKaras:
# Summary

Based on the [HELP-81729](https://jira.mongodb.org/browse/HELP-81729) ticket I investigated if our workloads align with `restricted` [Pod Security Standards](https://v1-32.docs.kubernetes.io/docs/concepts/security/pod-security-standards/) security level. Enforcing `restricted` cannot be guaranteed across our e2e setup — for example Istio injects `istio-init` containers that do not follow the restricted profile, and our `tests` pod does not meet the requirements either.

Because of this the `warn` mode for the `restricted` level is enabled instead of `enforce`. For the complex `e2e_om_ops_manager_backup_sharded_cluster` test, `restricted` is enforced on a single (non-multi) cluster so we can monitor PSS alignment. More on levels and modes [here](https://v1-32.docs.kubernetes.io/docs/concepts/security/pod-security-admission/).

Concretely, the more secure default `securityContext` settings applied are:
- Operator deployment: `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `seccompProfile.type: RuntimeDefault`
- Other workloads: `capabilities.drop: [ALL]` (container level), `seccompProfile.type: RuntimeDefault` (pod level)

If less restrictive settings are required, use `template` / `podTemplate` overrides.

- **Note on the latest revision**: commit `2d5ef4e` ("Apply remaining changes") inadvertently reset all 14 substantive files back to `master`, discarding the hardening and leaving only the changelog. This revision restores the full PSS work from parent `125366e`. The original request that triggered it — fixing lint issues — required no changes: `black`, `isort`, `shellcheck`, `gofmt`, and `go build` all pass clean on the PR-touched files (the remaining repo-wide `golangci-lint` and govulncheck findings are pre-existing in unmodified files, the latter tracked in #1189).

Three Security issues can be closed this PR:

- https://jira.mongodb.org/browse/ENGSEC-822
- https://jira.mongodb.org/browse/ENGSEC-821
- https://jira.mongodb.org/browse/ENGSEC-820

## Proof of Work

Passing CI for all tests that have warning + passing enforcement for `e2e_om_ops_manager_backup_sharded_cluster` test.

Example [warning](https://parsley-beta.corp.mongodb.com/evergreen/mongodb_kubernetes_e2e_multi_cluster_om_appdb_e2e_om_ops_manager_backup_tls_patch_58f35690937269c5e6028d08a3abe3f49d0d09a1_68d3e3c4a81aa80007f67b5f_25_09_24_12_27_51/0/task?bookmarks=0,34024&shareLine=33307) from the `e2e_om_ops_manager_backup_tls` test:
```
[2025/09/24 14:59:58.691] Warning: would violate PodSecurity "restricted:latest": unrestricted capabilities (container "istio-init" must not include "NET_ADMIN", "NET_RAW" in securityContext.capabilities.add), runAsNonRoot != true (container "istio-init" must not set securityContext.runAsNonRoot=false), runAsUser=0 (container "istio-init" must not set runAsUser=0)
```

## Checklist


controllers/operator/construct/database_construction_test.go
```diff
@@ -45,6 +45,9 @@ func Test_buildDatabaseInitContainer(t *testing.T) {
 		SecurityContext: &corev1.SecurityContext{
 			ReadOnlyRootFilesystem:   ptr.To(true),
 			AllowPrivilegeEscalation: ptr.To(false),
+			Capabilities: &corev1.Capabilities{
```
lsierant, line 48: is there any potential that adding this default will break *any* customer's workload (or rather prevent the operator from deploying or the workload sts from restarting) and will require some manual intervention? Just thinking about our semver guarantees.

MaciejKaras: I think this is our own deployment, that we manage. If the customer wants a managedSecurityContext they are allowed to, but otherwise we should be able to modify the one we provide.

cc @mircea-cosbuc

mircea-cosbuc: This is our defaults. Only problem I see with this is some customers now requiring explicitly setting capabilities.

MaciejKaras: Maybe this should be consider as a security fix? In that case we should be able to overwrite our defaults if they are not secure even if this forces customers to explicitly specify custom capabilities. What do you think?

lsierant: Still, do we force customers (which don't care about it) to do any manual fix when upgrading? If yes, we need to bump major.

MaciejKaras: I don't understand what will we break here. We are changing our default SecurityContext for operator and other pods created. If customer wants to have dedicated SecurityContext or PodSecurityContext they need to set `MANAGED_SECURITY_CONTEXT` env var and our defaults will be completely overwritten. If they don't set `MANAGED_SECURITY_CONTEXT` every change they make to SecurityContext manually will be overwritten by our defaults. 

Code that handles securityContext settings:
https://github.com/mongodb/mongodb-kubernetes/blob/917723bab0811cf016b4b4a9533189017d4f3031/mongodb-community-operator/pkg/kube/podtemplatespec/podspec_template.go#L312-L322

MaciejKaras: We have discussed with @lsierant that the change for the more strict Capabilities is only applied on the db/om containers, not the whole Pod. This will not affect other containers in the Pod i.e. security, istio sidecars that customer can have. The only change on the Pod level is adding `seccompProfile: type: RuntimeDefault`.  We can do two things with it:
- move the `seccompProfile: type: RuntimeDefault` to container level and don't specify it on pod level. We will have our containers with secure seccomp settings, but if customer will add any sidecar to it it will not have seccomp settings applied
- keep it as is and secure entire Pod

@mircea-cosbuc looking for guidance here how to proceed

mircea-cosbuc: I think it's best to set it at pod level. I think based on @lsierant this needs clarity on what customers might need to change on upgrade (if anything), outlining those scenarios and deciding if it's a breaking change.

MaciejKaras: I've checked what are the consequences for using `seccomp: type: RuntimeDefault` and it defaults to what container runtime is used. Containerd and docker for example have very similar default seccomp profile -> https://docs.docker.com/engine/security/seccomp/#significant-syscalls-blocked-by-the-default-profile

Based on what I have found in official [Kubernetes docs](https://kubernetes.io/blog/2021/08/25/seccomp-default/#seccompdefault-to-the-rescue):
> These profiles may differ between runtimes like [CRI-O](https://github.com/cri-o/cri-o/blob/fe30d62/vendor/github.com/containers/common/pkg/seccomp/default_linux.go#L45) or [containerd](https://github.com/containerd/containerd/blob/e1445df/contrib/seccomp/seccomp_default.go#L51). They also differ for its used hardware architectures. But generally speaking, those default profiles allow a common amount of syscalls while blocking the more dangerous ones, which are unlikely or unsafe to be used in a containerized application.

Additionally on Red Hat OpenShift Container Platform `RuntimeDefault` is often enforced by default via Security Context Constraints (SCCs).

To summarise it is unlikely that users of our Operator require more syscalls permissions **in MongoDB workloads** than what is allowed by `RuntimeDefault` seccomp. Nevertheless I should add comment in the changelog how to mitigate securityContext defaults by using `managedSecurityContext`.

@lsierant @mircea-cosbuc let me know if that justifies approving PR. I have already edited changelog.

changelog/20250924_fix_provide_more_secure_pod_and_container.md
```diff
@@ -0,0 +1,17 @@
...
+   - `allowPrivilegeEscalation: false`
+   - `capabilities.drop: [ ALL ]`
+   - `seccompProfile.type: RuntimeDefault`
+
+  Other workloads:
+   - `capabilities.drop: [ ALL ]` - container level
+   - `seccompProfile.type: RuntimeDefault` - pod level
+
+> **Note**: If you require less restrictive `securityContext` settings please use `template` or `podTemplate` overrides.
+> Detailed information about overrides can be found in [Modify Ops Manager or MongoDB Kubernetes Resource Containers](https://www.mongodb.com/docs/kubernetes/current/tutorial/modify-resource-image/).
```
vinilage: LGTM!

vinilage (approved):
LGTM!

mircea-cosbuc (approved):

MaciejKaras:
@copilot fix lint issues in the PR besides the go vulnerabilities. Those will be fixed in #1189

docker/mongodb-kubernetes-tests/tests/opsmanager/om_ops_manager_backup_sharded_cluster.py
```diff
@@ -33,15 +33,25 @@ def s3_bucket(aws_s3_client: AwsS3Client, namespace: str) -> Iterator[str]:
     yield from create_s3_bucket(aws_s3_client, "test-bucket-sharded-")
 
 
+@fixture(scope="module")
+def enforced_pss_namespace(namespace: str) -> str:
+    # Change pod-security mode from warn to enforce. This will make test fail if operator and deployments don't support enforce mode
+    # This will not work in multi-cluster, because Istio injects sidecar and that breaks restricted level
```
lucian-tosa, line 39: as a sanity check, we can enforce the restricted pss on a no_mesh multi-cluster test

helm_chart/tests/operator_security_context_test.yaml
```diff
@@ -87,8 +99,22 @@ tests:
...
+      - exists:
           path: spec.template.spec.containers[?(@.name=="mongodb-kubernetes-operator-multi-cluster")].securityContext
+      - equal:
+          path: spec.template.spec.containers[?(@.name=="mongodb-kubernetes-operator-multi-cluster")].securityContext.allowPrivilegeEscalation
+          value: false
+      - exists:
+          path: spec.template.spec.containers[?(@.name=="mongodb-kubernetes-operator-multi-cluster")].securityContext.capabilities.drop
+      - equal:
+          path: spec.template.spec.containers[?(@.name=="mongodb-kubernetes-operator-multi-cluster")].securityContext.capabilities.drop
+          value: [ALL]
```
lucian-tosa, line 117: I'm very confused as to how this changed (and it passes) if values-multi-cluster.yaml and the template remains unchanged

docker/mongodb-kubernetes-tests/kubetester/helm.py
```diff
@@ -36,7 +36,8 @@ def helm_template(
     args = ("helm", "template", *command_args, helm_chart_path)
     logger.info(" ".join(args))
 
-    yaml_file_name = "{}.yaml".format(str(uuid.uuid4()))
+    home = os.getenv("HOME")
+    yaml_file_name = os.path.join(home, "{}.yaml".format(str(uuid.uuid4())))
```
lucian-tosa: this would create conflicts if running on different worktrees

docker/mongodb-kubernetes-tests/tests/opsmanager/om_ops_manager_backup_sharded_cluster.py
MaciejKaras: I was able to actually enforce the pss restricted for mesh mc tests as well in the https://github.com/mongodb/mongodb-kubernetes/pull/1199/changes. The plan is to merge it afterwards.

docker/mongodb-kubernetes-tests/kubetester/helm.py
MaciejKaras: Nice catch!

helm_chart/tests/operator_security_context_test.yaml
MaciejKaras: That's weird, I'll investigate it further

MaciejKaras: It's really weird, the operator.yaml did change, but the PR diff in GitHub UI does not show it !

MaciejKaras: Ok, I see it now. The changes to security-context are present only in the values.yaml. If you pass an extra file like -f values-specific.yaml, Helm merges it on top of the default values. So values-specific.yaml is an override layer, not a replacement.

I've tested it locally and verified it.

docker/mongodb-kubernetes-tests/kubetester/helm.py
MaciejKaras: Reverted completely

lucian-tosa (approved):

PR #477, opened 2025-09-25
Fix `private_gke_code_snippets` + other small fixes

MaciejKaras:
# Summary

This PR fixes `private_gke_code_snippets` tests that could not properly set the context:

>scripts/dev/contexts/private_gke_code_snippets: line 13: /data/mci/33f77d5e70bc4d8e19bb3f674e0743b1/src/github.com/mongodb/mongodb-kubernetes/scripts/dev/contexts/variables/funcs/gke: No such file or directory

The reason was `om80` context file replaced the `script_dir` variable that was also set and used in `private_gke_code_snippets`. After that the `script_dir` pointed to `./scripts/dev/contexts/variables` instead of `./scripts/dev/contexts/` and the `gke` script could not be found. The solution was to replace `script_dir` usage in `om80` script with `PROJECT_DIR`. It was also replaced in other scripts which had long path based on `script_dir`.

The other thing that was fixed is `switch_context` function. Previously because we concatenated result from multiple bash commands, we didn't consume the error code from actual command:

```
      script: |
        echo "Switching context"
        scripts/dev/switch_context.sh "${build_variant}"
        echo "Finished switching context"
```

## Proof of Work

Passing CI + manual patch for [private_gke_code_snippets](https://spruce.mongodb.com/version/68d50ec1cafe8c000760009e) .

## Checklist

### DIFF
--- a/.evergreen-functions.yml
+++ b/.evergreen-functions.yml
@@ -301,7 +301,6 @@ functions:
   # the task configures the set of tools necessary for any task working with K8 cluster:
   # installs kubectl, jq, kind (if necessary), configures docker authentication
   download_kube_tools:
-    - *switch_context
     - *setup_kubectl
     - *setup_jq
     # we need aws to configure docker authentication


scripts/dev/contexts/funcs/gke
```diff
@@ -1,3 +1,7 @@
+#!/usr/bin/env bash
+
+set -Eeou pipefail
```
lsierant, line 3: this is not necessary as the file is only sourced, not executed directly

MaciejKaras: True! At the same time it helps mark the file in IDE as bash script ;)

m1kola: > True! At the same time it helps mark the file in IDE as bash script ;)

Not requesting to change this. Just joining the chat :)

From my experience adding a shebang is usually enough for IDEs.

scripts/dev/contexts/variables/om60
```diff
@@ -2,12 +2,9 @@
 
 set -Eeou pipefail
 
-script_name=$(readlink -f "${BASH_SOURCE[0]}")
-script_dir=$(dirname "${script_name}")
-
 CUSTOM_OM_PREV_VERSION=6.0.0
 export CUSTOM_OM_PREV_VERSION
-CUSTOM_OM_VERSION=$(grep -E "^\s*-\s*&ops_manager_60_latest\s+(\S+)\s+#" <"${script_dir}"/../../../../.evergreen.yml | awk '{print $3}')
+CUSTOM_OM_VERSION=$(grep -E "^\s*-\s*&ops_manager_60_latest\s+(\S+)\s+#" <"${PROJECT_DIR}"/.evergreen.yml | awk '{print $3}')
```
m1kola, line 7: Not something for this PR, but I think we should switch for yq/jq for reading yaml/json instead of awk.

m1kola (approved):

scripts/dev/contexts/funcs/gke
MaciejKaras: Yeah, I thought that Łukasz commented all added lines. But I agree pipefail is not needed, but also harmless ;)

PR #486, opened 2025-10-01
[DO NOT REVIEW] WIP

Julien-Ben:
# Summary

Examining the differences between the controllers of single- and multi-cluster replica sets.
Clean up the single cluster controller reconcile loop.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,3 +93,7 @@ docs/**/test.sh.run.log
 dist
 logs
 *.run.log
+
+# AI
+.mcp.json
+CLAUDE.md
--- a/controllers/om/replicaset/om_replicaset.go
+++ b/controllers/om/replicaset/om_replicaset.go
@@ -30,6 +30,16 @@ func BuildFromStatefulSetWithReplicas(mongoDBImage string, forceEnterprise bool,
 	return rsWithProcesses
 }
 
+// BuildFromMongoDBWithReplicas returns a replica set that can be set in the Automation Config
+// based on the given MongoDB resource directly without requiring a StatefulSet.
+func BuildFromMongoDBWithReplicas(mongoDBImage string, forceEnterprise bool, mdb *mdbv1.MongoDB, replicas int, fcv string, tlsCertPath string) om.ReplicaSetWithProcesses {
+	members := process.CreateMongodProcessesFromMongoDB(mongoDBImage, forceEnterprise, mdb, replicas, fcv, tlsCertPath)
+	replicaSet := om.NewReplicaSet(mdb.Name, mdb.Spec.GetMongoDBVersion())
+	rsWithProcesses := om.NewReplicaSetWithProcesses(replicaSet, members, mdb.Spec.GetMemberOptions())
+	rsWithProcesses.SetHorizons(mdb.Spec.GetHorizonConfig())
+	return rsWithProcesses
+}
+
 // PrepareScaleDownFromMap performs additional steps necessary to make sure removed members are not primary (so no
 // election happens and replica set is available) (see
 // https://jira.mongodb.org/browse/HELP-3818?focusedCommentId=1548348 for more details)
--- a/controllers/om/replicaset/om_replicaset.go
+++ b/controllers/om/replicaset/om_replicaset.go
@@ -97,8 +97,8 @@ func PrepareScaleDownFromMap(omClient om.Connection, rsMembers map[string][]stri
 	return nil
 }
 
-func PrepareScaleDownFromStatefulSet(omClient om.Connection, statefulSet appsv1.StatefulSet, rs *mdbv1.MongoDB, log *zap.SugaredLogger) error {
-	_, podNames := dns.GetDnsForStatefulSetReplicasSpecified(statefulSet, rs.Spec.GetClusterDomain(), rs.Status.Members, nil)
+func PrepareScaleDownFromMongoDB(omClient om.Connection, rs *mdbv1.MongoDB, log *zap.SugaredLogger) error {
+	_, podNames := dns.GetDNSNames(rs.Name, rs.ServiceName(), rs.Namespace, rs.Spec.GetClusterDomain(), rs.Status.Members, rs.Spec.DbCommonSpec.GetExternalDomain())
 	podNames = podNames[scale.ReplicasThisReconciliation(rs):rs.Status.Members]
 
 	if len(podNames) != 1 {
--- a/controllers/operator/mongodbmultireplicaset_controller.go
+++ b/controllers/operator/mongodbmultireplicaset_controller.go
@@ -152,6 +152,7 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 	if err != nil {
 		return r.updateStatus(ctx, &mrs, workflow.Failed(xerrors.Errorf("error establishing connection to Ops Manager: %w", err)), log)
 	}
+	// TODO: need to ensure supported OM version ?
 
 	log = log.With("MemberCluster Namespace", mrs.Namespace)
 
@@ -165,11 +166,9 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 	}
 
 	r.SetupCommonWatchers(&mrs, nil, nil, mrs.Name)
+	// TODO: Need equivalent of validateMongoDBResource and checkIfHasExcessProcesses here ?
 
-	publishAutomationConfigFirst, err := r.publishAutomationConfigFirstMultiCluster(ctx, &mrs, log)
-	if err != nil {
-		return r.updateStatus(ctx, &mrs, workflow.Failed(err), log)
-	}
+	// TODO: ensure feature controls ?
 
 	// If tls is enabled we need to configure the "processes" array in opsManager/Cloud Manager with the
 	// correct tlsCertPath, with the new tls design, this path has the certHash in it(so that cert can be rotated
@@ -192,6 +191,8 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 		}
 	}
 
+	// TODO: agentCertSecretSelector ?
+
 	agentCertSecretName := mrs.GetSecurity().AgentClientCertificateSecretName(mrs.GetName())
 	agentCertHash, agentCertPath := r.agentCertHashAndPath(ctx, log, mrs.Namespace, agentCertSecretName, "")
 
@@ -208,6 +209,12 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 		if automationConfigError != nil {
 			log.Errorf("Recovery failed because of Automation Config update errors, %w", automationConfigError)
 		}
+	}
+
+	// TODO: review conditions for publishing AC first
+	publishAutomationConfigFirst, err := r.publishAutomationConfigFirstMultiCluster(ctx, &mrs, log)
+	if err != nil {
+		return r.updateStatus(ctx, &mrs, workflow.Failed(err), log)
 	}
 
 	status := workflow.RunInGivenOrder(publishAutomationConfigFirst,
@@ -260,6 +267,10 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 	return r.updateStatus(ctx, &mrs, workflow.OK(), log, mdbstatus.NewPVCsStatusOptionEmptyStatus())
 }
 
+// TODO: in multi cluster, the below function only checks a subset of conditions compared to the shared "publishAutomationConfigFirst" in common_controller.go,
+//
+//	which is used for single cluster replica set. Why, and are these enough ?
+//
 // publishAutomationConfigFirstMultiCluster returns a boolean indicating whether Ops Manager
 // needs to be updated before the StatefulSets are created for this resource.
 func (r *ReconcileMongoDbMultiReplicaSet) publishAutomationConfigFirstMultiCluster(ctx context.Context, mrs *mdbmultiv1.MongoDBMultiCluster, log *zap.SugaredLogger) (bool, error) {
--- a/controllers/operator/mongodbmultireplicaset_controller.go
+++ b/controllers/operator/mongodbmultireplicaset_controller.go
@@ -530,6 +530,7 @@ func (r *ReconcileMongoDbMultiReplicaSet) reconcileStatefulSets(ctx context.Cont
 			}
 		}
 
+		// TODO: make a separate function similar to `buildStatefulSetOptions` in single rs controller?
 		opts := mconstruct.MultiClusterReplicaSetOptions(
 			mconstruct.WithClusterNum(clusterNum),
 			Replicas(replicasThisReconciliation),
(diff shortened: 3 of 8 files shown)


api/v1/search/zz_generated.deepcopy.go
```diff
@@ -159,7 +159,7 @@ func (in *MongoDBSearchSpec) DeepCopyInto(out *MongoDBSearchSpec) {
 		*out = new(v1.ResourceRequirements)
 		(*in).DeepCopyInto(*out)
 	}
-	out.Security = in.Security
+	in.Security.DeepCopyInto(&out.Security)
```
m1kola: This file seems to be generated. Why did it change?

.gitignore
```diff
@@ -93,3 +93,7 @@ docs/**/test.sh.run.log
 dist
 logs
 *.run.log
+
+# AI
+.mcp.json
+CLAUDE.md
```
m1kola: I would suggest to submit a separate tiny PR for this.

api/v1/search/zz_generated.deepcopy.go
Julien-Ben: Good catch, I had a controller gen version mismatch at some point. Reverted

.gitignore
Julien-Ben: Makes sense, reverted

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -210,6 +205,11 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 		}
 	}
 
+	publishAutomationConfigFirst, err := r.publishAutomationConfigFirstMultiCluster(ctx, &mrs, log)
```
Julien-Ben, line 208: Brought it closer to where it's actually used

controllers/om/replicaset/om_replicaset.go
```diff
@@ -65,30 +75,13 @@ func PrepareScaleDownFromMap(omClient om.Connection, rsMembers map[string][]stri
 		log.Debugw("Marked replica set members as non-voting", "replica set with members", rsMembers)
 	}
 
-	// TODO practice shows that automation agents can get stuck on setting db to "disabled" also it seems that this process
```
Julien-Ben, line 68: This comment was old (<2022)

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -108,6 +117,7 @@ func newReplicaSetReconciler(ctx context.Context, kubeClient client.Client, imag
 // Reconcile reads that state of the cluster for a MongoDbReplicaSet object and makes changes based on the state read
 // and what is in the MongoDbReplicaSet.Spec
 func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reconcile.Request) (res reconcile.Result, e error) {
+	// === 1. Initial Checks and setup
```
Julien-Ben, line 120: I added these as helpers for myself at the beginning of the refactoring, to keep track of where I was in the hundreds of lines of the reconcile loop

I'm okay to remove them now if they feel like noise more than useful comments

controllers/om/process/om_process_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/mongodb/mongodb-kubernetes/pkg/util/maputil"
+)
+
+const (
+	defaultMongoDBImage = "mongodb/mongodb-enterprise-server:7.0.0"
+	defaultFCV          = "7.0"
+	defaultNamespace    = "test-namespace"
+)
+
+func TestCreateMongodProcessesFromMongoDB(t *testing.T) {
```
m1kola, line 19: This test and `TestCreateMongodProcessesFromMongoDB_AdditionalConfig` seem to be testing `om.NewMongodProcess` (already has own unit tests) and `dns.GetDNSNames` (doesn't have own unit tests).

I think it is better to unit-test `om.NewMongodProcess` and `dns.GetDNSNames` separately. Once you have unit tests for these building blocks you wouldn't need to test the integration so thoroughly.

This also would (indirectly) cover `WaitForRsAgentsToRegisterByResource` and other places where these functions used.

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -155,153 +165,99 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
...
-
-	internalClusterCertPath := ""
-	if internalClusterCertHash != "" {
-		internalClusterCertPath = fmt.Sprintf("%s%s", util.InternalClusterAuthMountPath, internalClusterCertHash)
+	deploymentOpts := deploymentOptionsRS{
+		prometheusCertHash:   prometheusCertHash,
+		agentCertPath:        agentCertPath,
+		agentCertHash:        agentCertHash,
+		currentAgentAuthMode: currentAgentAuthMode,
 	}
```
m1kola, line 216: Nit: I think a long list of arguments is better than a struct. It is easier to pass a struct around, but arguments are more explicit: you only pass what you need in a given function. With plain arguments it is easier to spot unused arguments as well.

 There is also no question like - why some cert related stuff is in the struct, but `tlsCertPath` and `internalClusterCertPath` is not?

I would leave arguments as is.

lucian-tosa: I agree with Mikalai here, it is a bit confusing

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -333,6 +289,57 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
 	return r.updateStatus(ctx, rs, workflow.OK(), log, mdbstatus.NewBaseUrlOption(deployment.Link(conn.BaseURL(), conn.GroupID())), mdbstatus.MembersOption(rs), mdbstatus.NewPVCsStatusOptionEmptyStatus())
 }
 
+func publishAutomationConfigFirstRS(ctx context.Context, getter kubernetesClient.Client, mdb mdbv1.MongoDB, lastSpec *mdbv1.MongoDbSpec, currentAgentAuthMode string, sslMMSCAConfigMap string, log *zap.SugaredLogger) bool {
```
lucian-tosa, line 291: Why define a new function here? There is a very similar one in common controller?

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: That's a valid concern, I hesitated and did it mostly to match what we do in the sharded cluster controller:

```
type deploymentOptions struct {
	podEnvVars           *env.PodEnvVars
	currentAgentAuthMode string
	caFilePath           string
	agentCertPath        string
	agentCertHash        string
	certTLSType          map[string]bool
	finalizing           bool
	processNames         []string
	prometheusCertHash   string
}
```

If we decide to keep the struct, I will put more variables into it for sure while working on the unification, for more consistency.

I'm happy to hear the team's opinion about this

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -488,22 +617,22 @@ func (r *ReconcileMongoDbReplicaSet) updateOmDeploymentRs(ctx context.Context, c
...
 		secretsClient:      r.SecretClient,
 		namespace:          rs.GetNamespace(),
-		prometheusCertHash: prometheusCertHash,
+		prometheusCertHash: deploymentOptionsRS.prometheusCertHash,
 	}
 
 	err = conn.ReadUpdateDeployment(
 		func(d om.Deployment) error {
-			if shouldMirrorKeyfileForMongot {
+			if shouldMirrorKeyfile {
```
lucian-tosa: nit: Why did you rename this variable? It might be a bit too vague now

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -155,153 +165,99 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
...
-	)
-
-	caFilePath := fmt.Sprintf("%s/ca-pem", util.TLSCaMountPath)
-
-	if err := r.reconcileHostnameOverrideConfigMap(ctx, log, r.client, *rs); err != nil {
-		return r.updateStatus(ctx, rs, workflow.Failed(xerrors.Errorf("Failed to reconcileHostnameOverrideConfigMap: %w", err)), log)
+	prometheusCertHash, err := certs.EnsureTLSCertsForPrometheus(ctx, r.SecretClient, rs.GetNamespace(), rs.GetPrometheus(), certs.Database, log)
+	if err != nil {
+		log.Infof("Could not generate certificates for Prometheus: %s", err)
+		return r.updateStatus(ctx, rs, workflow.Failed(err), log)
```
anandsyncs:
```suggestion
		return r.updateStatus(ctx, rs, workflow.Failed(xerrors.Errorf("could not generate certificates for Prometheus: %w", err)), log)
```

other errors are more descriptive, is it better to clarify this one too?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -333,6 +289,57 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
...
+		}
+
+		log.Debugw(fmt.Sprintf("Error getting StatefulSet %s", namespacedName), "error", err)
+		return false
+	}
+
+	databaseContainer := container.GetByName(util.DatabaseContainerName, currentSts.Spec.Template.Spec.Containers)
+	volumeMounts := databaseContainer.VolumeMounts
+
+	if !mdb.Spec.Security.IsTLSEnabled() && wasTLSSecretMounted(ctx, getter, currentSts, mdb, log) {
```
anandsyncs, line 308: We probably need a nil check on `.Security`

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -449,36 +567,47 @@ func AddReplicaSetController(ctx context.Context, mgr manager.Manager, imageUrls
 
 // updateOmDeploymentRs performs OM registration operation for the replicaset. So the changes will be finally propagated
 // to automation agents in containers
-func (r *ReconcileMongoDbReplicaSet) updateOmDeploymentRs(ctx context.Context, conn om.Connection, membersNumberBefore int, rs *mdbv1.MongoDB, set appsv1.StatefulSet, log *zap.SugaredLogger, agentCertPath, caFilePath, tlsCertPath, internalClusterCertPath string, prometheusCertHash string, isRecovering bool, shouldMirrorKeyfileForMongot bool) workflow.Status {
+func (r *ReconcileMongoDbReplicaSet) updateOmDeploymentRs(ctx context.Context, conn om.Connection, membersNumberBefore int, rs *mdbv1.MongoDB, log *zap.SugaredLogger, tlsCertPath, internalClusterCertPath string, deploymentOptionsRS deploymentOptionsRS, shouldMirrorKeyfile bool, isRecovering bool) workflow.Status {
```
anandsyncs: `updateOmDeploymentRs` is carrying a lot of responsibilities: waiting for agents, TLS disable coordination, replica-set building, authentication sync, automation-config reconciliation, and cleanup. 
 
```go
func (r *ReconcileMongoDbReplicaSet) updateOmDeploymentRs(
ctx context.Context,
conn om.Connection,
membersBefore int,
rs *mdbv1.MongoDB,
log *zap.SugaredLogger,
tlsCertPath, internalClusterCertPath string,
opts deploymentOptionsRS,
shouldMirrorKeyfile, isRecovering bool,
) workflow.Status {
inputs, status := r.prepareOmDeploymentInputs(ctx, conn, membersBefore, rs, tlsCertPath, opts, shouldMirrorKeyfile, isRecovering, log)
if !status.IsOK() {
return status
}
 
if status := r.applyReplicaSetAutomation(ctx, conn, rs, inputs, internalClusterCertPath, log); !status.IsOK() {
return status
}
 
return r.finalizeOmDeployment(ctx, conn, rs, membersBefore, inputs, internalClusterCertPath, isRecovering, log)
}
```
 
Each helper focuses on one concern:
 
```go
type omDeploymentInputs struct {
replicasTarget           int
replicaSet               replicaset.ReplicaSet
processNames             []string
prometheusConfiguration  PrometheusConfiguration
additionalReconcileNeeded bool
}
 
func (r *ReconcileMongoDbReplicaSet) prepareOmDeploymentInputs(...) (omDeploymentInputs, workflow.Status) { /* wait for agents, handle TLS disable, build replicaSet */ }
 
func (r *ReconcileMongoDbReplicaSet) applyReplicaSetAutomation(...) workflow.Status { /* auth reconciliation + ReadUpdateDeployment */ }
 
func (r *ReconcileMongoDbReplicaSet) finalizeOmDeployment(...) workflow.Status { /* wait for ready, log rotate, host diff, backup */ }
```

controllers/operator/mongodbreplicaset_controller.go
anandsyncs: I would say that even this is too long, something like `applyOverrides` would be sufficient.
If you feel that this would not make the purpose of the variable clear enough it is because the function is too big and a lot is going on here.
It should ideally be broken into smaller chunks called by an orchestrator function so that flow of logic is easier to follow.

I have posted a comment above about this.

https://go.dev/wiki/CodeReviewComments#variable-names

controllers/operator/mongodbreplicaset_controller.go
anandsyncs: You can also use parameters instead of `omDeploymentInputs`

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: In the spirit of this PR, for two reasons: better separation of concerns, and to follow multi cluster controller pattern.

The existing function in common controller needs the `rsConfig` object, which we don't need to compute in the main reconcile loop, we can do it when we reconcile kubernetes resources, inside the dedicated function.

So I rewrote this one so that it only needs the mdb (`mdbv1.MongoDB`) resource (and last specs).

I followed the pattern that is in multi cluster rs.

On top of that it will give me more flexibility for the actual unification

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: I took it from the common controller, I haven't written it myself, so I will mark it as follow up, but I think we have guardrails against nil Security at initialization, otherwise we would have observed panics already 😅

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: Same remark as in the other comment, I'll mark it as follow up to not increase the size of the PR, since I just slightly modified the method here.
But if I have time during the unification implementation, I'll make sure to do it ! Thanks for the suggestion

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: Reverted

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: good catch, done

controllers/om/process/om_process_test.go
Julien-Ben: Great point, I tested the underlying function instead

controllers/operator/mongodbreplicaset_controller.go
m1kola: I don't find any initialisation. Most of the time we do `mdb.GetSecurity()` which returns an empty `Security` value if `mdb.Spec.Security` is `nil`.

PR #487, opened 2025-10-01
Fix linting `ct lint` issue in helm chart

viveksinghggits:
# Summary

Update:
Thanks to @MaciejKaras for the comment [here](https://github.com/mongodb/mongodb-kubernetes/pull/487#pullrequestreview-3293073690) and it looks like we were trying to lint the helm chart in parallel to other jobs that were trying to update the helm chart. This PR tries to change that and runs the helm chart linting after the updates to helm chart are made by `update_jobs` job. 

**Old Desc:**
The chart linting command that we run as part of lint_repo task

```
ct lint --charts="${PROJECT_DIR}/helm_chart/" \
    --chart-yaml-schema "${PROJECT_DIR}/helm_chart/tests/schemas/chart_schema.yaml" \
    --lint-conf "${PROJECT_DIR}/helm_chart/tests/schemas/lintconf.yaml"
```

seems to be flakey. It doesn't seem to fail regularly but we have seen issues that are [reported here](https://mongodb.slack.com/archives/CGLP6R2PQ/p1759148259540819).  If we look into the error 
```
[ERROR] templates/: template: mongodb-kubernetes/templates/secret-config.yaml:1:14: executing "mongodb-kubernetes/templates/secret-config.yaml" at <.Values.operator.vaultSecretBackend>: nil pointer evaluating interface
```
It looks like `.Values` object that is passed to the template file doesn't have the field `.Values.operator.vaultSecretBackend`. And the only places where the values.yaml doesn't have `vaultSecretBackend` defined are the files `values-openshift.yaml` and `values-multi-cluster.yaml`. 
And I suspect that 
- Either these values files are merged and then passed to `helm template`, and merge doesn't work properly and eventually `.Values.operator` has the value that is defined in either `values-openshift.yaml` or in `values-multi-cluster.yaml` (and it doesn't have `vaultSecretBackend`).
- Or `helm template` is run using the other values.yaml files which doesn't have `operator.vaultSecretBackend`. This is less likely because this could have caused consistent failure in CI.

One option that I considered to fix this is explicitly passing just the main values.yaml file to the `ct lint` command, using `--helm-extra-args` flag, but that didn't resolve the issues. I was able to reproduce it once if I ran 15 manual patches.

The other option was to have the `vaultSecretBackend` field in the other values files as well. And this is what I have done in this PR. I am not really sure if this actually fixed the problem but I was not able to reproduce it in 20 manual patches.  The other small change that this PR does is, running `helm template` before running `ct lint` so that if the test fails ever again we can check and see if `vaultSecretBackend`  is eventually generated  in the template.  

## Proof of Work

Output of successful run of `make precommit`: https://gist.github.com/viveksinghggits/45a6f80f2b5b85d26d7d21dcc3dfb56c
Output of failed run of `make precommit`: https://gist.github.com/viveksinghggits/7bb5a35e8bfe3a9be93e668ee8e734b9

Ran 20 manual evg patches and all of them are successful. 
image

## Checklist

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -234,8 +234,8 @@ pre_commit() {
 
     local lint_helm_chart_status=$?
     if [ "$lint_helm_chart_status" -eq 0 ]; then
-      return 0
       echo -e "${GREEN}pre-commit: All checks passed!${NO_COLOR}"
+      return 0
     else
       return 1
     fi
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -226,11 +226,19 @@ pre_commit() {
   run_job_in_background "python_formatting"
   run_job_in_background "check_erroneous_kubebuilder_annotations"
   run_job_in_background "validate_snippets"
-  run_job_in_background "lint_helm_chart"
 
   if wait_for_all_background_jobs; then
-    echo -e "${GREEN}pre-commit: All checks passed!${NO_COLOR}"
-    return 0
+    # lint_helm_chart must be run after all the background jobs are finished because one of the BG jobs (update_jobs)
+    # updates the helm chart. And lint_helm_chart requires the helm chart to be updated already.
+    lint_helm_chart
+
+    local lint_helm_chart_status=$?
+    if [ "$lint_helm_chart_status" -eq 0 ]; then
+      return 0
+      echo -e "${GREEN}pre-commit: All checks passed!${NO_COLOR}"
+    else
+      return 1
+    fi
   else
     return 1
   fi
--- a/scripts/dev/lint_helm_chart.sh
+++ b/scripts/dev/lint_helm_chart.sh
@@ -16,6 +16,11 @@ fi
 # the binaries yamale and yamllint required by ct are available at `${PROJECT_DIR}/venv/bin`
 export PATH=${PROJECT_DIR}/venv/bin:${PATH}
 
+helm template "${PROJECT_DIR}/helm_chart/" \
+  -f "${PROJECT_DIR}/helm_chart/values.yaml" \
+  -f "${PROJECT_DIR}/helm_chart/values-openshift.yaml" \
+  -f "${PROJECT_DIR}/helm_chart/values-multi-cluster.yaml"
+
 ct lint --charts="${PROJECT_DIR}/helm_chart/" \
     --chart-yaml-schema "${PROJECT_DIR}/helm_chart/tests/schemas/chart_schema.yaml" \
     --lint-conf "${PROJECT_DIR}/helm_chart/tests/schemas/lintconf.yaml"
--- a/scripts/dev/lint_helm_chart.sh
+++ b/scripts/dev/lint_helm_chart.sh
@@ -16,11 +16,6 @@ fi
 # the binaries yamale and yamllint required by ct are available at `${PROJECT_DIR}/venv/bin`
 export PATH=${PROJECT_DIR}/venv/bin:${PATH}
 
-helm template "${PROJECT_DIR}/helm_chart/" \
-  -f "${PROJECT_DIR}/helm_chart/values.yaml" \
-  -f "${PROJECT_DIR}/helm_chart/values-openshift.yaml" \
-  -f "${PROJECT_DIR}/helm_chart/values-multi-cluster.yaml"
-
 ct lint --charts="${PROJECT_DIR}/helm_chart/" \
     --chart-yaml-schema "${PROJECT_DIR}/helm_chart/tests/schemas/chart_schema.yaml" \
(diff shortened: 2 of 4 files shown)


scripts/dev/lint_helm_chart.sh
```diff
@@ -16,6 +16,11 @@ fi
 # the binaries yamale and yamllint required by ct are available at `${PROJECT_DIR}/venv/bin`
 export PATH=${PROJECT_DIR}/venv/bin:${PATH}
 
+helm template "${PROJECT_DIR}/helm_chart/" \
+  -f "${PROJECT_DIR}/helm_chart/values.yaml" \
+  -f "${PROJECT_DIR}/helm_chart/values-openshift.yaml" \
+  -f "${PROJECT_DIR}/helm_chart/values-multi-cluster.yaml"
```
m1kola: This doesn't look right: `values-openshift.yaml` and `values-multi-cluster.yaml` are not supposed to be used together, I believe.

viveksinghggits: This line is not necessarily required and I added it just to debug and see if multiple values files are merged and in the failure scenario the eventual `.Values.operator` doesn't have `vaultSecretBackend`. Because that's what seems to be causing our problem. Having all these values files together just increases our chance of catching the values.yaml files merge issue, if there is any.

MaciejKaras: If that is the reason for adding this line please comment it properly

viveksinghggits:
> I'm not convinced this will fix the helm linting issue definitely. More likely we are observing flakiness, because of running dependant tasks in parallel. `lint_helm_chart` func depends on the helm values generated by `update_values_yaml_files` func, but they are not running sequentially.
> 
> ```shell
> pre_commit() {
>   run_job_in_background "update_jobs"         # <- this rebuilds helm values
>   run_job_in_background "update_licenses"
>   run_job_in_background "lint_code"
>   run_job_in_background "start_shellcheck"
>   run_job_in_background "regenerate_public_rbac_multi_cluster"
>   run_job_in_background "python_formatting"
>   run_job_in_background "check_erroneous_kubebuilder_annotations"
>   run_job_in_background "validate_snippets"
>   run_job_in_background "lint_helm_chart"    # <- this requires helm values ready
> 
>   if wait_for_all_background_jobs; then
>     echo -e "${GREEN}pre-commit: All checks passed!${NO_COLOR}"
>     return 0
>   else
>     return 1
>   fi
> }
> 
> update_jobs() {
>      # Update release.json first in case there is a newer version
>     time update_release_json
>     # We need to generate the values files first
>     time update_values_yaml_files               # <- this rebuilds helm values
>     # The values files are used for generating the standalone yaml
>     time generate_standalone_yaml
> }
> ```
> 
> I think the solution should be to run `lint_helm_chart` in `update_jobs` function as the last step. Or maybe run it after all other jobs have finished?

Hi @MaciejKaras,
You are right. I somehow misread the code (that I wrote 😢) and thought that we are still running lint helm chart separately and not as part of pre-commit. I will make the changes and push them soon.

.githooks/pre-commit
```diff
@@ -226,11 +226,19 @@ pre_commit() {
...
+    # lint_helm_chart must be run after all the background jobs are finished because one of the BG jobs (update_jobs)
+    # updates the helm chart. And lint_helm_chart requires the helm chart to be updated already.
+    lint_helm_chart
+
+    local lint_helm_chart_status=$?
+    if [ "$lint_helm_chart_status" -eq 0 ]; then
+      return 0
+      echo -e "${GREEN}pre-commit: All checks passed!${NO_COLOR}"
+    else
+      return 1
```
MaciejKaras, line 240: Can you include the output of the precommit command when lint helm fails and also when it does not fail?

viveksinghggits: I have added it under proof of work @MaciejKaras.

.githooks/pre-commit
```diff
@@ -226,11 +226,19 @@ pre_commit() {
...
-    echo -e "${GREEN}pre-commit: All checks passed!${NO_COLOR}"
-    return 0
+    # lint_helm_chart must be run after all the background jobs are finished because one of the BG jobs (update_jobs)
+    # updates the helm chart. And lint_helm_chart requires the helm chart to be updated already.
+    lint_helm_chart
+
+    local lint_helm_chart_status=$?
+    if [ "$lint_helm_chart_status" -eq 0 ]; then
+      return 0
+      echo -e "${GREEN}pre-commit: All checks passed!${NO_COLOR}"
```
MaciejKaras, line 237: I think this line is not printed because it is after `return 0`

viveksinghggits:
@m1kola can you please have another look/approve?

m1kola (approved):

PR #490, opened 2025-10-02
CLOUDP-349087: Fix TLS disable + scale up test

Julien-Ben:
# Summary

This test has been broken for a while. Because there were two functions with the name `test_tls_is_disabled_and_scaled_up`, only one of them was running everytime.

I think this scenario has low chance of existing in production, hence why we never had a ticket related to this bug.

On top of that it performed the update in two separate step, whereas to test this behaviour it should be done in one update.

I uncovered it as part of the larger refactoring of the controller. But to keep the PR scope reasonable, I extracted the related changes as they are self contained
The bug may exist in multi-cluster as well. This bug was first discovered in 2021: https://jira.mongodb.org/browse/CLOUDP-80768 

The blocking mechanism will be implemented in a better way after the multi-cluster first refactor, since we will keep track of a global reconciler state, notably holding the target number of replicas for this reconciliation.

## Proof of Work

In the commit (https://github.com/mongodb/mongodb-kubernetes/pull/490/commits/1d8847e7bb1af22e36d685d1b30fb20476a4bf39) which just fixes the e2e test, it fails: [evg task](https://spruce.mongodb.com/task/mongodb_kubernetes_e2e_mdb_kind_ubi_cloudqa_e2e_disable_tls_scale_up_patch_dfb9424e9b34ddd048a725a9988114ca4032f9bf_68de79b2fccf070007a2c51e_25_10_02_13_10_12/tests?execution=4&sorts=STATUS%3AASC) Showing that the reconciler had to be fixed as well.

The PR is correct if CI is green again.

## Checklist

### DIFF
--- a/controllers/operator/mongodbreplicaset_controller.go
+++ b/controllers/operator/mongodbreplicaset_controller.go
@@ -254,8 +254,9 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
 
 	// Compute effective replicas for this reconciliation
 	var effectiveReplicas int32
-	if tlsWillBeDisabled {
-		// Lock at current member count during TLS disable
+	if tlsWillBeDisabled && rs.Status.Members > 0 {
+		// Lock at current member count during TLS disable, but only if there are existing members.
+		// For initial creation (Members == 0), use normal scaling logic.
 		effectiveReplicas = int32(rs.Status.Members)
 		log.Infof("TLS is being disabled, locking replicas at current member count: %d", rs.Status.Members)
 	} else {


docker/mongodb-kubernetes-tests/tests/tls/e2e_tls_disable_and_scale_up.py
```diff
@@ -38,18 +40,11 @@ def test_rs_is_running(replica_set: MongoDB):
 def test_tls_is_disabled_and_scaled_up(replica_set: MongoDB):
     replica_set.load()
     replica_set["spec"]["members"] = 5
```
Julien-Ben, line 42: The issue in the test is that it was doing the update in two steps (scale, and then disable tls), while the whole point is to change them at the same time. (on top of having a duplicate function name)

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -244,6 +244,29 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
 		}
 	}
 
+	// Check if TLS is being disabled. If so, we need to lock replicas at the current member count
+	// to prevent scaling during the TLS disable operation. This decision is made once here and
+	// applied to both the StatefulSet and OM automation config.
+	tlsWillBeDisabled, err := checkIfTLSWillBeDisabled(conn, rs, log)
```
lsierant, line 250: what if we just block this with validation and say that it's not possible to change both member count and disabling TLS at the same time?

Julien-Ben: Yes, after discussing it in DM, let's do that instead of adding complexity to the reconcile loop. This is not a common use case anyway.

PR #491, opened 2025-10-02
Reduce Evergreen retries to 1

Julien-Ben:
# Summary

Instead of 5 total tentatives for E2E tests, let's reduce it to 2.

## Proof of Work

Failing tests must restart only once.

### DIFF
--- a/scripts/evergreen/retry-evergreen.sh
+++ b/scripts/evergreen/retry-evergreen.sh
@@ -33,7 +33,7 @@ if [ -z "${EVERGREEN_API_KEY}" ]; then
 fi
 
 EVERGREEN_API="https://evergreen.mongodb.com/api"
-MAX_RETRIES="${EVERGREEN_MAX_RETRIES:-3}"
+MAX_RETRIES=1
 
 # Define build variants to exclude
 # We ignore openshift because they only run one test at a time.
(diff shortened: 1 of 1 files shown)


scripts/evergreen/retry-evergreen.sh
```diff
@@ -33,7 +33,7 @@ if [ -z "${EVERGREEN_API_KEY}" ]; then
 fi
 
 EVERGREEN_API="https://evergreen.mongodb.com/api"
-MAX_RETRIES="${EVERGREEN_MAX_RETRIES:-3}"
```
Julien-Ben, line 36: We had this env variable set to `2` in the Evergreen project settings, for some reason it was not picked up by the script, so I just left an hardcoded value.

scripts/evergreen/retry-evergreen.sh
```diff
@@ -49,7 +49,7 @@ BUILD_IDS=($(curl -s -H "Api-User: ${EVERGREEN_USER}" -H "Api-Key: ${EVERGREEN_A
 for BUILD_ID in "${BUILD_IDS[@]}"; do
   echo "Finding failed tasks in BUILD ID: ${BUILD_ID}"
   # shellcheck disable=SC2207
-  TASK_IDS=($(curl -s -H "Api-User: ${EVERGREEN_USER}" -H "Api-Key: ${EVERGREEN_API_KEY}" ${EVERGREEN_API}/rest/v2/builds/"${BUILD_ID}"/tasks | jq ".[] | select(.status == \"failed\" and .execution <= ${MAX_RETRIES})" | jq -r '.task_id'))
+  TASK_IDS=($(curl -s -H "Api-User: ${EVERGREEN_USER}" -H "Api-Key: ${EVERGREEN_API_KEY}" ${EVERGREEN_API}/rest/v2/builds/"${BUILD_ID}"/tasks | jq ".[] | select(.status == \"failed\" and .execution < ${MAX_RETRIES})" | jq -r '.task_id'))
```
Julien-Ben, line 52: `.execution` is the id from the last failing task. Executions are 0-indexed

anandsyncs (approved):

scripts/evergreen/retry-evergreen.sh
mircea-cosbuc: Don't we need to have it inside an `include_expansions_in_env` block for the task?

Julien-Ben: Probably, do we think we need an env variable for that ?

PR #495, opened 2025-10-06
CLOUDP-317886 - block removing cluster from MC Sharded deployment

MaciejKaras:
# Summary

During investigation of https://jira.mongodb.org/browse/CLOUDP-317886 we have found that the operator panics, when the cluster is removed from `clusterSpecList`. Initially we thought about allowing to do that, but later on I have discussed the issue with @lsierant and the better approach was chosen:
- when the user wants to remove cluster that already has 0 members, we allow this
- when the user wants to remove cluster that has more than 0 members, we block this, because it can lead to unexpected issues. This is described in the code comment:

https://github.com/mongodb/mongodb-kubernetes/blob/31c346ee4770c3caabb620f3e7ad3f38ebdc4d3c/controllers/operator/mongodbshardedcluster_controller.go#L840-L846

...and in the changelog:

**MultiClusterSharded**: Block removing non-zero member cluster from MongoDB resource. This prevents from scaling down member cluster without current configuration available, which can lead to unexpected issues. Previously operator was crashing in that scenario, after the fix it will mark reconciliation as `Failed` with appropriate message. Example unsafe scenario that is now blocked:
  * User has 2 member clusters: `main` is used for application traffic, `read-analytics` is used for read-only analytics
  * `main` cluster has 7 voting members
  * `read-analytics` cluster has 3 non-voting members
  * User decides to remove `read-analytics` cluster, by removing the `clusterSpecItem` completely
  * Operator scales down members from `read-analytics` cluster one by one
  * Because the configuration does not have voting options specified anymore and by default `priority` is set to 1, the operator will remove one member, but the other two members will be reconfigured as voting members
  * `replicaset` contains now 9 voting members, which is not [supported by MongoDB](https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Number-of-Voting-Members-of-a-Replica-Set)

## Proof of Work

Passing CI + new unit tests for `blockNonEmptyClusterSpecItemRemoval` function.

## Checklist

### DIFF
--- a/changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
+++ b/changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
@@ -3,4 +3,4 @@ kind: fix
 date: 2025-10-06
 ---
 
-* **MultiClusterSharded**: Block removing non-zero member cluster from MongoDB resource. This prevents from scaling down member cluster without current configuration available, which could lead to unexpected issues.
+* **MultiClusterSharded**: Blocked removing non-zero member cluster from MongoDB resource. This prevents from scaling down member cluster without current configuration available, which could lead to unexpected issues.


controllers/operator/authentication_test.go
```diff
@@ -770,15 +770,16 @@ func Test_NoAdditionalDomainsPresent(t *testing.T) {
 	// The default secret we create does not contain additional domains so it will not be valid for this RS
 	rs.Spec.Security.TLSConfig.AdditionalCertificateDomains = []string{"foo"}
 
-	reconciler, _, client, _, err := defaultClusterReconciler(ctx, nil, "", "", rs, nil)
-	require.NoError(t, err)
-	addKubernetesTlsResources(ctx, client, rs)
+	kubeClient, omConnectionFactory := mock.NewDefaultFakeClient(rs)
+	reconciler := newReplicaSetReconciler(ctx, kubeClient, nil, "", "", false, false, omConnectionFactory.GetConnectionFunc)
```
MaciejKaras, line 774: this was previously using shardedClusterReconciler instead of replicaset reconciler

api/v1/om/appdb_types.go
```diff
@@ -494,6 +494,9 @@ func (m *AppDBSpec) GetMemberClusterSpecByName(memberClusterName string) mdbv1.C
 
 	// In case the member cluster is not found in the cluster spec list, we return an empty ClusterSpecItem
 	// with 0 members to handle the case of removing a cluster from the spec list without a panic.
+	// TODO: this is not ideal, because we don't consider other fields that were removed i.e.MemberConfig.
+	// When scaling down we should consider the full spec that was used to create the cluster.
+	// Create a ticket
```
MaciejKaras: @MaciejKaras create Jira ticket for AppDB

changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
```diff
@@ -0,0 +1,13 @@
...
+
+* **MultiClusterSharded**: Block removing non-zero member cluster from MongoDB resource. This prevents from scaling down member cluster without current configuration available, which can lead to unexpected issues. Previously operator was crashing in that scenario, after the fix it will mark reconciliation as `Failed` with appropriate message. Example unsafe scenario that is now blocked:
+  * User has 2 member clusters: `main` is used for application traffic, `read-analytics` is used for read-only analytics
+  * `main` cluster has 7 voting members
+  * `read-analytics` cluster has 3 non-voting members
+  * User decides to remove `read-analytics` cluster, by removing the `clusterSpecItem` completely
+  * Operator scales down members from `read-analytics` cluster one by one
+  * Because the configuration does not have voting options specified anymore and by default `priority` is set to 1, the operator will remove one member, but the other two members will be reconfigured as voting members
+  * `replicaset` contains now 9 voting members, which is not [supported by MongoDB](https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Number-of-Voting-Members-of-a-Replica-Set)
```
lsierant: In practice there is a code in the operator that limits the number of voting members to 7 and mark the rest as non voting. But it's still a valid case of losing configuration and resulting in a non deterministic configuration of voting members.  

https://github.com/mongodb/mongodb-kubernetes/blob/master/controllers/om/deployment.go#L1206-L1218

viveksinghggits: q: are we trying to give multiple examples of the unsafe scenarios that are blocked or we are just trying to set background in initial points and then last point is the actual example that is blocked (`not supported by MongoDB)?

viveksinghggits:
```suggestion
  * `replicaset` now contains 9 voting members, which is not [supported by MongoDB](https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Number-of-Voting-Members-of-a-Replica-Set)
```

MaciejKaras: all of the points are steps in a scenario that can lead to unexpected issues, thus explaining the motivation for blocking such operation.

changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
```diff
@@ -0,0 +1,13 @@
+---
+kind: fix
+date: 2025-10-06
+---
+
+* **MultiClusterSharded**: Block removing non-zero member cluster from MongoDB resource. This prevents from scaling down member cluster without current configuration available, which can lead to unexpected issues. Previously operator was crashing in that scenario, after the fix it will mark reconciliation as `Failed` with appropriate message. Example unsafe scenario that is now blocked:
```
vinilage: This will go to the release notes but the text is very long. It is good to have this detail in the Jira ticket but we don't need to go so deep into details for the release notes. I would suggest to keep only the beginning of it:

"* **MultiClusterSharded**: Block removing non-zero member cluster from MongoDB resource. This prevents from scaling down member cluster without current configuration available."

vinilage (changes requested):

changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
MaciejKaras: I thought so as well, but needed that second feedback ;)

MaciejKaras: left a little more text in the end:

* **MultiClusterSharded**: Block removing non-zero member cluster from MongoDB resource. This prevents from scaling down member cluster without current configuration available, which could lead to unexpected issues.

changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
MaciejKaras: Talked with @vinilage about the changelog and now it is much shorter. Please check it.

changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
viveksinghggits: I have noticed that sometimes we start the release notes with **past tense** of the verb and we have not been consistent about it. 
Should we try to be consistent and just use past tense o the verb (Blocked)? It's also recommended to use past tense in release notes.
@vinilage I think it would be a great for a small dev document. 

How to write a good release note.

changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2025-10-06
+---
+
+* **MultiClusterSharded**: Block removing non-zero member cluster from MongoDB resource. This prevents from scaling down member cluster without current configuration available, which could lead to unexpected issues.
```
viveksinghggits:
```suggestion
* **MultiClusterSharded**: Blocked removing non-zero member cluster from MongoDB resource. This prevents from scaling down member cluster without current configuration available, which could lead to unexpected issues.
```

viveksinghggits (approved):

changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
MaciejKaras: I'm fine with both, waiting for @vinilage decision.

changelog/20251006_fix_block_removing_non_zero_member_cluster_from.md
vinilage: LGTM!

vinilage (approved):
LGTM!

PR #504, opened 2025-10-08
Update `dependabot` to allow patching Kubernetes related dependencies

MaciejKaras:
# Summary

## Proof of Work

N/A

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -50,7 +50,6 @@ updates:
        - dependency-name: kubernetes
          update-types:
           - version-update:semver-major
-          - version-update:semver-minor
     labels:
       - dependencies
       - python


.github/dependabot.yml
```diff
@@ -12,10 +12,25 @@ updates:
           - "*"
     ignore:
       - dependency-name: k8s.io/api
+        update-types:
+          - version-update:semver-major
+          - version-update:semver-minor
```
m1kola, line 17: @MaciejKaras  I had a look at the Dependabot docs. Do I understand correctly that with this change we say - "you must still not update major and minor versions for us so we can do that manually, but if you see a new patch version - please bump it"?

MaciejKaras: yes, that's what I understood as well. We basically block dependabot to think about updating major and minor versions, but allow him to update patches. This is beneficial for any security or plain bug fixes.

.github/dependabot.yml
```diff
@@ -33,6 +48,9 @@ updates:
           - "*"
     ignore:
        - dependency-name: kubernetes
+         update-types:
+          - version-update:semver-major
+          - version-update:semver-minor
```
m1kola: @MaciejKaras this package seems to be different. [Here](https://pypi.org/project/kubernetes/#history) I see 32.0.\*, 33.1.\* versions.

[Here](https://github.com/kubernetes-client/python?tab=readme-ov-file#compatibility-matrix-of-supported-client-versions) they say that major version is follows K8s. E.g. 34.y.z corresponds to K8s 1.34.

Should this be only `version-update:semver-major` here?

MaciejKaras: Good catch! Let me lookup other dependencies for similar issues. Versioning system for Kube dependencies is a mess 😢

MaciejKaras: Done, I found only this python version to be problematic

PR #509, opened 2025-10-08
E2e tests on published chart

viveksinghggits:
# Summary

## Checklist


scripts/release/oci_chart_info.py
```diff
@@ -0,0 +1,24 @@
...
+import os
+import sys
+from dataclasses import asdict
+
+from lib.base_logger import logger
+from scripts.release.build.build_info import load_build_info
+
+
+def main():
+    build_scenario = os.environ.get("BUILD_SCENARIO")
```
MaciejKaras: Let's not hide the dependency on "BUILD_SCENARIO" env var here and pass it directly as cmd flag.

scripts/evergreen/e2e/single_e2e.sh
```diff
@@ -142,6 +147,22 @@ deploy_test_app() {
         helm_params+=("--set" "omDebugHttp=true")
     fi
 
+    if [[ -n "${helm_oci_regisry:-}" ]]; then
```
MaciejKaras: are those values optional or should be expected to be present?

docker/mongodb-kubernetes-tests/tests/conftest.py
```diff
@@ -836,6 +839,29 @@ def _install_multi_cluster_operator(
...
+
+        helm_chart_path = chart_uri
+
+    if not custom_operator_version and helm_chart_path not in (MCK_HELM_CHART, LEGACY_OPERATOR_CHART):
+        # custom_operator_version is not set and chart is also neither MCK nor MEKO would mean we are trying to install
+        # the operator from OCI registry. The version based on build scenario (dev/staging) is set in `OPERATOR_VERSION` env var.
+        non_semver_custom_operator_version = os.environ.get(OPERATOR_VERSION_ENV_VAR_NAME)
+        # when we publish the helm chart we append `0.0.0+` in the chart version, details are
+        # here https://docs.google.com/document/d/1eJ8iKsI0libbpcJakGjxcPfbrTn8lmcZDbQH1UqMR_g/edit?tab=t.gg5ble8qlesq
+        custom_operator_version = f"0.0.0+{non_semver_custom_operator_version}"
```
MaciejKaras: this will not work when testing against released version which uses semantic version i.e. when we run e2e smoke tests after publishing images to production

docker/mongodb-kubernetes-tests/kubetester/operator.py
```diff
@@ -43,14 +48,40 @@ def __init__(
...
+                raise e
+
+            # figure out the registry URI, based on dev/staging scenario
+            chart_uri = f"oci://{registry}/{repository}"
+            helm_chart_path = chart_uri
+
+        # if operator_version is not specified and we are not installing the MCK or MEKO chart
+        # it would mean we want to install OCI published helm chart. Figure out respective version,
+        # it is set in env var `OPERATOR_VERSION` based on build_scenario.
+        if not operator_version and helm_chart_path not in (
```
MaciejKaras: I see this code in two places, it would make sense to encapsulate it in method

docker/mongodb-kubernetes-tests/kubetester/operator.py
```diff
@@ -43,14 +48,40 @@ def __init__(
         namespace: str,
         helm_args: Optional[Dict] = None,
         helm_options: Optional[List[str]] = None,
-        helm_chart_path: Optional[str] = "helm_chart",
+        helm_chart_path: Optional[str] = None,
```
MaciejKaras, line 47: q: how can we instruct this function to use local helm charts for local tests? Ideally we should be able to set this somewhere in the context files and override it by default locally, for example by specifying it in `local-defaults-context`.

docker/mongodb-kubernetes-tests/tests/conftest.py
viveksinghggits: thanks for mentioning this, I missed testing smoke tests while implementing this.

docker/mongodb-kubernetes-tests/kubetester/operator.py
viveksinghggits: Yes, that's the plan. I am working on the change that will make sure that we are able to run the tests locally. Will push the changes soon.

scripts/evergreen/e2e/single_e2e.sh
viveksinghggits: these values must be set for evergreen patches but will not be set for local tests. we are trying to run the tests image here that we don't do when we run the tests locally so we should be good in always expecting this to be set.

MaciejKaras: If that is the case we should fail in case they are missing and not omit setting them up, right?

viveksinghggits: Yes, that's right. I will make the change.

docker/mongodb-kubernetes-tests/tests/conftest.py
```diff
@@ -836,6 +840,8 @@ def _install_multi_cluster_operator(
     # The Operator will be installed from the following repo, so adding it first
     helm_repo_add("mongodb", "https://mongodb.github.io/helm-charts")
 
+    helm_chart_path, custom_operator_version = helm_chart_path_and_version(helm_chart_path, custom_operator_version)
```
viveksinghggits: we seem to be calling `Operator()` in line 854 below already figures out the chart path and version using the exact same function. So this call here seems unnecessary.

docker/mongodb-kubernetes-tests/Dockerfile
```diff
@@ -66,6 +67,11 @@ RUN curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -
     && chmod +x ./kubectl \
     && mv ./kubectl /usr/local/bin/kubectl
 
+# install aws, required to run helm registry login while running the tests
```
MaciejKaras: this is not going to work for multi arch test image

docker/mongodb-kubernetes-tests/kubetester/helm.py
```diff
@@ -145,6 +149,55 @@ def process_run_and_check(args, **kwargs):
...
+
+    try:
+        logger.info("Starting AWS ECR credential retrieval.")
+        aws_proc = subprocess.Popen(
+            aws_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True  # Treat input/output as text strings
+        )
+
+        logger.info("Starting Helm registry login.")
+        helm_proc = subprocess.Popen(
+            helm_command, stdin=aws_proc.stdout, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
```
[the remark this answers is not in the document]
MaciejKaras: /other test only

docker/mongodb-kubernetes-tests/kubetester/helm.py
```diff
@@ -1,15 +1,19 @@
...
 import uuid
 from typing import Dict, List, Optional, Tuple
 
+from kubetester.consts import *
 from tests import test_logger
 
 logger = test_logger.get_test_logger(__name__)
 
+# LOCAL_CRDs_DIR is the dir where local helm chart's CRDs are copied in tests image
+LOCAL_CRDs_DIR = "helm_chart/crds"
```
mircea-cosbuc: nit: Mixed casing

docker/mongodb-kubernetes-tests/kubetester/helm.py
```diff
@@ -1,15 +1,19 @@
 import glob
-import logging
 import os
 import re
 import subprocess
 import uuid
 from typing import Dict, List, Optional, Tuple
 
+from kubetester.consts import *
```
mircea-cosbuc: nit: Prefer explicit import for code navigation.

MaciejKaras: agree, missed that

mircea-cosbuc (approved):
LGTM

docker/mongodb-kubernetes-tests/Dockerfile
```diff
@@ -16,7 +16,12 @@ RUN apt-get -qq update \
 
 COPY requirements.txt requirements.txt
 
-RUN python3 -m venv /venv && . /venv/bin/activate && pip install --upgrade pip && GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=1 pip install -r requirements.txt
+# install aws, required to run helm registry login while running the tests
+RUN python3 -m venv /venv  \
```
lsierant, line 20: we have this already setup on the host - why we need to aws login again in pytest?

scripts/evergreen/e2e/single_e2e.sh
```diff
@@ -48,6 +48,16 @@ deploy_test_app() {
     BUILD_ID="${BUILD_ID:-default_build_id}"
     BUILD_VARIANT="${BUILD_VARIANT:-default_build_variant}"
 
+    if ! chart_info=$(scripts/dev/run_python.sh scripts/release/oci_chart_info.py --build-scenario "${BUILD_SCENARIO}" 2>&1); then
```
lsierant, line 51: if we have all of this executed on the host already, why to propagate it to the test pod at all? Couldn't we decide here which helm path to use and use a single HELM_CHART_PATH env var for that that we will use in tests?

docker/mongodb-kubernetes-tests/kubetester/helm.py
```diff
@@ -230,3 +301,35 @@ def _create_helm_args(helm_args: Dict[str, str], helm_options: Optional[List[str
...
+
+# helm_chart_path_and_version returns the chart path and version that we would like to install to run the E2E tests.
+# for local tests it returns early with local helm chart dir and for other scenarios it figures out the chart and version
+# based on the caller. In most of the cases we will install chart from OCI registry but for the tests where we would like
+# to install MEKO's specific version or MCK's specific version, we would expect `helm_chart_path` to set already.
+def helm_chart_path_and_version(helm_chart_path: str, operator_version: str) -> tuple[str, str]:
+    # these are imported here to resolve import cycle issue
+    from tests.conftest import LOCAL_HELM_CHART_DIR, local_operator
+
+    if local_operator():
```
lsierant: this is not sufficient - we can run the operator in IDE or in the pod. Also we might want to use oci helm chart even if running the operator locally. 

Could we have one env var that we could just point to a proper helm path that we could control in the context and not in pytest?

docker/mongodb-kubernetes-tests/Dockerfile
MaciejKaras: Because in pytests we are installing operator using helm charts and for those we need AWS/Quay credentials

scripts/evergreen/e2e/single_e2e.sh
MaciejKaras: This is required for:
- `helm_registry_login_to_ecr(registry, region)` where we need separate value rather than full path.
- using version during helm install if the OCI helm charts are used

docker/mongodb-kubernetes-tests/kubetester/helm.py
MaciejKaras: You've just reinforced my similar concern and I'm fixing it now. I'll add new env var and set default to local helm charts in `local-defaults-context`

docker/mongodb-kubernetes-tests/Dockerfile
lsierant: and it uses 

```
"--set" "aws.accessKey=${AWS_ACCESS_KEY_ID}"
"--set" "aws.secretAccessKey=${AWS_SECRET_ACCESS_KEY}"
``` 
that we already set in the test pod? 

Will this be also available locally?

MaciejKaras: AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are stored in private-context and it will be used in locally run tests

docker/mongodb-kubernetes-tests/kubetester/helm.py
MaciejKaras: Please take a look again, I've added `DEFAULT_HELM_CHART_PATH` env var

PR #510, opened 2025-10-09
CLOUDP-350339: Fix e2e_om_ops_manager_backup_restore_minio

Julien-Ben:
# Summary

This test started failing without code changes on the repo.
My hypothesis is Go 1.24.8 that introduced stricter validations to DNS names in x509 certificates https://go.dev/doc/devel/release#go1.24.minor

## Proof of Work

Tests must be green.
This test was slightly flaky before so it may still fail. But it should not fail systematically after 7 min anymore.

## Checklist


docker/mongodb-kubernetes-tests/tests/opsmanager/om_ops_manager_backup_restore_minio.py
```diff
@@ -94,12 +94,15 @@ def tenant_name() -> str:
 @fixture(scope="module")
 def tenant_domains() -> List[str]:
     return [
-        "tenant-0-pool-0-{0...3}.tenant-0-hl.tenant-0.svc.cluster.local",
```
lucian-tosa, line 97: I think this should still work. As far as I can see, only the leading period was enforced
https://github.com/cockroachdb/go/commit/36b4d8a1c294a521426c218b5e0c6ca1589e6fb6#diff-e3cea16ed73b68bdaa41bd72dffb3cbbabe2528712d72080de7e4def46c0dbf4R1211

m1kola: @Julien-Ben @lucian-tosa  I see that the CI is green now and we are ready to merge this. I suggest that we merge as is and this can be a separate follow up PR to revert back to `{0...3}` syntax, if one of you want to submit it.

What do you think?

Julien-Ben: I have the same understanding; but weirdly it doesn't work.
I started a patch with only the leading period fix (You can check the "changes" tab): https://spruce.mongodb.com/version/68e8d1b2f49f9b00070ced6f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

They all failed after 7min.
I'll go ahead and merge what I have then

Thanks for digging deeper

PR #511, opened 2025-10-09
CLOUDP-350375: `MongoDBMultiCluster`: Add annotation to disable reconciliation

m1kola:
# Summary

The new `mongodb.com/v1.disable-reconciliation` annotation allows disabling reconciliation for `MongoDBMultiCluster` objects.

This will be used later when migrating multi-cluster replica sets from `MongoDBMultiCluster` to `MongoDB`.

## Proof of Work

1. Add `mongodb.com/v1.disableReconciliation` to any `MongoDBMultiCluster`
2. Run `kubectl wait mongodbmulticluster NAME -n NAMESPACE --for=jsonpath='{.status.phase}'=Disabled`.
3. Check that the `kubectl wait` command prints `mongodbmulticluster.mongodb.com/NAME condition met`.

## Checklist

### DIFF
--- a/controllers/operator/mongodbmultireplicaset_controller.go
+++ b/controllers/operator/mongodbmultireplicaset_controller.go
@@ -135,6 +135,11 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 		return reconcileResult, err
 	}
 
+	if val, ok := mrs.Annotations[util.DisableReconciliation]; ok && val == util.DisableReconciliationValue {
+		log.Info("Reconciliation disabled")
+		return r.updateStatus(ctx, &mrs, workflow.Disabled(), log)
+	}
+
 	if !architectures.IsRunningStaticArchitecture(mrs.Annotations) {
 		agents.UpgradeAllIfNeeded(ctx, agents.ClientSecret{Client: r.client, SecretClient: r.SecretClient}, r.omConnectionFactory, GetWatchedNamespace(), true)
 	}
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -285,6 +285,9 @@ const (
 	LastAchievedSpec        = "mongodb.com/v1.lastSuccessfulConfiguration"
 	LastAchievedRsMemberIds = "mongodb.com/v1.lastAchievedRsMemberIds"
 
+	DisableReconciliation      = "mongodb.com/v1.disableReconciliation"
+	DisableReconciliationValue = "true"
+
 	// SecretVolumeName is the name of the volume resource.
 	SecretVolumeName = "secret-certs"
 
--- a/controllers/operator/mongodbmultireplicaset_controller_test.go
+++ b/controllers/operator/mongodbmultireplicaset_controller_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"sort"
 	"testing"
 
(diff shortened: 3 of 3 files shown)


controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -135,6 +135,11 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 		return reconcileResult, err
 	}
 
+	if val, ok := mrs.Annotations[util.DisableReconciliation]; ok && val == util.DisableReconciliationValue {
```
m1kola, line 138: It will be something our users will be using during migration, but as part of this PR I"m not going to document the annotation/add a release note. It will be included in the migration guide once we release a version which is ready for migration. Until then - let's leave it undocumented.

pkg/util/constants.go
```diff
@@ -285,6 +285,9 @@ const (
 	LastAchievedSpec        = "mongodb.com/v1.lastSuccessfulConfiguration"
 	LastAchievedRsMemberIds = "mongodb.com/v1.lastAchievedRsMemberIds"
 
+	DisableReconciliation      = "mongodb.com/v1.disableReconciliation"
+	DisableReconciliationValue = "true"
```
m1kola, line 289: Had to define a constant because linter was not happy.

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -135,6 +135,11 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
...
+		log.Info("Reconciliation disabled")
+		return r.updateStatus(ctx, &mrs, workflow.Disabled(), log)
```
viveksinghggits, line 140: q: It looks like disabled workflow was [introduced to](https://github.com/mongodb/mongodb-kubernetes/blob/master/controllers/operator/workflow/disabled.go#L7) notify if a **subresource** is enabled or disabled. would it be a good idea to use the same to notify that a resource is deprecated?

I was also wondering if we should call it disabled or deprecated?

viveksinghggits (commented):
PR looks good to me. Just left a small comment/question.

controllers/operator/mongodbmultireplicaset_controller.go
m1kola: @viveksinghggits I found this:

https://github.com/mongodb/mongodb-kubernetes/blob/219a3df6d3e35333fde8e8294f216b466222c19e/api/v1/status/phase.go#L20-L21

I believe these statuses are generic. It just happened so that we only use it currently to report that a backup is disabled. But I think there is no reason not to use it for anything else. I'm happy to update this comment to reflect this:
https://github.com/mongodb/mongodb-kubernetes/blob/219a3df6d3e35333fde8e8294f216b466222c19e/controllers/operator/workflow/disabled.go#L7-L10


> I was also wondering if we should call it disabled or deprecated?

We might be adding a deprecation message somewhere on the resource. But it will likely going to be a permanent message.

For the purpose of migration we need a status which can be used to programmatically check that the resource is not being reconciled.

viveksinghggits: I think just updating the comment should be ok. Just wanted to bring the attention to that comment, not a big issue.

viveksinghggits (approved):

m1kola:
Closing as stale

PR #522, opened 2025-10-10
fix split environment variables on helm chart template

edimarlnx:
# Summary

Fix environment variable parsing in Helm chart template to properly handle values containing equals signs.

The previous implementation used `split "=" .` which would incorrectly parse environment variables with values containing `=` characters (e.g., `VAR=value=with=equals`). Changed to use `splitn "=" 2 .` which limits the split to 2 parts, treating everything after the first `=` as the value.

## Proof of Work

The fix changes the parsing logic in `helm_chart/templates/operator.yaml:280-281`:
- Before: `split "=" .` - splits on all `=` characters
- After: `splitn "=" 2 .` - splits only on the first `=`, preserving `=` in values

This ensures custom environment variables with values containing `=` characters are correctly parsed when using `.Values.customEnvVars`.

**Validation:**

```bash
helm template helm_chart --set customEnvVars='CLUSTER_DOMAIN=mydomain.local&MY_VAR1=mylabel=abc&MY_VAR2=any'
```

The rendered output correctly shows:
```yaml
- name: CLUSTER_DOMAIN
  value: 'mydomain.local'
- name: MY_VAR1
  value: 'mylabel=abc'  # Correctly preserves '=' in the value
- name: MY_VAR2
  value: 'any'
```

## Checklist

### DIFF
--- a/changelog/20251010_fix_split_environment_variables_on_helm_chart.md
+++ b/changelog/20251010_fix_split_environment_variables_on_helm_chart.md
@@ -3,4 +3,4 @@ kind: fix
 date: 2025-10-10
 ---
 
-* Fix parsing of custom environment variables with values containing `=` characters in Helm chart template
+* Fixed parsing of the `customEnvVars` Helm value when values contain `=` characters.


changelog/20251010_fix_split_environment_variables_on_helm_chart.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2025-10-10
+---
+
+* Fixed parsing of the `customEnvVars` Helm value when values contain `=` characters.
```
vinilage, line 6: LGTM

PR #532, opened 2025-10-15
Use const for `helm_chart` local chart dir

viveksinghggits:
# Summary

The local helm chart directory (`helm_chart`) that we use to install the operator is being used at a lot of places and we were using the dir name in line without using the defined constant. This would make it harder to use any other dir as helm chart dir/URL, that we are going to do when we start installing helm chart from OCI registry.
This PR changes that and tries to use the const everywhere we used the local chart dir (`helm_chart`).

edit: To make this work I had to move all constants from `tests.conftest` to `tests.constants` file to remove circular dependency.

## Proof of Work

CI in the PR.

## Checklist

### DIFF
--- a/docker/mongodb-kubernetes-tests/kubetester/operator.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/operator.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import logging
 import time
 from typing import Dict, List, Optional
 
@@ -18,7 +17,7 @@ from kubetester.helm import (
     helm_upgrade,
 )
 from tests import test_logger
-from tests.conftest import LOCAL_HELM_CHART_DIR
+from tests.constants import LOCAL_HELM_CHART_DIR
 
 OPERATOR_CRDS = (
     "mongodb.mongodb.com",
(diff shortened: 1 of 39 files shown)


docker/mongodb-kubernetes-tests/kubetester/helm.py
```diff
@@ -7,13 +7,14 @@
 from typing import Dict, List, Optional, Tuple
 
 from tests import test_logger
+from tests.conftest import LOCAL_HELM_CHART_DIR
```
mircea-cosbuc: Shouldn't we instead ensure that we always call `helm_template` with the constant as a parameter for `helm_chart_path`?
I don't think we should have library code importing test configuration.

viveksinghggits: I agree. Do you think it makes sense to define the const it in `helm.py` itself, considering that we were using directory name `helm_chart` at quite a few places (in addition to helm template). 
Ideally we should not default to a chart dir here in `helm.py`.

viveksinghggits: Or I can every work without this PR.

mircea-cosbuc: I think we should have a central place to configure the helm chart path for all the tests. I just don't think we should import test code here. I'm on board with the other changes.

viveksinghggits: I just removed the changes from the `helm.py` file.

PR #539, opened 2025-10-20
CLOUDP-339241 - re-enable multi arch smoke tests

MaciejKaras:
# Summary

Key changes include:

- Added support for IBM Power and IBM Z architectures with podman builder configuration
- Updated repository URLs from `quay.io/mongodb/mongodb-kubernetes-tests` to `268558157000.dkr.ecr.us-east-1.amazonaws.com/staging/mongodb-kubernetes-tests` for release scenarios. This will prevent `meko-test` image to be available also for our customers
- Enhanced build pipeline to support both Docker and Podman builders for different architectures

## Proof of Work

Staging job is successful -> https://spruce.mongodb.com/version/69033a1e9701b900075d742f (ignore ibm z which is disabled in updated code)

Release job is successful -> https://spruce.mongodb.com/version/69036de79701b900075df4de (ignore GKE code snippets tasks, they need different fix. [create_chart_release_pr](https://spruce.mongodb.com/task/mongodb_kubernetes_create_chart_release_pr_create_chart_release_pr_patch_b47894435713aeb2e8c26f99034205150cfad247_69036de79701b900075df4de_25_10_30_13_53_44?execution=1) will fail due to missing tag)

We are not ready yet to enable smoke tests for IBM Z, because our evergreen machines are unstable -> https://jira.mongodb.org/browse/DEVPROD-23283

## Checklist

### DIFF
--- a/scripts/release/argparse_utils.py
+++ b/scripts/release/argparse_utils.py
@@ -1,5 +1,6 @@
 import argparse
 
+from scripts.release.build.build_info import BUILDER_DOCKER, BUILDER_PODMAN
 from scripts.release.build.build_scenario import BuildScenario
 from scripts.release.build.image_build_configuration import SUPPORTED_PLATFORMS
 from scripts.release.build.image_build_process import (
@@ -39,9 +40,9 @@ def get_platforms_from_arg(args_platforms: str) -> list[str] | None:
 
 
 def get_image_builder_from_arg(builder_name: str):
-    if builder_name == "docker":
+    if builder_name == BUILDER_DOCKER:
         return DockerImageBuilder()
-    elif builder_name == "podman":
+    elif builder_name == BUILDER_PODMAN:
         return PodmanImageBuilder()
     else:
         raise ValueError(f"Unsupported image builder '{builder_name}'. Supported builders: docker, podman")


scripts/release/atomic_pipeline.py
```diff
@@ -96,26 +98,59 @@ def build_image(
...
+            for key, value in build_args.items():
+                build_command.extend(["--build-arg", f"{key}={value}"])
+
+            result = subprocess.run(build_command, capture_output=True, text=True, check=True)
+            logger.debug(result.stdout)
+
+            for tag in tags:
+                push_command = ["sudo", "podman", "push", "--authfile=/root/.config/containers/auth.json", tag]
+                result = subprocess.run(push_command, capture_output=True, text=True, check=True)
+                logger.debug(result.stdout)
```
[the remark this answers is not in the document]
MaciejKaras: This is all printed when catching `subprocess.CalledProcessError` error

lucian-tosa: This logic seems better suited for `execute_docker_build`

scripts/release/atomic_pipeline.py
```diff
@@ -96,26 +98,59 @@ def build_image(
...
+            build_command = [
+                "sudo",
+                "podman",
+                "buildx",
+                "build",
+                "--progress",
+                "plain",
+                build_path,
+                "-f",
+                build_configuration.dockerfile_path,
```
lucian-tosa: Looks like the platform flag is not used (except for the tag). I know it is implied by the machine it runs on, but for clarity we should set it here

.evergreen-release.yml
```diff
@@ -180,56 +168,100 @@ buildvariants:
...
+    tasks:
+      - name: build_test_image
+
+  - name: init_smoke_tests_ibm_power_release
+    display_name: init_smoke_tests_ibm_power
+    max_hosts: -1
     tags: [ "release", "e2e_smoke_release_test_suite" ]
+    run_on:
+      - release-rhel9-power-small # This is required for CISA attestation https://jira.mongodb.org/browse/DEVPROD-17780
+      - release-rhel9-power-large # This is required for CISA attestation https://jira.mongodb.org/browse/DEVPROD-17780
```
lucian-tosa, line 226: it's a private test image, why?

scripts/dev/contexts/e2e_smoke_ibm_power_release
```diff
@@ -0,0 +1,15 @@
...
+script_dir=$(dirname "${script_name}")
+
+source "${script_dir}/root-context"
+source "${script_dir}/variables/e2e_ibm_smoke_base"
+
+CUSTOM_OM_VERSION=$(grep -E "^\s*-\s*&ops_manager_70_latest\s+(\S+)\s+#" <"${PROJECT_DIR}"/.evergreen.yml | awk '{print $3}')
+export CUSTOM_OM_VERSION
+
+# During release the meko-tests image is pushed to a staging registry. We don't want to push it to production registry
+export MEKO_TESTS_REGISTRY="268558157000.dkr.ecr.us-east-1.amazonaws.com/staging"
```
lucian-tosa, line 15: Shouldn't this logic be in evg-private-context? We will never use quay for test image, so I think this can be simplified

scripts/release/atomic_pipeline.py
MaciejKaras: I don't see the platform flag being used before:
```
sudo podman buildx build --progress plain . -f Dockerfile -t "${REGISTRY}/mongodb-kubernetes-tests:${OPERATOR_VERSION}-$(arch)" --build-arg PYTHON_VERSION="${PYTHON_VERSION}"
```

.evergreen-release.yml
MaciejKaras: well it's not a big deal, but we are using this test image to verify that our released images are passing smoke tests. In theory somebody from outside could interfere with test image build process and acquire stored credentials for quay.io for instance

scripts/dev/contexts/e2e_smoke_ibm_power_release
MaciejKaras: yeah, this can be also in `evg-private-context`. In the future `evg-private-context` should read repository settings from `build_info.json` instead of having them hardcoded. If you have nothing against this I would rather target this in separate PR and use `build_info.json` properly.

scripts/release/atomic_pipeline.py
MaciejKaras: Probably not `execute_docker_build` because it's using docker ;) but I agree this can be encapsulated in separate method with similar signature as `execute_docker_build`. Or you meant something else?

scripts/release/tests/build_info_test.py
```diff
@@ -152,6 +167,20 @@ def test_load_build_info_patch():
...
             ),
+            "meko-tests-ibm-power": ImageInfo(
+                repositories=["268558157000.dkr.ecr.us-east-1.amazonaws.com/dev/mongodb-kubernetes-tests"],
+                platforms=["linux/ppc64le"],
+                dockerfile_path="docker/mongodb-kubernetes-tests/Dockerfile",
+                builder=BUILDER_PODMAN,
+                architecture_suffix=True,
+            ),
+            "meko-tests-ibm-z": ImageInfo(
+                repositories=["268558157000.dkr.ecr.us-east-1.amazonaws.com/dev/mongodb-kubernetes-tests"],
```
anandsyncs, line 178: nit: It would be better to use constants for the repetitive `dev` and `staging` repos.

anandsyncs (approved):
LGTM, there is one nit and another issue the Lucian already mentioned:
https://github.com/mongodb/mongodb-kubernetes/pull/539/files#r2490868628

scripts/release/atomic_pipeline.py
lucian-tosa: That script and the respective evergreen function&task definitions were reused for the 2 different IBM architectures. In this case there are 2 distinct evergreen functions/tasks, entries in build_info, etc. It would be clearer and easier to debug if we set the platform flag. If we somehow edit the build_info file and swap the platform by mistake, we would result in an image with a platform different from the tag. It would also be harder to debug from the logs

scripts/release/atomic_pipeline.py
lucian-tosa: Yes, that's what I meant. I think it is implied that `ensure_docker_build` is not the right name for it. The point is that interactions with the container engine should be encapsulated under a common method, or even better, classes.

.evergreen-release.yml
lucian-tosa: I am not sure how large is the machine pool for these distros. Depending on that, using regular distros would speed up the total patch time. Also I don't think there are any quay credentials there, they are not needed in smoke tests where we use public images.
I will leave it up to you

scripts/dev/contexts/e2e_smoke_ibm_power_release
lucian-tosa: Fine with me if we have a ticket to track this

MaciejKaras: Not the ticket, but I've added bookmark in our AR docs -> https://docs.google.com/document/d/1eJ8iKsI0libbpcJakGjxcPfbrTn8lmcZDbQH1UqMR_g/edit?tab=t.45ig7xr3e3w4#bookmark=id.xiy7sriui1i7

scripts/release/tests/build_info_test.py
MaciejKaras: yes, we could but it's also nice to see all the repositories explicitly. Also each image type has different repository and that would mean creating x3 (per environment) constants. I don't think it's worth to do it.

.evergreen-release.yml
MaciejKaras: Actually `release-` distros are rarely used and the queue is much smaller. I would like to keep it that way, because there is currently no dowside.

scripts/release/atomic_pipeline.py
MaciejKaras: That makes sense. I've added the platform flag.

scripts/release/atomic_pipeline.py
MaciejKaras: Made changes like you suggested and used Python classes. Smoke tests for new code are passing -> https://spruce.mongodb.com/version/690b497052a78c0007bb13d9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

scripts/release/build/image_build_process.py
```diff
@@ -10,138 +10,201 @@
...
-        build_args = {k: str(v) for k, v in args.items()}
-
-        logger.info(f"Building image: {tags}")
-        logger.info(f"Platforms: {platforms}")
-        logger.info(f"Dockerfile: {dockerfile}")
-        logger.info(f"Build context: {path}")
-        logger.debug(f"Build args: {build_args}")
-
-        # Use buildx for multi-platform builds
+    @staticmethod
```
lucian-tosa, line 70: nit: not sure why static is needed. You are using an instance of the class anyway

lucian-tosa (approved):

scripts/release/build/image_build_process.py
MaciejKaras: This is what IDE suggested, because the method was only called internally from the class and `self` in the method signature was unused. I guess this is idiomatic way.

PR #540, opened 2025-10-20
Recommend to have `why` in PR desc

viveksinghggits:
# Summary

Adding why a particular PR is needed, is important in PR desc. This help reviewers understand the PR better. 
This PR encourages authors to do that by updating the PR template.

## Proof of Work
NA


.github/pull_request_template.md
```diff
@@ -1,6 +1,8 @@
 # Summary
 
-<!-- Enter your issue summary here.-->
+<!-- Enter your PR summary here. Try to emphasize on WHY this change is needed, followed by what's being done in the PR.
```
Julien-Ben: As a template, I think it should be much more concise. The first line is enough IMO.

Jira tickets are already mentioned in the Checklist

anandsyncs: I have a similar question, is this a substitution for absence of JIRA ticket in the PR?

viveksinghggits: So the point that @Julien-Ben raised is different from the question that you have @anandsyncs. 

@Julien-Ben's point was we don't have to mention `This is not a replacement for creating jira tickets.` this line because we already have have a check box in our PR template checklist for the jira ticket and that would mean a jira ticket is required.

And I think what you are asking is, if you mention *why* in PR desc, do you not have to create jira ticket anymore. And we knew that people might get into that confusion, that's why I explicitly mentioned, having why in PR desc doesnt mean that you don't have to create the respective jira ticket.

lsierant: +1, let's keep "Enter your PR summary here. Try to emphasize on WHY this change is needed, followed by what's being done in the PR." 
the rest is a bit obvious

Julien-Ben (approved):

MaciejKaras (approved):
LGTM

PR #507, opened 2025-10-08
Publish helm chart to OCI registry for PR patches

viveksinghggits:
# Summary

## Checklist

- [ ] Have you linked a jira ticket and/or is the ticket in the title?
- [ ] Have you checked whether your jira ticket required DOCSP changes?
- [ ] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

PR #542, opened 2025-10-21
[CLOUDP-352383] Add a build variant to release chart to OCI registry

viveksinghggits:
# Summary

In one of the previous PRs we made sure that we are publishing our helm chart to OCI container registry, as part of the work where we are going to eventually move to the OCI container registry for our helm chart repo.
After the chart is being published to to dev/staging registries, this PR makes sure that the chart get released to quay, as part of our release process.
We will keep publishing helm chart to the old github repo and quay container registry, for some time and will eventually move to just publishing it to OCI.

## Proof of Work

This evg patch is for dev scenario: https://spruce.mongodb.com/version/68f74e0d5420540007c32361/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
This one is for release, and we were able to publish to quay successfully: https://spruce.mongodb.com/version/68f75417f6cd5700079c0cb9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

## Checklist

### DIFF
--- a/scripts/release/helm_registry_login.py
+++ b/scripts/release/helm_registry_login.py
@@ -2,6 +2,7 @@ import argparse
 import os
 import subprocess
 import sys
+from shlex import quote
 
 from lib.base_logger import logger
 from scripts.release.build.build_info import load_build_info
@@ -88,12 +89,13 @@ def helm_registry_login_to_quay(registry):
     if not password:
         raise Exception(f"Env var {QUAY_PASSWORD_ENV_VAR} must be set with the quay password.")
 
-    command = ["helm", "registry", "login", "--username", username, "--password-stdin", registry]
+    # using quote will help us avoid command injectin issues, was reported by semggrep in PR
+    command = ["helm", "registry", "login", "--username", quote(username), "--password-stdin", quote(registry)]
 
     try:
         result = subprocess.run(
             command,
-            input=password,  # Pass the password as input bytes
+            input=quote(password),  # Pass the password as input bytes
             capture_output=True,
             text=True,
             check=False,  # Do not raise an exception on non-zero exit code


viveksinghggits:
@MaciejKaras @lucian-tosa can you please have a look into this PR as well.

scripts/release/helm_registry_login.py
```diff
@@ -67,7 +71,44 @@ def main():
 
     registry = build_info.helm_charts["mongodb-kubernetes"].registry
     region = build_info.helm_charts["mongodb-kubernetes"].region
-    return helm_registry_login(registry, region)
+
+    if build_scenario == BUILD_SCENARIO_RELEASE:
```
MaciejKaras: instead of using scenario as input we could use `registry` url itself. If it is quay, login to quay else to ecr. If anything else we should fail.

Also `helm_registry_login_to_release` should be `helm_registry_login_to_quay` and `helm_registry_login_to_nonrelease` -> `helm_registry_login_to_ecr`

scripts/release/publish_helm_chart.py
```diff
@@ -52,7 +50,17 @@ def update_chart_and_get_metadata(chart_dir: str) -> tuple[str, str]:
         chart_name = data.get("name")
         if not chart_name:
             raise ValueError("Chart.yaml is missing required 'name' field.")
+    except Exception as e:
+        raise Exception(f"Unable to load Chart.yaml from dir {chart_path}")
+
+    # if build_scenario is release, the chart.yaml would already have correct chart version
+    if build_scenario == BUILD_SCENARIO_RELEASE:
```
MaciejKaras, line 57: why not hide this logic in `build_info.json`? You could specify for example new field `version_prefix` as `0.0.0+` in `patch` and `staging` build scenarios. If it is specified just concatenate it with actual version_id, otherwise use plain version_id

scripts/release/publish_helm_chart.py
```diff
@@ -28,17 +29,14 @@ def run_command(command: list[str]):
 
 # update_chart_and_get_metadata updates the helm chart's Chart.yaml and sets the version
 # to either evg patch id or commit which is set in OPERATOR_VERSION.
-def update_chart_and_get_metadata(chart_dir: str) -> tuple[str, str]:
+def update_chart_and_get_metadata(chart_dir: str, build_scenario) -> tuple[str, str]:
     chart_path = os.path.join(chart_dir, "Chart.yaml")
     version_id = os.environ.get("OPERATOR_VERSION")
```
MaciejKaras: nit: we should not use `version_id` anymore in codebase. Use plain `version` name

scripts/release/publish_helm_chart.py
viveksinghggits: So, there are quite a few things here. First is, this change of figuring out the version of dev/staging is not really part of this PR, it was merged as part of the change in [this PR](https://github.com/mongodb/mongodb-kubernetes/pull/507/files#diff-3619355aa16d1843ce141ee17f4b484cfb842155d47f4f42bcf08eb9abb884b3R39). 
Having said that, I can make the change (in new PR) but I would like to understand the benefit of doing that. According to me if we have the build_scenario and with that if we can figure something out without looking into build_info.json, does it make sense to look into build_info.json?
Having the eventual version dependent on build scenario makes the code more readable, if it's dev/staging we are appending 0.0.0. But if we read the prefix from build_info, like you said we are hiding away why we are appending things based on prefix.

scripts/release/helm_registry_login.py
viveksinghggits: I can rename the function names to `helm_registry_login_to_quay ` and  `helm_registry_login_to_ecr`, but I still think we should compare with build_scenario and then login to respective registries. And the reason is usually we should not match with config file values in code because if we change the config file values, we will have to change the code. For example, in this case if we match with quay.io, and quay.io changes to something else (in this specific scenario it's not going to change but it's just an example), we will have to change the config file as well as code to reflect that.

scripts/release/publish_helm_chart.py
viveksinghggits: Or in other words, does it make sense to make something configurable, which doesn't need to be. We know that we are not going to change 0.0.0 to something else.

scripts/release/helm_registry_login.py
MaciejKaras: I think that we are not talking about the same thing. I'm saying is to ignore the `build_scenario` here completely and use `build_info.registry` and check if it starts with `quai.io` and then proceed with `quay` helm login. The same if we find `registry` to be ecr. If anything else, let's skip login and print warning. This behaviour is less dependant on the build scenario and we won't need to change anything in the future unless we add new registry type.

scripts/release/publish_helm_chart.py
MaciejKaras: >According to me if we have the build_scenario and with that if we can figure something out without looking into build_info.json, does it make sense to look into build_info.json?

If we follow this logic trail, we could just have the `build_scenario` variable and figure out everything in the code and don't read configuration in `build_info.json` at all. The reason for having `build_info.json` is to be able to configure different behaviour for different build scenarios and not have conditional logic in the code. Additional benefit is that we can read the `build_info.json` configuration and know how it is configured for each build scenario.

MaciejKaras: >Having the eventual version dependent on build scenario makes the code more readable, if it's dev/staging we are appending 0.0.0. But if we read the prefix from build_info, like you said we are hiding away why we are appending things based on prefix.

I think we can still add comment about this in `build_info.json` next to dev and staging scenarios.

MaciejKaras: >Or in other words, does it make sense to make something configurable, which doesn't need to be. We know that we are not going to change 0.0.0 to something else.

That's true, to build configuration will rarely change, but it doesn't mean this is the only benefit of having it - to change the configuration. The biggest gain in my opinion is to move abstract conditional logic from the code and make it simple, readable and testable.

scripts/release/helm_registry_login.py
viveksinghggits: no, I actually understood you the first time. What I have trying to say is, instead of comparing the registry comparing the build scenario is more readable and natural. It will essentially say that for release login into quay, otherwise login into ecr.
But I will go ahead and change this.

scripts/release/publish_helm_chart.py
viveksinghggits: ok, I can make the change. Is it ok if I do it in the new PR?
The reason is, I want to focus one PR in just one change, don't want to do multiple changes in single PR.

MaciejKaras: Sure, it can be done separately.

PR #544, opened 2025-10-21
CLOUDP-353180: Refactor replica set controller state handling, and use helper pattern

Julien-Ben:
# Summary

This PR refactors the ReplicaSet controller to use the helper pattern and separate state reading/writing, matching what we already do in the ShardedCluster and OpsManager controllers.

This is a no-op refactor: no behavior changes, just reorganizing code to prepare for multi-cluster support.

### Added ReplicaSetReconcilerHelper
- New helper struct that holds state for a single reconcile run
- Main reconcile logic moved from the controller to the helper
- Helper gets created fresh for each reconcile
- This is the pattern we used in our other reconcilers

### Separated State Operations
- `readState()` - reads deployment state from annotations
- `writeState()` - writes deployment state back to annotations ; write vault annotations only on successful reconciliation. In the same way it is done in the sharded controller. (note that vault is not supported for multi replica set at this point)
- `initialize()` - loads state during helper creation
- This is done to clearly show where we handle data persisted on the cluster, as opposed to reading and writing in multiple places during the reconciliation. This is important for multi-cluster support

### Helper's updateStatus() Override
- Writes deployment state after every status update
- Keeps state consistent even when we return early
- Same pattern as ShardedCluster controller

### Added ReplicaSetDeploymentState
- Holds the state we persist between reconciles
- Just has `LastAchievedSpec` and status member count for now
- The status is still read during the reconciliation loop, because of the scaler. Refactoring the scaler is also something that will be done as part of the epic

## Not in This PR

These will (potentially) come in follow-up PRs, in the main feature branch for multi cluster support:
- StateStore pattern (like ShardedCluster/OpsManager use)
- Use ConfigMap for state persistence
- State migration logic

## Proof of work

- Existing tests pass without changes

## Checklist

    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md

### DIFF
--- a/controllers/operator/mongodbreplicaset_controller.go
+++ b/controllers/operator/mongodbreplicaset_controller.go
@@ -130,32 +130,6 @@ func (r *ReplicaSetReconcilerHelper) readState() (*replicaSetDeploymentState, er
 	}, nil
 }
 
-// writeState writes the lastAchievedSpec annotation to the resource.
-// This should only be called on successful reconciliation.
-// The state is currently split between the annotations and the member count in status. Both should be migrated
-// to config maps
-// To avoid posting twice to the API server, we optionally include the vault annotations here, even though they are not
-// considered as proper state. The reconciler does not rely on them, they are write-only
-func (r *ReplicaSetReconcilerHelper) writeState(ctx context.Context, vaultAnnotations map[string]string) error {
-	// Get lastAchievedSpec annotation
-	annotationsToAdd, err := getAnnotationsForResource(r.resource)
-	if err != nil {
-		return err
-	}
-
-	for k, val := range vaultAnnotations {
-		annotationsToAdd[k] = val
-	}
-
-	// Write to CR
-	if err := annotations.SetAnnotations(ctx, r.resource, annotationsToAdd, r.reconciler.client); err != nil {
-		return err
-	}
-
-	r.log.Debugf("Successfully wrote deployment state for ReplicaSet %s/%s", r.resource.Namespace, r.resource.Name)
-	return nil
-}
-
 // getVaultAnnotations gets vault secret version annotations to write to the CR.
 func (r *ReplicaSetReconcilerHelper) getVaultAnnotations() map[string]string {
 	if !vault.IsVaultSecretBackend() {
@@ -329,9 +303,21 @@ func (r *ReplicaSetReconcilerHelper) Reconcile(ctx context.Context) (reconcile.R
 		return r.updateStatus(ctx, workflow.Pending("Continuing scaling operation for ReplicaSet %s, desiredMembers=%d, currentMembers=%d", rs.ObjectKey(), rs.DesiredReplicas(), scale.ReplicasThisReconciliation(rs)), mdbstatus.MembersOption(rs))
 	}
 
-	// Write state and vault annotations on successful reconciliation
-	if err := r.writeState(ctx, r.getVaultAnnotations()); err != nil {
-		return r.updateStatus(ctx, workflow.Failed(xerrors.Errorf("failed to write state: %w", err)))
+	// Get lastspec, vault annotations when needed and write them to the resource.
+	// These operations should only be performed on successful reconciliations.
+	// The state of replica sets is currently split between the annotations and the member count in status. Both should
+	// be migrated to config maps
+	annotationsToAdd, err := getAnnotationsForResource(r.resource)
+	if err != nil {
+		return r.updateStatus(ctx, workflow.Failed(xerrors.Errorf("could not get resource annotations: %w", err)))
+	}
+
+	for k, val := range r.getVaultAnnotations() {
+		annotationsToAdd[k] = val
+	}
+
+	if err := annotations.SetAnnotations(ctx, r.resource, annotationsToAdd, r.reconciler.client); err != nil {
+		return r.updateStatus(ctx, workflow.Failed(xerrors.Errorf("could not update resource annotations: %w", err)))
 	}
 
 	log.Infof("Finished reconciliation for MongoDbReplicaSet! %s", completionMessage(conn.BaseURL(), conn.GroupID()))


controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -76,109 +78,182 @@ type ReconcileMongoDbReplicaSet struct {
 	databaseNonStaticImageVersion     string
 }
 
+type ReplicaSetDeploymentState struct {
+	LastAchievedSpec  *mdbv1.MongoDbSpec `json:"lastAchievedSpec"`
+	MemberCountBefore int                `json:"memberCountBefore"`
```
m1kola: Are we marshalling this struct?

m1kola: q/nit: What does `Before` in `MemberCountBefore` refer to? Is it actual state?

Edit: below I see that this gets assigned `Status.Members`, so I think the answer is - yes, it is actual state. Right?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -76,109 +78,182 @@ type ReconcileMongoDbReplicaSet struct {
...
+		for _, s := range secrets {
+			path := fmt.Sprintf("%s/%s/%s", r.reconciler.VaultClient.DatabaseSecretMetadataPath(), r.resource.Namespace, s)
+			vaultMap = merge.StringToStringMap(vaultMap, r.reconciler.VaultClient.GetSecretAnnotation(path))
+		}
+		path := fmt.Sprintf("%s/%s/%s", r.reconciler.VaultClient.OperatorScretMetadataPath(), r.resource.Namespace, r.resource.Spec.Credentials)
+		vaultMap = merge.StringToStringMap(vaultMap, r.reconciler.VaultClient.GetSecretAnnotation(path))
+		for k, val := range vaultMap {
+			annotationsToAdd[k] = val
+		}
+	}
```
m1kola: This is in `writeState` method. Is vault integration related to deployment state?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -76,109 +78,182 @@ type ReconcileMongoDbReplicaSet struct {
...
-	if reconcileResult, err := r.prepareResourceForReconciliation(ctx, request, rs, log); err != nil {
-		if errors.IsNotFound(err) {
-			return workflow.Invalid("Object for reconciliation not found").ReconcileResult()
-		}
-		return reconcileResult, err
+// updateStatus overrides the common controller's updateStatus to ensure that the deployment state
+// is written after every status update. This ensures state consistency even on early returns.
+// Vault annotations are only included when status.IsOK() is true, meaning reconciliation succeeded.
+// This function must be executed only once per reconcile (with a return)
+func (r *ReplicaSetReconcilerHelper) updateStatus(ctx context.Context, status workflow.Status, statusOptions ...mdbstatus.Option) (reconcile.Result, error) {
```
m1kola, line 167: I find it confusing - a method called `updateStatus` which updates resource's status, but also updates the resource itself (annotations)?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -238,54 +312,85 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
...
+// +kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=create;get;list;watch;delete;update,namespace=placeholder
+
+// MongoDB Resource
+// +kubebuilder:rbac:groups=mongodb.com,resources={mongodb,mongodb/status,mongodb/finalizers},verbs=*,namespace=placeholder
+
+// Setting up a webhook
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;create;update;delete
+
+// Certificate generation
+// +kubebuilder:rbac:groups=certificates.k8s.io,resources=certificatesigningrequests,verbs=get;create;list;watch
```
m1kola, line 361: Nit: Do we actually use kubebuilder annotations to generate RBAC? Maybe we should drop them?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -190,22 +265,22 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
...
 
 	agentCertSecretName := rs.GetSecurity().AgentClientCertificateSecretName(rs.Name)
-	agentCertHash, agentCertPath := r.agentCertHashAndPath(ctx, log, rs.Namespace, agentCertSecretName, databaseSecretPath)
+	agentCertHash, agentCertPath := reconciler.agentCertHashAndPath(ctx, log, rs.Namespace, agentCertSecretName, databaseSecretPath)
 
-	prometheusCertHash, err := certs.EnsureTLSCertsForPrometheus(ctx, r.SecretClient, rs.GetNamespace(), rs.GetPrometheus(), certs.Database, log)
+	prometheusCertHash, err := certs.EnsureTLSCertsForPrometheus(ctx, reconciler.SecretClient, rs.GetNamespace(), rs.GetPrometheus(), certs.Database, log)
 	if err != nil {
-		return r.updateStatus(ctx, rs, workflow.Failed(xerrors.Errorf("could not generate certificates for Prometheus: %w", err)), log)
+		return r.updateStatus(ctx, workflow.Failed(xerrors.Errorf("Could not generate certificates for Prometheus: %w", err)))
```
m1kola: Nit: errors normally do not start from a capital letter because. Most often then not they get concatenated.

Also below we still have a mix of lower- and upper-case errors.

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: Not for now, but if we store it in a config map at some point then it will be serialized to JSON

Julien-Ben: `MemberCountBefore` refer to what the member count was at the end of the last reconciliation
Before this PR we did the following:
- Read this in the status in a few places
- Update the status once, at the end of every reconcile with the new member count

So we do an equivalent thing, but we read it only once and then keep it in the state data structure
Eventually, we want to never rely on the status, as we do for e.g sharded clusters

This is an intermediary step to at least have a single data structure and a single read

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: I agree, I followed the pattern from other reconciler here
Do you think a rename is enough ?

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: Good question

This is the diff being a bit weird here but I haven't added them, you can see the same lines above being deleted. They were just moved around with the refactoring

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: done: https://github.com/mongodb/mongodb-kubernetes/pull/544/commits/ac8fb339060e6e7705ddd1254b5066060cb142ae

We should probably set the linter to catch this

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: We were storing it in annotations on successful reconciliations previously so I treated it as state:
https://github.com/mongodb/mongodb-kubernetes/blob/582d248c7b25ca2242a0e7ecf02430ae6e7bc200/controllers/operator/mongodbreplicaset_controller.go#L269

controllers/operator/mongodbreplicaset_controller.go
m1kola: We only generate CRDs, so this is safe to remove. But it is irrelevant. Resolving this thread.

controllers/operator/mongodbreplicaset_controller.go
m1kola: Before posting this review I checked and realised it is the same pattern as in sharded so I made an edit to this comment, but it looks look I didn't save it 🤦‍♂️ .

I don't like it but I also think it is better to be consistent across controllers. So resolving this thread.

controllers/operator/mongodbreplicaset_controller.go
m1kola: @Julien-Ben maybe `lastAchievedMemberCount` would be a better name to follow `lastAchievedSpec`, `lastAchievedRsMemberIds`? Or just `memberCount` in the context of reconciliation & deployment state should be clear that it is about current.

controllers/operator/mongodbreplicaset_controller.go
m1kola: @Julien-Ben I don't know a lot about Vault integration, but it seems unrelated to the deployment state to me. I also see that a similar piece of code in the sharded controller is is part of `ShardedClusterReconcileHelper.Reconcile` method:

https://github.com/mongodb/mongodb-kubernetes/blob/582d248c7b25ca2242a0e7ecf02430ae6e7bc200/controllers/operator/mongodbshardedcluster_controller.go#L928-L940

Should we follow the same pattern?

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: I renamed it to lastReconcileMemberCount: https://github.com/mongodb/mongodb-kubernetes/pull/544/commits/d5bdd182edce0e60f9e9ae910bf9e39d95dcd83e

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: I separated them, you were right.
Vault just needs to read annotations, and they are not state as we don't rely on them ourselves. They should be written only upon successful reconcile

https://github.com/mongodb/mongodb-kubernetes/commit/a777b2f428919b15182e62899afc6e7fddff0087 

I also changed how we write to lastAchievedSpec
It is actually an annotation we should change only when we achieve Running phase, not on every reconcile.

controllers/operator/mongodbreplicaset_controller_test.go
```diff
@@ -858,6 +892,130 @@ func TestHandlePVCResize(t *testing.T) {
 	testPVCFinishedResizing(t, ctx, memberClient, p, reconciledResource, statefulSet, logger)
 }
 
+// ===== Test for state and vault annotations handling in replicaset controller =====
```
Julien-Ben, line 862: @m1kola FYI after your review I added these unit tests

api/v1/mdb/mongodb_types.go
```diff
@@ -243,6 +243,9 @@ func GetLastAdditionalMongodConfigByType(lastSpec *MongoDbSpec, configType Addit
 
 // GetLastAdditionalMongodConfigByType returns the last successfully achieved AdditionalMongodConfigType for the given component.
 func (m *MongoDB) GetLastAdditionalMongodConfigByType(configType AdditionalMongodConfigType) (*AdditionalMongodConfig, error) {
```
Julien-Ben, line 245: This function is now used by the standalone controller only.
The reason we don't want it any more is because it acceeds the spec with m.GetLastSpec() 

We now do it only once per reconcile

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -238,54 +311,94 @@ func (r *ReconcileMongoDbReplicaSet) Reconcile(ctx context.Context, request reco
...
-		vaultMap := make(map[string]string)
-		for _, s := range secrets {
-			path := fmt.Sprintf("%s/%s/%s", r.VaultClient.DatabaseSecretMetadataPath(), rs.Namespace, s)
-			vaultMap = merge.StringToStringMap(vaultMap, r.VaultClient.GetSecretAnnotation(path))
-		}
-		path := fmt.Sprintf("%s/%s/%s", r.VaultClient.OperatorScretMetadataPath(), rs.Namespace, rs.Spec.Credentials)
-		vaultMap = merge.StringToStringMap(vaultMap, r.VaultClient.GetSecretAnnotation(path))
-		for k, val := range vaultMap {
-			annotationsToAdd[k] = val
+	if err := r.writeVaultAnnotations(ctx); err != nil {
```
m1kola: Both `writeState` and `writeVaultAnnotations` call `annotations.SetAnnotations` which gets the object and patches it.

If we can reduce write operations to the API server - that would be great.

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -76,109 +78,181 @@ type ReconcileMongoDbReplicaSet struct {
...
+	}
+
+	// Read current member count from Status once at initialization. This provides a stable view throughout
+	// reconciliation and prepares for eventually storing this in ConfigMap state instead of ephemeral status.
+	memberCountBefore := r.resource.Status.Members
+
+	return &ReplicaSetDeploymentState{
+		LastAchievedSpec:         lastAchievedSpec,
+		LastReconcileMemberCount: memberCountBefore,
+	}, nil
```
m1kola, line 130: nit: Old name in the var `memberCountBefore`. Alternatively we can just get rid of it:

```suggestion
	return &ReplicaSetDeploymentState{
		LastAchievedSpec: lastAchievedSpec,
		
		// Read current member count from Status once at initialization. This provides a stable view throughout
		// reconciliation and prepares for eventually storing this in ConfigMap state instead of ephemeral status.		
		LastReconcileMemberCount: r.resource.Status.Members,
	}, nil
```

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -76,109 +78,181 @@ type ReconcileMongoDbReplicaSet struct {
 	databaseNonStaticImageVersion     string
 }
 
+type ReplicaSetDeploymentState struct {
+	LastAchievedSpec         *mdbv1.MongoDbSpec `json:"lastAchievedSpec"`
+	LastReconcileMemberCount int                `json:"memberCountBefore"`
+}
```
m1kola: Nit: does it need to be exportable?

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: I fixed this and we write only once now

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: Good catch, fixed

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: I removed it for now, will add it if we serialize to JSON later

PR #547, opened 2025-10-22
CLOUDP-295785 - Fix multi-arch builds by migrating away from goreleaser

MaciejKaras:
# Summary

⚠️ Some of the files were moved and this messes up GitHub preview. It's better to review the last commit with all the changes https://github.com/mongodb/mongodb-kubernetes/pull/547/commits/ec01373debc783ad444fa27b4f65389508cb5727

Completely removed `goreleaser` usage in the project. It was preventing us to build multiarch `kubect-mongodb` plugin binaries. There were also other reasons to drop `goreleaser` i.e. it was only used to build binaries, nothing else was used, but required us to keep additional, static `.goreleaser.yaml` that was not compatible with `build_info.json`. 

This PR also improves quite a lot of other things:

- moves building of `kubectl-mongodb` plugin to a separate evergreen task
- `kubectl-mongodb` plugin is now only dowloaded when building test image. This means we need to wait with building test image for `kubectl-plugin` to be available in s3
- `promote_kubectl_plugin.py` now uses temp_dir instead of hardcoded `./artifacts` dir. Previously we couldn't re-run the promotion pipeline if the signatures were already present on the machine: `Signature already exists. Displaying proof` 
- simplified and unified the [build_kubectl_plugin.py‎](https://github.com/mongodb/mongodb-kubernetes/pull/547/files#diff-a6df2ec746c87a5d07b813bb16612734474e6c40f6f60a38f194802403dd4fe7) and [download_kubectl_plugin.py](https://github.com/mongodb/mongodb-kubernetes/pull/547/files#diff-eac03d2b21304201e625cb87a6701449d4ba1cfa0c3dbaa129db2dbd6522f2d4) scripts
- removed all references to `goreleaser` 
- `kubectl-mongodb` also pushes binaries to s3 release bucket so they can be used for e2e smoke tests
- `kubectl-mongodb` is now pushed to staging and released with two new architectures:
  -  `linux/s390x`
  -  `linux/ppc64le`

## Proof of Work

Passing CI for PR.

Passing CI for [smoke arm on staging scenario](https://spruce.mongodb.com/version/68fb37de9e149f0007e8638a) (with arm smoke tests enabled)

Passing CI for [release scenario](https://spruce.mongodb.com/version/68fa1b1b9b6f3600075262c0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (`release_kubectl_mongodb_plugin` fails on adding assets to GH release, which is expected since there is no release `rc-1.3.0`)

## Checklist


scripts/release/argparse_utils.py
```diff
@@ -10,3 +13,22 @@ def str2bool(v):
...
+    except ValueError as e:
+        raise ValueError(f"Invalid scenario '{args_scenario}': {e}")
+
+
+def get_platforms_from_arg(args_platforms: str) -> list[str] | None:
+    if not args_platforms:
+        return None
+
+    platforms = [p.strip() for p in args_platforms.split(",")]
+    if any(p not in SUPPORTED_PLATFORMS for p in platforms):
```
lucian-tosa, line 30: SUPPORTED_PLATFORMS is still limited to amd64 and arm64. We might want to extend that, especially since we need darwin/arm64 locally

.evergreen-functions.yml
lucian-tosa, line 1: This is inconsistent. Building and downloading the binary use a one-line bash script, while promoting uses `run_python.sh`. I personally prefer using run_python directly instead of having many one-line scripts

.evergreen.yml
lucian-tosa, line 1733: this needs a dependency as well

scripts/release/argparse_utils.py
MaciejKaras: Good catch!

.evergreen.yml
MaciejKaras: that is true, but I have not enabled it yet. I can add dependency either way. Edit: missed that!

.evergreen-functions.yml
MaciejKaras: I prefer run_python directly as well, but the issue with that is we cannot use our context variables directly in such call. For example having:
```
binary: scripts/dev/run_python.sh scripts/release/kubectl_mongodb/download_kubectl_plugin.py --build-scenario "${BUILD_SCENARIO}" --version "${OPERATOR_VERSION}"
```

... will not work because `BUILD_SCENARIO` and `OPERATOR_VERSION` are not exposed in evergreen and only exported when sourcing the .generated/context.env file

scripts/release/kubectl_mongodb/promote_kubectl_plugin.py
viveksinghggits, line 1: If we used `git mv` to move these files around, i think we won't see this as a new file.

.evergreen-functions.yml
MaciejKaras: `promote_kubectl_plugin.py` is using env variables directly in the code and we can call directly from evg. Other python scripts need bash wrapper.

scripts/release/kubectl_mongodb/promote_kubectl_plugin.py
MaciejKaras: yeah, I think it would be better. I'm not sure if it is easy to do it now

scripts/release/kubectl_mongodb/kubectl_mac_notarize.sh
```diff
@@ -17,30 +17,38 @@
...
 
 # Notarize generated binaries with Apple and replace the original binary with the notarized one
-# This depends on binaries being generated in a goreleaser manner and gon being set up.
-# goreleaser should already take care of calling this script as a hook.
 
 if [ -z "${1-}" ]; then
-  echo "Error: Missing required argument <version> as first positional parameter to script"
-  echo "Usage: ./kubectl_mac_notarize.sh <version>"
+  echo "Error: Missing required argument <temp_dir> as first positional parameter to script"
+  echo "Usage: ./kubectl_mac_notarize.sh <temp_dir> <version>"
```
viveksinghggits: Using temp dir is not really required right? Irrespective of whether we are able to update cosign or not.

scripts/release/kubectl_mongodb/promote_kubectl_plugin.py
viveksinghggits, line 1: If this file's content are not changed, can we mention this here for reviewers help. We can mention that it's already in master we are just moving it to diff location.

viveksinghggits: I just read the entire file and it looks like we are changing it slightly.

scripts/release/kubectl_mongodb/promote_kubectl_plugin.py
```diff
@@ -0,0 +1,304 @@
...
+        try:
+            s3_client.upload_file(local_file_path, release_scenario_bucket_name, s3_key)
+            logger.debug(f"{local_file_path} was promoted to s3://{release_scenario_bucket_name}/{s3_key} successfully")
+        except ClientError as e:
+            raise Exception(f"failed to upload the file {local_file_path}: {e}")
+
+    logger.info("All artifacts were promoted to release bucket successfully")
+
+
+def is_expected_artifact_for_promotion(file_path: str) -> bool:
```
viveksinghggits, line 132: is this required because now we have the compressed as well as decompressed file in staging? I am a little sceptical about the idea because it would make things confusing, like we discussed on call.
But if we HAVE to do this, we should add a comment here explaining that.

viveksinghggits: Would it be problematic if run `tar` locally and don't upload copies or artifacts in S3.

.goreleaser.yaml
viveksinghggits, line 1: should we keep this file for some time, to make sure we can use this to release plugin in case of issues?

scripts/release/kubectl_mongodb/kubectl_mac_notarize.sh
MaciejKaras: Yes, it's not required for this change. I can try to revert it and use ./artifacts dir, but I don't think it makes difference other than seeing less changes in the PR.

scripts/release/kubectl_mongodb/promote_kubectl_plugin.py
MaciejKaras: >is this required because now we have the compressed as well as decompressed file in staging?

only in release bucket, in staging we still have only decompressed ones.

>if we HAVE to do this, we should add a comment here explaining that

yes, we have to put both compressed as well as decompressed files. Compressed are used for release assets backup while uncompressed are used for e2e testing. I wanted to keep the simple logic when calling `download_multi_cluster_binary` that is used when building test image and not having to decompress binaries, when downloading from release bucket. It's much simpler to just push both flavours.

.goreleaser.yaml
MaciejKaras: we can always revert it back or copy it from git history. If it is unused then lets not keep it in repo

scripts/release/kubectl_mongodb/kubectl_mac_notarize.sh
viveksinghggits: seeing less changes in PR and it would also not redo something that is already working and is in master. so I am in favor or keeping it has it was before.

MaciejKaras: yeah, makes sense. Let me revert it back

scripts/release/kubectl_mongodb/promote_kubectl_plugin.py
MaciejKaras: Done!

MaciejKaras: I've used git mv command and moved the files in the first commit and changed in another, but this didn't help with the GitHub preview :/

MaciejKaras: Reviewing only the latest commit helps! https://github.com/mongodb/mongodb-kubernetes/pull/547/commits/5eb86a921c4f2c6f43e903e8bbe3fa67ec5db051

.evergreen-functions.yml
lucian-tosa: Exactly, could we have all scripts follow the same design? `promote...` can read `OPERATOR_VERSION` from env, but the others need an explicit flag.

MaciejKaras: I had a comment from Nam in previous PRs (and I agree with him) that the python script should not hide arguments that are passed to it using environment variables. It's harder to tests and run locally, that's why we have bash wrappers

lucian-tosa: Then should we use shell.exec instead of subprocess.exec and include the bash script inline?
Also, if we decide we don't want to use env variables, then `promote...` should also use explicit flags

MaciejKaras: Based on discussions with Evergreen team and their [documentation](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Best-Practices#subprocessexec) shell.exec should be avoided. Mainly because it's harder to debug, more error prone (we already had multiple issues with silenced errors) and makes our scripts part of .evegreen.yaml configuration.

I'm not saying that having bash wrapper script is ideal, but I have not came up with nothing better yet. 

For unifying the `promote_kubectl_plugin` - this PR was not supposed to make big changes to `promote_kubectl_plugin.py` that Vivek wrote, but I can do it in separate PR.

PR #549, opened 2025-10-24
CLOUDP-349087: Block simultaneous TLS disabling and scaling for ReplicaSets

Julien-Ben:
# Summary

We have code in the replica set controller dedicated to handling a scenario where the user disable TLS, and ascale their cluster at the same time. The test for this behaviour was broken. Because there were two functions with the name `test_tls_is_disabled_and_scaled_up`, so pytest would run only one of them.

When I fixed the test, it failed: https://spruce.mongodb.com/task/mongodb_kubernetes_e2e_mdb_kind_ubi_cloudqa_e2e_disable_tls_scale_up_patch_dfb9424e9b34ddd048a725a9988114ca4032f9bf_68de79b2fccf070007a2c51e_25_10_02_13_10_12/tests?execution=4&sorts=STATUS%3AASC

Which means the code to handle it was incorrect. My opinion is that it is better to block this change, rather than introducing complexity to handle it. It is not a common use case.

A changelog entry was added. Question: should it be explicitly mentioned in our public documentation ?

Old related PR (2021): https://github.com/10gen/ops-manager-kubernetes/pull/1444 
And ticket: https://jira.mongodb.org/browse/CLOUDP-80768

## Proof of Work

Unit and e2e test for blocking the operation should pass.
No regression due to the change in `updateOmDeploymentRs` should be observed.

## Checklist

### DIFF
--- a/controllers/operator/mongodbreplicaset_controller_test.go
+++ b/controllers/operator/mongodbreplicaset_controller_test.go
@@ -68,7 +68,7 @@ func TestCreateReplicaSet(t *testing.T) {
 
 	connection := omConnectionFactory.GetConnection()
 	connection.(*om.MockedOmConnection).CheckDeployment(t, deployment.CreateFromReplicaSet("fake-mongoDBImage", false, rs), "auth", "ssl")
-	connection.(*om.MockedOmConnection).CheckNumberOfUpdateRequests(t, 2)
+	connection.(*om.MockedOmConnection).CheckNumberOfUpdateRequests(t, 1)
 }
 
 func TestReplicaSetRace(t *testing.T) {


controllers/operator/mongodbreplicaset_controller_test.go
```diff
@@ -68,7 +68,7 @@ func TestCreateReplicaSet(t *testing.T) {
 
 	connection := omConnectionFactory.GetConnection()
 	connection.(*om.MockedOmConnection).CheckDeployment(t, deployment.CreateFromReplicaSet("fake-mongoDBImage", false, rs), "auth", "ssl")
-	connection.(*om.MockedOmConnection).CheckNumberOfUpdateRequests(t, 2)
+	connection.(*om.MockedOmConnection).CheckNumberOfUpdateRequests(t, 1)
```
Julien-Ben, line 71: Reduced to one since we removed one update step in `updateOmDeploymentRs`

changelog/20251024_fix_block_simultaneous_tls_disabling_and_scaling.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2025-10-24
+---
+
+* **ReplicaSet**: Blocked disabling TLS and changing member count simultaneously. These operations must now be applied separately to prevent configuration inconsistencies.
```
vinilage, line 6: LGMT!

api/v1/mdb/mongodb_validation.go
```diff
@@ -369,6 +369,25 @@ func noTopologyMigration(newObj, oldObj MongoDbSpec) v1.ValidationResult {
 	return v1.ValidationSuccess()
 }
 
+func noSimultaneousTLSDisablingAndScaling(newObj, oldObj MongoDbSpec) v1.ValidationResult {
```
m1kola, line 372: Looks like a good candidate for CEL validation on the API server so changes like this don't get even admitted to the storage.

Julien-Ben: That's definitely something we can add to a Discuss agenda, or at least discuss in the team channel !

PR #551, opened 2025-10-24
CLOUDP-349078: custom roles regression tests

lsierant:
# Summary

This PR adds regression test for custom roles ensuring we will keep backwards compatibility wrt handling empty strings vs omitted fields. 

In the subsequent #553 the serialization rules for custom roles will be changed. 

## Proof of Work
Tests passing.

PR #553, opened 2025-10-25
CLOUDP-350197: allow to create a custom role with cluster: true

lsierant:
# Summary

This PR fixes a problem which was preventing creating cluster-wide mongodb custom roles. 

**Context**
As described in [MongoDB docs](https://www.mongodb.com/docs/manual/reference/system-roles-collection/#mongodb-data-admin.system.roles.privileges-n-.resource) the resource object is either: 
```
{ db: <database>, collection: <collection> }
```
or
```
{ cluster : true }
```

So that means we cannot provide empty strings for `db` and `collection` fields as empty string means "any db/collection". 

At the same time our previous serialization rules were always sending empty strings even if not set. This was the source of the problem - it wasn't possible to specify `cluster: true` because the operator was also sending empty db and collections. 

**Backwards compatibility**
Making `db` and `collection` fields just as omitempty `*string` is not sufficient, because that would change the semantics of the resource and would be potentially a breaking change. In order to preserve backwards compatibility we need additional logic (see [normalizePrivilegeResource](https://github.com/mongodb/mongodb-kubernetes/pull/553/files#diff-457d90107d5cbfbfb07a276cc6d58bd1f53c710f58ce19d6ec54daa7ab8e08aaR174) that will maintain the same behavior for non-cluster-wide resources (sending empty strings even if the field is not set in yaml).

## Proof of Work

Tests passing.

# Based on PR #551

## Chain of upstream PRs as of 2025-10-25

* PR #551:
  `master` ← `lsierant/custom-roles-regression-tests`

  * **PR #553 (THIS ONE)**:
    `lsierant/custom-roles-regression-tests` ← `lsierant/custom-roles`

### DIFF
--- a/changelog/20251027_fix_cloudp_350197_allow_to_create_a_custom_role_with.md
+++ b/changelog/20251027_fix_cloudp_350197_allow_to_create_a_custom_role_with.md
@@ -3,4 +3,4 @@ kind: fix
 date: 2025-10-27
 ---
 
-* Fixed a problem that prevented specifying cluster-wide privileges in database custom roles.
+* Fixed inability to specify cluster-wide privileges in custom roles.


changelog/20251027_fix_cloudp_350197_allow_to_create_a_custom_role_with.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2025-10-27
+---
+
+* Fixed a problem that prevented specifying cluster-wide privileges in database custom roles.
```
vinilage: I propose to change to: "Fixed inability to specify cluster-wide privileges in custom roles."

lsierant: updated, thanks!

changelog/20251027_fix_cloudp_350197_allow_to_create_a_custom_role_with.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2025-10-27
+---
+
+* Fixed inability to specify cluster-wide privileges in custom roles.
```
vinilage, line 6: LGTM!

PR #566, opened 2025-10-31
CLOUDP-328217: Automation agent password secret

filipcirtog:
# Summary

If a deployment is moved to a different project, the automation agent password will be re-generated, triggering a password change in the automation plan.

This will cause a deadlock in a sharded cluster due to the multiple components requiring automation. However, it will not cause issues in replicasets.

This is a blocker for migrating projects in sharded deployments.

## Proof of Work

For SCRAM (the only auth mechanism who re-generates a pwd), we now save the automation agent's password in a secret. During migration, the stored secret is utilized to preserve the password ,ensuring project migration possible.

## Observed problems
For LDAP (Sharded + Replica), the following tests are failing, even though the only modification made is updating the MongoDB resource's project reference, with no other changes applied. To help further investigation, I have commented out certain code in the tests (which can make them fail) so the issue can be consistently reproduced. While the deployment returns to the "running" state, the users are missing from the automation configuration.

Additionally, certain flaky parts in the SHA1 and SHA256 tests, as well as the LDAP and X509 project switch tests, have been commented out/disabled due to the lack of support for project migrations. These tests should be activated once project migration functionality has been implemented.

## Checklist


controllers/operator/authentication/authentication.go
```diff
@@ -82,7 +86,7 @@ type UserOptions struct {
 
 // Configure will configure all the specified authentication Mechanisms. We need to ensure we wait for
 // the agents to reach ready state after each operation as prematurely updating the automation config can cause the agents to get stuck.
-func Configure(conn om.Connection, opts Options, isRecovering bool, log *zap.SugaredLogger) error {
+func Configure(client kubernetesClient.Client, ctx context.Context, mdbNamespacedName *types.NamespacedName, conn om.Connection, opts Options, isRecovering bool, log *zap.SugaredLogger) error {
```
lsierant: nit: ctx should always be first arg

controllers/operator/authentication/authentication.go
```diff
@@ -151,7 +155,7 @@ func Configure(conn om.Connection, opts Options, isRecovering bool, log *zap.Sug
 
 // Disable disables all authentication mechanisms, and waits for the agents to reach goal state. It is still required to provide
 // automation agent username, password and keyfile contents to ensure a valid Automation Config.
-func Disable(conn om.Connection, opts Options, deleteUsers bool, log *zap.SugaredLogger) error {
+func Disable(client kubernetesClient.Client, ctx context.Context, mdbNamespacedName *types.NamespacedName, conn om.Connection, opts Options, deleteUsers bool, log *zap.SugaredLogger) error {
```
lsierant: `types.NamespacedName` is usually not passed by pointer. Is there a reason it's a pointer here? Is passing nil here a valid case?

docker/mongodb-kubernetes-tests/tests/authentication/fixtures/switch-project/replica-set-scram-sha-1-switch-project.yaml
```diff
@@ -0,0 +1,23 @@
...
+    configMapRef:
+      name: my-project
+  credentials: my-credentials
+  logLevel: DEBUG
+  persistent: false
+  security:
+    authentication:
+      agents:
+        # This may look weird, but without it we'll get this from OpsManager:
+        # Cannot configure SCRAM-SHA-1 without using MONGODB-CR in te Agent Mode","reason":"Cannot configure SCRAM-SHA-1 without using MONGODB-CR in te Agent Mode
```
lsierant: could you pls fix the typo in "te" in the mentioned validation btw.?

docker/mongodb-kubernetes-tests/tests/authentication/fixtures/switch-project/replica-set-scram-sha-1-switch-project.yaml
```diff
@@ -0,0 +1,23 @@
...
+  opsManager:
+    configMapRef:
+      name: my-project
+  credentials: my-credentials
+  logLevel: DEBUG
+  persistent: false
+  security:
+    authentication:
+      agents:
+        # This may look weird, but without it we'll get this from OpsManager:
```
lsierant: nit: remove "This may look weird" - let's state the _why_ objectively.

lsierant: btw. SCRAM-SHA-1 is deprecated and it requires some additional legacy enablement with that MONGODB-CR IIRC

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_ldap_switch_project.py
```diff
@@ -0,0 +1,176 @@
...
+
+@pytest.fixture(scope="module")
+def replica_set(
+    openldap: OpenLDAP,
+    issuer_ca_configmap: str,
+    ldap_mongodb_agent_user: LDAPUser,
+    server_certs: str,
+    namespace: str,
+) -> MongoDB:
+    resource = MongoDB.from_yaml(find_fixture(f"switch-project/{MDB_FIXTURE_NAME}.yaml"), namespace=namespace)
```
lsierant: could we use some basic fixture to not create redundant yamls? I see you're configuring all the security and auth in the test anyway

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+@pytest.fixture(scope="module")
+def project_name_prefix(namespace: str) -> str:
+    """
+    Generates a random Kubernetes project name prefix based on the namespace.
+
+    Ensures test isolation in a multi-namespace test environment.
+    """
+    return random_k8s_name(f"{namespace}-project-")
```
lsierant: we don't need to randomize it - namespace is randomized in evg anyway. And randomizing it is just making local runs difficult and not possible to re-run.

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
```diff
@@ -0,0 +1,166 @@
...
+
+CONFIG_MAP_KEYS = {
+    "BASE_URL": "baseUrl",
+    "PROJECT_NAME": "projectName",
+    "ORG_ID": "orgId",
+}
+
+
+@pytest.fixture(scope="module")
+def replica_set(namespace: str) -> MongoDB:
```
lsierant, line 20: resource fixture should be scoped to function and have if try_load(resource) applied (look for it in other tests). 

We don't have this pattern applied across the board, but we try to use it for newer tests.

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
lsierant: if you want to have different project names for different deployments then just add a resource name to it: {namespace}-{mdb.name} - it will suffice for making them unique for the test

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
```diff
@@ -0,0 +1,166 @@
...
+        Ensure Ops Manager state is correctly updated in the original replica set.
+        """
+        tester = replica_set.get_automation_config_tester()
+        tester.assert_authentication_mechanism_enabled("MONGODB-CR")
+        tester.assert_authoritative_set(True)
+        tester.assert_authentication_enabled(2)
+        tester.assert_expected_users(0)
+
+    def test_create_secret(self):
+        print(f"creating password for MongoDBUser {self.USER_NAME} in secret/{self.PASSWORD_SECRET_NAME} ")
```
lsierant: is this necessary? normally we have the progress visible when running the test, i.e. which test step is currently executing. There is nothing more than that so I think it's redundant

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
```diff
@@ -0,0 +1,166 @@
...
+            namespace=namespace,
+            name=new_project_name,
+            data={
+                CONFIG_MAP_KEYS["BASE_URL"]: original_configmap[CONFIG_MAP_KEYS["BASE_URL"]],
+                CONFIG_MAP_KEYS["PROJECT_NAME"]: new_project_name,
+                CONFIG_MAP_KEYS["ORG_ID"]: original_configmap[CONFIG_MAP_KEYS["ORG_ID"]],
+            },
+        )
+
+        replica_set.load()
```
lsierant: when you use the pattern with function scope and try_load, you don't need to load it manually in the tests avoiding flakiness errors due to stale object

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
```diff
@@ -0,0 +1,166 @@
...
+        replica_set["spec"]["opsManager"]["configMapRef"]["name"] = new_project_configmap
+        replica_set.set_version(custom_mdb_version)
```
lsierant: the version should be only set in the resource's fixture unless the changing version is part of the test.

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
```diff
@@ -0,0 +1,166 @@
...
+        replica_set.update()
+
+        replica_set.assert_reaches_phase(Phase.Running, timeout=600)
+
+    def test_moved_replica_set_connectivity(self):
+        """
+        Verify connectivity to the replica set after switching projects.
```
lsierant: the comment is redundant, just name the function name so it's self describing (it's already good)

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        ReplicaSetTester(MDB_RESOURCE_NAME, 3).assert_connectivity()
+
+    def test_ops_manager_state_correctly_updated_in_moved_replica_set(self, replica_set: MongoDB):
+        """
+        Ensure Ops Manager state is correctly updated in the moved replica set after the project switch.
```
lsierant: this comment also is not adding much over the already good function name

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
```diff
@@ -0,0 +1,166 @@
...
+from kubetester.mongodb import MongoDB
+from kubetester.mongodb_user import MongoDBUser
+from kubetester.mongotester import ReplicaSetTester
+from kubetester.phase import Phase
+
+# Constants
+MDB_RESOURCE_NAME = "replica-set-scram-sha-1-switch-project"
+MDB_FIXTURE_NAME = MDB_RESOURCE_NAME
+
+CONFIG_MAP_KEYS = {
```
lsierant: is this necessary? Those keys won't be ever changed so we can just inline them.

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_x509_switch_project.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    Generates a random Kubernetes project name prefix based on the namespace.
+
+    Ensures test isolation in a multi-namespace test environment.
+    """
+    return random_k8s_name(f"{namespace}-project-")
+
+
+@pytest.mark.e2e_replica_set_x509_switch_project
+class TestReplicaSetCreationAndProjectSwitch(KubernetesTester):
```
lsierant, line 51: are the three or four test classes here any different? Do you think we could extract it as a reusable test class parametrized with resource and auth mechanism? 

The way to do this is to have a generic test helper (important: without the pytest.mark annotation!)

```
class ReplicaSetCreationAndProjectSwitchTestHelper: 
     def test_create_replica_set
     def test_ops_manager_state_correctly_updated_in_initial_replica_set
[...]
```

And in the actual test files we could have only test functions that simply delegate to the common code: 

```
@pytest.fixture
def test_helper() -> ReplicaSetCreationAndProjectSwitchTestHelper:
    ... configure and return test helper

@pytest.mark.e2e_replica_set_x509_switch_project
def test_create_replica_set(test_helper: ReplicaSetCreationAndProjectSwitchTestHelper):
        test_helper.test_create_replica_set()

... etc. 
```


I'm a bit worried that we've added just too much of duplicated code. We've already have a code duplication problem - let's try to not exacerbate it further.

lsierant: example of using test helper to reduce duplication: 
https://github.com/mongodb/mongodb-kubernetes/blob/78e9a13bc51168e78574d3f46671e8a8a9dc6c55/docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py#L216

unfortunately for now we cannot reuse easily whole test classes with the testing steps. The test functions/classes must be defined in each file separately, but we can organize the code to minimize the duplication

docker/mongodb-kubernetes-tests/tests/authentication/sharded_cluster_scram_sha_256_switch_project.py
```diff
@@ -0,0 +1,163 @@
...
+        Ensure Ops Manager state is correctly updated in the original cluster.
+        """
+        tester = sharded_cluster.get_automation_config_tester()
+        tester.assert_authentication_mechanism_enabled("SCRAM-SHA-256")
+        tester.assert_authentication_enabled()
+
+    # def test_create_secret(self):
+    #     print(f"creating password for MongoDBUser {self.USER_NAME} in secret/{self.PASSWORD_SECRET_NAME} ")
+
+    #     create_or_update_secret(
```
lsierant, line 63: either remove or uncomment commented code; if necessary to leave it - explain why is commented

lsierant (changes requested):
Awesome you've added so many e2e tests! But let's try to think how we could minimize the code duplication in there. It looks like all the tests are almost identical.

controllers/operator/authentication/authentication.go
filipcirtog: done! thank you

controllers/operator/authentication/authentication.go
filipcirtog: no it is not a case in here. I have now changed that. thank you!

docker/mongodb-kubernetes-tests/tests/authentication/fixtures/switch-project/replica-set-scram-sha-1-switch-project.yaml
filipcirtog: sure! i have done that!

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_ldap_switch_project.py
filipcirtog: I have now utilized the existing fixtures from the auth module, with modifying only their names. would this be ok?

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
filipcirtog: done! thank you!

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
filipcirtog: it was indeed redundant! I have now changed that! thank you!

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
filipcirtog: sure! now everything uses try_load. thank you!

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
filipcirtog: done! thank you!

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
filipcirtog: I have deleted the redundant comments. thank you!

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
filipcirtog: same as above. deleted! than you

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
filipcirtog: i have inlined them. thanks

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_x509_switch_project.py
filipcirtog: I've created `ReplicaSetCreationAndProjectSwitchTestHelper` and `ShardedClusterCreationAndProjectSwitchTestHelper`, to enable code reuse across the tests. Thank you!

docker/mongodb-kubernetes-tests/tests/authentication/sharded_cluster_scram_sha_256_switch_project.py
filipcirtog: The commented tests are theoretically expected to work in scenarios where moving deployments across projects is supported. For instance, in the case of LDAP, the cluster does reach a running state but the automation config ends up with an empty users array. 
If more investigation will be done in this topic, I believe the code will be useful for reproducing the issues

controllers/om/automation_config.go
```diff
@@ -1,19 +1,29 @@
...
+	"github.com/mongodb/mongodb-kubernetes/mongodb-community-operator/pkg/kube/secret"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/generate"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/maputil"
 )
 
+// The constants for the authentication secret
+const (
+	autoPwdSecretKey = "automation-agent-password"
+)
```
lucian-tosa: This could be moved to our `constants.go` file

controllers/om/automation_config.go
```diff
@@ -426,19 +436,72 @@ func (ac *AutomationConfig) EnsureKeyFileContents() error {
...
+	if err == nil {
+		if val, ok := data[autoPwdSecretKey]; ok && len(val) > 0 {
+			password = val
+		}
+	} else if secret.SecretNotExist(err) {
+		if ac.Auth.AutoPwd != "" && ac.Auth.AutoPwd != util.InvalidAutomationAgentPassword {
+			password = ac.Auth.AutoPwd
+		}
+
+		err := EnsureEmptySecret(ctx, k8sClient, secretNamespacedName)
```
lucian-tosa: Why create an empty secret then update it? It can be done in one go when the password variable is set

controllers/operator/authentication/authentication.go
```diff
@@ -82,7 +86,7 @@ type UserOptions struct {
 
 // Configure will configure all the specified authentication Mechanisms. We need to ensure we wait for
 // the agents to reach ready state after each operation as prematurely updating the automation config can cause the agents to get stuck.
-func Configure(conn om.Connection, opts Options, isRecovering bool, log *zap.SugaredLogger) error {
+func Configure(ctx context.Context, client kubernetesClient.Client, mdbNamespacedName types.NamespacedName, conn om.Connection, opts Options, isRecovering bool, log *zap.SugaredLogger) error {
```
lucian-tosa: nit: the `Options` struct could be reused to include the name of the resource

docker/mongodb-kubernetes-tests/tests/authentication/helper_sharded_cluster_switch_project.py
```diff
@@ -0,0 +1,62 @@
...
+            data={
+                "baseUrl": original_configmap["baseUrl"],
+                "projectName": new_project_name,
+                "orgId": original_configmap["orgId"],
+            },
+        )
+
+        self.sharded_cluster["spec"]["opsManager"]["configMapRef"]["name"] = new_project_configmap
+        self.sharded_cluster.update()
+        self.sharded_cluster.assert_reaches_phase(Phase.Running, timeout=800)
```
lucian-tosa: nit: the replicaset and sharded helpers have a few methods in common such as this one

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
```diff
@@ -0,0 +1,114 @@
...
+    return ReplicaSetCreationAndProjectSwitchTestHelper(
+        replica_set=replica_set,
+        namespace=namespace,
+        authentication_mechanism="MONGODB-CR",
+        expected_num_deployment_auth_mechanisms=2,
+    )
+
+
+@pytest.mark.e2e_replica_set_scram_sha_1_switch_project
+class TestReplicaSetCreationAndProjectSwitch(KubernetesTester):
```
lucian-tosa, line 43: as we talked on the call, please add a method to assert that the automation agent password in the AC has not changed after migrating projects

docker/mongodb-kubernetes-tests/tests/authentication/sharded_cluster_scram_sha_256_switch_project.py
```diff
@@ -0,0 +1,121 @@
...
+        self, test_helper: ShardedClusterCreationAndProjectSwitchTestHelper
+    ):
+        test_helper.test_sharded_cluster_connectivity(1)
+
+    def test_ops_manager_state_correctly_updated_after_switch(
+        self, test_helper: ShardedClusterCreationAndProjectSwitchTestHelper
+    ):
+        test_helper.test_ops_manager_state_with_expected_authentication(expected_users=1)
+
+    def test_ops_manager_state_with_users_correctly_updated_after_switch(
```
lucian-tosa: Make sure that this is not flaky. If you suspect it is, then disable it, and add a comment

.evergreen.yml
```diff
@@ -713,6 +713,14 @@ task_groups:
...
       - e2e_sharded_cluster_scram_x509_ic_manual_certs
       - e2e_sharded_cluster_external_access
+      - e2e_sharded_cluster_scram_sha_256_switch_project
+      - e2e_sharded_cluster_scram_sha_1_switch_project
+      - e2e_sharded_cluster_x509_switch_project
+      - e2e_replica_set_scram_sha_256_switch_project
+      - e2e_replica_set_scram_sha_1_switch_project
+      - e2e_replica_set_x509_switch_project
+      - e2e_replica_set_ldap_switch_project
+      - e2e_sharded_cluster_ldap_switch_project
```
lucian-tosa: Since these don't use the password secret, better to disable them. For now, they only test whether project migrations work, but we don't fully support that yet.

controllers/om/automation_config.go
filipcirtog: thanks! added

controllers/om/automation_config.go
filipcirtog: Thanks! I had that logic before because I was adding both the key and the password in the same secret. Now, with just the pwd, this works better. I've changed it.

controllers/operator/authentication/authentication.go
filipcirtog: thanks! now the code is cleaner and makes much more sense.

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
filipcirtog: thanks! i have added this.

.evergreen.yml
filipcirtog: thanks! i have commented those and also added a comment for reasoning.

docker/mongodb-kubernetes-tests/tests/authentication/sharded_cluster_scram_sha_256_switch_project.py
filipcirtog: thanks. i have disabled that parts and add a comment for that

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_1_switch_project.py
filipcirtog: thanks! i have implemented this suggestions.

docker/mongodb-kubernetes-tests/tests/authentication/fixtures/switch-project/replica-set-scram-sha-1-switch-project.yaml
filipcirtog: thanks. i have refactored the comments.

controllers/operator/authentication/ldap.go
```diff
@@ -15,7 +18,7 @@ func (l *ldapAuthMechanism) GetName() MechanismName {
 	return LDAPPlain
 }
 
-func (l *ldapAuthMechanism) EnableAgentAuthentication(conn om.Connection, opts Options, log *zap.SugaredLogger) error {
+func (l *ldapAuthMechanism) EnableAgentAuthentication(ctx context.Context, client kubernetesClient.Client, conn om.Connection, opts Options, log *zap.SugaredLogger) error {
```
lucian-tosa: nit: since `client` is not used, you can mark it with `_`

lucian-tosa (approved):

controllers/operator/authentication/ldap.go
filipcirtog: done. thanks!!

docker/mongodb-kubernetes-tests/tests/authentication/sharded_cluster_scram_sha_256_switch_project.py
Julien-Ben: It's good to keep them then, just add a TODO to uncomment them when we support that, and link a Jira ticket related to that task (moving deployments across projects).

changelog/20251124_fix_resolved_automation_agent_password_secret_deadlock.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2025-11-24
+---
+
+* Resolved automation agent password secret deadlock issue in sharded clusters during project migration
```
vinilage: Since this goes to the release notes and it is a customer facing text, I would rephrase it:

"Fixed an issue that caused automation agent password secrets to block project migration in sharded clusters."

filipcirtog: Thank you. On second thought, should we even mention project migrations since they aren't currently supported? Including them might be misleading and there are cases in which we still get some errors.

filipcirtog: "Backed up the agent password in a secret for SCRAM authentication to prevent unnecessary password rotations."
WDYT?

vinilage: Good from my side Filip!!!

changelog/20251127_fix_backed_up_the_agent_password_in_a_secret_for_scram.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2025-11-27
+---
+
+* Backed up the agent password in a secret for SCRAM authentication to prevent unnecessary password rotations.
```
vinilage, line 6: LGTM!

docker/mongodb-kubernetes-tests/tests/authentication/helper_sharded_cluster_switch_project.py
```diff
@@ -0,0 +1,71 @@
+from kubetester import (
+    create_or_update_configmap,
+    read_configmap,
+)
+from kubetester.mongodb import MongoDB
+from kubetester.mongotester import ShardedClusterTester
+from kubetester.phase import Phase
+
+
+class ShardedClusterSwitchProjectHelper:
```
lsierant: nit: How different is the implementation between ShardedClusterSwitchProjectHelper and ReplicaSetSwitchProjectHelper? Looks identical at first glance

lsierant (approved):

docker/mongodb-kubernetes-tests/tests/authentication/helper_sharded_cluster_switch_project.py
filipcirtog: Thanks. They are very similar. However, one uses `ReplicaSetTester` and the other `ShardedClusterTester` but for this we can create two methods depending on the resource. I will merge them as it would be cleaner.

filipcirtog: done. thank you!

PR #527, opened 2025-10-13
CLOUDP-353160 [Search] Implement gRPC and mTLS

fealebenpae:
# Summary
This updates the `mongod`->`mongot` communication to use the gRPC protocol as the Wire Protocol server is deprecated in `mongot`.

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

PR #568, opened 2025-11-03
CLOUDP-356376 Allow Search to work with x509 cluster auth

fealebenpae:
# Summary
With the switch to gRPC for mongot in #527 it's now possible for Search to be deployed against clusters that use x509 internal authentication. This removes the validation check for the internal cluster authentication mode and adds a new e2e test based on the existing `search_enterprise_tls` test that verifies existing functionality against a x509 cluster. 

## Proof of Work
New test.

## Checklist


docker/mongodb-kubernetes-tests/tests/search/search_enterprise_x509_cluster_auth.py
```diff
@@ -0,0 +1,261 @@
...
+        return resource
+
+    resource.configure(om=get_ops_manager(namespace), project_name=MDB_RESOURCE_NAME)
+    resource.configure_custom_tls(issuer_ca_configmap, "certs")
+    resource["spec"]["security"]["authentication"] = {
+        "enabled": True,
+        "modes": ["X509", "SCRAM"],
+        "agents": {"mode": "X509"},
+        "internalCluster": "X509",
+    }
```
fealebenpae, line 57: This is the only addition to the code already in the `e2e_search_enterprise_tls` test.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_x509_cluster_auth.py
```diff
@@ -0,0 +1,261 @@
...
+    ops_manager.update()
+    ops_manager.om_status().assert_reaches_phase(Phase.Running, timeout=1200)
+    ops_manager.appdb_status().assert_reaches_phase(Phase.Running, timeout=600)
+
+
+@mark.e2e_search_enterprise_x509_cluster_auth
+def test_install_tls_secrets_and_configmaps(namespace: str, mdb: MongoDB, mdbs: MongoDBSearch, issuer: str):
+    create_agent_tls_certs(issuer, namespace, mdb.name, "certs")
+    create_x509_mongodb_tls_certs(issuer, namespace, mdb.name, f"certs-{mdb.name}-clusterfile")
+    create_x509_mongodb_tls_certs(issuer, namespace, mdb.name, f"certs-{mdb.name}-cert", mdb.get_members())
```
fealebenpae, line 145: The other change, necessary to populate the secrets expected by the MongoDB CR change above.

changelog/20251103_feature_mongodbsearch_mongodb_deployments_using_x509.md
```diff
@@ -0,0 +1,7 @@
+---
+kind: feature
+date: 2025-11-03
+---
+
+* **MongoDBSearch**: MongoDB deployments using X509 internal cluster authentication are now supported. Previously MongoDB Search required SCRAM authentication among members of a MongoDB replica set.
```
lsierant: I think it's worth mentioning that mongot itself still cannot authenticate using x509, so SCRAM auth must be enabled for clients to connect (but it might be both x509, but SCRAM is mandatory for search)

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_x509_cluster_auth.py
```diff
@@ -0,0 +1,261 @@
...
+
+    search_service_name = f"{mdbs.name}-search-svc"
+    create_tls_certs(
+        issuer,
+        namespace,
+        f"{mdbs.name}-search",
+        replicas=1,
+        service_name=search_service_name,
+        additional_domains=[f"{search_service_name}.{namespace}.svc.cluster.local"],
```
lsierant, line 154: Is this service FQDN actually validated by mongod when connecting to mongot? It should, right? 

Also when we switch to L7 our proxy component should probably present the same cert, or at least cert issued for the same service FQDN?

lsierant (approved):

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_x509_cluster_auth.py
fealebenpae: I'm actually not a 100% if mongod performs hostname validation. I'll comment this out and see what happens.

re. L7 proxies - the proxy will terminate the TLS connection from mongod, and establish its own TLS connection to the mongots it represents. It can use the same TLS certificate as mongot, or its own (though it too will have to be signed by the same CA). We'll sort this out separately with load balancing, however, and adjust as necessary. This is just a copy of the existing TLS test we have.

changelog/20251103_feature_mongodbsearch_mongodb_deployments_using_x509.md
```diff
@@ -0,0 +1,7 @@
+---
+kind: feature
+date: 2025-11-03
+---
+
+* **MongoDBSearch**: MongoDB deployments using X509 internal cluster authentication are now supported. Previously MongoDB Search required SCRAM authentication among members of a MongoDB replica set. Note: SCRAM client authentication is still required, this change merely relaxes the requirements on internal cluster authentication.
+
```
vinilage, line 7: LGTM!

vinilage (approved):
LGTM!

controllers/searchcontroller/enterprise_search_source_test.go
```diff
@@ -215,35 +213,6 @@ func TestEnterpriseResourceSearchSource_Validate(t *testing.T) {
 			authModes:    []string{"SCRAM", "SCRAM-SHA-1", "SCRAM-SHA-256"},
 			expectError:  false,
 		},
-		// Internal cluster authentication tests
-		{
-			name:                "X509 internal cluster auth not supported",
```
MaciejKaras, line 220: shouldn't we keep this test case and just verify that we don't throw an error? As I understand there are three options for `internalClusterAuth` right now:
 - SCRAM
 - X5909
 - empty

Shouldn't we validate if received `internalClusterAuth` that is outside of the accepted values?

fealebenpae: The existing e2e tests cover the SCRAM internal auth cases, and the new e2e test covers x509 so I thought since the Validate method no longer concerns itself with internal cluster auth at all those unit test cases are no longer relevant, but I agree - it costs nothing to have them still.

fealebenpae:
@MaciejKaras exactly - not supporting x509 internal cluster authentication was a mongot limitation that was lifted with the change in #527, so this PR just removes our validation check for x509 auth and adds a test.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_x509_cluster_auth.py
fealebenpae: @lsierant it does seem that when mongod connects to mongot the hostname is validated against mongot's TLS certificate. Commenting out the `additional_domains` specification here causes the `createSearchIndex` command to fail with a `Error connecting to Search Index Management service` message.

PR #578, opened 2025-11-07
CLOUDP-321076: expose prometheus metrics settings in MongoDBSearch

lsierant:
# Summary

Expose prometheus settings in `MongoDBSearch` spec in a new `spec.prometheus` field. 

* By default, if `spec.prometheus` field is not provided then metrics endpoint in mongot is disabled. **This is a breaking change**. Previously the metrics endpoing was always enabled on port 9946.
* To enable prometheus metrics endpoint specify empty `spec.prometheus:` field. It will enable metrics endpoint on a default port (9946). To change the port, set it in `spec.prometheus.port` field.

## Proof of Work

Tested in unit and e2e tests.

For e2e prometheus endpoint checks were hooked into existing `e2e_search_enterprise_tls`.


api/v1/search/mongodbsearch_types.go
```diff
@@ -30,6 +30,14 @@ func init() {
 	v1.SchemeBuilder.Register(&MongoDBSearch{}, &MongoDBSearchList{})
 }
 
+type Prometheus struct {
+	// Port where metrics endpoint will be exposed on. Defaults to 9216.
```
fealebenpae:
```suggestion
	// Port where metrics endpoint will be exposed on. Defaults to 9946.
```

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -385,10 +387,18 @@ func createMongotConfig(search *searchv1.MongoDBSearch, db SearchSourceDBResourc
 				},
 			}
 		}
-		config.Metrics = mongot.ConfigMetrics{
-			Enabled: true,
-			Address: fmt.Sprintf("0.0.0.0:%d", search.GetMongotMetricsPort()),
+
+		if prometheus := search.GetPrometheus(); prometheus != nil {
+			port := search.GetMongotMetricsPort()
```
fealebenpae: Perhaps we should remove `GetMongotMetricsPort()` from the `MongoDBSearch` struct and implement it on the `Prometheus` struct so this will read something like:
```go
if prometheus := search.GetPrometheus(); prometheus != nil {
    port := prometheus.GetPort()
    config.Metrics = mongot.ConfigMetrics{...}
}
``` 

Same thing in `buildSearchHeadlessService()` above.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
```diff
@@ -247,3 +285,42 @@ def get_user_sample_movies_helper(mdb):
...
+        assert (
+            ports["prometheus"] == expected_port
+        ), f"Prometheus port should be {expected_port} but was {ports['prometheus']}"
+    else:
+        assert "prometheus" not in ports, "Prometheus port should not be in service when prometheus is disabled"
+
+
+def assert_search_pod_prometheus_endpoint(mdbs: MongoDBSearch, should_be_accessible: bool, port: int = 9946):
+    pod_name = f"{mdbs.name}-search-0"
+    url = f"http://localhost:{port}/metrics"
```
fealebenpae: If we hit `http://{mdbs.name}-search-svc:{port}` directly we could omit explicitly asserting properties of the Service object in `assert_search_service_prometheus_port()` altogether, I think. I don't know if it's worth it to maintain the distinction between "the port was configured on the Service" and "prometheus was enabled in mongot configuration" since the former is already tested in the search reconcile helper unit test.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
lsierant: excellent suggestion! Thanks!

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_tls.py
lsierant: Agree 💯 I've reworked it a bit. I'm deploying a tools/bastion pod (community image like in snippets) to connect to the probe. Might seem like an overkill for now, but I wanted to take that opportunity and test this approach in practice as we're planning to get rid of the e2e pod entirely and we'll need some sort of proxy for the connectivity checks soon anyway.

controllers/operator/appdbreplicaset_controller_test.go
```diff
@@ -54,15 +54,14 @@ func init() {
 
 // getReleaseJsonPath searches for a specified target directory by traversing the directory tree backwards from the current working directory
 func getReleaseJsonPath() (string, error) {
-	repositoryRootDirName := "mongodb-kubernetes"
```
lsierant, line 57: this failed tests when running from a different dir, e.g. a git worktree

controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -205,22 +202,16 @@ func TestMongoDBSearchReconcile_Success(t *testing.T) {
 			mdbc := newMongoDBCommunity("mdb", mock.TestNamespace)
 			reconciler, c := newSearchReconciler(mdbc, search)
 
-			res, err := reconciler.Reconcile(
-				ctx,
-				reconcile.Request{NamespacedName: types.NamespacedName{Name: search.Name, Namespace: search.Namespace}},
-			)
-			expected, _ := workflow.OK().ReconcileResult()
```
lsierant, line 210: thanks to the fix [here](https://github.com/mongodb/mongodb-kubernetes/pull/578/files#diff-9b1183581cc6af6723f6445d7a14ae4ae7b435dc3b93ca5d82247cda13f7ca39R79) it started to fail.

In order to reconcile successfully it must be always done by two steps: first reconcile is Pending (sts not ready), second time we simulate sts readiness and then it's Running. This is what checkSearchReconcileSuccessful is doing here.

changelog/20251030_feature_update_mongodb_search_2nd_preview.md
lsierant, line 1: @vinilage I've aggregated all the search-related change logs for this release under one point as those were scattered randomly across release notes. We might think about adding some additional grouping per CRD for example to make it automatic.

lsierant: The new entry for this PR is the one starting from "Exposed configuration settings..."

api/v1/search/mongodbsearch_types.go
```diff
@@ -30,6 +30,23 @@ func init() {
...
+	// Port where metrics endpoint will be exposed on. Defaults to 9946.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65535
+	Port int `json:"port,omitempty"`
+}
+
+func (p *Prometheus) GetPort() int32 {
+	if p.Port == 0 {
+		return MongotDefaultPrometheusPort
```
s-urbaniak: q: why not defaulting on the CRD level?

s-urbaniak: The current implementation will persist 0 but yield 9946 internally which is confusing to the user.

anandsyncs: I will make the CRD default 9946 but keep the runtime fallback.

s-urbaniak: @anandsyncs rather than leaving the runtime fallback the best practice is rather setting a `// +kubebuilder:default={}` on the Prometheus top-level property spec. This will make the default explicit and will appear even if the user does not specify the `prometheus` property in the first place.

In this case a
```
apiVersion: mongodb.com/v1
kind: MongoDBSearch
metadata:
  name: search-example
spec: {}
```
will yield
```
k get mongodbsearch search-example -oyaml
apiVersion: mongodb.com/v1
kind: MongoDBSearch
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"mongodb.com/v1","kind":"MongoDBSearch","metadata":{"annotations":{},"name":"search-example","namespace":"default"},"spec":{}}
  creationTimestamp: "2025-11-10T16:58:57Z"
  generation: 1
  name: search-example
  namespace: default
  resourceVersion: "889785"
  uid: de7dd459-4ab0-4345-83a0-6077207f4798
spec:
  prometheus:
    port: 9946
```

s-urbaniak: If you want to ensure runtime fallbacks they should be done in the webhook, not in the operator, as the operator does not mutate the spec in the first place.

s-urbaniak: and if you for some reason insist on keeping the "runtime" fallback please document that
```
spec:
  prometheus:
    port: 0
```
will be interpreted as 9946 and add a rationale why this implicit interpretation is necessary. (i can't think of a reason so leaving to you)

anandsyncs: Thanks for the feedback :) 

Actually, neither fallback nor the Prometheus default is necessary here, because we want to interpret nil Prometheus value as disabled.

s-urbaniak: ah, that changes things :-) in this case indeed neither default nor runtime fallback are necessary.

PR #580, opened 2025-11-10
Flaky test fixes

lucian-tosa:
# Summary

This PR aims to reduce the flakiness of the following tests:

### e2e_multi_cluster_sharded_snippets
Increased the timeout of `test_running`, since in failing [runs](https://spruce.mongodb.com/task/mongodb_kubernetes_e2e_multi_cluster_kind_e2e_multi_cluster_sharded_snippets_7272a430f4150a5fa67753ae6dc4bcce3ba293e8_25_11_10_08_47_17/files?execution=0&sorts=STATUS%3AASC), by the time the diagnostics are collected, the resources become ready.

### e2e_multi_cluster_appdb_upgrade_downgrade_v1_27_to_mck
Increased the timeout of `test_scale_appdb`. Similarly, the assertion on appdb status fails, but by the time diagnostics are collected, the resource becomes ready.

### e2e_appdb_tls_operator_upgrade_v1_32_to_mck
 In this test we have a race condition.
```
om-appdb-upgrade-tls   1          7.0.18    Running              Pending         Disabled         17m
om-appdb-upgrade-tls   1          7.0.18    Running              Running         Disabled         17m
om-appdb-upgrade-tls   1          7.0.18    Pending              Running         Disabled         17m
om-appdb-upgrade-tls   1          7.0.18    Pending              Running         Disabled         18m
om-appdb-upgrade-tls   1          7.0.18    Pending              Running         Disabled         18m
om-appdb-upgrade-tls   1          7.0.18    Running              Running         Disabled         19m
```
There is a moment during the operator upgrade where the resource has the status of AppDB and OM set to running. This happens very briefly before the operator starts reconciling OM and sets the OM status to Pending. In that moment, the test will very quickly pass both assertions and move on to assert healthiness by connecting to OM. This will fail since OM was not actually ready.
```
Reaching phase Running for resource AppDbStatus took 216.2561867237091s
Reaching phase Running for resource OmStatus took 0.0025169849395751953s
```
To fix this, I added a `persist_for` flag in our assertion methods. This makes sure that the phase we are currently asserting is reached and persists for a number of retries.

## Proof of Work

Retried the above tests a few times, and all pass
https://spruce.mongodb.com/version/6911c25146ed0e00077796e3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

## Checklist

### DIFF
--- a/docker/mongodb-kubernetes-tests/kubetester/mongodb_common.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/mongodb_common.py
@@ -12,6 +12,19 @@ TRACER = trace.get_tracer("evergreen-agent")
 class MongoDBCommon:
     @TRACER.start_as_current_span("wait_for")
     def wait_for(self, fn, timeout=None, should_raise=True, persist_for=1):
+        """
+        Waits for the given function `fn` to return True, retrying until the timeout is reached.
+        If persist_for > 1, the function must return True for that many consecutive checks.
+        Optionally raises an exception if the condition is not met within the timeout.
+
+        Args:
+            fn: A callable that returns a boolean.
+            timeout: Maximum time to wait in seconds (default: 600).
+            should_raise: If True, raises an Exception on timeout (default: True).
+            persist_for: Number of consecutive successful checks required (default: 1).
+        Returns:
+            True if the condition is met within the timeout, otherwise raises Exception if `should_raise` is True.
+        """
         if timeout is None:
             timeout = 600
         initial_timeout = timeout


docker/mongodb-kubernetes-tests/kubetester/mongodb_common.py
```diff
@@ -11,20 +11,26 @@
 
 class MongoDBCommon:
     @TRACER.start_as_current_span("wait_for")
-    def wait_for(self, fn, timeout=None, should_raise=True):
+    def wait_for(self, fn, timeout=None, should_raise=True, persist_for=1):
```
m1kola: Can we separate "waiting for something" from "something staying in the same state for a period of time" please? It will keep the testing framework simple and understanding of tests easier: no need to dig for the meaning of `persist_for` argument.

Here is an example of how it is done in Gomega:
* https://onsi.github.io/gomega/#eventually - our `wait_for`
* https://onsi.github.io/gomega/#consistently - what you are trying to incorporate into `wait_for`

m1kola (changes requested):
> There is a moment during the operator upgrade where the resource has the status of AppDB and OM set to running. This happens very briefly before the operator starts reconciling OM and sets the OM status to Pending.

Great observation! I think we should address this root cause instead of hacking around and making tests pass.

It will likely take more time so I suggest to separate this work from other flake fixes (where we rightfully increase timeouts).

lucian-tosa:
> I think we should address this root cause instead of hacking around and making tests pass.

@m1kola Do you think it is worth investing the time to look into the code for setting different status values? Is it something critical for customers? Otherwise, I would opt for fixing the test, which is a cheap but important fix for our CI stability

docker/mongodb-kubernetes-tests/kubetester/mongodb_common.py
lsierant, line 14: is persist_for essentially equivalent to "required consecutive successes to pass"?

lsierant: if yes, then pls leave a brief comment or rename it, it took me a bit thinking to get the meaning of this param

lsierant: and one of the reason was that when looking for "persist for" suggest that the "problem persists" or the goal is not achieve yet and the situation persists. When something succeeds it's not often described as a situation that "persists". But I might be nitpicking here.

lucian-tosa: > is persist_for essentially equivalent to "required consecutive successes to pass"?

Yes

Not the best name, I know, I couldn't find a better one. I can add a comment. But do you have a better option? `no_of_successful_passes`?

m1kola: While I'm not a fan of Gomega and Ginkgo I think we should separate "wait for something" from "consistently passes meets something" like they do:
* Wait for a condition - https://onsi.github.io/gomega/#eventually
* Ensure consistent pass of a condition - https://onsi.github.io/gomega/#consistently (edited: corrected the link)

lucian-tosa: As we discussed, I created a new method `assert_persist_phase`, and opened a ticket to track the root issue
please take another look

Julien-Ben: Thanks ! I like that we have the two high level methods `assert_reaches_phase` and `assert_persist_phase` separated now.

Julien-Ben:
I'm in flavor of merging this because we have too many flaky tests at the moment and this is annoying. I agree that the proper thing to do is to fix the root cause, but it's unsure when we will have time dedicated to that.

docker/mongodb-kubernetes-tests/kubetester/mongodb_common.py
m1kola: I also like how we now have `assert_reaches_phase` and `assert_persist_phase`, but still find `wait_for` to be a bit confusing. It's difficult to get what it does just by looking at the signature.

I'm not going to block on this, but if you can think of a way so we can leave `wait_for` as it was before and introduce another low level function with a simple signature - that would be awesome.

anandsyncs (approved):

docker/mongodb-kubernetes-tests/kubetester/mongodb_common.py
lucian-tosa: I added a comment in that function to explain the signature. Creating a new low level function would mean some  duplicated code

PR #588, opened 2025-11-12
Run mongorestore in a tools pod

lsierant:
# Summary

This PR moves execution of mongorestore into the tools pod which removes problems with the hostnames and connectivity when executed locally.

## Proof of Work

TBD


fealebenpae (approved):
This looks very good! I especially like the streaming tar solution.

I think the python imports can be optimized, though - we import `get_tools_pod` several  times in the same file, even though there's also a global import. The `mongodb_tools_pod` import looks unused too.

PR #602, opened 2025-11-17
Separate MongoDB Multi-Cluster tests and encapsulate shared functionalities for reuse

filipcirtog:
# Summary

**Overview:**
In order to replicate E2E test coverage from MongoDBMulti to MongoDB within a multicluster topology, we are transitioning to library-based approach. This PR marks the first step in this transition by focusing on the implementation of MongoDBMultiCluster using the shared library pattern. Future implementations for the MongoDB resource will adopt this same methodology.

**File structure:**
`../multicluster/shared/example-test.py` → Contains the core test logic shared across implementations.
`../multicluster/mongodbmulticluster/mongodbmulticluster-example-test.py` → Includes test fixtures and pytest calls to invoke methods from the shared library.

**Exceptions to the Library-Based Approach:**
Some tests do not utilize the shared library due to the increased complexity it would introduce. Unlike the multicluster folder, which includes 40 resource-specific tests leveraging the shared library approach, these tests are unique within their respective folders:
`multicluster_om/multicluster_mongodbmulticluster_om_appdb_no_mesh.py`
`multicluster_appdb/multicluster_mongodbmulticluster_appdb_s3_based_backup_restore.py`

## Proof of Work
**Shared library:**
- The shared library contains the actual test logic and can be reused. This library does not include fixtures, as fixtures are resource-specific (MongoDB vs MongoDBMulti).

**Test implementation**
- For each test present in the shared library, a pytest is implemented in the MongoDBMultiCluster folder.

**Naming convention for distinction:**
- Fixtures and evergreen tests related to MongoDBMultiCluster are prefixed/named mongodbmulticluster for differentiation.

## Checklist


.evergreen-tasks.yml
```diff
@@ -1174,18 +1174,23 @@ tasks:
...
+      - func: e2e_test
+
+  - name: e2e_mongodbmulticluster_multi_cluster_pvc_resize
     tags: [ "patch-run" ]
     commands:
       - func: e2e_test
 
   # this test is run, with an operator with race enabled
-  - name: e2e_om_reconcile_race_with_telemetry
+  - name: e2e_mongodbmulticluster_om_reconcile_race_with_telemetry
```
m1kola: I think we should only modify tests that test `MongoDBMultiCluster` CRD. This one seems to be only about Ops Manager. I don't think we should touch this test.

Or am I missing something?

filipcirtog: There are some tests which use MongoDBMulti in a different way. For example, they aren't centered on testing MongoDBMulti  but rather as a supporting component. These tests use MongoDBMulti as a resource to test functionalities such as backup and restore workflows, Ops Manager integration, or multi-cluster configurations without focusing on MongoDBMulti itself. I’ve attached such tests below (description using AI) :

`multicluster/multi_cluster_reconcile_races.py `- This test validates the MongoDB Kubernetes Operator's ability to reconcile resources reliably across multi-cluster environments while detecting data races and ensuring telemetry correctness.

`multicluster_om/multicluster_om_appdb_no_mesh.py` - This test validates the Multi-Cluster MongoDB Kubernetes Operator when deployed without Service Mesh, focusing on its ability to manage MongoDBMulti resources with TLS security, S3 backups, external app database monitoring, external connectivity, data restoration via restores, and telemetry validation.

`multicluster_appdb/multicluster_appdb_s3_based_backup_restore.py` - This test validates the MongoDB Kubernetes Operator in a multi-cluster environment with S3-backed backups and restores, ensuring Ops Manager is properly configured with cluster-level TLS, external connectivity, and custom DNS settings while verifying metadata stores, restore functionality, and backup consistency for MongoDBMulti resources.

While leaving them alone and just replacing MongoDBMultiCluster with MongoDB once the feature branch is merged is one possibility, I would consider replicating them as well in order to be on the safe side. As a result, it’s safer to leave these tests untouched (aside from changes such as renaming and library inclusion for multicluster/multi_cluster_reconcile_races.py) while creating new, identical tests for MongoDB. This approach also makes it easier to delete them once the resource is deprecated.

Additionally, `multicluster_om/multicluster_om_appdb_no_mesh.py` and `multicluster_appdb/multicluster_appdb_s3_based_backup_restore.py` will not adopt the library-based approach, as these tests are unique within their respective folders (being the only tests utilizing MongoDBMulti) and will simply be replicated as is.

m1kola: * `multicluster/multi_cluster_reconcile_races.py` - For this test I think it makes sense to add `MongoDB`-based multi-cluster RS as an additional resource into that test (in the feature branch) and not convert it into using a library.

* `multicluster_om/multicluster_om_appdb_no_mesh.py` and `multicluster_appdb/multicluster_appdb_s3_based_backup_restore.py` - I agree that it is better to be safe and have a respective tests for MongoDB-based multi-cluster RS. As an option we can still have shared folders (one in `multicluster_om` and one in `multicluster_appdb`) and add a new file on the same level as existing test for MongoDB-based multi-cluster RS. But I'm fine with just copying them.

docker/mongodb-kubernetes-tests/tests/multicluster/shared/multi_cluster_validation.py
```diff
@@ -1,44 +1,42 @@
...
-    def test_unique_cluster_names(self, central_cluster_client: kubernetes.client.ApiClient):
-        resource = yaml.safe_load(open(yaml_fixture("mongodb-multi-cluster.yaml")))
+    def test_unique_cluster_names(central_cluster_client: kubernetes.client.ApiClient, fixture: str):
+        resource = yaml.safe_load(open(yaml_fixture(fixture)))
         resource["spec"]["clusterSpecList"].append({"clusterName": "kind-e2e-cluster-1", "members": 1})
 
-        self.create_custom_resource_from_object(
-            self.get_namespace(),
+        KubernetesTester.create_custom_resource_from_object(
+            KubernetesTester.get_namespace(),
```
m1kola: Just curious - why replace `self` with `KubernetesTester`? Also - would `super()` be better/worse here?

docker/mongodb-kubernetes-tests/tests/multicluster/mongodbmulticluster/mongodbmulticluster_multi_cluster_validation.py
```diff
@@ -0,0 +1,25 @@
...
+from kubetester.kubetester import fixture as yaml_fixture
+from kubetester.operator import Operator
+
+from ..shared import multi_cluster_validation as testhelper
+
+MDBM_RESOURCE = "mongodbmulticluster-multi-cluster.yaml"
+
+
+@pytest.mark.e2e_mongodbmulticluster_multi_cluster_validation
+class TestWebhookValidation(KubernetesTester):
```
m1kola, line 11: Wondering if something like this make it less boilerplate for class-based tests?

```@pytest.mark.e2e_mongodbmulticluster_multi_cluster_validation
class TestWebhookValidation(testhelper.TestWebhookValidation):
    pass
```

Or even maybe (as a more generic for both class and function based tests)

```
TestWebhookValidation = pytest.mark.e2e_mongodbmulticluster_multi_cluster_validation(testhelper.TestWebhookValidation)
```

Or does it not work because of fixtures?

docker/mongodb-kubernetes-tests/tests/authentication/mongodbmulticluster_custom_roles.py
```diff
@@ -234,7 +234,7 @@ def mc_replica_set(
     return resource
 
 
-@mark.e2e_mongodb_custom_roles
+@mark.e2e_mongodbmulticluster_custom_roles
```
m1kola, line 175: I think we will also need an equivalent test for `MongoDB`-based multi-cluster RS. So would it make sense to also make it shared?

docker/mongodb-kubernetes-tests/tests/multicluster/shared/multi_cluster_validation.py
filipcirtog: thanks for highlighting this. i have reverted such changes now.

docker/mongodb-kubernetes-tests/tests/multicluster/mongodbmulticluster/mongodbmulticluster_multi_cluster_validation.py
filipcirtog: i would say this does not work since there is no way to inject the other fixtures/arguments

docker/mongodb-kubernetes-tests/tests/authentication/mongodbmulticluster_custom_roles.py
filipcirtog: i have made a shared library for this as well. thank you!

anandsyncs (approved):
Thanks for explaining the crux on the call.

I am fine with these changes considering that the shared test approach will be revisited when MongoDBMultiCluster is removed.

lsierant (approved):
The overall approach LGTM, but due to sheer volume it wasn't able to verify all the diffs. 

Let's ensure we have the same number of e2e tests running and their running time is similar to the "previous" runs. This way we'll ensure we're running the same tests after extracting them to the shared library code.

PR #604, opened 2025-11-19
Add the kubectl plugin build task to the gke snippets dependencies

fealebenpae:
# Summary
The `download_multi_cluster_binary` function expends to find the multicluster binary on S3 using the current patch id. Adding a dependency on the `build_kubectl_mongodb_plugin` task to the multicluster code snippets tests ensures there is a multicluster binary at the expected path for the current patch.

## Proof of Work
[The multi cluster code snippet test tasks](https://spruce.mongodb.com/version/691b2082d0bea90007785020/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) don't fail (at least not with a missing `kubectl-mongodb` error).

## Checklist


.evergreen-snippets.yml
```diff
@@ -143,6 +143,9 @@ buildvariants:
     allowed_requesters: [ "patch" ]
     run_on:
       - ubuntu2404-small
+    depends_on:
```
MaciejKaras: This makes sense for private snippets, but for public we should already have the `kubectl_mongodb_plugin` build and pushed to correct S3 path.

MaciejKaras (changes requested):

.evergreen-snippets.yml
m1kola: @MaciejKaras I've noticed that our snippets job tries to pull dev S3 instead of staging, if we run the patch as per release guide:

```
evergreen patch -p mongodb-kubernetes -v public_gke_code_snippets,public_kind_code_snippets -t all -d "Release 1.6.1: public gke snippets" -u -f -y --browse
```

m1kola (changes requested):
Let's put a hold on this for now. I think we are just trying to pull the binary from the wrong place.

.evergreen-snippets.yml
MaciejKaras: Well in fact it should try to pull the binary from the release bucket instead. We could either run the command with additional param `--param BUILD_SCENARIO=release` or make sure we set proper build_scenario in the `public_gke_code_snippets` variant somehow. WDYT @viveksinghggits ?

Also I'm really close to making code snippets part of automated release workflow, so we won't need to fix it soon.

viveksinghggits: I was also in favor of making sure proper `build_scenario` gets set in the `public_gke_code_snippets`. That would also make things more clear that we are using build scenario in the script/variant.

MaciejKaras:
https://github.com/mongodb/mongodb-kubernetes/pull/604#discussion_r2626214467

@m1kola I see that this is already fixed in the latest [Release Guide](https://docs.google.com/document/d/1QWpYu-YOrxwLgyHhSKQCAf6byJp2M6cq7C1FBxr3rbE/edit?tab=t.6hxkfucr4bko) template:
```
​​evergreen patch -p mongodb-kubernetes -v public_gke_code_snippets,public_kind_code_snippets -t all -d "Release x.y.z: public gke snippets" -u -f -y --browse --param BUILD_SCENARIO=release --param OPERATOR_VERSION=x.y.z
```

For the `private_gke_code_snippets` it makes sense to add `build_kubectl_mongodb_plugin` dependency so the GKE tests wait for the kubectl plugin to be present in S3. For the `public_gke_code_snippets` it does not make sense, because those are run manually. Later on they will be also automated and happen after `image promotion`, right @josvazg?

If everybody agrees (together with @fealebenpae and @viveksinghggits ) I can update the PR and hopefully we can merge it.

josvazg:
> [#604 (comment)](https://github.com/mongodb/mongodb-kubernetes/pull/604#discussion_r2626214467)
> 
> @m1kola I see that this is already fixed in the latest [Release Guide](https://docs.google.com/document/d/1QWpYu-YOrxwLgyHhSKQCAf6byJp2M6cq7C1FBxr3rbE/edit?tab=t.6hxkfucr4bko) template:
> 
> ```
> ​​evergreen patch -p mongodb-kubernetes -v public_gke_code_snippets,public_kind_code_snippets -t all -d "Release x.y.z: public gke snippets" -u -f -y --browse --param BUILD_SCENARIO=release --param OPERATOR_VERSION=x.y.z
> ```
> 
> For the `private_gke_code_snippets` it makes sense to add `build_kubectl_mongodb_plugin` dependency so the GKE tests wait for the kubectl plugin to be present in S3. For the `public_gke_code_snippets` it does not make sense, because those are run manually. Later on they will be also automated and happen after `image promotion`, right @josvazg?
> 
> If everybody agrees (together with @fealebenpae and @viveksinghggits ) I can update the PR and hopefully we can merge it.

I feel I lack context here, but I noticed something might be wrong. I recap the invariants:

1. At PR time, running all tests is optional (if only evergreen made our lives easy there)
1. At **merge time all tests must run**, without exception. This is the only way to verify the commit and images produced are releseable.
1. At **release time nothing is rebuilt nor retested**: The tested assets are just published.

So this sentence sounds wrong to me: 
> ... so the GKE tests wait for the kubectl plugin to be present in S3. ... Later on they will be also automated and happen after image promotion, right @josvazg?

We cannot tests things fir the first time after a release, we need to validate. them before we release.

Other than that I am not sure what those tests cover and what is the difference between public and private. In any case, I assume they should or may gate the release when failing, so they should happen before release.

MaciejKaras:
@josvazg your intuition is correct that all tests should be run before the release, not after. `public_gke_code_snippets` are specific tests that are used to produce output files that are then in our public documentation, not to test that all things work. They are not regular tests _per se_.

josvazg:
> @josvazg your intuition is correct that all tests should be run before the release, not after. `public_gke_code_snippets` are specific tests that are used to produce output files that are then in our public documentation, not to test that all things work. They are not regular tests _per se_.

Why are they called "tests" then? Why not instead of call them, "publish snippet samples"?

Tests are things that can fail, if the failure case is covered by the private tests, and these are just publishing something that is already tested, we should rename them. The idea is that if these fail it should be the publishing that failed, we are not actually running those snippets.

MaciejKaras:
> > @josvazg your intuition is correct that all tests should be run before the release, not after. `public_gke_code_snippets` are specific tests that are used to produce output files that are then in our public documentation, not to test that all things work. They are not regular tests _per se_.
> 
> Why are they called "tests" then? Why not instead of call them, "publish snippet samples"?
> 
> Tests are things that can fail, if the failure case is covered by the private tests, and these are just publishing something that is already tested, we should rename them. The idea is that if these fail it should be the publishing that failed, we are not actually running those snippets.

So it's more complicated than that. I'll try to provide more context. `public_gke_code_snippets` are e2e tests that run on the GKE cluster(s). They use input files from e.g. `public/architectures/ra-06-ops-manager-multi-cluster/code_snippets` and during the tests we also capture the outputs that are later published in Public Docs e.g. https://www.mongodb.com/docs/kubernetes/current/reference-architectures/multi-cluster/deploy-operator/#deploy-the-kubernetes-operator-1. For those outputs to have correct image versions and container urls, **they have to be run against quay.io production images**. 

I'm not sure how can we overcome this problem, maybe there is a way, but this should be a separate discussion from this PR. Here we only try to solve dependency issue for `kubectl plugin`.

josvazg:
> > > @josvazg your intuition is correct that all tests should be run before the release, not after. `public_gke_code_snippets` are specific tests that are used to produce output files that are then in our public documentation, not to test that all things work. They are not regular tests _per se_.
> > 
> > 
> > Why are they called "tests" then? Why not instead of call them, "publish snippet samples"?
> > Tests are things that can fail, if the failure case is covered by the private tests, and these are just publishing something that is already tested, we should rename them. The idea is that if these fail it should be the publishing that failed, we are not actually running those snippets.
> 
> So it's more complicated than that. I'll try to provide more context. `public_gke_code_snippets` are e2e tests that run on the GKE cluster(s). They use input files from e.g. `public/architectures/ra-06-ops-manager-multi-cluster/code_snippets` and during the tests we also capture the outputs that are later published in Public Docs e.g. https://www.mongodb.com/docs/kubernetes/current/reference-architectures/multi-cluster/deploy-operator/#deploy-the-kubernetes-operator-1. For those outputs to have correct image versions and container urls, **they have to be run against quay.io production images**.
> 
> I'm not sure how can we overcome this problem, maybe there is a way, but this should be a separate discussion from this PR. Here we only try to solve dependency issue for `kubectl plugin`.

If that is the case, I think I cannot provide much input to this discussion. My goal with the new release process is to keep the invariants I mention, we will discuss these public tests separately.

.evergreen-snippets.yml
MaciejKaras: I've removed the `build_kubectl_mongodb_plugin` from the `public_gke_code_snippets` since the Release Guide was updated to include the `build_scenario`

MaciejKaras (approved):

m1kola (approved):

viveksinghggits (approved):

PR #622, opened 2025-12-02
[CLOUDP-361632] Add information about latest OM and agent in release.json

viveksinghggits:
# Summary

We have been getting some request from the customers where they wanted to know the combination of the container images that they require to run a specific version of operator and Mongo DB instance. As part of that work we are going to add a new file (release_info.json) as the assets in the GitHub release. 
That file is going to have details about the latest released images. These images are going to be the images that MCK publishes as well as the latest OM and Agent version mappings. To enable that, as part of this PR we are adding a new field in the `release.json` that has the mapping between OM version and agent version of latest OM releases. This new field in the `release.json` is going to have the opsManaget to agent image version mapping for every major opsmanager version.
 We are adding this information in the `release.json` so that it can easily be used later (in next PR), to create the release_info.json file.
This PR also updated the precommit hook so that when new version of OM is released, the newly added field of release.json would properly be updated with newly release OM/agent.

## Proof of Work

Manual: 
- Mimic new release of OM by adding a new version (8.0.17) to `supportedImages.ops-manager.versions` and respective entry in `mongodb-agent.opsManagerMapping`. Run `make precommit` and verify that the `supportedImages.latestOpsManagerAgentMapping` has been updated for major version 8 with updated versions.
- Mimic the new major version release of OM by adding new version (9.0.1) to  `supportedImages.ops-manager.versions` and respective entry in `mongodb-agent.opsManagerMapping`. Run `make precommit` and verify that a new entry for major version 9 has been added to `supportedImages.latestOpsManagerAgentMapping`.

## Checklist

- [ ] Have you added changelog file? (will show up later)

### DIFF
--- a/scripts/evergreen/release/update_release.py
+++ b/scripts/evergreen/release/update_release.py
@@ -31,8 +31,8 @@ def update_release_json():
     newest_operator_version = data["mongodbOperator"]
     update_operator_related_versions(data, newest_operator_version)
 
-    if not update_latest_om_agent_mapping(data, newest_om_version):
-        logger.info("Failed updating release.json for latest om-agent mapping")
+    # Adds mapping between latest major version of OM and agent to the release.json
+    update_latest_om_agent_mapping(data, newest_om_version)
 
     with open(release, "w") as f:
         json.dump(


scripts/evergreen/release/update_release.py
```diff
@@ -31,6 +31,9 @@ def update_release_json():
     newest_operator_version = data["mongodbOperator"]
     update_operator_related_versions(data, newest_operator_version)
 
+    if not update_latest_om_agent_mapping(data, newest_om_version):
```
MaciejKaras: `update_latest_om_agent_mapping` doesn't need to return anything and in case of issues just raise an Exception with proper message. 

Currently the errors are printed using `debug` level which is inadequate for errors.

release.json
```diff
@@ -17,6 +17,26 @@
     "version": "0.55.0"
   },
   "supportedImages": {
+    "latestOpsManagerAgentMapping": [
```
MaciejKaras: I wouldn't put thins in the `supportedImages` node. It is not a new type of image.

Maybe better place would be in the `supportedImages.ops-manager`, `supportedImages.mongodb-agent` or on the root level?

viveksinghggits: I like the idea to move it at the root lever better, I will do that.

release.json
```diff
@@ -16,6 +16,26 @@
   "search": {
     "version": "0.55.0"
   },
+  "latestOpsManagerAgentMapping": [
```
nammn, line 19: do you think we can replace the existing code here to use what you've implemented here? https://github.com/mongodb/mongodb-kubernetes/blob/f0050b8942545701e8cb9e42d54d14f0cb58ee6a/pkg/agentVersionManagement/get_agent_version.go#L57-L77

nammn (approved):

PR #624, opened 2025-12-03
[CLOUDP-361632] Add functionality to upload release info to GitHub release assets

viveksinghggits:
# Summary

As discussed in the [base PR](https://github.com/mongodb/mongodb-kubernetes/pull/622), customers have been asking us to provide some details that can be used to figure out the container images that they need to deploy a specific version of MCK and MongoDB instance.
This PR does that. It add the functionality to add a new file called `release_info_<release-version>.json` to the GitHub release assets that would contain the details about MCK released images, OM, Agent, Search and the mapping between OM and agent's latest versions.

This is how the file loos like

https://gist.github.com/viveksinghggits/560e6179a3c7d9b11b0df68f02b15ccc

## Proof of Work

Run the patch manually 

```
evergreen patch --path .evergreen.yml -p mongodb-kubernetes -v add_releaseinfo_to_github_assets -t add_releaseinfo_to_github_assets -f -y -u --browse -d "Test add release info to github assets"
```
and make sure the new file is added to the release.

https://github.com/user-attachments/assets/5e22d7fa-edaf-4d46-98f3-cce431c49f38

https://github.com/user-attachments/assets/b1a6d0a2-965e-4135-a15f-5d2550c773cf

## Checklist

- [ ] Have you added changelog file? (will show up later)

### DIFF
--- a/scripts/release/release_info.py
+++ b/scripts/release/release_info.py
@@ -111,6 +111,7 @@ def convert_to_release_info_json(build_info: BuildInfo, release_json_path: str, 
         latest_search_version(release_data),
     )
 
+    # add enterprise server image info
     add_image_info(
         release_info_output,
         MONGODB_ENTERPRISE_SERVER_IMAGE,


.evergreen-release.yml
```diff
@@ -151,6 +157,18 @@ buildvariants:
...
 
+  - name: add_releaseinfo_to_github_assets
+    display_name: add_releaseinfo_to_github_assets
+    tags: ["release"]
+    run_on:
+      - ubuntu2404-small
+    allowed_requesters: ["patch", "github_tag"]
+    # depends_on:
+    #   - name: "*"
+    #     variant: release_images
```
viveksinghggits: this will be removed.

scripts/release/release_info.py
```diff
@@ -1,90 +1,149 @@
...
+    )
+
+    # add upgrade hook image info
+    upgradehook_build_info = build_info.images[UPGRADE_HOOK_IMAGE]
+    add_image_info(
+        release_info_output,
+        UPGRADE_HOOK_IMAGE,
+        upgradehook_build_info.repositories,
+        upgradehook_build_info.platforms,
+        latest_upgrade_hook_version(release_data),
```
MaciejKaras, line 92: I see `upgradehook` but `readiness probe` is still missing. 

This should show up in the tests as well, you might need to update it.

.evergreen-release.yml
```diff
@@ -151,6 +157,18 @@ buildvariants:
       - name: release_readiness_probe
       - name: release_version_upgrade_hook
 
+  - name: add_releaseinfo_to_github_assets
+    display_name: add_releaseinfo_to_github_assets
+    tags: ["release"]
+    run_on:
+      - ubuntu2404-small
+    allowed_requesters: ["patch", "github_tag"]
+    # depends_on:
```
MaciejKaras: why this is still commented out?

scripts/release/release_info.py
```diff
@@ -1,90 +1,149 @@
...
+        latest_search_version(release_data),
+    )
+
+    release_info_output = add_om_agent_mappings(release_data, release_info_output)
+
+    return release_info_output
+
+
+def add_image_info(release_info_output, name, repositories, platforms, version):
+    release_info_output["images"][name] = {"repositories": repositories, "platforms": platforms, "version": version}
```
MaciejKaras: we have discussed in the thread possibility to join `repository` and `version` into single `url` field. Is this going to be handled in a separate PR?

.evergreen-release.yml
viveksinghggits: I will remove this, I have commented this out because I don't want to mistakenly release the images while testing this.

scripts/release/release_info.py
viveksinghggits: I have added readiness probe image as well, I didn't notice that it's version is also mentioned in the release.json.

Which tests are you talking about here?

scripts/release/release_info.py
viveksinghggits: I have not implemented that change yet, because there are quite a few discussion happening there. Having two files one json and one raw, having just one raw file etc.
I am planning to change the format eventually in this PR itself once we reach to a conclusion in that thread. Do you think that's ok?

scripts/release/release_info.py
```diff
@@ -1,90 +1,186 @@
...
+    add_image_info(
+        release_info_output,
+        SEARCH_IMAGE,
+        SEARCH_IMAGE_REPOSITORY,
+        ["linux/arm64", "linux/amd64"],
+        latest_search_version(release_data),
+    )
+
+    release_info_output = add_om_agent_mappings(release_data, release_info_output)
+
```
viveksinghggits, line 124: ToDo:
Add mongodb-enterprise-server image here. I am doing it right now.

scripts/release/release_info.py
viveksinghggits: It looks like we are better off without the `release_info_test.py`. And the reason is, if we want to test the `release_info.py`, the test file is not just going to test logic/functions, it will have actual logic (that the file that we are testing has), to figure things out. 
For example, it (the test file) will have to figure a lot of versions out from `release.json` and most importantly it will have to figure out the digest of the images and then use that in json that we are going to test against. 
So, in my opinion we should get rid of release_info_test.py, what do you think?

scripts/release/release_info.py
MaciejKaras: That's ok, but maybe in that case move the PR to draft? Otherwise reviewers, myself included, receive notifications about updates in the PR.

.evergreen-release.yml
MaciejKaras: Sure, understand. Let's keep the thread unresolved until this is uncommented.

scripts/release/release_info.py
viveksinghggits: After we got the confirmation on the format from Joe and we agreed on the format @MaciejKaras I made the change and it's already addressed now.

scripts/release/release_info.py
MaciejKaras: I would still love to have test for this logic, because it is not only about retrieving the correct values from `release.json`, but also about proper output file structures, list of artifacts that are going to be published, getting sha etc.

What I would recommend is to have a separate `test_release.json` file in `testdata` and use it instead. This way we can hardcode specific values in the tests and do not worry about logic for retrieving them in the test. 

Other option is to pass dictionary of `image_name -> version` to `create_release_info_json()` or `convert_to_release_info_json()` function and move the version retrieval logic outside of the this function. I think this is less preferable, because you won't test version retrieval from `test_release.json`.

scripts/release/build/image_build_process.py
```diff
@@ -109,6 +117,25 @@ def check_if_image_exists(self, image_tag: str) -> bool:
...
+    def pull_image(self, image):
+        docker_cmd = python_on_whales.docker
+        try:
+            docker_cmd.image.pull(image, quiet=True)
+        except Exception as e:
+            logger.error(f"Failed pulling image. {e}")
+            raise e
+
+    def get_manfiest_list_digest(self, image) -> Optional[str]:
+        self.pull_image(image)
```
MaciejKaras: Pulling Ops Manager (which is huge) and other images just to get the digest seems like an overkill. Are we sure there is no other way?

scripts/release/build/image_build_process.py
```diff
@@ -109,6 +117,25 @@ def check_if_image_exists(self, image_tag: str) -> bool:
...
+
+        docker_cmd = python_on_whales.docker
+        try:
+            manifest = docker_cmd.image.inspect(image)
+            return manifest.id
+        except Exception as e:
+            logger.error(f"Failed inspecting the image to get manifest list digest. {e}")
+            raise e
```
MaciejKaras:
```suggestion
        except Exception as e:
            raise Exception("Failed inspecting the image to get manifest list digest") from e
```

This is idiomatic way to chain exceptions in Python -> https://docs.python.org/3/tutorial/errors.html#exception-chaining. I also see you logging the error immediately when caught, but this is not needed if we re-throw it or chain with new exception. The exception will be logged by the program either way.

scripts/release/kubectl_mongodb/utils.py
```diff
@@ -34,3 +40,47 @@ def kubectl_plugin_name(os_name: str, arch_name: str) -> str:
...
+def upload_assets_to_github_release(asset_paths: list[str], release_version: str):
+    if not GITHUB_TOKEN:
+        logger.info("ERROR: GITHUB_TOKEN environment variable not set.")
+        sys.exit(1)
+
+    try:
+        g = Github(GITHUB_TOKEN)
+        repo = g.get_repo(GITHUB_REPO)
+    except GithubException as e:
+        logger.info(f"ERROR: Could not connect to GitHub or find repository '{GITHUB_REPO}', Error {e}.")
```
MaciejKaras: Same principle here, sys.exit(1) is not needed if you re-throw the error. The syntax is shorter and function can be reused in other scenarios - not necessarily causing program exit.

MaciejKaras: Please make the changes in similar places as well.

scripts/release/build/image_build_process.py
viveksinghggits: I looked into these ways of figuring out manifest list digest. 

### docker_cmd.buildx.imagetools.inspect(image_tag)

The python client that we are using doesn't return the manifest list digest, even though the respective CLI returns the required digest. So we can not use this way. I raised [an issue](https://github.com/gabrieldemarmiesse/python-on-whales/issues/698) upstream. 

### docker manifest inspect

Does not return the manifest list digest. 

### docker_cmd.image.inspect(image)

Works well and expects image to be pulled first.

### docker inspect

Works well and expects image to be pulled locally.

viveksinghggits: The overall process of making the release_info file is taking less than 10 (~6) seconds.

viveksinghggits: The other option is to call HTTP API of container registry to figure the information out.

viveksinghggits: Now that I was thinking about it again, the other option is to not use the python module and just run `docker buildx imagetools inspect` command directly. The small problem with this approach is, in the other functions we will be using the python package but in one method we will be running `docker` command directly.

scripts/release/release_info.py
viveksinghggits: > What I would recommend is to have a separate test_release.json file in testdata and use it instead. This way we can hardcode specific values in the tests and do not worry about logic for retrieving them in the test.

The problem with this approach is, we will have to update this test file as well along with `release.json`, when new versions of OM/Agent are released.
And we will have to somehow also mention the digest of the container images in this test file.

If we just want to make sure that all the container images are part of the release_info.json, should we just try to test that. As in, should we create the release_info and just verify that all the images are present in that file, without verifying their version and sha?

MaciejKaras: >The problem with this approach is, we will have to update this test file as well along with release.json, when new versions of OM/Agent are released.

Actually we won't need to update `test_release.json` ever, this is the main benefit of having separate file, to test only the logic, not the actual contents of the `release.json` file. We don't care about new versions released and I don't understand why we would need to update `test_release.json` when this happens. Can you explain why you think so?

scripts/release/build/image_build_process.py
MaciejKaras: Why not use https://github.com/containers/skopeo to inspect remote images? It's the ideal tool for this case.

viveksinghggits: I have used skopeo now to get the manifest list digest.

scripts/release/tests/release_info_test.py
```diff
@@ -1,65 +1,150 @@
...
+            "ops-manager": {
+                "repoURL": "quay.io/mongodb/mongodb-enterprise-ops-manager-ubi",
+                "platforms": [
+                    "linux/amd64"
+                ],
+                "tag": "8.0.16",
+                "digest": "sha256:ca4aad523f14d68fccb60256f9ce8909c66ebb5b321ee15e5abf9ac5738947f9"
+            },
+            "agent": {
+                "repoURL": "quay.io/mongodb/mongodb-agent",
```
nammn, line 51: you've changed repositories from a slice to a string. 
The agents are released - as of right now - to 2 different repositories, how would we reflect that now?

scripts/release/release_info.py
```diff
@@ -1,90 +1,205 @@
...
-            "version": DUMMY_VERSION,
-        }
+        add_image_info(release_info_output, name, image.repositories[0], image.platforms, operator_version)
+
+    # add OPS manager image info
+    om_build_info = build_info.images[OPS_MANAGER_IMAGE]
+    add_image_info(
+        release_info_output,
+        OPS_MANAGER_IMAGE,
+        om_build_info.repositories[0],
```
MaciejKaras: b: We should change the `build_info.json` instead and instead of having array of repositories there, rather something like this:
```json
"release": {
  "sign": true,
  "olm-tag": true,
  "skip-if-exists": true,
  "repository": "quay.io/mongodb/mongodb-agent",
  "secondary-repositories": ["quay.io/mongodb/mongodb-agent-ubi"],
  "platforms": [
    "linux/arm64",
    "linux/amd64",
    "linux/s390x",
    "linux/ppc64le"
  ]
}
```

cc @nammn

scripts/release/tests/release_info_test.py
MaciejKaras: Good question! The idea is to share information only about main repository. The other repository we push agent to `mongodb-agent-ubi` is used for backward compatibility issue. Related comment -> https://github.com/mongodb/mongodb-kubernetes/pull/624/files#r2597674694

scripts/release/release_info.py
viveksinghggits: Hi @MaciejKaras,
Even though I agree with the change, I don't think this change should be part of the current PR. If you agree, I have raised this CLOUDP ticket https://jira.mongodb.org/browse/CLOUDP-363995 and we can work on this separately.

If you still think the change should be part of this PR, I can make the change.

MaciejKaras: It doesn't have and even shouldn't be part of this change. I would rather see this is a prerequisite, separate change. I wouldn't want to have `xx_build_info.repositories[0]` and hardcoded `AGENT_IMAGE_REPOSITORY` values in master. My recommendation would be to create a separate PR that splits the `repositories` field and merge it first (or stack the current PR on top)

viveksinghggits: ok, so do you think we can merge this PR as it is and then can work on the CLOUDP ticket that I have raised, or do you think we should work on the CLOUDP ticket first and then the merge this PR after refactor?

viveksinghggits: I have raised this PR https://github.com/mongodb/mongodb-kubernetes/pull/633 that changes the format of `build_info.json` like discussed above.

scripts/release/build/image_build_process.py
```diff
@@ -167,6 +195,10 @@ def check_if_image_exists(self, image_tag: str) -> bool:
             f"PodmanImageBuilder does not support checking if image exists remotely. Skipping check for {image_tag}.")
         return False
 
+    def get_manfiest_list_digest(self, image) -> Optional[str]:
+        logger.warning(f"PodmanImageBuilder does not support getting digest for manifest list, returning empty digest.")
```
MaciejKaras: nit: I believe we won't run Podman builder to create `release-info.json` so can as well just raise Exception here

scripts/release/build/image_build_process.py
```diff
@@ -109,6 +115,28 @@ def check_if_image_exists(self, image_tag: str) -> bool:
...
+                capture_output=True,
+                text=True,
+                check=True
+            )
+            return result.stdout.strip()
+        except subprocess.CalledProcessError as e:
+            raise Exception(f"Failed to run skopeo inspect using 'docker run' for image {image}. Error: {e.stderr.strip()}") from e
+        except FileNotFoundError:
+            raise Exception("docker is not installed on the system.")
+        except Exception as e:
```
MaciejKaras: nit: if you are catching `Exception e` here and then immediately re-throwing it without adding more context, then you could just skip the entire `except Exception as e` clause.

scripts/release/kubectl_mongodb/promote_kubectl_plugin.py
```diff
@@ -65,7 +62,10 @@ def main():
 
     if os.environ.get("SKIP_GITHUB_RELEASE_UPLOAD", "false").lower() == "false":
         github_artifacts = artifacts_tar + [checksum_file]
-        upload_assets_to_github_release(github_artifacts, release_version)
+        try:
+            upload_assets_to_github_release(github_artifacts, release_version)
+        except Exception as e:
```
MaciejKaras: nit: similar comment to https://github.com/mongodb/mongodb-kubernetes/pull/624/files#r2601544444

scripts/release/release_info.py
```diff
@@ -1,90 +1,205 @@
...
+
+
+# manifest_list_digest_for_image returns manifest list digest for the passed image. Returns
+# empty string if there was an error figuring that out.
+def manifest_list_digest_for_image(image: str) -> str:
+    builder = DockerImageBuilder()
+    try:
+        digest = builder.get_manfiest_list_digest(image)
+    except Exception as e:
+        logger.error(f"There was an error, figuring out manifest list digest for image {image}. Error: {e}")
```
MaciejKaras: q: so with this implementation we are fine to produce `release-info.json` that is missing some of the digests for images?

viveksinghggits: For this release I think it's ok to not mention the digest for the image versions that are mentioned under OMAgentMapping. We can add that in the next release if required. Is that what you are talking about? I don't think we are missing digest in any other image.

MaciejKaras: Let me clarify. `manifest_list_digest_for_image` is used by all images and if for some reason retrieving the digest fails, it will produce `release-info.json` with empty `digest` field. Now I'm asking if we are we fine with this? 

Do we expect some images don't have the proper digest we can retrieve and the request will fail? You've mentioned something about `OMAgentMapping`, but I didn't get it.

viveksinghggits: ohh, I understand you now. That's what my plan was, that if we are not able to find out the digest for an image we can have empty digest in release-info.json. Do you think we should fail instead?

MaciejKaras: I don't know either. Can this actually happen with our current images? 

I would ask @vinilage and @mircea-cosbuc for guidance

viveksinghggits: > I don't know either. Can this actually happen with our current images?

It's highly unlikely but not impossible. One case that I can think of is, network issue and because of that `skopeo` is not able to connect to container registry.

changelog/20251209_other_added_a_new_asset_in_github_releases_that_would.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: other
+date: 2025-12-09
+---
+
+* Added the functionality to add a new asset named `release_info_<version>.json`, to further GitHub releases, that would contain detailed information about released MCK (MongoDB controllers for Kubernetes) and other relevant container images.
```
MaciejKaras: nit: small text improvement

```suggestion
*  Future GitHub releases will include new asset, `release_info_<version>.json`, with detailed information about the MCK (MongoDB controllers for Kubernetes) release and other related container images.
```

MaciejKaras (approved):
Nice work! 🤩

changelog/20251209_other_added_a_new_asset_in_github_releases_that_would.md
viveksinghggits: > Added new GitHub release asset

I think this doesn't exactly convey what we did right? We are not adding the asset just to this release but we are adding a feature because of which this new asset will be added to all the upcoming releases. 

If we just write `Added new GitHub release asset` it looks like we are adding asset to the current release.

MaciejKaras: You're right. Still users don't care about the mechanism that adds the new asset, but what will change in the GitHub Release assets. How about now?

viveksinghggits:
@nammn when you get some time, can you please have another look into this.

changelog/20251209_other_added_a_new_asset_in_github_releases_that_would.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: other
+date: 2025-12-09
+---
+
+* Future GitHub releases will include new asset, `release_info_<version>.json`, with detailed information about the MCK (MongoDB controllers for Kubernetes) release and other related container images.
```
vinilage: Proposal:

"* Future releases will include a new asset, `release_info_<version>.json`, which provides detailed information about each MongoDB Controllers for Kubernetes release, including a clear list of all container images. This will help customers, especially those running in air-gapped environments, easily identify all required images for a given release."

vinilage (changes requested):
I left one comment proposing an improvement for the release notes.

scripts/release/release_info.py
```diff
@@ -1,90 +1,194 @@
...
-
-    for name, chart in build_info.helm_charts.items():
-        output["helm-charts"][name] = {
-            "registry": chart.registry,
-            "repository": chart.repository,
-            "version": DUMMY_VERSION,
-        }
+        add_image_info(release_info_output, name, image.repository, image.platforms, operator_version)
+
+    # add OPS manager image info
```
nammn, line 65: lets keep the comment everywhere or remove it everywhere.

nammn (approved):

scripts/release/release_info.py
viveksinghggits: added the comment to enterprise-server image.

changelog/20251209_other_added_a_new_asset_in_github_releases_that_would.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: other
+date: 2025-12-09
+---
+
+* Future releases will include a new asset, release_info_<version>.json, which provides detailed information about each MongoDB Controllers for Kubernetes release, including a clear list of all container images. This will help customers, especially those running in air-gapped environments, easily identify all required images for a given release.
```
vinilage, line 6: LGTM!

PR #625, opened 2025-12-04
CLOUDP-355710 Fix deletion of external roles

lucian-tosa:
# Summary

This PR fixes an issue where roles added to an OM project outside the operator (through UI or the API) are overwritten by the operator every reconciliation. This is now consistent with the user behaviour, where users can be defined in the UI and will not be removed.

To accomplish this, we are now tracking the roles we configure with the operator in an annotation, and state configmap (where applicable). This was necessary due to possible usage of the `ClusterMongoDBRole` resource. If the reference to a role resource is removed at the same time as removing the resource itself, then it is not possible to determine which role needs to be removed from the automation config without knowing the previous set of roles.

## Proof of Work

Unit tests have been added to verify that the `ensureRoles` method which is reused across all controllers behaves correctly. Additionally, the unit tests verify that each controller keeps track of the previously configured roles in an annotation.

## Checklist

### DIFF
--- a/controllers/operator/mongodbshardedcluster_controller_multi_test.go
+++ b/controllers/operator/mongodbshardedcluster_controller_multi_test.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/ghodss/yaml"
-	"github.com/mongodb/mongodb-kubernetes/pkg/kube"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/yudai/gojsondiff"
@@ -38,6 +37,7 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/mongodb-community-operator/api/v1/common"
 	kubernetesClient "github.com/mongodb/mongodb-kubernetes/mongodb-community-operator/pkg/kube/client"
 	"github.com/mongodb/mongodb-kubernetes/mongodb-community-operator/pkg/kube/configmap"
+	"github.com/mongodb/mongodb-kubernetes/pkg/kube"
 	"github.com/mongodb/mongodb-kubernetes/pkg/multicluster"
 	"github.com/mongodb/mongodb-kubernetes/pkg/test"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util"


controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -954,6 +956,17 @@ func (r *ShardedClusterReconcileHelper) Reconcile(ctx context.Context, log *zap.
 			annotationsToAdd[k] = val
 		}
 	}
+
+	// Set annotation and state for previously configured roles
+	roleAnnotation, roleStrings, err := r.commonController.getRoleAnnotation(ctx, r.sc.Spec.DbCommonSpec, r.enableClusterMongoDBRoles, kube.ObjectKeyFromApiObject(r.sc))
```
lsierant: why do we rely on annotation where we have deployment state available? I understand this is necessary for mongodb and mdbmc as we don't have the state yet,  but here?

lsierant: OK, I see that we actually rely on the LastConfiguredRoles in the logic, so the annotation is to align with other controllers?

lucian-tosa: Yes. I can remove the annotation from the sharded cluster, I was unsure whether I should add it to both the state and the annotation

controllers/operator/common_controller_test.go
```diff
@@ -528,6 +528,56 @@ func TestCheckEmptyStringsInPrivilegesEquivalentToNotPassingFields(t *testing.T)
...
+	mockOm.AddRole(externalRole)
+
+	// Reconcile again - role created from the UI should still be there
+	roleStrings, _ := controller.getRoleStrings(ctx, rs.Spec.DbCommonSpec, true, kube.ObjectKeyFromApiObject(rs))
+	controller.ensureRoles(ctx, rs.Spec.DbCommonSpec, true, mockOm, kube.ObjectKeyFromApiObject(rs), roleStrings, zap.S())
+
+	roles = mockOm.GetRoles()
+	require.Len(t, roles, 2)
+
+	// Delete embedded role, only the external should remain
```
nammn, line 665: q: why did the embedded role get removed?

controllers/operator/common_controller.go
```diff
@@ -189,6 +230,33 @@ func normalizePrivilegeResource(resource mdbv1.Resource) mdbv1.Resource {
 	return resource
 }
 
+// mergeRoles merges the deployed roles with the current roles and previously configured roles.
+// It ensures that roles configured outside the operator are not removed.
+// This is achieved by removing currently configured roles from the deployed roles.
+// To ensure that roles removed from the spec are also removed from OM, we also remove the previously configured roles.
+// Finally, we add back the currently configured roles.
+func mergeRoles(deployed []mdbv1.MongoDBRole, current []mdbv1.MongoDBRole, previous []string) []mdbv1.MongoDBRole {
```
nammn, line 238: blocking: lets add a dedicated unit test for mergeRoles

controllers/operator/common_controller_test.go
lucian-tosa: re-initialized `rs` to a replicaset without roles on line 573

nammn: ahhhh "deleting" by re-initialisation. Can you update the comment to clarify that ?

controllers/operator/common_controller_test.go
```diff
@@ -528,6 +528,56 @@ func TestCheckEmptyStringsInPrivilegesEquivalentToNotPassingFields(t *testing.T)
...
+	externalRole := mdbv1.MongoDBRole{
+		Role: "external-role",
+		Db:   "admin",
+		Roles: []mdbv1.InheritedRole{{
+			Db:   "admin",
+			Role: "read",
+		}},
+	}
+	mockOm.AddRole(externalRole)
+
```
nammn, line 653: can you - for the sake of clarity - add here:

```
	roles = mockOm.GetRoles()
	require.Len(t, roles, 2)
``` 

as well?

controllers/operator/common_controller_test.go
lucian-tosa: no, you're right, I'll just set the roles field to an empty array to make it clear

changelog/20251204_fix_roles_configured_via_ops_manager_ui_or_api_will_no.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2025-12-04
+---
+
+* Roles configured via Ops Manager UI or API will no longer be removed by the operator
```
vinilage, line 6: LGTM!

controllers/operator/mongodbshardedcluster_controller.go
lucian-tosa: I removed the annotation

controllers/operator/common_controller_test.go
lucian-tosa: done

controllers/operator/common_controller.go
lucian-tosa: done

controllers/operator/common_controller_test.go
lucian-tosa: done

PR #628, opened 2025-12-05
[CLOUDP-349630] Don't generate scramshacreds if `MongoDBUser`'s password is not changed

viveksinghggits:
# Summary

When the reconciliation happens for the `MongoDBUser` resource (because of controller restart or because of any other reason), for `SCRAM` authentication mechanism, the scram sha creds are generated and then they are set in the automation config which results into another version getting created for Automation Config.
This workflow can be improved to generated the scram sha creds only when the password of the user is changed. If we do that we will not be changing the user's scram creds and the automation config will not be updated.
To achieve this we are checking if the scram sha creds that are in automation config for a user, match with the scram sha creds generated with the stored salt and new password. If the scram sha creds match it means that the password for the user is not changed and we don't have to generate the new scram sha creds, otherwise new scram sha creds will be generated.

## Proof of Work

Unit test:

```
~/work/opensource/mongodb-kubernetes/controllers/operator/authentication (dont-rotate-scram-creds-mdbu) » go test -timeout 30s -run ^Test_isPasswordChanged$ -v
=== RUN   Test_isPasswordChanged
--- PASS: Test_isPasswordChanged (0.01s)
PASS
ok      github.com/mongodb/mongodb-kubernetes/controllers/operator/authentication       0.472s
```

I am yet to do some more tests to actually be confident that the automation config is not changed when reconciliation happens.

## Checklist

### DIFF
--- a/controllers/operator/mongodbuser_controller_test.go
+++ b/controllers/operator/mongodbuser_controller_test.go
@@ -74,12 +74,12 @@ func TestUserIsAdded_ToAutomationConfig_OnSuccessfulReconciliation(t *testing.T)
 
 // Not making this (DeepCopy) a method of type AutomationConfig because I want to be
 // explicit that this method is just for test code.
-func DeepCopy(origianl *om.AutomationConfig) (*om.AutomationConfig, error) {
-	if origianl == nil {
+func DeepCopy(original *om.AutomationConfig) (*om.AutomationConfig, error) {
+	if original == nil {
 		return nil, nil
 	}
 
-	b, err := json.Marshal(origianl)
+	b, err := json.Marshal(original)
 	if err != nil {
 		return nil, err
 	}
@@ -113,7 +113,7 @@ func TestNoChange_InAC_After_Same_User_Reconciliation(t *testing.T) {
 	ac, _ := omConnectionFactory.GetConnection().ReadAutomationConfig()
 	// since underlying implmentation of ReadAutomationConfig just has reference to automation config
 	// it's better to deep copy this version of AC so that it can be used to compare later.
-	origianAC, err := DeepCopy(ac)
+	originalAC, err := DeepCopy(ac)
 	assert.Nil(t, err)
 
 	// the automation config should have been updated during reconciliation
@@ -132,7 +132,7 @@ func TestNoChange_InAC_After_Same_User_Reconciliation(t *testing.T) {
 
 	acAfterSecondReconcile, _ := omConnectionFactory.GetConnection().ReadAutomationConfig()
 	// verify that the automation cnofig has not been changed because we ran reconciliation second time with the same user
-	assert.True(t, acAfterSecondReconcile.EqualsWithoutDeployment(*origianAC), "Automation config before the second reconciliation and after the second reconciliation should be same")
+	assert.True(t, acAfterSecondReconcile.EqualsWithoutDeployment(*originalAC), "Automation config before the second reconciliation and after the second reconciliation should be same")
 }
 
 func TestReconciliationSucceed_OnAddingUser_FromADifferentNamespace(t *testing.T) {


controllers/operator/authentication/scramsha_credentials.go
```diff
@@ -28,10 +28,57 @@ const (
...
+		}
+		sha1Salt, err := base64.StdEncoding.DecodeString(acUser.ScramSha1Creds.Salt)
+		if err != nil {
+			return true, nil
+		}
+		// generate scramshacreds with (new) password but with old salt to verify if given password
+		// is actually changed
+		newScramSha256Creds, err := computeScramShaCreds(user.Username, password, sha256Salt, ScramSha256)
+		if err != nil {
+			return false, xerrors.Errorf("error generating scramSah256 creds to verify with already present user on automation config %w", err)
```
lsierant: typo in scramSah256 and below

controllers/operator/mongodbuser_controller.go
lucian-tosa, line 1: I would add here a unit test where we reconcile the same user twice and expect a single update to the AC

viveksinghggits: do you think the test should be added in the unit test as well as in one of the E2Es? Or just adding in unit test should be good.

lucian-tosa: IMO a unit test should be enough here since the bug was purely with the AC that the operator pushes to OM.

viveksinghggits: I just added a small unit test that tries to do that, can you please have a look again.

controllers/operator/mongodbuser_controller_test.go
```diff
@@ -71,6 +71,45 @@ func TestUserIsAdded_ToAutomationConfig_OnSuccessfulReconciliation(t *testing.T)
...
+	createPasswordSecret(ctx, client, user.Spec.PasswordSecretKeyRef, "password")
+
+	actual, err := reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: kube.ObjectKey(user.Namespace, user.Name)})
+
+	okReconcileResult, _ := workflow.OK().ReconcileResult()
+
+	assert.Nil(t, err, "there should be no error on successful reconciliation")
+	assert.Equal(t, okReconcileResult, actual, "there should be a successful reconciliation if the password is a valid reference")
+
+	ac, _ := omConnectionFactory.GetConnection().ReadAutomationConfig()
```
lucian-tosa, line 113: this is a pointer, so the test will always pass

viveksinghggits: sorry, I don't think I understood the issue correctly. 
Assuming that `ReadAutomationConfig()` works correctly and it returns the updated automation config after reconciliation, even though `ac` is a pointer we are passing `*ac` to `EqualsWithoutDeployment`, so this should do what's expected right?
And even if `ac` was passed as reference (it is not right now) to `EqualsWithoutDeployment`, I think `EqualsWithoutDeployment` would be able to handle this right?

lucian-tosa: The way `ReadAutomationConfig` works in our mocked OM is that it returns the pointer to a stored automation config. But since we are using the same mocked OM object (therefore the same underlying AC object), after the second reconcile, the underlying automation config will change, and since `ac` points to the same AC object, it points to the newest AC version.
If you debug the test, you will see that `ac` and `acAfterSecondReconcile` point to the same memory location. Meaning that EqualsWithoutDeployment will compare the second AC with itself. The solution is to make a deep copy of the AC at this point and compare it with the latest version at the end.

controllers/operator/mongodbuser_controller_test.go
```diff
@@ -71,6 +72,69 @@ func TestUserIsAdded_ToAutomationConfig_OnSuccessfulReconciliation(t *testing.T)
...
+	// since underlying implmentation of ReadAutomationConfig just has reference to automation config
+	// it's better to deep copy this version of AC so that it can be used to compare later.
+	origianAC, err := DeepCopy(ac)
```
lucian-tosa: nit: typo `originalAC`

controllers/operator/mongodbuser_controller.go
```diff
@@ -421,18 +421,18 @@ func (r *MongoDBUserReconciler) handleScramShaUser(ctx context.Context, user *us
 }
 
 func (r *MongoDBUserReconciler) handleExternalAuthUser(ctx context.Context, user *userv1.MongoDBUser, conn om.Connection, log *zap.SugaredLogger) (reconcile.Result, error) {
```
viveksinghggits, line 423: Hi @lucian-tosa,
Before I merge this, I would like to highlight that because we are changing signature of `toOmUser` it's also affecting `handleExternalAuthUser`. Even though we were mainly planning to handle the scram sha authentication as part of this change.
Do you think it's an acceptable change even for external users? If yes, I will go ahead and merge the PR.

lucian-tosa: fine with me, it's a noop
if you want to, you can add a new method `toScramOMUser` which adds the credentials on top of the `toOmUser` result
you can also pass a nil AC in `handleExternalAuthUser`

PR #648, opened 2025-12-16
CLOUDP-366183: Bump Ops Manager Container Image version to 7.0.21

mms-build-account:
_Opened by Private Cloud Tools (PCT)_.

# Ticket
[CLOUDP-366183](https://jira.mongodb.org/browse/CLOUDP-366183)

# Description
Bump Ops Manager container image version to 7.0.21.

# Reviewer Checklist

Before merging this PR, verify the following:
- [ ] the following tasks are passing in Evergreen:
  - `publish_ops_manager` task (variant: `publish_om70_images`)
- [ ] the `agent_version` was updated correctly
- [ ] the `tools_version` was updated correctly


release.json
```diff
@@ -217,6 +218,10 @@
             "agent_version": "108.0.16.8895-1",
             "tools_version": "100.13.0"
           },
+          "7.0.21": {
```
viveksinghggits, line 221: one thing that I noticed is, this mapping of OM version and agent/tools version should be sorted as we can see the the older version is already sorted. 
So ideally `7.0.21` should be before `8.0.1` and after `7.0.20` above in the list.

m1kola: Keys in objects are are unordered in JS. I think we've been manually ordering these before (I certainly did move `7.0.20` up when I was rebasing 7.0.20 bump PR).

But I think it is just for human convenience. Order should not affect anything else.

m1kola: Ha, interesting. I see you also created https://github.com/mongodb/mongodb-kubernetes/pull/668

viveksinghggits: > But I think it is just for human convenience. Order should not affect anything else.

Actually we do rely on the order of the elements at some places in our code, not for this exact field (`mongodb-agent.opsManagerMapping.ops_manager`) but for example `supportedImages.ops-manager.versions`. [This is](https://github.com/mongodb/mongodb-kubernetes/blob/master/scripts/evergreen/release/update_release.py#L26) one example.

m1kola: @viveksinghggits `supportedImages.ops-manager.versions` is different - it is a list and it is ok to rely on order there (assuming that we update the order correctly.

But in our scripts we should not rely on the ordering of keys in objects even if they are ordered in json. And, as far as I understand, it is what you fixed in https://github.com/mongodb/mongodb-kubernetes/pull/668 with sorting.

nammn:
fixed by https://github.com/mongodb/mongodb-kubernetes/pull/671

PR #737, opened 2026-01-30
CLOUDP-319138: Add hostname override for backup agents with external domain [Replica Sets]

Julien-Ben:
# Summary

When using `externalDomain`, backup agents were reporting internal cluster hostnames while automation and monitoring agents correctly reported external hostnames. This caused hostname mismatches in Ops Manager.

The fix passes the `-backupOverrideLocalHost` flag to the automation agent when:
1. An external domain is configured
2. No backup agents are already registered (to avoid breaking existing deployments)

## Scope

To keep the scope of this PR manageable and agree on the design, **this fix is applied to replica set only**. A follow up ticket and PR are required for multi cluster replica sets, and sharded clusters (single/multi).

## Why not apply to existing deployments?

Adding the `-backupOverrideLocalHost` flag to a deployment that already has backup agents registered would cause problems:
- **Ghost agents**: Ops Manager would show duplicate agents (old internal hostname + new external hostname)
- **Corrupted state**: The mismatch between registered agents and actual hostnames could break backup operations

A manual opt-in mechanism (e.g., a new field in the CRD) was considered as an alternative. Instead, we query the Ops Manager API to check if backup agents already exist. This makes the fix automatic for new deployments while protecting existing ones.

## What changed

- Added `ReadBackupAgents` method to query backup agents from Ops Manager API
- Refactored `ReadAutomationAgents` and `ReadBackupAgents` to share code via `readAgents`
- Controller sets `MDB_BACKUP_HOSTNAME_OVERRIDE=true` env var when conditions are met
- `agent-launcher.sh` passes `-backupOverrideLocalHost` flag when the env var is set
- Added `BackupHostnameOverrideEnv` constant for the env var name
- Optimized StatefulSet reads: added `getStatefulSetIfExists()` helper to fetch StatefulSet once per reconcile instead of twice (for `backupEnvVarAlreadyInSts` and `publishAutomationConfigFirstRS`)

### Agent Crash Risk
- The `-backupOverrideLocalHost` flag was added in agent v13.26.0 (shipped with OM 7.0)
- **Passing this flag to an older agent will crash the agent**
- Version gate in `isBackupHostnameOverrideSupported()` ensures we only enable for OM 7.0+ or Cloud Manager
- If this check is wrong or bypassed, users will experience agent crashes

### New OM API Call in Reconciliation
- `ReadBackupAgents(1)` is called during reconciliation when `externalDomain` is configured, this adds latency and a potential failure point to every reconcile

**Failure modes to consider:**
- If we incorrectly enable on existing deployment → ghost agents, corrupted backup state
- If we incorrectly disable on deployment that had it → env var removed, agents start reporting internal hostnames, mismatch with registered agents
- StatefulSet is the source of truth once enabled - we never disable if the env var exists

### Architecture Differences
- **Non-static**: Env var goes in `mongodb-enterprise-database` container (runs `agent-launcher.sh`)
- **Static**: Env var goes in `mongodb-agent` container (runs `agent-launcher-shim.sh` → `agent-launcher.sh`)
- `backupEnvVarAlreadyInSts()` checks the correct container based on architecture annotation

### StatefulSet Read Optimization
- StatefulSet is fetched once via `getStatefulSetIfExists()` and passed to both:
  - `backupEnvVarAlreadyInSts()`
  - `publishAutomationConfigFirstRS()`

## Proof of Work

New unit tests, e2e test: `test_backup_agents_use_external_hostnames` in `om_ops_manager_backup_tls_custom_ca.py`

## Checklist


changelog/20260130_fix_backup_agents_hostname_with_external_domain.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Fix backup agents hostname with external domain
+kind: fix
+date: 2026-01-30
+---
+
+* Fixed an issue where backup agents reported internal cluster hostnames instead of external hostnames when `externalDomain` was configured.
```
vinilage: LGTM!

vinilage (approved):

changelog/20260130_fix_backup_agents_hostname_with_external_domain.md
nammn: should we note what could happen if thats the case?

controllers/om/mockedomclient.go
```diff
@@ -490,6 +491,15 @@ func (oc *MockedOmConnection) ReadAutomationAgents(pageNum int) (Paginated, erro
 	return AutomationAgentStatusResponse{AutomationAgents: results}, nil
 }
 
+func (oc *MockedOmConnection) ReadBackupAgents(pageNum int) (Paginated, error) {
+	oc.addToHistory(reflect.ValueOf(oc.ReadBackupAgents))
+	if oc.ReadBackupAgentsFunc != nil {
+		return oc.ReadBackupAgentsFunc(pageNum)
+	}
+	// Default: return empty response (no backup agents)
```
nammn: lets remove that comment

controllers/operator/construct/database_construction.go
```diff
@@ -68,6 +68,9 @@ const (
 	LogFileMongoDBEnv                = "MDB_LOG_FILE_MONGODB"
 	LogFileAgentMonitoringEnv        = "MDB_LOG_FILE_MONITORING_AGENT"
 	LogFileAgentBackupEnv            = "MDB_LOG_FILE_BACKUP_AGENT"
+
+	// BackupHostnameOverrideEnv triggers the -backupOverrideLocalHost flag in agent-launcher.sh
```
nammn: lets describe it differently instead of "triggers". Something along the lines of "sets up the agent flag ..."

controllers/operator/common_controller.go
```diff
@@ -838,6 +838,17 @@ func isPrometheusSupported(conn om.Connection) bool {
 	return err == nil && omVersion.GTE(semver.MustParse("5.9.0"))
 }
 
+// isBackupHostnameOverrideSupported checks if the OM version supports the
+// -backupOverrideLocalHost agent flag (added in agent v13.26.0, available in OM 7.0+).
```
nammn: every om8 version supports it?

changelog/20260130_fix_backup_agents_hostname_with_external_domain.md
nammn: Additionally, shouldn't we note that this fix will cause a rolling restart if applied?

docker/mongodb-kubernetes-init-database/content/agent-launcher.sh
```diff
@@ -121,6 +121,10 @@ override_file="/opt/scripts/config/${hostpath}"
 if [[ -f "${override_file}" ]]; then
   override="$(cat "${override_file}")"
   agentOpts+=("-overrideLocalHost=${override}")
+  # The backup agent host must also be overridden, but only for new deployments (see CLOUDP-319138).
```
nammn, line 124: nit: i hate that we have to do that due to how the agent works

controllers/operator/mongodbreplicaset_controller_test.go
```diff
@@ -1561,9 +1561,221 @@ func TestPublishAutomationConfigFirstRS(t *testing.T) {
...
+				break
+			}
+		}
+		assert.True(t, foundEnvVar,
+			"reconcile %d: %s env var should be present", reconcileNum, construct.BackupHostnameOverrideEnv)
+	}
+
+	// First reconciliation: no backup agents, env var should be added
+	checkReconcileSuccessful(ctx, t, reconciler, rs, kubeClient)
+	verifyEnvVarPresent(1)
```
nammn: q: i don't what `reconcileNum` gives me in this test, is it visibility if it fails to know when it failed?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -616,11 +642,51 @@ func (r *ReplicaSetReconcilerHelper) buildStatefulSetOptions(ctx context.Context
...
+// If sts is nil, it means the StatefulSet doesn't exist yet.
+func (r *ReplicaSetReconcilerHelper) backupEnvVarAlreadyInSts(sts *appsv1.StatefulSet) bool {
+	if sts == nil {
+		return false
+	}
+
+	// Determine which container to check based on architecture.
+	// In non-static, agent-launcher.sh runs in the database container.
+	// In static, agent-launcher-shim.sh (which exec's agent-launcher.sh) runs in the agent container.
+	containerName := util.DatabaseContainerName
```
nammn, line 698: semi-nit: do/should we have a function that returns you the correctContainerName? I am sure we have this duplicated somewhere.

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -266,11 +267,40 @@ func (r *ReplicaSetReconcilerHelper) Reconcile(ctx context.Context) (reconcile.R
 			return r.updateStatus(ctx, workflow.Failed(xerrors.Errorf("failed to prepare Replica Set for scaling down using Ops Manager: %w", err)))
 		}
 	}
+
+	// Fetch StatefulSet once for use by multiple functions below
```
nammn, line 271: from the pr title i can see its only for rs, do we plan to add the other controllers?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -266,11 +267,40 @@ func (r *ReplicaSetReconcilerHelper) Reconcile(ctx context.Context) (reconcile.R
...
+	existingSts, err := r.getStatefulSetIfExists(ctx)
+	if err != nil {
+		return r.updateStatus(ctx, workflow.Failed(xerrors.Errorf("failed to fetch existing StatefulSet: %w", err)))
+	}
+
+	// Check if we override backup agent hostname
```
nammn: semi-nit: i feel its nicer if we fully extract the `enableBackupHostnameOverride` logic into a method + add a log so customers know that it was detected and what we will do

nammn (commented):
LGTM! Very clear changeset. Thanks! A few comments and q

controllers/operator/mongodbreplicaset_controller_test.go
Julien-Ben: Yeah exactly, just for logging. In my first implementation the fix would be applied on initial reconcile, and then deleted by the subsequent one.
So I used this to track when the test would fail before fixing that

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: I considered doing it, I didn't find an existing helper function
I think in the follow up PR of this one we should create it.

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: > To keep the scope of this PR manageable and agree on the design, this fix is applied to replica set only. A follow up ticket and PR are required for multi cluster replica sets, and sharded clusters (single/multi).

I mentioned in the description that we should follow up with other controllers. I prefered to agree on the approach first

changelog/20260130_fix_backup_agents_hostname_with_external_domain.md
Julien-Ben: Rolling restart won't happen since this flag should only be added on new deployments.

> should we note what could happen if thats the case?
I don't know if it wouldn't be too verbose. With this changelog only affected customers will understand that it is fixed, maybe that enough.

I don't really know best practices for changelogs.
Thoughts @vinilage ?

controllers/om/mockedomclient.go
Julien-Ben: Done

controllers/operator/construct/database_construction.go
Julien-Ben: Done

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: You are right, I extracted it and it's much nicer ! Improved logging too

controllers/operator/mongodbreplicaset_controller.go
nammn: got it! I think once we work on the other controllers we might need to refactor a bit to see common functions between them. But not blocking for me

controllers/operator/mongodbreplicaset_controller.go
nammn: sure, then good as follow-up (as long as we note this down) for me! Resolve this comment then

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -659,6 +644,46 @@ func (r *ReplicaSetReconcilerHelper) getStatefulSetIfExists(ctx context.Context)
...
+
+	// For new or updated deployments, only enable if no backup agents are registered yet.
+	// Enabling on existing deployments with registered agents causes hostname mismatches.
+	backupAgents, err := conn.ReadBackupAgents(1)
+	if err != nil {
+		return false, xerrors.Errorf("failed to check registered backup agents in Ops Manager: %w", err)
+	}
+
+	if backupAgents.ItemsCount() == 0 {
+		log.Infof("Enabling backup hostname override: external domain is configured and no backup agents are registered yet")
```
nammn, line 680: We might want to give the info that this will cause a rolling restart now

controllers/operator/mongodbreplicaset_controller.go
nammn: way better, thanks!

controllers/operator/mongodbreplicaset_controller_test.go
nammn: nit: hm - can we somehow deduct this instead of having this as a parameter? 

If not, good to resolve - feels like an overhead which I don't like in a test

controllers/operator/mongodbreplicaset_controller.go
nammn: oh yeah - i think - if its not too complex - we should have a conditional. Basically from customer pov, if my cluster restarted I would like to know the reason.

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: https://jira.mongodb.org/browse/CLOUDP-378172
Follow up ticket

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: Since it is only for a brand new deployment, there shouldn't be any additional rolling restart, this will just be the initial agent parameters set (I should edit the changelog again, it is wrong then)

controllers/operator/mongodbreplicaset_controller_test.go
Julien-Ben: Removed

controllers/operator/mongodbreplicaset_controller.go
nammn: ah got it, yea!

controllers/operator/common_controller.go
Julien-Ben: It was added in November 2024: https://github.com/10gen/mms-automation/commit/f1ee20c2e1

I assumed yes, same for the OM7 versions that our customers would use if they use our latest operator version.

docker/mongodb-kubernetes-init-database/content/agent-launcher.sh
Julien-Ben: I can only agree :'(

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: No actually the rolling restart is a new behaviour introduced by this PR:

- Enabling backup currently doesn't involve a rolling restart
- With this PR, enabling backup on an existing deployment, with external hostnames, will involve a rolling restart to pass the correct flag to the agent

We need to decide what is worse:

- Configuring wrong external hostname
- Introducing rolling restart when backups are enabled

controllers/om/omclient.go
```diff
@@ -524,6 +533,16 @@ func (oc *HTTPOmConnection) ReadAutomationAgents(pageNum int) (Paginated, error)
 	return resp, apierror.New(err)
 }
 
+// ReadAutomationAgents returns the state of the automation agents registered in Ops Manager
+func (oc *HTTPOmConnection) ReadAutomationAgents(pageNum int) (Paginated, error) {
```
viveksinghggits, line 537: nit: does it make sense to rename this to `ReadAutomationAgentsState`? I know that historically this is how things were but if we agree that `ReadAutomationAgentsState` is better than `ReadAutomationAgents`, I think we should rename.  It clearly says that we are reading state and not the agents.

controllers/operator/common_controller.go
```diff
@@ -838,6 +838,17 @@ func isPrometheusSupported(conn om.Connection) bool {
...
+func isBackupHostnameOverrideSupported(conn om.Connection) bool {
+	if conn.OpsManagerVersion().IsCloudManager() {
+		return true
+	}
+
+	omVersion, err := conn.OpsManagerVersion().Semver()
+	return err == nil && omVersion.GTE(semver.MustParse("7.0.0"))
```
viveksinghggits: should this `7.0.0` be a const? I saw it being used somewhere else as well.

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -616,11 +627,91 @@ func (r *ReplicaSetReconcilerHelper) buildStatefulSetOptions(ctx context.Context
...
+
+// shouldEnableBackupHostnameOverride determines if backup agents should use external hostnames.
+// This is needed when spec.externalDomain is configured, so backup agents register with
+// resolvable hostnames instead of internal pod names.
+//
+// Important: This must only be enabled for NEW deployments. Enabling on existing deployments
+// with already-registered backup agents causes hostname mismatches (see CLOUDP-260397).
+//
+// The feature requires Ops Manager 7.0+ (agent v13.26.0+) which supports the -backupOverrideLocalHost flag.
+func (r *ReplicaSetReconcilerHelper) shouldEnableBackupHostnameOverride(conn om.Connection, existingSts *appsv1.StatefulSet, log *zap.SugaredLogger) (bool, error) {
```
viveksinghggits, line 656: nit: If we read the comment `shouldEnableBackupHostnameOverride determines if backup agents should use external hostnames.`, I think the name of the method should be  `shouldBackupAgentUseExternalDomain`. 
This also aligns with what's happening in the method, i.e., (loosely) if resource has `ExternalDomain` set, return true. 
In my opinion, it makes things more readable, for example we can start with resource if it has externalDomain set, we can figure our that agent should use the external domain and then in agent-launcher.sh we can pass the correct flag.
If we do it the other way around, we read the externalDomain from resource and then set backup hostname override which according to me breaks the flow. We suddenly went from externalDomain to hostname override.

viveksinghggits: If you agree with above we will have to also change the name of the env var.

controllers/operator/construct/database_construction.go
```diff
@@ -68,6 +68,9 @@ const (
 	LogFileMongoDBEnv                = "MDB_LOG_FILE_MONGODB"
 	LogFileAgentMonitoringEnv        = "MDB_LOG_FILE_MONITORING_AGENT"
 	LogFileAgentBackupEnv            = "MDB_LOG_FILE_BACKUP_AGENT"
+
+	// BackupHostnameOverrideEnv sets up the -backupOverrideLocalHost flag in agent-launcher.sh
+	BackupHostnameOverrideEnv = "MDB_BACKUP_HOSTNAME_OVERRIDE"
```
viveksinghggits, line 73: related to the comment that I put [here](https://github.com/mongodb/mongodb-kubernetes/pull/737/changes#r2763656357), if we just read the const defined above, I would not be able to easily understand that it's figured out from the resource's external domain field. But if the env var is named let's `MDB_HOST_EXTERNAL_DOMAIN` (just as an example), we would know that it's somehow being figured out from the external domain.

changelog/20260203_fix_backup_agents_hostname_with_external_domain.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Fix backup agents hostname with external domain
+kind: fix
+date: 2026-02-03
+---
+
+* Replica sets: Fixed an issue where backup agents reported internal cluster hostnames instead of external hostnames when `externalDomain` was configured.
```
viveksinghggits: nit: should we also configure which resource `externalDomain` is configured using? I think that would make reader's life easier.

controllers/om/omclient.go
Julien-Ben: It could make sense. I'm just moving things around here and not creating the method.

I'd also argue that the Ops Manager method can be read as reading agents (although it's true it is their state):
`GET /groups/{PROJECT-ID}/agents ` https://www.mongodb.com/docs/ops-manager/current/reference/api/agents-get-all/

controllers/operator/common_controller.go
Julien-Ben: This was addressed in commit a022575 ("Proper version-gating"). The inline strings follow the same pattern as `isPrometheusSupported` above (`semver.MustParse("5.9.0")`), and their meaning is clear from the immediate surrounding context (the switch on `omVersion.Major`). The versions are also used as boundary values in tests (`7.0.12 → false`, `7.0.13 → true`), where hardcoded literals are preferable to sharing a const — otherwise a typo in the const would silently pass all tests.

controllers/operator/mongodbreplicaset_controller.go
Julien-Ben: The method doesn't decide whether to "use external domain" . They are enabled regardless via `-overrideLocalHost`. This method decides whether to also enable the `-backupOverrideLocalHost` flag, which is a separate concern with its own version-gating and safety logic (checking registered backup agents, OM version, etc.).
Do you find it confusing ?

controllers/operator/construct/database_construction.go
Julien-Ben: `MDB_HOST_EXTERNAL_DOMAIN` would be too generic — it doesn't indicate this is backup-specific. The env var controls the `-backupOverrideLocalHost` agent flag in agent-launcher.sh, so `MDB_BACKUP_HOSTNAME_OVERRIDE` accurately describes its purpose: override the backup agent's hostname.

changelog/20260203_fix_backup_agents_hostname_with_external_domain.md
Julien-Ben: Do you mean adding something like: 
`when spec.externalAccessConfiguration.externalDomain was configured on the MongoDB resource` ?

controllers/operator/common_controller.go
Julien-Ben: I added proper version gating here: https://github.com/mongodb/mongodb-kubernetes/pull/737/changes/a022575a22f01c396c487a9c3670e4783dd1409a

But it is based on OM versions and not agent versions. As discussed in DM we need to decide if we are happy or not with that

changelog/20260203_fix_backup_agents_hostname_with_external_domain.md
viveksinghggits: yes.

controllers/operator/mongodbreplicaset_controller.go
viveksinghggits: It's not confusing, what I meant was, in that method we are trying to figure out if resource as externalDomain then return true. IF that's the case why not name it like, enableExternalDomainOverride or something similar? 
That's all.

viveksinghggits: But I just wanted to bring this to your attention, we can leave it as it is as well.

Julien-Ben: We are more deciding if we should enable the agent flag for overriding. Deciding if a resource has external domain is done elsewhere, and usually just inferred from the spec.

changelog/20260203_fix_backup_agents_hostname_with_external_domain.md
Julien-Ben: Done

controllers/operator/mongodbreplicaset_controller.go
viveksinghggits: ok

changelog/20260203_fix_backup_agents_hostname_with_external_domain.md
viveksinghggits: you might have forgotten mentioning the resource.

PR #755, opened 2026-02-09
Enable development on local kind clusters

lucian-tosa:
# Summary

This PR makes development on local kind clusters easier.
The `make switch` target now has a shortcut (`ctrl + t`) for toggling between your evergreen machine and local kind clusters. This switches the `KUBECONFIG` variable to a local kubeconfig.

After switching to local, create your clusters by running `scripts/dev/recreate_kind_clusters.sh`.
Afterwards the operator and tests will all run against your local kind clusters.

## Checklist


docker/mongodb-enterprise-ops-manager/Dockerfile
```diff
@@ -70,11 +67,34 @@ RUN microdnf install --disableplugin=subscription-manager -y \
...
+  description="MongoDB Enterprise Ops Manager"
+
 RUN curl --fail -L -o ops_manager.tar.gz ${om_download_url} \
   && tar -xzf ops_manager.tar.gz \
   && rm ops_manager.tar.gz \
   && mv mongodb-mms* "${MMS_HOME}"
 
+ARG TARGETPLATFORM
+ARG jdk_arm_download_url
+RUN if [[ ${TARGETPLATFORM} == "linux/arm64" ]]; then \
```
lsierant, line 88: qq: isn't jdk part of the om.tar.gz?

scripts/dev/recreate_kind_clusters.sh
```diff
@@ -23,42 +23,55 @@ docker_cleanup 2>&1| prepend "docker_cleanup"
 docker_create_kind_network
 docker_run_local_registry "kind-registry" "5000"
 
+CLUSTER_1="e2e-cluster-1"
```
lsierant, line 26: hero! :clap:

scripts/dev/switch_context.sh
```diff
@@ -65,6 +80,11 @@ else
       echo "Using override file: ${override_context_file}. If you do not want to use one, remove the file or its contents."
       base_command+=" && source ${override_context_file}"
   fi
+
+  if [[ $(cat "${local_kind_file}") == "local" ]]; then
+    echo "Using local kind cluster, setting default KUBECONFIG to ~/.kube/config"
+    base_command+=" && export KUBECONFIG=~/.kube/config"
```
lsierant, line 86: nit: this is a bit magical hardcoding it here, don't you think it would be better to just write that export to an additional override file and have it included in the context switch this way?

But probably it doesn't matter in the end

lsierant (approved):

fealebenpae (approved):
That's a very cool use of fzf here! Would it make sense to add an argument for scripted access so we could do something like `make switch context=foo kind=local/evergreen`?

filipcirtog (approved):
Very cool! good job!!

PR #776, opened 2026-02-11
[DO NOT MERGE] Draft devbox setup for MCK

josvazg:
# Summary

Sample devbox env for MCK.

## Proof of Work

```bash
$ devbox shell
Starting a devbox shell...
✅ OpenLDAP headers found
✅ Cyrus-SASL headers found
--- Environment Ready ---
To (re)install Python deps: pip install -r requirements.txt
...
/
$ SHELL=$(which fish) devbox shell
Info: Ensuring packages are installed.
✓ Computed the Devbox environment.
Starting a devbox shell...
✅ OpenLDAP headers found
✅ Cyrus-SASL headers found
--- Environment Ready ---
To (re)install Python deps: pip install -r requirements.txt
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
/
$ ./scripts/evergreen/check_precommit.sh && echo PASS
...
PASS
```

## Checklist

- [X] Have you added changelog file? Use `skip-changelog` label if not needed

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -1,0 +1,41 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.16.0/.schema/devbox.schema.json",
+  "packages": [
+    "git@latest",
+    "python@3.13",
+    "go@1.24.12",
+    "openldap@latest",
+    "cyrus_sasl@latest",
+    "pkg-config@latest",
+    "kubernetes-helm@latest",
+    "chart-testing@latest",
+    "shellcheck@latest"
+  ],
+  "env": {
+    "DEVBOX_SKIP_VENV_CONFIG": "1",
+    "PROJECT_DIR": "${HOME}/mongodb/mongodb-kubernetes"
+  },
+  "shell": {
+    "init_hook": [
+      "# Go Toolchain Control",
+      "export GOTOOLCHAIN=auto",
+      "export NIX_GOROOT=$(go env GOROOT)",
+      "export PATH=$NIX_GOROOT/bin:$PATH",
+      "unset GOROOT",
+      "export GOWORK=off",
+
+      "# Python Environment Setup",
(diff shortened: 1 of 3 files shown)


devbox.json
```diff
@@ -0,0 +1,41 @@
...
+    "init_hook": [
+      "# Go Toolchain Control",
+      "export GOTOOLCHAIN=auto",
+      "export NIX_GOROOT=$(go env GOROOT)",
+      "export PATH=$NIX_GOROOT/bin:$PATH",
+      "unset GOROOT",
+      "export GOWORK=off",
+
+      "# Python Environment Setup",
+      "if [ ! -d .venv ]; then python -m venv .venv; fi",
```
filipcirtog: q: Do we need this two times? 

```
"if [ ! -d .venv ]; then python -m venv .venv; fi",
      "source .venv/bin/activate",
```

devbox.json
```diff
@@ -0,0 +1,41 @@
...
+    "shellcheck@latest"
+  ],
+  "env": {
+    "DEVBOX_SKIP_VENV_CONFIG": "1",
+    "PROJECT_DIR": "${HOME}/mongodb/mongodb-kubernetes"
+  },
+  "shell": {
+    "init_hook": [
+      "# Go Toolchain Control",
+      "export GOTOOLCHAIN=auto",
```
filipcirtog: can `GOTOOLCHAIN` and `GOWORK` also be set in `"env"` ?

josvazg: I guess they can, will move then there

devbox.json
josvazg: I guess I can remove this one and leave the other only

josvazg:
This is too old to keep around, I do have a devbox setup I use personally and keep up to date and working. If we ever need or agree to put it in the repo, I can just recreate the PR easily.

PR #836, opened 2026-02-26
[PoC] Sample conditional evg runs

josvazg:
# Summary

Sample scrip to enable conditional runs of scopes in evergreen CI.

## Proof of Work

TBD

## Checklist

- [X] Have you added changelog file? used `skip-changelog` label


docs/ci-path-filtering-quick-start.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Integration Options
+
+### Option A: Variant-Level (Recommended for E2E)
+
+Add to variant definition in `.evergreen.yml`:
+
+```yaml
+- name: e2e_mdb_community
+  activate:
```
MaciejKaras, line 47: I believe `activate` is only true or false, not a complex object like showed below. I think AI hallucinated here

docs/ci-path-filtering-quick-start.md
```diff
@@ -0,0 +1,160 @@
...
+        working_dir: src/github.com/mongodb/mongodb-kubernetes
+        script: scripts/evergreen/should_run.sh e2e_tests
+  # ... rest of variant config
+```
+
+**When to use**: Expensive variants (e2e tests, image builds)
+
+### Option B: Task-Level Early Exit
+
+Add to task commands:
```
MaciejKaras, line 62: The biggest concern is that we will still spin up the worker just to exit. It's waste of resources and not elegant, but otherwise easy to implement.

docs/ci-path-filtering-quick-start.md
```diff
@@ -0,0 +1,160 @@
...
+
+```yaml
+- func: run_task_conditionally
+  vars:
+    condition_script: scripts/evergreen/should_run.sh e2e_tests
+    variant: e2e_mdb_community
+    task: e2e_mdb_community_task_group
+```
+
+**When to use**: When you need to conditionally add tasks to variants
```
MaciejKaras, line 93: you can even create variants from scratch with [`tasks.generate`](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Commands#generatetasks), so no need to have empty variants in the evergreen.yaml

docs/ci-path-filtering-quick-start.md
```diff
@@ -0,0 +1,160 @@
...
+        if ! scripts/evergreen/should_run.sh e2e_tests; then
+          echo "Skipping - only docs changed"
+          exit 0
+        fi
+  # ... rest of commands
+```
+
+**When to use**: Cheaper tasks (unit tests, lint)
+
+### Option C: Dynamic Task Addition
```
MaciejKaras, line 81: The biggest downside of this option is that we will move our static, yaml configuration into .json file which will not be validated when pushed. This will be validated on execution only. In other words we can break evergreen configuration without knowing it and learn about that only when it's actually executed.

We could minimise the impact of this by always using the variant, task generation **also for master merges**. This way any configuration issue will show itself during master builds. It's not perfect and we can still push faulty code. Other option is to use script to generate all tasks and call `evergreen validate` afterwards.

MaciejKaras: On the other side, this option provides minimal overhead to Evergreen workers usage + we only start the jobs we actually need.

docs/ci-path-filtering-quick-start.md
nammn: i would even go with something like shrub https://github.com/mongodb/mongodb-kubernetes/blob/bb8d3ba67b6505112e79cd35b7c271e264627a1b/scripts/evergreen/e2e/mco/create_mco_tests.py#L1

docs/ci-path-filtering-quick-start.md
josvazg: Yes, I do not like this one either. I would like to have as much logic away from the CI as possible, but spinning resources for nothing must be avoided. We need to figure our the smallest integration so that we avoid launching things that should not run.

docs/ci-path-filtering-quick-start.md
josvazg: My preference is that we can debug things locally. Before I even send the PR I should be able to find out what will run and why.

That means I should be able to run the scripts and know in the CLI and not depend on something that only runs in the CI. I have been burn many times with debugging GitHub issues so I'd rather move as much logic to our own code than relaying to heavily on CI tricks.

nammn: its about dynamically creating tasks, right? In that case shrub does that for you. You can run it locally and verify that it generates the tasks it would generate in the CI.

josvazg: In my mind it is not about creating tasks dynamically, the tree of tests to run is always the same, but we skip some branches in some cases before PR merging, and never after merge.

Ideally we have a script that tells us what would run or skip in the CLI. That same script is used by evergreen, ideally to run or SKIP things completely, not spinning a VM to do nothing and cleaning up right away.

nammn: conditionally adding a tasks is - from my understanding - either adding a task, which will run a host or not add a task and thus not run a host. 

Just to be on the same page terminology-wise. In evergreen we would always run a task on a host when its been added to the overall yaml tree. 

> Ideally we have a script that tells us what would run or skip in the CLI

Thats what I meant by dynamically creating tasks. Just create the task we need, otherwise don't create the ones we don't need. If you have `a` and `b` in the yaml, evergreen will spawn hosts for `a` and `b`. With shrub or any evergreen task generator framework you can dynamically generate tasks - in this example only `a` if file `x` has changed.

docs/ci-path-filtering-quick-start.md
josvazg: This is actually the closest option to what I had in mind. The script acts as a Gandalf that sometimes goes "though shall not pass!" and that branch of tests or variant is skipped completely.

> The biggest downside of this option is that we will move our static, yaml configuration into .json file which will not be validated when pushed. 

I do not think we are doing that. This proposal does not replace YAML with JSON config. The JSON file is an additional filter that select which parts (if any) of the test tree can be skipped on PRs (before merging)

If we are worried about the JSON integrity (which I do not think we would touch too often), we could include a quick validity test that runs always and before anything else. Or the script itself could do the JSON validation and give spit a proper error message in Evergreen or in the CLI.

The goal is not avoiding making mistakes, we will always make mistakes. The goal is to be able to detect mistakes ASAP, even before we send the changes to CI, from our local CLI.

MaciejKaras: Can you explain how would you achieve dynamic tasks addition to variants? Some variants will be empty and the tasks defined in evergreen.yml, but unused? And then your script will just link the tasks to variants (which were empty earlier) depending on the configuration?

I wonder if evergreen will allow empty variants or it will validate unused tasks the same way it does regular tasks.

josvazg: I understand thus runs somewhere:

- func: run_task_conditionally
  vars:
    condition_script: scripts/evergreen/should_run.sh e2e_tests
    variant: e2e_mdb_community
    task: e2e_mdb_community_task_group
    
 But if the condition is not met, then the task "e2e_mdb_community_task_group" is skipped.
   
 The code in the repo does not change, what changes is what gets to run.
 
 Maybe my mental model of how evergreen runs things is too naive right now.
 
 We might need to sync on this

MaciejKaras: function `run_task_conditionally` underneath actually runs `generate.tasks` and prepares `evergreen_tasks.json` file based on the parameters passed, namely `task`:

```yaml
  run_task_conditionally:
    - command: shell.exec
      params:
        shell: bash
        script: |
          if ${condition_script}; then
            echo "Adding ${task} task to ${variant} variant"
            scripts/evergreen/add_evergreen_task.sh ${variant} ${task}
          else
            echo "skipping task ${task} due to ${condition_script} result: $?"
          fi
    - command: generate.tasks
      params:
        files:
          - evergreen_tasks.json
        optional: true
```

I'm not saying this is bad or anything, but just wanted to highlight this. If you want to have single logic (`should_run.sh`) that decides what to run, you would have to have single Evergreen `should_run` task that does all of this and more what we have in `run_task_conditionally` function e.g. deciding about adding which task to which variant, produce resulting `evergreen_tasks.json` file and so on. This will be additional logic and we cannot reuse `run_task_conditionally` as is.

josvazg: Not necessarily, the `should_run.sh` works with "scopes", each scope is one decision point and is associated with a set of files that trigger such scope. For instance: `should_run.sh e2e-tests` is different from `should_run.sh build-image` and might reply differently on PRs. Which means we could have different run_task_conditionals using different scopes.

For me the most important one would be e2e tests, which I expect to be the most expensive.

josvazg:
We do not seem to be following this path, at least for now

PR #897, opened 2026-03-14
Use ty for statically check python types

lsierant:
# Summary

Expand static Python type checking in `docker/mongodb-kubernetes-tests/` from the search tests to every `tests/` file. Before this PR the `ty` pre-commit hook only ran against `tests/search/` and `tests/common/search/`; everything else was unchecked.

## Proof of Work

Green [evg](https://spruce.mongodb.com/version/69e9a71e8b720100074d7b76)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -131,22 +131,12 @@ repos:
 
   - repo: local
     hooks:
-      - id: ty-search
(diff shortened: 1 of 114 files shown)


MaciejKaras (approved):

Julien-Ben (approved):
Awesome !

PR #900, opened 2026-03-16
Make http proxy test more strict

fealebenpae:
# Summary
This change makes the operator proxy test more strict in order to better catch failures in configuring the automation agent for proxy access. It also runs them against Ops Manager, not just Cloud Manager to test the Ops Manager agent behavior as well.

This is achieved by a NetworkPolicy that allows the tested replica set pods to access only each other and the proxy pod. Failure to reconcile means that some component didn't pick up the proxy settings.

Currently this fails because the automation agent doesn't seem to respect the proxy environment variables defined by the operator. Adding the `-httpProxy` agent flag in the replica set CR fixes that.

## Proof of Work
[Successful* patch](https://spruce.corp.mongodb.com/version/69b6a9dc707b620007ab59a4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) with `-httpProxy` set.
[Failing patch](https://spruce.corp.mongodb.com/version/69b6b611a37b080007fb3aad/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) with just the proxy environment variables.

\* The non-static CloudQA variant fails because of a regression in the Cloud Manager automation agent's handling of proxy settings when downloading binaries.

## Checklist


filipcirtog (approved):

docker/mongodb-kubernetes-tests/tests/operator/operator_proxy.py
```diff
@@ -38,22 +38,23 @@ def check_svc_endpoints():
...
 
 
 @fixture(scope="module")
-def replica_set(namespace: str, custom_mdb_version: str) -> MongoDB:
+def replica_set(namespace: str, custom_mdb_version: str, squid_proxy: str) -> MongoDB:
     resource = MongoDB.from_yaml(_fixture("replica-set-basic.yaml"), namespace=namespace, name=MDB_RESOURCE)
     resource.set_version(ensure_ent_version(custom_mdb_version))
     resource.set_architecture_annotation()
+    resource.configure(om=get_ops_manager(namespace), project_name=resource.name)
+    # Uncomment the following to pass the -httpProxy flag to the agent at startup
```
MaciejKaras: Is this comment still relevant?

fealebenpae: Yes - there is a separate issue that the automation agent doesn't seem to honor the `HTTP_PROXY` environment variables set on the database pods. The tests need to be able to pass with those lines commented out as well. I can invert the comment to make this more obvious.

MaciejKaras: It's not clear, because there is nothing to uncomment below. If you could reword this it would be great.

fealebenpae: I reworded that, thanks for pointing it out.

docker/mongodb-kubernetes-tests/tests/operator/operator_proxy.py
```diff
@@ -67,20 +68,58 @@ def test_install_operator_with_proxy(
...
+                client.V1NetworkPolicyEgressRule(
+                    ports=[
+                        client.V1NetworkPolicyPort(port=53, protocol="UDP"),
+                        client.V1NetworkPolicyPort(port=53, protocol="TCP"),
+                    ]
+                ),
+            ],
+        ),
+    )
+    client.NetworkingV1Api().create_namespaced_network_policy(namespace, policy)
```
MaciejKaras, line 116: nit: can we somehow easily verify that there is no access to other resources than the ones specified in the NetworkPolicy? I know it should and will work right now, but the thing is we want to catch issues if someone changes the NetworkPolicy in the tests.

Or another, simpler idea. Run the tests without the squid proxy deployed, but with `httpProxy` config and assert that it fails. After that deploy squid proxy and check it all works.

docker/mongodb-kubernetes-tests/tests/operator/operator_proxy.py
MaciejKaras: Oh, and if you could link the related issue 🙈

fealebenpae: We don't have a related issue yet, I'm hoping to use this test as the justification to create one 😈

docker/mongodb-kubernetes-tests/tests/operator/operator_proxy.py
fealebenpae: That's a good idea! I'll see what's the best way to assert a failure in the absence of a proxy first.

PR #941, opened 2026-03-27
CLOUDP-239791: Replace log.Fatal calls in main.go with run() error pattern

m1kola:
# Summary

Extract operator startup logic into a `run()` function that returns an `error`, so deferred cleanup (tracer shutdown, span end) runs on failure instead of being bypassed by `os.Exit` via `log.Fatal`.

## Proof of Work

CI must be green

## Checklist

### DIFF
--- a/main.go
+++ b/main.go
@@ -110,6 +110,13 @@ func (c *crdsToWatch) String() string {
 }
 
 func main() {
+	if err := run(); err != nil {
+		log.Error(err)
+		os.Exit(1)
+	}
+}
+
+func run() error {
 	flag.Parse()
 	// If no CRDs are specified, we set default to non-multicluster CRDs
 	if len(crds) == 0 {
@@ -200,13 +207,13 @@ func main() {
 
 	mgr, err := ctrl.NewManager(cfg, managerOptions)
 	if err != nil {
-		log.Fatal(err)
+		return err
 	}
 	log.Info("Registering Components.")
 
 	// Setup Scheme for all resources
 	if err := apiv1.AddToScheme(scheme); err != nil {
-		log.Fatal(err)
+		return err
 	}
 
 	// memberClusterObjectsMap is a map of clusterName -> clusterObject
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -110,6 +110,13 @@ func (c *crdsToWatch) String() string {
 }
 
 func main() {
+	if err := run(); err != nil {
+		log.Error(err)
+		os.Exit(1)
```
josvazg, line 115: q: why not making this a fatal? I would also wrap the error with some info.

```suggestion
		log.Fatalf("Fatal error: %v", err)
```

main.go
```diff
@@ -199,13 +206,13 @@ func main() {
 
 	mgr, err := ctrl.NewManager(cfg, managerOptions)
 	if err != nil {
-		log.Fatal(err)
+		return err
```
josvazg, line 210: nit: it is usually best to wrap the error context. E.g:

```suggestion
		return fmt.Errorf("failed to create manager: %w", err)
```

main.go
m1kola: I normally prefer explicit over implicit. And `log.Fatalf` has an implicit exit.

But I don't mind changing it to `log.Fatalf`, if you prefer.

main.go
m1kola: This is startup code in `main`, not a library/package. The errors from `ctrl.NewManager`, `apiv1.AddToScheme`, etc. already return descriptive messages. Wrapping them adds boilerplate without giving the caller anything useful. And there's no caller that needs to `errors.Is`/`errors.As` these, and the log output in `main` will print the original message anyway.

The point of the ticket was to fix the `os.Exit` bypass, not to improve error messages.

main.go
josvazg: I'd say `log.Fatalf` is more idiomatic and standard

josvazg (approved):

Julien-Ben (approved):
I like the change, it's more readable this way 👍

m1kola:
Had to rebase to resolve conflicts: `setupMongoDBUserCRD` and `setupMongoDBCRD` got a new `backupEnableDelay` argument and this change was conflicting with changes in the PR.

PR #948, opened 2026-03-30
fix(community): fix MongoDBCommunity scale-up deadlock when TLS/mTLS …

princeba84:
…is enabled

When MongoDBCommunity has TLS enabled and spec.members is increased, the reconciliation loop gets stuck indefinitely with the StatefulSet never scaling up.

Root cause: In shouldRunInOrder(), the TLS ordering check fires before the scale-up check. During scale-up the StatefulSet is not yet ready which causes the TLS not-ready check to return true (AC first) before the scale-up check is ever evaluated. The AutomationConfig is updated with the new member count while the StatefulSet still has the old pod count. The agent waits forever for pods that do not exist yet causing an infinite reconciliation loop.

Fix: Reorder checks so scale-up is always evaluated before TLS checks. ensure TLS is applied to all members at once after scaling completes, not at an intermediate member count. This also correctly handles the case where TLS is enabled simultaneously with a scale-up.

Tested on OpenShift with MongoDB Community Operator v1.7.0:
- Scale up 3->5 with mTLS already enabled (original bug) PASS
- Enable TLS + scale up 3->5 simultaneously PASS
- Scale down 5->3 with mTLS enabled PASS

# Summary

## Checklist

### DIFF
--- a/mongodb-community-operator/controllers/replica_set_controller.go
+++ b/mongodb-community-operator/controllers/replica_set_controller.go
@@ -397,19 +397,24 @@ func (r *ReplicaSetReconciler) deployAutomationConfig(ctx context.Context, mdb m
 
 // shouldRunInOrder returns true if the order of execution of the AutomationConfig & StatefulSet
 // functions should be sequential or not. A value of false indicates they will run in reversed order.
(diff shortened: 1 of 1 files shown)


princeba84:
Could you please have a look this issue!

PR #950, opened 2026-03-30
CLOUDP-385917: Add installation method telemetry

roothorp:
# Summary

Track the installation method of MCK (Helm, OLM, or YAML) and return that result alongside the existing telemetry we send.

Installation method is determined based on annotations we attach to the operator, which we then retrieve as environment variables via the downward API.

## Checklist

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -18,6 +18,8 @@ spec:
         app.kubernetes.io/component: controller
         app.kubernetes.io/name: mongodb-kubernetes-operator
         app.kubernetes.io/instance: mongodb-kubernetes-operator
+      annotations:
+        mongodb.com/installation-method: "helm"
     spec:
       serviceAccountName: mongodb-kubernetes-operator
       containers:
@@ -102,6 +104,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: MCK_INSTALLER
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.annotations['mongodb.com/installation-method']
             - name: OPERATOR_NAME
               value: mongodb-kubernetes-operator
             # Community Env Vars Start
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -19,7 +19,7 @@ spec:
         app.kubernetes.io/name: mongodb-kubernetes-operator
         app.kubernetes.io/instance: mongodb-kubernetes-operator
       annotations:
-        mongodb.com/installation-method: "olm"
+        mongodb.com/installation-method: "yaml"
     spec:
       serviceAccountName: mongodb-kubernetes-operator
       containers:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -372,4 +372,3 @@ spec:
       volumes:
         - name: webhook-server-dir
           emptyDir: {}
-
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -19,7 +19,7 @@ spec:
         app.kubernetes.io/name: mongodb-kubernetes-operator
         app.kubernetes.io/instance: mongodb-kubernetes-operator
       annotations:
-        mongodb.com/installation-method: "helm"
+        mongodb.com/installation-method: "olm"
     spec:
       serviceAccountName: mongodb-kubernetes-operator
       containers:
@@ -372,3 +372,4 @@ spec:
       volumes:
         - name: webhook-server-dir
           emptyDir: {}
+
--- a/helm_chart/templates/operator.yaml
+++ b/helm_chart/templates/operator.yaml
@@ -19,9 +19,10 @@ spec:
         app.kubernetes.io/component: controller
         app.kubernetes.io/name: {{ .Values.operator.name }}
         app.kubernetes.io/instance: {{ .Values.operator.name }}
+      annotations:
+        mongodb.com/installation-method: {{ .Values.operator.installationMethod | default "helm" | quote }}
 {{- if .Values.operator.vaultSecretBackend }}
   {{- if .Values.operator.vaultSecretBackend.enabled }}
-      annotations:
         vault.hashicorp.com/agent-inject: "true"
         vault.hashicorp.com/role: "mongodbenterprise"
         {{- if .Values.operator.vaultSecretBackend.tlsSecretRef }}
@@ -220,6 +221,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: MCK_INSTALLER
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.annotations['mongodb.com/installation-method']
             - name: OPERATOR_NAME
               value: {{ .Values.operator.name }}
             # Community Env Vars Start
--- a/pkg/telemetry/collector.go
+++ b/pkg/telemetry/collector.go
@@ -175,6 +175,7 @@ func collectOperatorSnapshot(ctx context.Context, memberClusterMap map[string]Co
 		uid := getKubernetesClusterUUID(ctx, uncachedClient)
 		kubeClusterUUIDList = append(kubeClusterUUIDList, uid)
 	}
+	installMethod := envvar.GetEnvOrDefault("MCK_INSTALLER", "")
 
 	slices.Sort(kubeClusterUUIDList)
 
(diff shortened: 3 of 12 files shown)


scripts/evergreen/operator-sdk/prepare-openshift-bundles.sh
```diff
@@ -27,6 +27,10 @@ echo "Aligning metadata.annotations.containerImage version with deployment's ima
 operator_deployment_image=$(yq '.spec.install.spec.deployments[0].spec.template.spec.containers[0].image' < "${bundle_csv_file}")
 yq e ".metadata.annotations.containerImage = \"${operator_deployment_image}\"" -i "${bundle_csv_file}"
 
+echo "Setting installation-method annotation to 'olm' in deployment pod template in ${bundle_csv_file}"
+yq e '.spec.install.spec.deployments[0].spec.template.metadata.annotations."mongodb.com/installation-method" = "olm"' \
```
m1kola: @roothorp nit: it would be great to not depend on index in `.spec.install.spec.deployments[0]`. This could lead to a difficult to spot and troubleshoot bug if we add another deployment and it takes this index.

Can we filter deployments by `name`? It would also be great to make this script fail, if we can't find the right deployment (e.g. if we rename deployment name for whatever reason).

roothorp: I've filtered by name using yq, in all of the places where we previously used `.deployments[0]` in this script.

config/manager/manager.yaml
```diff
@@ -18,6 +18,8 @@ spec:
         app.kubernetes.io/component: controller
         app.kubernetes.io/name: mongodb-kubernetes-operator
         app.kubernetes.io/instance: mongodb-kubernetes-operator
+      annotations:
+        mongodb.com/installation-method: "helm"
```
m1kola: Hm. Why is it hardcoded to helm? Is there a chance of it ending up in static yaml files (e.g. via `kustomize`)?

pkg/telemetry/collector.go
```diff
@@ -175,6 +179,7 @@ func collectOperatorSnapshot(ctx context.Context, memberClusterMap map[string]Co
 		uid := getKubernetesClusterUUID(ctx, uncachedClient)
 		kubeClusterUUIDList = append(kubeClusterUUIDList, uid)
 	}
+	installMethod := envvar.GetEnvOrDefault(InstallerEnvVar, "")
```
m1kola: Just noticed this. We normally read env vars in `main.go` and pass it down via arguments. Otherwise it is difficult to track inventory of env vars. We are not in there yet to have all env var reads in main, but we are trying not to add new cases.

We even have [a lint rule](https://github.com/mongodb/mongodb-kubernetes/blob/d89262c9bd16182b4ccadca13706cc8c10ab2df2/.golangci.yml#L47-L49) for that, but apparently it is not working.

m1kola (commented):
A few minor comments, but other than that looks good to me.

config/manager/manager.yaml
```diff
@@ -366,3 +372,4 @@ spec:
       volumes:
         - name: webhook-server-dir
           emptyDir: {}
+
```
m1kola: It seems like the linter is not happy with this new line for some reason...

Edit: Ignore me. It was before you pushed the change it seeems.

m1kola (approved):

josvazg:
Could you please add a brief summary on what this is and why it is needed? A couple of phrases may suffice.

pkg/telemetry/types.go
```diff
@@ -22,6 +22,7 @@ type OperatorUsageSnapshotProperties struct {
 	OperatorType         OperatorType `json:"operatorType"`         // MEKO, MCK, MCO (here meko)
 	OperatorArchitecture string       `json:"operatorArchitecture"` // Architecture of the operator binary (amd64, arm64, s390x, ppc64le)
 	OperatorOS           string       `json:"operatorOS"`           // Operating system of the operator binary (linux, darwin, windows)
+	OperatorInstaller    string       `json:"operatorInstaller"`    // Installation method of the operator (helm, olm, yaml, unknown)
```
josvazg, line 25: > Empty string vs "unknown" — documentation inconsistency (minor, but worth aligning)
> 
>   types.go:25 documents the field as // Installation method of the operator (helm, olm, yaml, unknown), but main.go:317 uses env.ReadOrDefault(telemetry.InstallerEnvVar, "") — the fallback is "", not "unknown". If a customer uses a custom deployment that doesn't set the annotation, Segment will receive an empty string. This makes querying/filtering harder than a sentinel like "unknown".

helm_chart/templates/operator.yaml
```diff
@@ -19,9 +19,10 @@ spec:
         app.kubernetes.io/component: controller
         app.kubernetes.io/name: {{ .Values.operator.name }}
         app.kubernetes.io/instance: {{ .Values.operator.name }}
+      annotations:
+        mongodb.com/installation-method: {{ .Values.operator.installationMethod | default "helm" | quote }}
```
josvazg, line 23: >installationMethod not declared in values.yaml (minor)

>   helm_chart/templates/operator.yaml:23 references .Values.operator.installationMethod but there's no entry for it in helm_chart/values.yaml or any of the variant values files. It works at runtime because of the | default "helm" fallback, but it's undocumented — Helm lint may warn, and users inspecting the values file won't see it.

public/mongodb-kubernetes-multi-cluster.yaml
```diff
@@ -95,6 +95,7 @@ subjects:
   - kind: ServiceAccount
     name: mongodb-kubernetes-operator-multi-cluster
     namespace: mongodb
+
```
josvazg: >   3. Trailing space in generated ServiceAccount definitions (cosmetic / CI lint risk)

> In all three public/*.yaml files, the ServiceAccount namespace line now has a trailing space: namespace: mongodb  
> This was introduced by the generate_files.sh run. It's harmless functionally but can cause YAML linters or diff-based CI checks to flag the files on future regeneration.

josvazg: >  The space between }} and {{/* leaks into rendered output as namespace: mongodb  in all three public/*.yaml files. Fix:
>  -{{- end }} {{/* if .Values.operator.createOperatorServiceAccount */}}
>  +{{- end -}} {{/* if .Values.operator.createOperatorServiceAccount */}}
>  Then regenerate the public YAMLs.

pkg/telemetry/collector_test.go
```diff
@@ -1454,6 +1454,7 @@ func TestCollectOperatorSnapshot(t *testing.T) {
 				OperatorType:         MEKO,
 				OperatorArchitecture: runtime.GOARCH,
 				OperatorOS:           runtime.GOOS,
+				OperatorInstaller:    "yaml",
```
josvazg, line 1457: >  4. Test coverage only covers "yaml" installer — collector_test.go

>  No test cases for "helm", "olm", or the empty/unknown fallback. Worth adding table-driven cases to TestCollectOperatorSnapshot to cover the full enum.

pkg/telemetry/types.go
roothorp: This is intentional; by "unknown" I meant "anything else" - including users explicitly unsetting the value (or setting it to something else)

helm_chart/templates/operator.yaml
roothorp: This was also intentional; we don't really want users setting this.

pkg/telemetry/collector_test.go
roothorp: I will add these, I think it's trivial.

public/mongodb-kubernetes-multi-cluster.yaml
roothorp: I think this is what was causing the weird linting issues - good catch.

roothorp: Looks like a bug with Helm; https://github.com/helm/helm/issues/32163

pkg/telemetry/collector_test.go
roothorp: Added

public/mongodb-kubernetes-multi-cluster.yaml
roothorp: When using helm@v3, works fine. I'll continue with this until we deliberately update to using v4.

josvazg (approved):

PR #970, opened 2026-04-07
Integrate connectivity validation dry-run into ReplicaSet controller

nammn:


controllers/operator/migration/jobrunner.go
```diff
@@ -0,0 +1,104 @@
...
+	var pods corev1.PodList
+	if err := kubeClient.List(ctx, &pods,
+		client.InNamespace(job.Namespace),
+		client.MatchingLabels{"job-name": job.Name},
+	); err != nil {
+		return migration.ExitUnknown, time.Time{}
+	}
+	for i := range pods.Items {
+		for _, cs := range pods.Items[i].Status.ContainerStatuses {
+			if cs.Name == "connectivity-validator" && cs.State.Terminated != nil {
```
filipcirtog: q: should we use a constant for `connectivity-validator` ?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -298,6 +304,22 @@ func (r *ReplicaSetReconcilerHelper) Reconcile(ctx context.Context) (reconcile.R
 		}
 	}
 
+	// 5a. Connectivity dry-run: launch a validation Job without touching OM or StatefulSets.
```
filipcirtog, line 307: as the validation dry-run step is placed after `EnsureFeatureControls`, what would happen if the validation fails? Should we `ClearFeatureControls` to allow UI OM changes?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -804,6 +827,80 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+		if updateResult, updateErr := r.updateStatus(ctx,
+			workflow.Failed(fmt.Errorf("%s: %s", result.Reason, result.Message)),
+			mdbstatus.NewMigrationConditionOption(mdbstatus.MigrationCondition(
+				mdbstatus.MigrationPhaseConnectivityCheckFailed, result.Reason, result.Message,
+			)),
+		); updateErr != nil {
+			return updateResult, updateErr
+		}
+		// Requeue after 5 minutes to retry once connectivity issues may be resolved.
+		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
```
filipcirtog, line 931: the requeue is every `5 minutes` but the job's time to live is `10 minutes`. if the previous job returned as failed (and won't be collected by the GC), would the next requeue work with the previous job already present or will just return its previous iternation state?

controllers/operator/mongodbreplicaset_controller.go
filipcirtog: I was having the impression we are not performing any OM changes but I may understand the workflow incorrectly.

filipcirtog (approved):

controllers/operator/mongodbreplicaset_controller.go
nammn: yeah we discussed we should open fc in dry-run, let me do it!

controllers/operator/migration/jobrunner.go
nammn: will do!

controllers/operator/mongodbreplicaset_controller.go
nammn: it would just return failed again and requeue in 5 minutes until the job is GC'ed so e.g. in 15 minutes after the third retry. This unit test should reflect that: https://github.com/mongodb/mongodb-kubernetes/blob/787fc6f7fb0ceabe9f3772d09f05b3425fcb72ba/controllers/operator/migration/jobrunner_test.go#L209

config/manager/manager.yaml
```diff
@@ -52,6 +52,9 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            # Must match the image actually running (OLM/certified installs: set to the same ref the catalog pulls, e.g. Red Hat registry).
+            - name: MDB_OPERATOR_IMAGE
+              value: "quay.io/mongodb/mongodb-kubernetes:1.7.0"
```
josvazg, line 57: **q**: why are these config values wired in as code?

nammn: MDB_OPERATOR_IMAGE contains the validator binary which we deploy in the job (its described in a bit better detail in the TD): https://docs.google.com/document/d/1AUUk5ehKKyyFiPN8G6QN6cRcpEmUFuNWxFsUgdLJJxA/edit?tab=t.z8gok0h29svj#bookmark=id.uisllulhhl1p

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -804,6 +827,80 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+// reachable from within the cluster. No StatefulSets or Ops Manager config are modified.
+// The Job is built from the same StatefulSet spec (buildStatefulSetOptions + DatabaseStatefulSet)
+// so it uses the same credentials volumes and mounts as the STS.
+//
+// While the Job is still running the resource phase is set to Pending and reconciliation
+// is requeued after 30s. On success the resource phase is set to Running with a
+// ConnectivityCheckPassed migration status. On failure the phase is set to Failed with a
+// ConnectivityCheckFailed migration status and reconciliation is requeued after 5 minutes.
+// TODO: extract to common controllers
+func (r *ReplicaSetReconcilerHelper) runConnectivityValidationDryRun(ctx context.Context, conn om.Connection, projectConfig mdbv1.ProjectConfig, externalMemberProcessNames []mdbv1.ExternalMember, rs *mdbv1.MongoDB, deploymentOpts deploymentOptionsRS, operatorImage, internalClusterCertPath string, log *zap.SugaredLogger) (reconcile.Result, error) {
```
josvazg: q: this method mutates deploymentOpts which is a copy to a struct, those changes will not survive this call. Was that expected?

josvazg: wrong place

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -717,7 +740,7 @@ func AddReplicaSetController(ctx context.Context, mgr manager.Manager, imageUrls
 
 // updateOmDeploymentRs performs OM registration operation for the replicaset. So the changes will be finally propagated
 // to automation agents in containers
-func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, conn om.Connection, membersNumberBefore int, tlsCertPath, internalClusterCertPath string, deploymentOptions *deploymentOptionsRS, shouldMirrorKeyfileForMongot bool, isRecovering bool) workflow.Status {
+func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, conn om.Connection, membersNumberBefore int, tlsCertPath, internalClusterCertPath string, deploymentOptions deploymentOptionsRS, shouldMirrorKeyfileForMongot bool, isRecovering bool) workflow.Status {
```
josvazg, line 752: **q**: this method mutates deploymentOpts which is a copy to a struct, those changes will not survive this call. Was that expected?

nammn: i've modified this by accident in prior stack, on master its without, since we didn't plan to modify it there or rather rely on that. 

https://github.com/mongodb/mongodb-kubernetes/blob/master/controllers/operator/mongodbreplicaset_controller.go#L291

controllers/operator/migration/jobrunner.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// jobPodOutcome lists the Job's pods once and returns the connectivity-validator container's exit code and finished time.
+func jobPodOutcome(ctx context.Context, kubeClient client.Client, job *batchv1.Job) (exitCode int32, finishedAt time.Time) {
+	var pods corev1.PodList
+	if err := kubeClient.List(ctx, &pods,
+		client.InNamespace(job.Namespace),
+		client.MatchingLabels{"job-name": job.Name},
+	); err != nil {
+		return migration.ExitUnknown, time.Time{}
```
josvazg: **q**: Maybe we should log here in case this is hiding an RBAC issue?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -804,6 +827,80 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+// runConnectivityValidationDryRun launches (or polls) a connectivity-validator Kubernetes Job
+// that checks whether all external MongoDB members in the current Ops Manager deployment are
+// reachable from within the cluster. No StatefulSets or Ops Manager config are modified.
+// The Job is built from the same StatefulSet spec (buildStatefulSetOptions + DatabaseStatefulSet)
+// so it uses the same credentials volumes and mounts as the STS.
+//
+// While the Job is still running the resource phase is set to Pending and reconciliation
+// is requeued after 30s. On success the resource phase is set to Running with a
+// ConnectivityCheckPassed migration status. On failure the phase is set to Failed with a
+// ConnectivityCheckFailed migration status and reconciliation is requeued after 5 minutes.
```
josvazg: q: this does not seem to be accurate right?


```suggestion
// The MongoDB resource's status.phase remains ConnectivityValidation for the dry-run path. The
// NetworkConnectivityVerification condition is updated to reflect the Job: ConnectivityCheckRunning
// (condition status Unknown) while the Job runs, ConnectivityCheckPassed (True) on success, or
// ConnectivityCheckFailed (False) on failure. Reconciliation is requeued every 30s while the Job
// is in progress; on failure the status is updated to Failed and reconciliation is requeued after
// 5 minutes. Remove the migration dry-run annotation to run full reconciliation.
```

controllers/operator/migration/jobrunner.go
nammn: done!

fealebenpae (approved):

PR #973, opened 2026-04-07
feat: add migrate mongodb and users subcommands with validation to kubectl-mongodb plugin

filipcirtog:
This PR adds `kubectl mongodb migrate mongodb` and `kubectl mongodb migrate users` subcommands to the `kubectl-mongodb` plugin. Both subcommands connect to Ops Manager using the same ConfigMap and Secret format as the operator, read the Automation Config, and run compatibility validation before producing any output. CR generation will be added on top of this foundation in a follow-up.  
  
Why this change is n**eeded**?

Migrating a MongoDB replica set from Ops Manager to the Kubernetes operator requires manually reconstructing a CR from an Automation Config. The migration workflow requires generating two distinct resource types: a `MongoDB` CR from the replica set configuration and `MongoDBUser` CRs from the users in the Automation Config. These are split into separate subcommands because some customers may choose to migrate only the MongoDB resource and manage users separately. Each subcommand runs validation against operator defaults before producing any output, so that incompatibilities are surfaced before the operator begins reconciling.

**What this PR does**?

- This PR introduces `kubectl mongodb migrate` with two subcommands: 
    - `mongodb` for generating a `MongoDB` or `MongoDBMultiCluster` CR and `users` for generating `MongoDBUser` CRs. Rather than generating CRs immediately, both subcommands validate the Automation Config against operator defaults before producing any output, surfacing actionable messages instead of letting the operator expose problems during reconciliation.
- A validation step which is split into two parts:
    - **Errors (migration blockers):** things the operator will override
    - **Warnings (manual steps required post-migration):** things the user must handle outside the CR such as creating TLS secrets and reviewing per-member config drift that will be flattened to a single source process.
- The connection layer reuses the same ConfigMap and Secret format the operator already uses (`baseUrl`, `orgId`, `projectName`, `publicKey`, `privateKey`). If there is already a running operator deployment for the project, no new credentials or configuration are needed to run the command. CR generation is deferred to the next PR in the stack.

## Checklist


filipcirtog:
* **#974** <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/974?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#973** <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/973?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/973?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#977** <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/977?utm_source=stack-comment-icon" target="_blank">Graphite</a>


<h2></h2>

cmd/kubectl-mongodb/migrate-to-mck/validation.go
```diff
@@ -0,0 +1,542 @@
+package migrate
+
+import (
+	"fmt"
+	"slices"
+
+	"github.com/mongodb/mongodb-kubernetes/controllers/om"
+	"github.com/mongodb/mongodb-kubernetes/controllers/operator/ldap"
```
nammn, line 9: does this import lead to a largely increased binary size of the cli?

cmd/kubectl-mongodb/migrate/connection.go
```diff
@@ -0,0 +1,141 @@
...
+	if config.ProjectName == "" {
+		return mdbv1.ProjectConfig{}, mdbv1.Credentials{}, fmt.Errorf("ConfigMap %s/%s does not contain a projectName", namespace, configMapName)
+	}
+	secretClient := secrets.SecretClient{KubeClient: kubeClient}
+	credentials, err := project.ReadCredentials(ctx, secretClient, kube.ObjectKey(namespace, secretName), log)
+	if err != nil {
+		return mdbv1.ProjectConfig{}, mdbv1.Credentials{}, fmt.Errorf("error reading credentials secret: %w", err)
+	}
+	return config, credentials, nil
+}
```
lucian-tosa: nit: there is already a function doing this? Why duplicate it?

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,536 @@
...
+	if agentSSL.CAFilePath != "" && agentSSL.CAFilePath != defaultCAFilePath {
+		results = append(results, ValidationResult{
+			Severity: SeverityError,
+			Message:  fmt.Sprintf("tls.CAFilePath %q differs from the operator default %q and will be overwritten.", agentSSL.CAFilePath, defaultCAFilePath),
+		})
+	}
+	if agentSSL.CAFilePath != "" {
+		results = append(results, ValidationResult{
+			Severity: SeverityWarning,
+			Message:  "TLS is enabled. Create a kubernetes.io/tls Secret named \"<certsSecretPrefix>-<resourceName>-cert\" with keys \"tls.crt\" and \"tls.key\" before applying the Custom Resource.",
```
lucian-tosa: Is the tls secret required for this or the CA configmap?

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,536 @@
...
+		host := m.Name()
+		proc, ok := processMap[host]
+		if !ok {
+			continue
+		}
+		dbPath := proc.DbPath()
+		if dbPath != "" && dbPath != util.PvcMountPathData {
+			return []ValidationResult{{
+				Severity: SeverityWarning,
+				Message:  fmt.Sprintf("Process %q has dbPath %q. The operator uses %q; the path will change when the member becomes operator-managed.", host, dbPath, util.PvcMountPathData),
```
lucian-tosa: will it change? Isn't this managed per process?

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,536 @@
...
+		sections := proc.NetTLSSections()
+
+		for tlsKey, tlsSection := range sections {
+			certKey, _ := tlsSection["certificateKeyFile"].(string)
+			pemKey, _ := tlsSection["PEMKeyFile"].(string)
+
+			if certKey != "" && certKey != util.PEMKeyFilePathInContainer {
+				results = append(results, ValidationResult{
+					Severity: SeverityError,
+					Message:  fmt.Sprintf("Process %q has net.%s.certificateKeyFile %q; the operator defaults to %q. The certificate path will change after migration.", name, tlsKey, certKey, util.PEMKeyFilePathInContainer),
```
lucian-tosa: this is also per process, we shouldn't validate it because we won't change it

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,536 @@
...
+
+func validateTLSPaths(d om.Deployment) []ValidationResult {
+	var results []ValidationResult
+	for _, proc := range d.GetProcesses() {
+		name := proc.Name()
+		sections := proc.NetTLSSections()
+
+		for tlsKey, tlsSection := range sections {
+			certKey, _ := tlsSection["certificateKeyFile"].(string)
+			pemKey, _ := tlsSection["PEMKeyFile"].(string)
```
lucian-tosa: `PEMKeyFile` doesn't exist in net.tls (as far as the documentation says). Where does this come from?

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,536 @@
...
+}
+
+func validateShardingClusterRole(d om.Deployment) []ValidationResult {
+	var results []ValidationResult
+	for _, proc := range d.GetProcesses() {
+		role := proc.ShardingClusterRole()
+		if role != "" {
+			results = append(results, ValidationResult{
+				Severity: SeverityError,
+				Message:  fmt.Sprintf("Process %q has sharding.clusterRole %q. Only standalone replica sets are supported.", proc.Name(), role),
```
lucian-tosa: "standalone" replica sets can sound confusing

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,536 @@
...
+		}
+	}
+	return results
+}
+
+func validateOneDeploymentPerProject(d om.Deployment) []ValidationResult {
+	if len(d.GetShardedClusters()) > 0 {
+		return []ValidationResult{{
+			Severity: SeverityError,
+			Message:  "Sharded clusters are not supported. Only standalone replica sets are supported.",
```
lucian-tosa: same as above

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,536 @@
...
+	if count <= 1 {
+		return nil
+	}
+	return []ValidationResult{{
+		Severity: SeverityError,
+		Message:  fmt.Sprintf("The project contains %d deployments. The operator requires exactly one deployment per project. Split the project before migrating.", count),
+	}}
+}
+
+func validateReplicaSetProtocolVersion(d om.Deployment) []ValidationResult {
```
lucian-tosa: nit: there is only one accepted value for protocol version, not sure if we need to validate it as well

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,536 @@
...
+	if len(d.GetReplicaSets()) == 0 {
+		return []ValidationResult{{
+			Severity: SeverityError,
+			Message:  "No replica sets found. Only replica set deployments can be migrated.",
+		}}
+	}
+	return nil
+}
+
+func validateMembersReferenceProcesses(d om.Deployment, processMap map[string]om.Process) []ValidationResult {
```
lucian-tosa: If I understand correctly, this validates whether a replicaset member in the AC has a corresponding process in the AC. Is this needed? I doubt OM will accept this AC

cmd/kubectl-mongodb/migrate-to-mck/validation.go
filipcirtog: As I have observed until now the size does not significantly. (less than 20%)

cmd/kubectl-mongodb/migrate/connection.go
filipcirtog: Thanks! Updated! I am now reusing it.

cmd/kubectl-mongodb/migrate/validation.go
filipcirtog: Both a secret and a configMap are required.

cmd/kubectl-mongodb/migrate/validation.go
filipcirtog: Agreed. Removed! Thank you!

cmd/kubectl-mongodb/migrate/validation.go
filipcirtog: Agreed. Removed the function. Thank you!

cmd/kubectl-mongodb/migrate/validation.go
filipcirtog: Updated the messages to say the following message. Thank you!

> Only replica set deployments can be migrated.

cmd/kubectl-mongodb/migrate/validation.go
filipcirtog: `PEMKeyFile` is the legacy field under `net.ssl,` which is now deprecated. NetTLSSections() returns both `net.tls` and `net.ssl `sections when present, so this handles deployments that still use the old net.ssl config.

cmd/kubectl-mongodb/migrate/validation.go
filipcirtog: This is true. Dropping these assertions. Thank you!

cmd/kubectl-mongodb/migrate/validation.go
filipcirtog: This is true. Dropping these assertions. Thank you!

cmd/kubectl-mongodb/migrate-to-mck/connection.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	if proj == nil {
+		return nil, fmt.Errorf("project %q not found in organization %s (%q)", config.ProjectName, org.ID, org.Name)
+	}
+
+	conn.ConfigureProject(proj)
+	return conn, nil
+}
+
+func newKubeClient() (kubernetesClient.Client, error) {
```
lucian-tosa, line 94: there is already a GetKubernetesClient method used in the setup subcommand of the plugin. Why is this needed?

cmd/kubectl-mongodb/migrate/users.go
```diff
@@ -0,0 +1,63 @@
...
+	RunE: runGenerateUsers,
+}
+
+func init() {
+	UsersCmd.Flags().StringVar(&uFlags.configMapName, "config-map-name", "", "Name of the ConfigMap containing the OM connection details (baseUrl, orgId, projectName) (required)")
+	UsersCmd.Flags().StringVar(&uFlags.secretName, "secret-name", "", "Name of the Secret containing the OM API credentials (publicKey, privateKey) (required)")
+	UsersCmd.Flags().StringVar(&uFlags.namespace, "namespace", defaultNamespace, "Namespace of the ConfigMap and Secret")
+	UsersCmd.Flags().StringVarP(&uFlags.outputFile, "output", "o", "", "Write generated CRs to this file instead of stdout")
+	UsersCmd.Flags().StringVar(&uFlags.usersSecretsFile, "users-secrets-file", "", "CSV file mapping 'username:database,secret-name' for SCRAM users. When provided, customer-owned Secrets are referenced instead of generated and interactive prompts for user passwords are suppressed")
+	UsersCmd.Flags().StringVar(&uFlags.resourceNameOverride, "resource-name-override", "", "Name of the MongoDB CR that users will reference (mongodbResourceRef.name). Defaults to the normalized replica set name from the automation config")
```
lucian-tosa: This could be confused with the name of the MongoDBUser CR.

cmd/kubectl-mongodb/migrate-to-mck/validation.go
```diff
@@ -0,0 +1,425 @@
...
+			Severity: SeverityError,
+			Message:  fmt.Sprintf("auth.autoUser %q has no matching entry in auth.usersWanted (database: %q). Agent authentication will fail after migration.", auth.AutoUser, util.DefaultUserDatabase),
+		}}
+	}
+	return nil
+}
+
+// validateX509 warns when MONGODB-X509 agent auth is configured.
+func validateX509(auth *om.Auth) []ValidationResult {
+	if auth == nil || auth.Disabled {
```
lucian-tosa, line 179: if the scram validation checks id auto user has a matching user, doesn't this need it as well?

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,425 @@
...
+	if agentSSL == nil {
+		return nil
+	}
+
+	var results []ValidationResult
+
+	if agentSSL.AutoPEMKeyFilePath != "" {
+		results = append(results, ValidationResult{
+			Severity: SeverityError,
+			Message:  fmt.Sprintf("tls.autoPEMKeyFilePath %q will be overwritten by the operator during reconciliation.", agentSSL.AutoPEMKeyFilePath),
```
lucian-tosa: this isn't true anymore

cmd/kubectl-mongodb/migrate/validation.go
```diff
@@ -0,0 +1,425 @@
...
+	for _, rs := range d.GetReplicaSets() {
+		rsID := rs.Name()
+
+		for _, m := range rs.Members() {
+			host := m.Name()
+
+			if delay := m.SecondaryDelaySecs(); delay > 0 {
+				results = append(results, ValidationResult{
+					Severity: SeverityWarning,
+					Message:  fmt.Sprintf("Member %q (replica set %q) has secondaryDelaySecs %d. This is preserved while the member is external and is lost when operator-managed.", host, rsID, delay),
```
lucian-tosa: The comment is not very clear. Maybe say something like: "This field will not be applied to new Kubernetes members"

cmd/kubectl-mongodb/migrate/validation.go
lucian-tosa: Indeed they are required for TLS. I think the comment is fine. But functionality wise, the TLS secret is not used to populate the CAFilePath field in the AC

cmd/kubectl-mongodb/migrate/validation.go
filipcirtog: Updated. Thank you!


`Message:  fmt.Sprintf("Member %q (replica set %q) has secondaryDelaySecs %d. This setting is not supported by the operator and will not be applied to new Kubernetes members managed by the operator.", host, rsID, delay)`

cmd/kubectl-mongodb/migrate/validation.go
filipcirtog: Updated. Thank you!

cmd/kubectl-mongodb/migrate-to-mck/validation.go
filipcirtog: The function now includes the autoUser matching check for $external database mirroring what validateScram does. Thank you!

cmd/kubectl-mongodb/migrate/users.go
filipcirtog: Updated to `Name of the MongoDB replica set CR that the generated MongoDBUser resources will reference via mongodbResourceRef.name.` Thank you!

cmd/kubectl-mongodb/migrate-to-mck/connection.go
filipcirtog: `common.GetKubernetesClient` returns a `KubeClient` (built on kubernetes.Clientset), but `project.ReadConfigAndCredentials` requires a `kubernetesClient.Client `from controller-runtime. They are different types and cannot be substituted.

cmd/kubectl-mongodb/migrate-to-mck/validation.go
```diff
@@ -0,0 +1,432 @@
...
+
+// validateAgentTLS checks project-level agent TLS paths against operator defaults.
+func validateAgentTLS(agentSSL *om.AgentSSL) []ValidationResult {
+	if agentSSL == nil {
+		return nil
+	}
+
+	var results []ValidationResult
+
+	if agentSSL.CAFilePath != "" && agentSSL.CAFilePath != defaultCAFilePath {
```
lucian-tosa, line 212: not needed anymore right?

cmd/kubectl-mongodb/migrate-to-mck/validation.go
```diff
@@ -0,0 +1,432 @@
...
+		})
+	}
+
+	return results
+}
+
+// validateProjectOptions checks options.downloadBase against the operator default.
+func validateProjectOptions(d om.Deployment) []ValidationResult {
+	downloadBase := d.DownloadBase()
+	if downloadBase == "" || downloadBase == defaultDownloadBase {
```
lucian-tosa, line 286: Also not needed

cmd/kubectl-mongodb/migrate-to-mck/validation.go
filipcirtog: we need to merge those first.

cmd/kubectl-mongodb/migrate-to-mck/validation.go
filipcirtog: same as the comment above. we need to merge those first, I will remove them then and add tests for it.

PR #974, opened 2026-04-07
feat: implement MongoDB and MongoDBUser CR generation for kubectl-mongodb migrate

filipcirtog:
# Based on PR #973

* PR #973:
  `graphite-base/973` ← `04-07-feat_kubectl-mongodb_migrate_command_connection_validation`

  * **PR #974 (THIS ONE)**:
    `04-07-feat_kubectl-mongodb_migrate_command_connection_validation` ← `04-07-feat_ac-to-cr_spec_generation_logic`

This PR delivers the generation step for the `kubectl-mongodb migrate-to-mck` command. Given a validated Automation Config, both the `mongodb` and `users` subcommands now produce a complete, apply-ready YAML bundle that the operator can consume without manual editing. Nothing is applied to the cluster directly. The output is written to stdout or to a file via `--output` and the apply step remains the responsibility of the user.
**Why this PR is needed?**
Migrating a MongoDB replica set from Ops Manager to the Kubernetes operator requires manually reconstructing a CR from an Automation Config. The migration workflow requires generating two distinct resource types: a `MongoDB` CR from the replica set configuration and `MongoDBUser` CRs from the users in the Automation Config. These are split into separate subcommands because some users may choose to migrate only the MongoDB resource and manage users separately.
**What this PR does?**
- `kubectl-mongodb migrate-to-mck mongodb` generates a `MongoDB` CR along with any supporting ConfigMaps for LDAP CA and Prometheus TLS. All existing replica set members are placed in `spec.externalMembers` so the operator treats them as VM-managed during the initial handover.
- `kubectl-mongodb migrate-to-mck users` generates `MongoDBUser` CRs with password Secrets for all SCRAM users and the operator-managed `mongodbResourceRef` already populated.
    - **Password collection and verification:** The operator manages users through `MongoDBUser` CRs that reference a Kubernetes Secret holding the plaintext password. Since existing OM users have their credentials stored only as SCRAM hashes in the Automation Config with no API to recover the original plaintext, every supplied password is verified against the AC hashes before any output is produced. An incorrect password would otherwise cause the operator to silently overwrite the existing SCRAM hashes on its first reconciliation, invalidating all clients using those credentials.
    - For interactive environments the command prompts for each SCRAM user secret name at generation time. For non-interactive environments such as CI pipelines or GitOps workflows, the `--users-secrets-file` flag accepts a CSV mapping each user to a pre-created Kubernetes Secret. Referenced Secrets are validated against the AC hashes before being included in the generated CRs, providing the same correctness guarantee without relying on stdin. TLS and Prometheus credential collection follows the same dual-mode pattern: interactive prompts for local use and flag-supplied Secret references for automated pipelines.
## Checklist


filipcirtog:
* **#974** <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/974?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/974?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#973** <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/973?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#977** <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/977?utm_source=stack-comment-icon" target="_blank">Graphite</a>


<h2></h2>

cmd/kubectl-mongodb/migrate/command.go
```diff
@@ -45,37 +60,377 @@ func init() {
...
+		if err != nil {
+			return "", fmt.Errorf("failed to marshal %T: %w", obj, err)
+		}
+		sb.WriteString(y)
+	}
+	return sb.String(), nil
+}
+
+// generateExtraResources returns the supporting Secrets/ConfigMaps for LDAP and Prometheus.
+func generateExtraResources(ac *om.AutomationConfig, opts GenerateOptions) []client.Object {
```
lucian-tosa: nit: this doesn't seem to belong here, generator.go might be a better place

cmd/kubectl-mongodb/migrate/command.go
```diff
@@ -45,37 +60,377 @@ func init() {
...
+		if err != nil {
+			return fmt.Errorf("failed to read certsSecretPrefix: %w", err)
+		}
+		prefix = p
+	}
+	if prefix == "" {
+		return fmt.Errorf("security.certsSecretPrefix is required when TLS is enabled; use --certs-secret-prefix or enter it interactively")
+	}
+	if len(k8svalidation.IsDNS1123Subdomain(prefix)) > 0 {
+		return fmt.Errorf("security.certsSecretPrefix %q is not a valid Kubernetes resource name", prefix)
```
lucian-tosa: Would it be a better UX to allow retrying the input? If we fail, they have to start all over

cmd/kubectl-mongodb/migrate/spec_builder.go
```diff
@@ -0,0 +1,379 @@
...
+		return util.X509, nil
+	case "keyFile":
+		return "", fmt.Errorf("clusterAuthMode %q is not supported by the operator (only x509 is supported). Migrate the deployment to x509 internal cluster authentication before using the operator", mode)
+	default:
+		return "", fmt.Errorf("unsupported clusterAuthMode %q in automation config. Only x509 is supported by the operator", mode)
+	}
+}
+
+// isTLSEnabled returns true if any member has an explicit TLS section with a non-disabled mode.
+func isTLSEnabled(processMap map[string]om.Process, members []om.ReplicaSetMember) (bool, error) {
```
lucian-tosa: why do we need the replica set members here? We can find this out only from the processes

cmd/kubectl-mongodb/migrate/command.go
```diff
@@ -45,37 +60,377 @@ func init() {
...
+	}
+	return strings.TrimSpace(scanner.Text()), nil
+}
+
+func ensureTLS(ac *om.AutomationConfig, opts *GenerateOptions, scanner *bufio.Scanner, certsSecretPrefix string) error {
+	rss := ac.Deployment.GetReplicaSets()
+	if len(rss) == 0 {
+		return nil
+	}
+	tlsEnabled, err := isTLSEnabled(ac.Deployment.ProcessMap(), rss[0].Members())
```
lucian-tosa: navigating this code is a bit convoluted. I suggest reorganizing the functions. `isTLSEnabled` is in spec_builder.go, but only called here

cmd/kubectl-mongodb/migrate/generator.go
```diff
@@ -0,0 +1,318 @@
...
+		return nil, "", fmt.Errorf("no replica sets found in the automation config")
+	}
+	rs := replicaSets[0]
+
+	rsName := rs.Name()
+	externalMembers, version, fcv := om.ExtractMemberInfo(rs.Members(), ac.Deployment.ProcessMap())
+
+	resourceName := opts.ReplicaSetNameOverride
+	if resourceName == "" {
+		if userv1.NormalizeName(rsName) != rsName {
```
lucian-tosa: Isn't this from the mdb community package?

cmd/kubectl-mongodb/migrate/generator.go
```diff
@@ -0,0 +1,318 @@
...
+			return nil, "", fmt.Errorf("replica set name %q is not a valid Kubernetes resource name. Use --replicaset-name-override to provide a valid name (spec.replicaSetNameOverride will be set automatically)", rsName)
+		}
+		resourceName = rsName
+	} else if userv1.NormalizeName(resourceName) != resourceName {
```
lucian-tosa: It would be simpler/more readable if we just check if the string is a valid DNS name

cmd/kubectl-mongodb/migrate/generator.go
lucian-tosa: same as below. Just use `IsDNS1123Subdomain`

cmd/kubectl-mongodb/migrate/generator.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+	rs := replicaSets[0]
+
+	rsName := rs.Name()
+	externalMembers, version, fcv := om.ExtractMemberInfo(rs.Members(), ac.Deployment.ProcessMap())
+
+	resourceName := opts.ReplicaSetNameOverride
+	if resourceName == "" {
+		if userv1.NormalizeName(rsName) != rsName {
+			return nil, "", fmt.Errorf("replica set name %q is not a valid Kubernetes resource name. Use --replicaset-name-override to provide a valid name (spec.replicaSetNameOverride will be set automatically)", rsName)
```
lucian-tosa: I think it's worth considering if we can normalize the replicaset name, use it as resource name, and set replicaSetNameOverride to it. Thus we would not have to force the user to set this flag.
Example: "myReplicaSet_main" -> MongoDB "myReplicaSet-main" and `spec.replicaSetNameOverride: myReplicaSet_main"

cmd/kubectl-mongodb/migrate/command.go
```diff
@@ -45,37 +60,377 @@ func init() {
 	MigrateCmd.Flags().StringVar(&flags.multiClusterNames, "multi-cluster-names", "", "Comma-separated list of target cluster names (e.g., east1,west1); generates a MongoDBMultiCluster CR")
 	MigrateCmd.Flags().StringVarP(&flags.outputFile, "output", "o", "", "Write generated CRs to this file instead of stdout")
 	MigrateCmd.Flags().StringVar(&flags.replicaSetNameOverride, "replicaset-name-override", "", "Kubernetes resource name for the generated CR; required when the replica set name is not a valid Kubernetes name (sets spec.replicaSetNameOverride automatically)")
```
lucian-tosa: as I wrote in the TD, this would be better called `resource-name-override`

cmd/kubectl-mongodb/migrate/generator.go
```diff
@@ -0,0 +1,318 @@
...
+
+func generateReplicaSet(ac *om.AutomationConfig, opts GenerateOptions) (client.Object, string, error) {
+	replicaSets := ac.Deployment.GetReplicaSets()
+	if len(replicaSets) == 0 {
+		return nil, "", fmt.Errorf("no replica sets found in the automation config")
+	}
+	rs := replicaSets[0]
+
+	rsName := rs.Name()
+	externalMembers, version, fcv := om.ExtractMemberInfo(rs.Members(), ac.Deployment.ProcessMap())
```
lucian-tosa: where is `ExtractMemberInfo` implemented?

cmd/kubectl-mongodb/migrate/spec_builder.go
```diff
@@ -0,0 +1,379 @@
...
+
+// buildReplicaSetSpec assembles the MongoDbSpec for a single-cluster replica set.
+func buildReplicaSetSpec(version, fcv string, externalMembers []mdbv1.ExternalMember, rsName, resourceName string, opts GenerateOptions, ac *om.AutomationConfig) (mdbv1.MongoDbSpec, error) {
+	common, err := buildCommonSpec(ac, version, fcv, rsName, resourceName, externalMembers, opts)
+	if err != nil {
+		return mdbv1.MongoDbSpec{}, err
+	}
+	return mdbv1.MongoDbSpec{
+		DbCommonSpec: common,
+		Members:      len(externalMembers),
```
lucian-tosa: Members defines the amount of kube pods we deploy. If they have a replicaset with 50 members, we will immediately deploy 50 pods. This might not be a good idea. We should set a sensible default, or even set it to 0 and let the user manage the expansion

cmd/kubectl-mongodb/migrate/spec_builder.go
```diff
@@ -0,0 +1,379 @@
...
+		MemberConfig: buildMemberConfig(ac.Deployment.GetReplicaSets()[0].Members()),
```
lucian-tosa: Same here. We don't want to copy the member configs from existing rs members. Ideally, customers should expand the RS with non-voting kube members.

cmd/kubectl-mongodb/migrate/spec_builder.go
```diff
@@ -0,0 +1,379 @@
...
+	certsSecretPrefix string,
+) (*mdbv1.Security, error) {
+	security := &mdbv1.Security{}
+	hasSettings := false
+
+	if certsSecretPrefix != "" {
+		// certsSecretPrefix enables TLS (tls.enabled is deprecated, see RELEASE_NOTES_MEKO.md 1.15).
+		security.CertificatesSecretsPrefix = certsSecretPrefix
+		hasSettings = true
+	}
```
lucian-tosa: where do we set the CA configmap?

cmd/kubectl-mongodb/migrate/spec_builder.go
```diff
@@ -0,0 +1,379 @@
...
+		}
+	}
+	return "", nil
+}
+
+func mapClusterAuthMode(mode string) (string, error) {
+	switch mode {
+	case "x509":
+		return util.X509, nil
+	case "keyFile":
```
lucian-tosa: this is the default, we support it

lucian-tosa: We just don't set it explicitly. As long as auth is enabled, keyfile authentication is used if x509 is not explicitly set

cmd/kubectl-mongodb/migrate/command.go
filipcirtog: Done! Moved generateExtraResources to generator.go.

cmd/kubectl-mongodb/migrate/command.go
filipcirtog: Done. `isTLSEnabled` is now in `mongodb.go` since it is only called from `ensureTLS` there.

cmd/kubectl-mongodb/migrate/command.go
filipcirtog: Done. If `--certs-secret-prefix` (flag) is provided but invalid, the command now returns an error immediately. The interactive retry loop runs when the flag is absent and the input provided is invalid.

filipcirtog: (same for the other cases in which we have the same logic)

cmd/kubectl-mongodb/migrate/spec_builder.go
filipcirtog: I have simplified the function signature to only take `processMap` without the members parameter. Thank you!

cmd/kubectl-mongodb/migrate/generator.go
filipcirtog: Yes, but both of them used the same code. Now I have changed and we rely on the `util.NormalizeName` from the utilities package. I have then updated the community and enterprise users to also re-use the same function from the utils.

cmd/kubectl-mongodb/migrate/generator.go
filipcirtog: Switched to using `k8svalidation.IsDNS1123Subdomain(resourceName).` Thank you!

cmd/kubectl-mongodb/migrate/command.go
filipcirtog: Done. The flag is now --resource-name-override.

cmd/kubectl-mongodb/migrate/generator.go
filipcirtog: It's in controllers/om/replicaset.go.

cmd/kubectl-mongodb/migrate/spec_builder.go
filipcirtog: Done. Members is set to 0. The operator starts with no new Kubernetes-managed pods. Existing VM nodes are tracked as externalMembers. The customer expands the replica set incrementally when ready.

cmd/kubectl-mongodb/migrate/spec_builder.go
filipcirtog: Done. MemberConfig is no longer set in the generated CR.

cmd/kubectl-mongodb/migrate/spec_builder.go
filipcirtog: Done. keyFile maps to an empty string. Thank you!

cmd/kubectl-mongodb/migrate/spec_builder.go
filipcirtog: It's set in buildSecurity now:
`security.TLSConfig = &mdbv1.TLSConfig{CA: fmt.Sprintf("%s-ca", resourceName)},` following the operator's default CA ConfigMap naming convention.

I have also updated the warning: `TLS is enabled. Create a ConfigMap with key \"ca-pem\" containing the CA certificate, and a kubernetes.io/tls Secret named \"<certsSecretPrefix>-<resourceName>-cert\" with keys \"tls.crt\" and \"tls.key\" before applying the Custom Resource. The generated CR sets security.tls.ca to \"<resourceName>-ca\" by default. Change it if your ConfigMap has a different name.`

cmd/kubectl-mongodb/migrate/generator.go
filipcirtog: Done. `util.NormalizeName` auto-normalizes the RS name to a valid Kubernetes name. When the normalized name differs from the original, spec.replicaSetNameOverride is set to the original RS name so the operator can locate the existing replica set.

api/v1/mdb/mongodb_types.go
```diff
@@ -1481,38 +1481,91 @@ func (m *MongoDB) ObjectKey() client.ObjectKey {
...
+	// Fields requiring custom conversion logic.
+	{
+		func(c *Ldap, a *ldap.Ldap) {
+			a.ValidateLDAPServerConfig = true
+			if c.ValidateLDAPServerConfig != nil {
+				a.ValidateLDAPServerConfig = *c.ValidateLDAPServerConfig
+			}
+		},
+		func(a *ldap.Ldap, c *Ldap) { c.ValidateLDAPServerConfig = &a.ValidateLDAPServerConfig },
+	},
```
lucian-tosa: Shouldn't this be passed as arguments to `ldapField`?

cmd/kubectl-mongodb/migrate-to-mck/common_spec.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	list := make(mdbv1.ClusterSpecList, len(clusterNames))
+	for i, name := range clusterNames {
+		list[i] = mdbv1.ClusterSpecItem{
+			ClusterName: name,
+			Members:     0,
+		}
+	}
+	return list
+}
```
lucian-tosa: Is this called anywhere?

PR #800, opened 2026-02-17
wip - migration feature branch 

nammn:
# Summary

This pull request introduces support for two new fields, `externalMembers` and `replicaSetNameOverride`, in the MongoDB custom resource definitions (CRDs) and associated Go types, and updates the handling of replica set merging logic to accommodate these changes. The PR also adds a utility method for agent authentication and updates test utilities and test cases to align with the new method signatures and structures.

**Key changes:**

### API and CRD Enhancements

* Added `externalMembers` (array of strings) and `replicaSetNameOverride` (string) fields to the `DbCommonSpec` Go struct and to the MongoDB and MongoDBMultiCluster CRDs, allowing users to specify external replica set members and override the replica set name. [[1]](diffhunk://#diff-412c6fb7d7d5e0a038c09d2d0b1a9dc1996fb1f107a58abfca060caf39427287R434-R437) [[2]](diffhunk://#diff-0eed4e61b4f19c1bdb4acbf8e9a6174b974fd4f251e980f716091ec292099f2eR858-R861) [[3]](diffhunk://#diff-0eed4e61b4f19c1bdb4acbf8e9a6174b974fd4f251e980f716091ec292099f2eR1403-R1404) [[4]](diffhunk://#diff-4fb58ed5f7878f03c31c5c851a264c660e771eda2b496cc20f366b566a78f584R591-R594) [[5]](diffhunk://#diff-4fb58ed5f7878f03c31c5c851a264c660e771eda2b496cc20f366b566a78f584R663-R664)

### Deep Copy and Utility Methods

* Updated the autogenerated `DeepCopyInto` method for `DbCommonSpec` to correctly handle the new `externalMembers` field.
* Added a `GetAutomationUserName` method to `AgentAuthentication` to provide a default username if none is specified.

### Deployment and Replica Set Merge Logic

[...]

PR #976, opened 2026-04-07
test: vm-migration to kubernetes replicaset CI coverage 

filipcirtog:
# Based on PR #975

## Chain of upstream PRs as of 2026-06-03

* PR #973:
  `graphite-base/973` ← `04-07-feat_kubectl-mongodb_migrate_command_connection_validation`

  * PR #974:
    `04-07-feat_kubectl-mongodb_migrate_command_connection_validation` ← `04-07-feat_ac-to-cr_spec_generation_logic`

    * PR #975:
      `04-07-feat_ac-to-cr_spec_generation_logic` ← `04-07-test_golden_testdata_for_migration_spec_generation`

      * **PR #976 (THIS ONE)**:
        `04-07-test_golden_testdata_for_migration_spec_generation` ← `04-07-test_e2e_tests_and_evergreen_ci_for_vm_migration`

This PR adds an E2E test suite that exercises the complete VM-to-Kubernetes migration workflow against a live Ops Manager project, along with the Evergreen CI configuration to run it.

**Why this change is needed?**

The migration tool has been validated through unit and golden file tests, but those tests operate against fixture JSON. They do not exercise the real path: a live Ops Manager project, an agent actually running on a VM StatefulSet, a generated CR being applied and reconciled by the operator.

**What this PR does?**

Each test deploys a VM-based replica set via a Kubernetes StatefulSet running the MongoDB agent in automation mode, configures it through the Ops Manager API, runs `kubectl-mongodb migrate mongodb` and `kubectl-mongodb migrate users` against it, applies the generated CRs, and asserts the operator reaches `Running`. Where users are present, the tests apply the generated password Secrets and `MongoDBUser` CRs, assert each user reaches `Phase.Updated`, and verify that both `scramSha256Creds` and `scramSha1Creds` are present in the Automation Config. Password rotation is tested by updating the referenced Kubernetes Secret and polling the Automation Config until the `scramSha256` salt changes, then asserting both credential types are still present.

- `vm_migration_helpers.py` centralises StatefulSet deployment, migrate invocation, promote-and-prune lifecycle, Secret application, user CR assertion, and password rotation logic. This is needed because every test scenario follows the same structure and the duplication would otherwise obscure the per-scenario differences.

The scenarios covered are:

- `vm_migration_generate_scram_full.py`: SCRAM-SHA-256 with custom roles, audit logging, log rotation, member tags, and FCV. Passwords are provided to the migrate tool via stdin.
[...]

PR #975, opened 2026-04-07
test: add fixture and unit tests for kubectl-mongodb migrate

filipcirtog:
Based on PR #974
Chain of upstream PRs as of 2026-06-03
- PR #973:
graphite-base/973 ← 04-07-feat_kubectl-mongodb_migrate_command_connection_validation
- PR #974:
04-07-feat_kubectl-mongodb_migrate_command_connection_validation ← 04-07-feat_ac-to-cr_spec_generation_logic
- PR #975 (THIS ONE):
04-07-feat_ac-to-cr_spec_generation_logic ← 04-07-test_golden_testdata_for_migration_spec_generation

This PR adds fixture-based golden tests for the kubectl mongodb migrate-to-mck generation path.

Why this change is needed?  
The migration generator translates Ops Manager Automation Configs into Kubernetes YAML. The output is nested and combines multiple configuration areas such as authentication, TLS, LDAP, OIDC, Prometheus, agent settings, external members, and additional mongod config. Golden fixtures make those generated YAML documents reviewable as complete user-facing output, instead of only validating scattered individual fields.
What this PR does?
Adds fixture tests for migration spec generation.
- Adds Automation Config JSON input fixtures under cmd/kubectl-mongodb/migrate-to-mck/testdata/.
- Adds committed YAML golden files for generated MongoDB and MongoDBUser resources.
- Exercises the generator output path used by the CLI so the fixtures match the YAML users would receive from the command.
- Covers replica set scenarios including complex auth/TLS/LDAP/OIDC/Prometheus config, custom mongod config, TLS modes, disabled auth, SCRAM users, X.509 users, and member options.
- Supports fixture regeneration with go test -run TestFixtureMatch -update-fixtures.

Checklist
- Have you linked a jira ticket and/or is the ticket in the title?
- Have you checked whether your jira ticket required DOCSP changes?
- Have you added changelog file?
- use skip-changelog label if not needed
- refer to Changelog files and Release Notes (https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md (http://CONTRIBUTING.md) for more details

PR #977, opened 2026-04-07
feat: reset OM user mechanisms on VM migration

filipcirtog:
# Based on PR #800

* PR #800:

  * **PR #977 (THIS ONE)**:
    `vm-migration-feature-branch` ← `04-07-feat_operator_additions_for_migration_logic`

This PR updates the operator's SCRAM credential pipeline to correctly handle users that already exist in the Automation Config with mechanisms set. This is the case that arises when migrating from VM deployments.

**Why this change is needed?**

The previous `ConfigureScramCredentials` assumed it owned both SHA-1 and SHA-256 credentials for every user. For users imported from a VM deployment where Ops Manager already owns their credentials and mechanisms, this caused two problems.

- If only one algorithm was present, the other was silently injected, which the OM API rejects without the following `mechanism` being set.
- The`mechanisms` array was also overwritten with an empty slice, which OM rejects when mechanisms are already set.

**What this PR does?**

- `ConfigureScramCredentials` now handles three cases:
    - **New users (not in AC):** generate both SHA-256 and SHA-1 with new salts. Unchanged behaviour.
    - **Imported users (mechanisms set in AC):** validate the supplied password matches existing credentials, copy only the algorithms already present, and set `InitPassword`. This causes OM to empty the mechanisms field so the operator can finalise credentials on the follow-up reconcile. Returns `needsFollowUp=true` to trigger that requeue.
    - **K8s-managed users (in AC, no mechanisms):** check each algorithm independently and only regenerate the one whose credentials no longer match the current password.
- `isPasswordChanged` is replaced by `isCredChanged`, which checks a single algorithm at a time, fixing the false-positive regeneration when only one credential type was present.
- **E2E test coverage:** Extended SCRAM connectivity tests with imported-user scenarios and a new `kubetester/scram.py` helper for seeding pre-computed credentials in OM and asserting post-migration state.

## Checklist


filipcirtog:
* **#974** <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/974?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#973** <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/973?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#977** <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/977?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mongodb/mongodb-kubernetes/977?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>


<h2></h2>

controllers/operator/authentication/oidc.go
```diff
@@ -114,6 +114,35 @@ func MapOIDCProviderConfigs(oidcProviderConfigs []mdbv1.OIDCProviderConfig) []oi
 	return result
 }
 
+// MapACOIDCToProviderConfigs converts AC OIDC provider configs to the CR's
+// OIDCProviderConfig representation. This is the reverse of MapOIDCProviderConfigs.
+func MapACOIDCToProviderConfigs(configs []oidc.ProviderConfig) []mdbv1.OIDCProviderConfig {
```
filipcirtog: from @nammn on https://github.com/mongodb/mongodb-kubernetes/pull/972  
same here

controllers/operator/authentication/authentication_mechanism.go
```diff
@@ -123,6 +123,26 @@ func convertToMechanismOrPanic(mechanismModeInCR string, ac *om.AutomationConfig
 	panic(xerrors.Errorf("unknown mechanism name %s", mechanismModeInCR))
 }
 
+// MapMechanismToAuthMode converts an automation config mechanism string to
+// the corresponding CR AuthMode. This is the reverse of convertToMechanismOrPanic.
+func MapMechanismToAuthMode(mech string) (string, bool) {
```
filipcirtog: from @nammn on https://github.com/mongodb/mongodb-kubernetes/pull/972  
  
lets refactor this and the convertToMechanismOrPanic to use a shared map instead

that should make things cleaner

controllers/om/replicaset.go
```diff
@@ -338,6 +346,31 @@ func findDifference(leftMap map[string]ReplicaSetMember, rightMap map[string]Rep
 	return ans
 }
 
+// ExtractMemberInfo reads version, FCV, and per-member metadata from the
+// given replica set members and process map. Each member becomes an
+// mdbv1.ExternalMember suitable for the CR's spec.externalMembers.
+func ExtractMemberInfo(members []ReplicaSetMember, processMap map[string]Process) ([]mdbv1.ExternalMember, string, string) {
+	firstProc := processMap[members[0].Name()]
```
filipcirtog: from @nammn on https://github.com/mongodb/mongodb-kubernetes/pull/972  
  
just to be safe; members can never be an empty slice, right?

controllers/om/process.go
```diff
@@ -299,9 +299,86 @@ func (p Process) ProcessType() MongoType {
...
+func (p Process) AuthSchemaVersion() int {
+	return cast.ToInt(p["authSchemaVersion"])
+}
+
+// SystemLogMap returns the systemLog sub-map from args2_6, or nil if absent.
+func (p Process) SystemLogMap() map[string]interface{} {
+	return maputil.ReadMapValueAsMap(p.Args(), "systemLog")
+}
+
+// AdditionalMongodConfig is the inverse of WithAdditionalMongodConfig: it reads args2_6 back
```
filipcirtog: from @nammn on https://github.com/mongodb/mongodb-kubernetes/pull/972  
  
can you elaborate on this comment? doesn't seem apparent to me what this means  
i would prefer if comment on why we need this

controllers/om/deployment.go
```diff
@@ -858,6 +880,21 @@ func (d Deployment) GetAllProcessNames() (names []string) {
...
+func (d Deployment) GetProcesses() []Process {
+	return d.getProcesses()
+}
+
+// ProcessMap returns a map of process name → Process for O(1) lookups.
+func (d Deployment) ProcessMap() map[string]Process {
+	processes := d.getProcesses()
+	m := make(map[string]Process, len(processes))
+	for _, p := range processes {
+		m[p.Name()] = p
```
filipcirtog: from @nammn on https://github.com/mongodb/mongodb-kubernetes/pull/972  
  
the name must be unique, right?

controllers/om/backup_agent_config.go
```diff
@@ -79,6 +81,29 @@ func (bac *BackupAgentConfig) UnsetLdapGroupDN() {
 	bac.BackupAgentTemplate.LdapGroupDN = util.MergoDelete
 }
 
+// ReadLogRotate extracts logRotate from the BackingMap and returns it as a
```
filipcirtog: from @nammn on https://github.com/mongodb/mongodb-kubernetes/pull/972  
  
we don't seem to use this function here, I assume we use this in further down the line stacks, i think we should try to group those together, wdyt?

api/v1/user/mongodbuser_types.go
```diff
@@ -85,6 +85,7 @@ type MongoDBResourceRef struct {
 	Namespace string `json:"namespace,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="!(has(self.migratedFromVm) && self.migratedFromVm && (!has(oldSelf.migratedFromVm) || !oldSelf.migratedFromVm))",message="migratedFromVm cannot be set to true after initial creation"
```
filipcirtog: from @nammn on https://github.com/mongodb/mongodb-kubernetes/pull/972  
  
do we inline validations like this in the code somewhere else? I personally wouldn't like this due to readability and missing code validations.  
  
is self based on the prior etcd resource version?

api/v1/mdb/mongodb_types.go
```diff
@@ -1537,6 +1537,46 @@ func GetTransportSecurity(mdbLdap *Ldap) TransportSecurity {
 	return transportSecurity
 }
 
+// ConvertACLdapToCR converts an AC LDAP config to the CR representation.
+// This is the reverse of MongoDB.GetLDAP. Callers must create the
```
nammn: like in the other comments, lets consider alternativse to reverse methods spread around the code base. Some ideas:

- a map that is co-located
- some logic that auto reverses
- probably more

api/v1/user/mongodbuser_types.go
```diff
@@ -95,6 +96,11 @@ type MongoDBUserSpec struct {
 	PasswordSecretKeyRef SecretKeyRef `json:"passwordSecretKeyRef"`
 	// +optional
 	ConnectionStringSecretName string `json:"connectionStringSecretName"`
+	// MigratedFromVM preserves only the original SCRAM mechanisms from the OM
```
nammn: could the customer generate a resource with the tool and set migratedFromVM to false? What about relying on externalMembers instead? Do we have a section in the TD on the API changes? We should discuss this part especially

controllers/om/backup_agent_config.go
nammn: i am fine with linking the further usage in the stack here as a link in this thread

controllers/om/omclient.go
```diff
@@ -259,10 +261,34 @@ func (oc *HTTPOmConnection) UpdateProcessLogRotation(logRotateSetting automation
 	return oc.put(fmt.Sprintf("/api/public/v1.0/groups/%s/automationConfig/systemLogRotateConfig", oc.GroupID()), logRotateSetting)
 }
 
+func (oc *HTTPOmConnection) ReadProcessLogRotation() (*automationconfig.AcLogRotate, error) {
```
nammn: q: why do we read it from the ac as well as from the api?

api/v1/user/mongodbuser_types.go
filipcirtog: after pruning is done, we do not have any external members anymore, but the `MongoDBUsers` still need to be handled differently. (i.e. populate mechanism array, only set SHA1/SHA256). I will develop more on this.

controllers/om/omclient.go
filipcirtog: The AC populates this per-process and by calling the API we get one for the project, similar to the backup and monitoring endpoints. We could instead use these two from the first process as we source the additionalMongodConfig. WDYT?

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_mechanisms.py
```diff
@@ -0,0 +1,249 @@
+"""
+E2E tests for SCRAM mechanisms preservation behaviour.
```
nammn: should we have a dedicated test for this or merge this with our existing migration tests?

controllers/om/omclient.go
nammn: i still don't fully understand why we can't just rely on the project one for instance.

controllers/om/process.go
```diff
@@ -299,9 +299,86 @@ func (p Process) ProcessType() MongoType {
...
+
+// SystemLogMap returns the systemLog sub-map from args2_6, or nil if absent.
+func (p Process) SystemLogMap() map[string]interface{} {
+	return maputil.ReadMapValueAsMap(p.Args(), "systemLog")
+}
+
+// AdditionalMongodConfig recovers user-supplied mongod config from args2_6 during VM migration.
+// Operator-owned fields (systemLog, storage.dbPath, replication.replSetName, default net.port,
+// TLS cert paths and default mode) are stripped. Returns nil when nothing user-relevant remains.
+func (p Process) AdditionalMongodConfig() *mdbv1.AdditionalMongodConfig {
```
nammn: Is and will this function only be used for migration? If yes, we should name/move the file accordingly, otherwise its in a good place

controllers/operator/authentication/scramsha_credentials.go
```diff
@@ -28,78 +28,134 @@ const (
 	rfc5802MandatedSaltSize = 4
 )
 
-func isPasswordChanged(user *om.MongoDBUser, password string, acUser *om.MongoDBUser) (bool, error) {
-	// if something unexpected happens return true to make sure that the scramShaCreds will be generated again.
-	if acUser == nil || acUser.ScramSha256Creds == nil || acUser.ScramSha1Creds == nil {
+func IsPasswordChanged(user *om.MongoDBUser, password string, acUser *om.MongoDBUser) (bool, error) {
```
nammn: wdyt about readability to split this function into 2 sub functions:  
\- checkSHA256Creds and check SHA1Creds?

controllers/om/omclient.go
filipcirtog: but this is the project one, right?

api/v1/user/mongodbuser_types.go
nammn: there is an open discussion on this here: https://docs.google.com/document/d/1AUUk5ehKKyyFiPN8G6QN6cRcpEmUFuNWxFsUgdLJJxA/edit?tab=t.avzeg81j66b3#heading=h.v3na2nd7jwxg

controllers/om/omclient.go
nammn: i thought to remember that I see somewhere we had both, if not then this is fine, feel free to resolve this one!

api/v1/mdb/mongodb_types.go
```diff
@@ -1481,38 +1481,80 @@ func (m *MongoDB) ObjectKey() client.ObjectKey {
 	return kube.ObjectKey(m.Namespace, m.Name)
 }
 
+// ldapFieldMapping holds both directions of a single field conversion so they stay co-located.
+type ldapFieldMapping struct {
+	toAC func(*Ldap, *ldap.Ldap)
+	toCR func(*ldap.Ldap, *Ldap)
+}
+
+var ldapFieldMappings = []ldapFieldMapping{
```
nammn: nit either: 

- use generics to map them 
- group them

api/v1/user/mongodbuser_types.go
```diff
@@ -157,14 +157,12 @@ func (u MongoDBUser) GetConnectionStringSecretName() string {
 		database = strings.TrimPrefix(database, "$")
 	}
 
-	return normalizeName(fmt.Sprintf("%s%s-%s", resourceRef, u.Name, database))
+	return NormalizeName(fmt.Sprintf("%s%s-%s", resourceRef, u.Name, database))
 }
 
-// normalizeName returns a string that conforms to RFC-1123.
+// NormalizeName returns a string that conforms to RFC-1123.
 // This logic is duplicated in the community operator in https://github.com/mongodb/mongodb-kubernetes/mongodb-community-operator/blob/master/api/v1/mongodbcommunity_types.go.
```
nammn, line 164: nit: we now share the code-base, should we re-use that method instead and delete that one or vice versa? Plus the comment is not relevant anymore?

controllers/om/process.go
```diff
@@ -299,9 +299,94 @@ func (p Process) ProcessType() MongoType {
...
+	if len(p.Args()) == 0 {
+		return nil
+	}
+	m, err := util.MapDeepCopy(p.Args())
+	if err != nil {
+		return nil
+	}
+
+	delete(m, "systemLog")
+	maputil.DeleteMapValue(m, "storage", "dbPath")
```
nammn: I am afraid as long as we don't centralize this somewhere, this can be error prone. Imagine we add/change/remove that somewhere else, this will be broken. Can you consider a safer way?

controllers/om/process.go
```diff
@@ -523,7 +608,7 @@ func (p Process) setClusterAuthMode(authMode string) Process {
 }
 
 func (p Process) authSchemaVersion() int {
-	return p["authSchemaVersion"].(int)
+	return p.AuthSchemaVersion()
```
lucian-tosa: do we still need this then?

controllers/om/shardedcluster.go
```diff
@@ -139,6 +139,11 @@ func (s Shard) mergeFrom(operatorShard Shard) {
 	s.setRs(operatorShard.rs())
 }
 
+// Shards returns the shards in the sharded cluster.
+func (s ShardedCluster) Shards() []Shard {
+	return s.shards()
+}
+
```
lucian-tosa: Why not make the other method public?

controllers/om/shardedcluster.go
```diff
@@ -221,6 +226,11 @@ func (s ShardedCluster) getAllReplicaSets() []string {
 	return ans
 }
 
+// Id returns the shard's _id.
+func (s Shard) Id() string {
+	return s.id()
+}
+
```
lucian-tosa: same here

controllers/operator/authentication/authentication_mechanism.go
filipcirtog: updated. thank you!

controllers/om/shardedcluster.go
filipcirtog: thank you! updated and pushed this change to the stack it is used in.

controllers/om/shardedcluster.go
filipcirtog: thank you! updated and pushed this change to the stack it is used in.

controllers/om/process.go
filipcirtog: thank you! updated and pushed this change to the stack it is used in.

api/v1/user/mongodbuser_types.go
filipcirtog: Moved the method in util.go and reused it in `community` and `enterprise`.  Also, I have moved the code to the stack it was actually being used in. Thank you!

api/v1/mdb/mongodb_types.go
filipcirtog: Grouping them makes sense and I have done that. Thank you! I have also moved the code in the stack it was actually being used in.

controllers/om/process.go
filipcirtog: Done. Defined `infrastructureFieldPaths` and `infrastructureTLSCertKeys` as package-level vars at the top of process.go for better visibility.

controllers/om/deployment.go
filipcirtog: Yes, process names are unique within a deployment. https://www.mongodb.com/docs/ops-manager/current/tutorial/add-existing-mongodb-processes/

controllers/om/backup_agent_config.go
filipcirtog: I have moved everything in the stack which calls each and every method. thank you!

api/v1/user/mongodbuser_types.go
filipcirtog: resolving this since this is now stale. thank you!

api/v1/user/mongodbuser_types.go
filipcirtog: resolving this since this is now stale. thank you!

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_mechanisms.py
filipcirtog: I have updated our existing tests to assert the new logic. thank you!

controllers/om/process.go
filipcirtog: Yes. This function is only used by migration. I have now moved it in the stack this actually gets used. Thank you!

controllers/operator/authentication/scramsha_credentials.go
filipcirtog: Done. Introduced `isCredChanged` which handles the per-algorithm check as it takes the stored `ScramShaCreds` and the `MechanismName` and returns whether the password no longer reproduces those credentials. Thank you!

controllers/operator/authentication/scramsha_credentials.go
```diff
@@ -28,78 +28,132 @@ const (
...
+	sha1Changed, err := isCredChanged(user.Username, password, acUser.ScramSha1Creds, MongoDBCR)
+	if err != nil {
+		return false, err
 	}
 
-	scram256Salt, err := generateSalt(sha256.New)
-	if err != nil {
-		return xerrors.Errorf("error generating scramSha256 salt: %w", err)
+	if len(acUser.Mechanisms) > 0 {
+		// Imported user: reject any password that does not match existing creds.
```
lucian-tosa: But what if it's a regular operator-managed user. We are able to change the password at the moment and it should regenerate the credentials. This would prevent password changing for all users, not just the imported users, right?

filipcirtog: the operator-managed user would not get through this path. `len(acUser.Mechanisms) > 0` is only for users which have mechanisms set and only om does that

lucian-tosa: You could add that as a comment for clarity

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_256_connectivity.py
```diff
@@ -245,6 +306,153 @@ def test_credentials_can_connect_to_db_with_connection_string_secret(
...
+def test_seed_sha256_user_in_ac(replica_set: MongoDB):
+    _seed_sha256_user_in_ac(replica_set)
+
+
+@mark.e2e_replica_set_scram_sha_256_user_connectivity
+def test_om_user_sha256_created(namespace: str):
+    resource = _build_sha256_user_in_k8s(namespace)
+    resource.update()
+    resource.assert_reaches_phase(Phase.Updated)
+
```
lucian-tosa, line 355: It's worth adding a comment for what this does. It's a bit difficult to follow

lucian-tosa (approved):

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_256_connectivity.py
filipcirtog: thanks! updated!

lucian-tosa (approved):

controllers/operator/mongodbuser_controller.go
```diff
@@ -410,7 +410,7 @@ func (r *MongoDBUserReconciler) handleScramShaUser(ctx context.Context, user *us
 	// OM derives from it. Stays Pending until OM processes initPwd.
 	if needsFollowUp {
 		log.Info("initPwd synced for imported user, requeuing to finalize mechanisms and credentials")
-		return r.updateStatus(ctx, user, workflow.Pending("finalizing mechanisms after initPwd sync").WithRetry(3), log)
+		return r.updateStatus(ctx, user, workflow.Pending("finalizing mechanisms after initPwd sync").WithRetry(10), log)
```
lucian-tosa, line 478: any specific reason for increasing this?

lucian-tosa (approved):

controllers/operator/mongodbuser_controller.go
filipcirtog: The other Pending exit in this function already uses WithRetry(10), nothing more than this, just wanted to bring more consistency.

Issue #343, opened 2025-08-13
Cannot create connectionStringSecret in different namespace


thombohlk:
**What did you do to encounter the bug?**
1. Create a MongoDBCommunity resource in namespace X
2. Add a user with a connectionStringSecret in namespace Y
    ```
      users:
        - name: test
          db: test
          passwordSecretRef:
            name: mongo-password
          connectionStringSecretNamespace: some_namespace
          roles:
            - name: dbOwner
              db: test
    ```
3. Notice the secret being created and deleted every few seconds

**What did you expect?**
For a stable secret to be present in namespace Y containing the connection string.

**What happened instead?**
The secret is created and immediately removed again. This is caused by the ownerReference of the secret, which points to the MongoDBCommunity resource in namespace X. This is not allowed, and the secret is therefor garbage collected [link](https://github.com/mongodb/mongodb-kubernetes/issues).

**Operator Information**
 - Operator Version: 0.13.0
 - MongoDB Image used: 8.0.4

**Kubernetes Cluster Information**
 - Distribution: RKE2
 - Version: v1.32.5+rke2r1
 - Image Registry location: quay

realtimetodie:
@nammn @lsierant @mircea-cosbuc @MaciejKaras

vinilage:
Hi @realtimetodie, thanks for raising this. 

This is actually a feature request for our Community Operator, as it currently doesn’t support connectionStringSecret in a different namespace. I’ve opened an internal feature request for it, but please note it may take some time to release due to our current priorities.

In the meantime, I’d recommend using our Enterprise Operator, which is designed for enterprise use cases and includes this functionality.

realtimetodie:
Fixed by https://github.com/mongodb/mongodb-kubernetes/pull/987

PR #987, opened 2026-04-11
fix: cross-namespace connection string secret ownership

realtimetodie:
## Summary

This change fixes an issue when `connectionStringSecretNamespace` points to a different namespace than the `MongoDBCommunity` resource (see reported bug in <https://github.com/mongodb/mongodb-kubernetes/issues/343>).

The underlying problem is that the operator still attaches an **[owner reference](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications)** to the `MongoDBCommunity` object on the generated connection string secret.

That owner reference is invalid because namespaced resources cannot own dependents in a different namespace. Kubernetes rejects those secrets with `OwnerRefInvalidNamespace`, so cross-namespace connection string secret generation does not work even though the field is exposed by the API.

Example - Captured Kubernetes event log

```
kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace
```

```
...
Warning   OwnerRefInvalidNamespace   secret/scram-credentials-secret-username
ownerRef [mongodbcommunity.mongodb.com/v1/MongoDBCommunity, namespace: other-namespace, name: example-scram-mongodb,
uid: 44e6d4a3-09a1-4a70-ab2f-58eefee4d549] does not exist in namespace "other-namespace"
...
```

For example, the following resource is intended to create the generated connection string secret in `other-namespace`:

```yaml
apiVersion: mongodbcommunity.mongodb.com/v1
kind: MongoDBCommunity
metadata:
  name: example-scram-mongodb
  namespace: my-namespace
spec:
  type: ReplicaSet
  members: 3
  security:
    authentication:
      modes: ["SCRAM"]
  users:
    - name: username
      db: admin
      passwordSecretRef:
        name: db-user-secret
      roles:
        - name: role-1
          db: role-1-database
      scramCredentialsSecretName: scram-credentials-secret
   ⛔ connectionStringSecretNamespace: other-namespace ⚠️ this is broken
```

Before this change, the operator attempts to create the connection string secret in `other-namespace` but still attaches an **[owner reference](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications)** to the `MongoDBCommunity` resource in namespace `my-namespace`. Kubernetes rejects that ownership relationship as invalid.

This change keeps the existing owner reference behavior for same-namespace connection string secrets, but skips owner references for cross-namespace connection string secrets. For the cross-namespace case, the operator now adds **annotations** identifying the owning `MongoDBCommunity` resource so that these secrets can still be recognized as operator-managed.

New annotations added to generated connection string secrets:

- `mongodbcommunity.mongodb.com/connection-string-owner-namespace`
- `mongodbcommunity.mongodb.com/connection-string-owner-name`
- `mongodbcommunity.mongodb.com/connection-string-owner-uid`

This change also adds a `MongoDBCommunity` resource [finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) to clean up cross-namespace connection string secrets on resource deletion, since those secrets cannot rely on Kubernetes garbage collection through owner references and therefore must be deleted explicitly.

This change takes care of cleanup for connection string secrets:

- removes cross-namespace connection string secrets when the desired secret namespace changes (`connectionStringSecretNamespace` changes)
- removes cross-namespace connection string secrets when the desired secret name changes while the namespace stays the same (`scramCredentialsSecretName` changes and `connectionStringSecretNamespace` is set but unchanged)
- removes cross-namespace connection string secrets when a user is removed and `connectionStringSecretNamespace` was set
- removes cross-namespace connection string secrets when the `MongoDBCommunity` resource is deleted

## Proof of Work

Reproduced with a `MongoDBCommunity` resource where `spec.users[].connectionStringSecretNamespace` points to a different namespace.

Before this change, Kubernetes emits `OwnerRefInvalidNamespace` warnings for the generated connection string secret, and the secret is not usable as intended.

After this change:

- same-namespace connection string secrets still use owner references
- cross-namespace connection string secrets are created without owner references
- cross-namespace connection string secrets are annotated with the owning `MongoDBCommunity` resource namespace, name, and UID
- controller tests pass, including the new regression test for the cross-namespace case

## Checklist


realtimetodie:
I added a MongoDBCommunity resource [finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) to clean up cross-namespace connection string secrets on resource deletion, since those secrets cannot rely on Kubernetes garbage collection through owner references and therefore must be deleted explicitly.

realtimetodie:
I successfully tested the changes using Helm and a custom MongoDB community operator image in https://github.com/users/realtimetodie/packages/container/package/mongodb-kubernetes

Example

```yaml
helm upgrade --install mongodb-kubernetes-operator mongodb/community-operator \
  --namespace my-namespace \
  --create-namespace \
  --set operator.watchNamespace="my-namespace,other-namespace" \
  --set operator.operatorImageName="mongodb-kubernetes" \
  --set operator.version="fix-connection-string-secret-namespace" \
  --set database.registry.operator="ghcr.io/realtimetodie"
  ```

realtimetodie:
@nammn @lsierant @mircea-cosbuc @MaciejKaras

I can also join your standup to explain the changes etc just let me know

PR #988, opened 2026-04-13
CLOUDP-396749: Unify RS and sharded search reconciliation for multicluster readiness

anandsyncs:
## Context

Search reconciliation has two parallel code paths — `reconcileNonSharded()` and `reconcileSharded()` — that share most of their logic but duplicate service builders, naming resolution, and loop structure. This makes it harder to add multicluster search support later (which would add a cluster dimension on top of shards). This PR unifies both paths into a single topology-agnostic reconcile loop.

### Architecture: Before vs After

> **BEFORE:** `reconcile()` forks into two separate functions (`reconcileNonSharded` / `reconcileSharded`) with duplicated setup logic and 4 separate service builders.
>
> **AFTER:** `reconcile()` does common setup once, then `buildReconcilePlan()` produces a `reconcilePlan{units, manageProxySvc, preflight, cleanup}` (1 unit for RS, N for sharded). A single topology-agnostic loop processes every unit with 2 unified service builders. Sharded-only side effects (per-shard TLS validation, stale shard cleanup) are encoded as `preflight` / `cleanup` closures on the plan, so the reconcile body stays unbranched.

### New: `reconcileUnit` and `reconcilePlan` types
- `reconcileUnit` carries every per-unit difference (STS / Service / ConfigMap names, pod labels, additional service labels, headless-port extras, TLS resource, mongot config function) so the loop body never branches on topology. The "app" selector value is read directly from `podLabels[appLabelKey]`.
- `reconcilePlan` wraps `[]reconcileUnit` plus topology-wide knobs (`manageProxySvc`) and lifecycle hooks (`preflight`, `cleanup`). All sharded-vs-RS differences are resolved by the plan factory; the loop body itself contains no `isSharded` check.
- `buildReconcilePlan()` dispatches to `buildReplicaSetPlan()` (1 unit, no-op preflight/cleanup) or `buildShardedPlan()` (N units, `preflight = validatePerShardTLSSecrets`, `cleanup = cleanupStaleShardResources`) based on the `SearchSourceShardedDeployment` type assertion.
- ReplicaSet units use the existing naming methods (`StatefulSetNamespacedName()`, `SearchServiceNamespacedName()`, etc.) to preserve backward-compatible resource names.

### Unified reconcile loop (`reconcile()`)
- Replaces `reconcileNonSharded()` and `reconcileSharded()` with a single sequence: validate spec → build plan → run `plan.preflight` → compute topology-agnostic modifications (password auth, egress TLS, x509, embedding) once → loop over `plan.units` → run `plan.cleanup` → check LB readiness.
- Per-unit modifications (ingress TLS via `unit.tlsResource`) are computed inside the loop — for RS this is `mdbSearch` itself, for sharded it's the existing `perShardTLSResource` adapter.
- Proxy service creation is gated by `plan.manageProxySvc` (true unless the user opted into unmanaged LB mode for that topology).

### Unified service builders
- `buildHeadlessService(search, unit)` replaces `buildSearchHeadlessService()` + `buildSearchHeadlessServiceForShard()`.
- `buildProxyService(search, unit)` replaces `buildProxyService()` + `buildProxyServiceForShard()`.
- All behavioral differences (`PublishNotReadyAddresses`, wireproto port on the headless Service, shard label, selector source) are derived from `unit` fields.

### Tests
- `TestReconcileReplicaSet_CreatesResources` — new full reconciliation test for RS: runs `helper.reconcile()` against a fake k8s API, marks the STS ready, verifies all created Services/StatefulSets/ConfigMaps with correct names, labels, selectors, and ports.
- `TestReconcileSharded_CreatesPerShardResources` and `TestReconcileSharded_CertificateKeySecretRefRejected` — updated to call `reconcile()` instead of the removed `reconcileSharded()`.
- Proxy service builder tests updated to use `newTestRSUnit` / `newTestShardUnit` helpers with the unified `buildProxyService(search, unit)`.

## Test plan

- [x] All Go unit tests pass (`go test ./controllers/searchcontroller/...`)
- [x] Initial refactor — full PR Evergreen run on `2325e6903`: [succeeded](https://spruce.mongodb.com/version/69dccf7b8c3bc70007b9aa7a) (683 tasks, 0 failures)
- [x] Latest tip (`reconcilePlan` + `manageProxySvc` + `appLabel` cleanup) — search E2E + snippets on `5859e3eb3`: [succeeded](https://spruce.mongodb.com/version/69ee1c2c76c8070007a28e0e) (41/41 tasks: `e2e_mdb_kind_ubi_cloudqa_large` + `private_kind_code_snippets`)


Julien-Ben (approved):
Clean PR and description 👏 
Love the new test of the full reconcile loop 💯

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -94,6 +94,118 @@ func NewMongoDBSearchReconcileHelper(
...
+		preflight:    func(context.Context, *zap.SugaredLogger) workflow.Status { return workflow.OK() },
+		cleanup:      func(context.Context, *zap.SugaredLogger) {},
+	}, nil
+}
+
+func (r *MongoDBSearchReconcileHelper) buildShardedPlan(shardedSource SearchSourceShardedDeployment) (reconcilePlan, error) {
+	shardNames := shardedSource.GetShardNames()
+	units := make([]reconcileUnit, 0, len(shardNames))
+
+	for shardIdx, shardName := range shardNames {
```
lsierant, line 174: can we have a deterministic reconcile order here, e.g. by sorting the shardNames? Or if we're sure the order we'll have from mdb ref or external shard is deterministic enough then it's also fine

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -1175,21 +1097,34 @@ func baseMongotConfig(search *searchv1.MongoDBSearch, hostAndPorts []string) mon
...
-				Address: fmt.Sprintf("0.0.0.0:%d", search.GetMongotWireprotoPort()),
-				Authentication: &mongot.ConfigAuthentication{
-					Mode:    "keyfile",
-					KeyFile: TempKeyfilePath,
-				},
-				TLS: &mongot.ConfigWireprotoTLS{
-					Mode: mongot.ConfigTLSModeDisabled,
-				},
-			}
+		config.Server.Wireproto = &mongot.ConfigWireproto{
```
lsierant, line 1107: we should probably get rid of wireproto entirely; we don't test for it, the code will be cleaner.
not that it should be part of this PR, we can have a ticket for that.

lsierant (approved):

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: Order is deterministic by construction:

- `ShardedExternalSearchSource.GetShardNames()` preserves the order of `spec.ShardedCluster.Shards` ([sharded_external_search_source.go:99-103](https://github.com/mongodb/mongodb-kubernetes/blob/CLOUDP-search-unify-rs-sharded/controllers/searchcontroller/sharded_external_search_source.go#L99-L103)).
- `ShardedInternalSearchSource.GetShardNames()` returns `ShardRsNames()`, which generates `<rsName>-<i>` for `i ∈ [0, ShardCount)` — also stable.

Both derive from the spec, so iteration order is the same across reconciles. Leaving as-is.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: Filed [KUBE-15](https://jira.mongodb.org/browse/KUBE-15) under the [MCK Search GA epic (KUBE-4)](https://jira.mongodb.org/browse/KUBE-4) to remove wireproto entirely.

PR #995, opened 2026-04-14
Introduce kubernetes support version tracking

josvazg:
# Summary

Introduce Kubernetes support version tracking.

This PR introduces the Kubernetes file for version tracking and the script that uses it to check when we should bump according to our Kubernetes version support policy.

Currently this is unenforced and totally disconnected from reality. Next PRs should:
- Use the version file as source of truth for what versions of kind use in tests.
- Run checks against the OpenShift cluster to ensure it is at the version expected by the version file.

## Proof of Work

```shell
$ ./scripts/check-kube-versions.sh
--------------------------------------------------------
Checking Support Policy Compliance
Operator Release Date: 2026-04-14
Buffer: 1 months
--------------------------------------------------------
Policy Cutoff (New Versions): Releases before 2026-03-14 must be supported.

### Checking Kubernetes Range (1.33 - 1.35)
✓ Max version is correct (1.35)
✓ Min version 1.33 is valid (EOL: 2026-06-28)

### Checking OpenShift Version (4.20)
✓ OpenShift version is correct (4.20)


✓ All version checks passed!
```

## Checklist


docs/dev/kubernetes-version-support.md
```diff
@@ -0,0 +1,34 @@
+# Kubernetes and OpenShift version support
+
+**`kubernetes-versions.json`** at the repo root is the engineering source of truth for declared support:
+
+- **`kubernetes.min`** / **`kubernetes.max`**: version range we support the current version as it was tested with such version range.
+- **`openshift`**: OpenShift **minor** i sthe version we expect for our shared test cluster to be at. This is **not** the same as **`release.json`**’s OLM minimum, as explained later.
```
MaciejKaras:
```suggestion
- **`openshift`**: OpenShift **minor** is the version we expect for our shared test cluster to be at. This is **not** the same as **`release.json`**’s OLM minimum, as explained later.
```

docs/dev/kubernetes-version-support.md
```diff
@@ -0,0 +1,34 @@
...
+- **`CONFIG_FILE`**
+- **`OPERATOR_RELEASE_DATE`** (`YYYY-MM-DD`, default today).
+
+## OpenShift
+
+We do not enforce the cluster version from the repo. **`openshift`** records what we **intend** to run against; automation should **fail** if the live cluster does not match so someone either **updates the cluster** or **the file**.
+
+## `release.json` vs this file
+
+**`openshift.minimumSupportedVersion`** in **`release.json`** feeds OLM bundle metadata (e.g. `com.redhat.openshift.versions`)—the **customer certification** floor. **`kubernetes-versions.json` → `openshift`** is the **tracked** line for our test infrastructure. They are independent.
```
MaciejKaras, line 33: Shouldn't we combine the **`openshift.minimumSupportedVersion`** in **`release.json` with **`kubernetes-versions.json` → `openshift`**? I don't fully understand why we should keep the testing infrastructure version different from what we support. I would expect to see similar structure for openshift as for kubernetes in this file:

```json
{
  "kubernetes": { "min": "1.33.7", "max": "1.35.0" },
  "openshift": { "min": "4.6", "max": "4.20" }
}
```

The `openshift.min` would go to the bundle and `openshift.max` will be the current test infrastructure version. Keeping openshift versions split across multiple files makes it hard to understand what each version field is used for.

MaciejKaras (approved):
I really like the idea and it's a great improvement if we can automate this! 🥇

mircea-cosbuc (approved):
LGTM with addressing @MaciejKaras 's comments.

docs/dev/kubernetes-version-support.md
josvazg: I lack enough context to understand the implications. It seems that `release.json` scope is different and not too overlapping with this file.
This file tracks min and max Kubernetes according to our policy. For OpenShift we cannot spin up a cluster of the version we want as easily, so we just track the version of the cluster we test on. It is not "max" precisely, but it kinda means that. 

`openshift.min` on the other hand, not sure how we ensure that support. Are we testing on an old version somehow or we just aim to be compatible with it but we have no hard proof?

josvazg: Going back to the suggestion itself. I would rather introduce this without ties to `release.json`, and make the move separately later, if we all agree is for the best.

MaciejKaras: > It is not "max" precisely, but it kinda means that.

For `kubernetes.max` it's the version of the Kubernetes cluster in the test infrastructure. Why this cannot mean the same for `openshift.max`?

>`openshift.min` on the other hand, not sure how we ensure that support. Are we testing on an old version 
somehow or we just aim to be compatible with it but we have no hard proof?

No we are not, it is just a version that goes to the bundle. We could in the future test against multiple Openshift versions, but that is not the case right now. But basically it says that we support Openshift from version `openshift.min` so it's a min version. And then `openshift.max` is the latest version we have run our tests against.

josvazg: > For kubernetes.max it's the version of the Kubernetes cluster in the test infrastructure. Why this cannot mean the same for openshift.max?

`kubernetes.max/min` will be our tests with kind or similar, any vanilla like kube cluster.
But yes, `openshift.max` could be the test cluster, but feels weird if there is no `openshift.min` being tested.

In other words, the meaning of this file is to **explicitly say why we tested with**. And we did not test in `openshift.min`, so IMHO I would leave things as they are. The ´release.json´ claims that we support from version min, but it is not the same as saying as our tests included version.

`kubernetes-versions.json` precisely tracks what we test, not what we claim we support. In Atlas Kubernetes both things are the same:

https://github.com/mongodb/mongodb-atlas-kubernetes/releases/tag/v2.13.2
> Supported Kubernetes versions updated in this release: 1.33 - 1.35
> Supported OpenShift versions updated: 4.20

 But I see in MCK they are not really a exact match.

MaciejKaras: I see what you mean now. In MCK we only test against single Kubernetes cluster version, that will be `kubernetes.max`. So `kubernetes.min` is not going to be tested against, similar to `openshift.min`.

josvazg: uhm, do you think it is best to avoid `.min` at enforcement time
I would rather leave it there and start testing the min version eventually, at least once before release time.

If we make the tests read `kubernetes.max` we can also make them ready `kubernetes.min` to run the tests in some other config.

MaciejKaras: Ok, so if we agree that this file represents all the versions that we test against, that is acceptable. I was confused because the name of the file says `kubernetes-version-support.md` which is not the same as versions we test against - and OLM version in `release.json` is actually the one we support but do not test against.

PR #1000, opened 2026-04-17
CLOUDP-390500: Enforce govulncheck

josvazg:
# Summary

This adds **govulncheck** as part of the CI checks, enforcing bumping Go or libraries patch versions on fixable issues.

## Proof of Work

✅ [New check passes in CI](https://github.com/mongodb/mongodb-kubernetes/actions/runs/24562652226/job/71814865854?pr=1000)

## Checklist

### DIFF
--- a/.evergreen-functions.yml
+++ b/.evergreen-functions.yml
@@ -679,6 +679,15 @@ functions:
         working_dir: src/github.com/mongodb/mongodb-kubernetes
         script: scripts/evergreen/unit-tests-sbom.sh
 
+  govulncheck:
+    - command: subprocess.exec
+      type: test
+      params:
+        working_dir: src/github.com/mongodb/mongodb-kubernetes
+        add_to_path:
+          - ${workdir}/bin
+        binary: scripts/evergreen/govulncheck.sh
+
   generate_perf_tests_tasks:
     - command: shell.exec
       type: setup
--- a/.evergreen-functions.yml
+++ b/.evergreen-functions.yml
@@ -679,15 +679,6 @@ functions:
         working_dir: src/github.com/mongodb/mongodb-kubernetes
         script: scripts/evergreen/unit-tests-sbom.sh
 
-  govulncheck:
-    - command: subprocess.exec
-      type: test
-      params:
-        working_dir: src/github.com/mongodb/mongodb-kubernetes
-        add_to_path:
-          - ${workdir}/bin
-        binary: scripts/evergreen/govulncheck.sh
-
   generate_perf_tests_tasks:
     - command: shell.exec
       type: setup
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -294,11 +294,6 @@ tasks:
     commands:
       - func: lint_repo
 
-  - name: govulncheck
-    tags: [ "unit_tests" ]
-    commands:
-      - func: "govulncheck"
-
   # Runs on every merge - detects release.json changes and releases appropriate images
   - name: release_om_and_agents
     allowed_requesters: [ "patch" , "commit"]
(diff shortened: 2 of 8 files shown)


.evergreen-functions.yml
```diff
@@ -679,6 +679,15 @@ functions:
...
         script: scripts/evergreen/unit-tests-sbom.sh
 
+  govulncheck:
+    - command: subprocess.exec
+      type: test
+      params:
+        working_dir: src/github.com/mongodb/mongodb-kubernetes
+        add_to_path:
+          - ${workdir}/bin
+        binary: scripts/evergreen/govulncheck.sh
```
mircea-cosbuc: Wouldn't it be better to have govulncheck run as a GH Action to get feedback in PRs?

josvazg: I can move it to a Github action if that is preferable.

josvazg: Done:
[New check passes in CI](https://github.com/mongodb/mongodb-kubernetes/actions/runs/24562652226/job/71814865854?pr=1000)

MaciejKaras: What do you think about adding this check as part of our pre-commit checks? https://github.com/WilliamDeKeyzer/govulncheck-precommit

No CI job will be required, because we already run pre-commit in our lint-repo task.

If we rather want this as GHA then it should be added as required task in repo settings.

josvazg: You mean in addition to or instead of?

> If we rather want this as GHA then it should be added as required task in repo settings.
I do not get this, the GHA alrady runs on PRs due to its config:
```yaml
on:
  pull_request:
  ...
```

MaciejKaras: yes, but this should be added as required task in GH settings. Example required tasks we have already:

Screenshot 2026-04-17 at 13 57 45

mircea-cosbuc: I wanted to have clear visibility of the task. I like having it as a GHA, but I guess we have the same visibility if we add it as a required pass on PRs, like we did with `evergreen/unit_tests`. We don't need it as an EVG job if we have it as a GHA. It's also useful to have it in the pre-commit.

mircea-cosbuc (approved):
LGTM!

.evergreen-functions.yml
josvazg: why required? what is the benefit? 
Eg. If I changed a doc I don't not expect this to run, why should I?

MaciejKaras: I believe that if it is skipped (no changes to go files) and required, it will still pass

josvazg: but then what is the benefit?

josvazg: BTW I would run it in both places.
- Pre-commit catches the issue early while it is being developed.
- CI run catches it when a CVE is published after merge. This happens more often than it would seem.

josvazg: @MaciejKaras are you ok with the current change state now?

filipcirtog (approved):

.evergreen-functions.yml
MaciejKaras: I have still question about the second point:

> - CI run catches it when a CVE is published after merge. This happens more often than it would seem.

This is not the case right now, because this GHA action is configured to only run for Pull Requests. I would either remove it (it's duplicate for what is already checked with `lint-repo` evergreen test) or make it running on some schedule. Since we are frequently committing to this repo I don't see a value to run scheduled job, it would also require configuring Slack notifications etc.

josvazg: Right! I could easily add:
```
on:
  push:
    branches:
      - 'master'
 ```
 
Would that eliminate the concern?

MaciejKaras: sure, if you want to keep the job, yes.

MaciejKaras (approved):

PR #1009, opened 2026-04-21
fix: use retry flakiness metric for accurate flaky test detection

nammn:
# Summary

The flaky test notification script was producing inaccurate results by using a cross-run pass/fail comparison approach. 

This PR replaces that approach with Honeycomb's `mck.retry_flakiness_percent` metric from the `evergreen-agent` dataset, which measures **true flakiness**: the same code passing on retry after an initial failure. This is the same metric used by the existing Honeycomb board.

**Why this approach is better:**
- Measures genuine flakiness (same commit, different outcome)
- Single query instead of 4+ queries
- Full coverage including OpenShift variants
- Aligns with the existing Honeycomb board query
- Reduces script from ~400 lines to ~200 lines

**Trade-off:** Loses per-test granularity (now reports at task level), but the previous per-test data was noisy and incomplete anyway.

## Proof of Work

Dry run output matches the Honeycomb board:
```
| Flakiness % | Variant | Task |
|-------------|---------|------|
| 40.0% | e2e_mdb_openshift_search_ubi_cloudqa | e2e_search_sharded_openshift_external |
| 25.0% | e2e_mdb_openshift_ubi_cloudqa | e2e_sharded_cluster_pv |
| 20.8% | e2e_static_multi_cluster_kind | e2e_multi_cluster_replica_set_deletion |
| 19.5% | e2e_mdb_openshift_ubi_cloudqa | e2e_replica_set_pv |
| 16.7% | e2e_mdb_kind_ubi_cloudqa | e2e_disable_tls_and_scale |
```

https://spruce.corp.mongodb.com/version/69e749859d4cf20007401082/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

## Checklist

    - Marking skip-changelog as this is an internal tooling change

### DIFF
--- a/scripts/evergreen/notify_flaky_tests.py
+++ b/scripts/evergreen/notify_flaky_tests.py
@@ -116,9 +116,7 @@ def get_flaky_tasks(headers: dict) -> list[FlakyTask]:
             {"column": "evergreen.version.requester", "op": "=", "value": "gitter_request"},
             {"column": "evergreen.project.id", "op": "=", "value": PROJECT_ID},
         ],
-        "havings": [
-            {"calculate_op": "AVG", "column": "mck.retry_flakiness_percent", "op": ">", "value": 0}
-        ],
+        "havings": [{"calculate_op": "AVG", "column": "mck.retry_flakiness_percent", "op": ">", "value": 0}],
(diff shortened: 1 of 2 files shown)


MaciejKaras (approved):

mircea-cosbuc (approved):
Nice!

PR #1010, opened 2026-04-21
CLOUDP-398902: Reduce mongot CPU request in e2e_search_sharded_openshift_external

Julien-Ben:
# Summary

`e2e_search_sharded_openshift_external` has been failing most of the time on master since it was added. The test runs 4 mongot pods and reserves 1 CPU each, so it asks for 4 CPU. The shared OpenShift CloudQA cluster only has 2 schedulable workers (the other nodes are tainted master/infra), and with other tasks running concurrently there isn't enough room, so pods stay `Pending` with `FailedScheduling: Insufficient cpu`.

Observed mongot CPU usage during passing runs is well under 1 CPU per pod, so the reservation was much larger than the real demand. Dropping it to 250m leaves plenty of headroom for the JVM while freeing 3 CPU of phantom reservation across the cluster. Memory stays at 2 Gi because `search_construction.go` derives mongot's `-Xmx` from it.

## Proof of Work

The test was failing 100% of the time so a passing run won't prove anything. I'll monitor it this week.

## Checklist

### DIFF
--- a/docker/mongodb-kubernetes-tests/tests/search/search_sharded_openshift_external.py
+++ b/docker/mongodb-kubernetes-tests/tests/search/search_sharded_openshift_external.py
@@ -219,8 +219,15 @@ def mdbs(namespace: str, mdb: MongoDB, helper: SearchDeploymentHelper) -> MongoD
     # mongod connects via Route hostnames instead.
     if _route_hostnames:
         resource["spec"]["loadBalancer"]["managed"]["externalHostname"] = derive_lb_endpoint_template(_route_hostnames)
+    # We run on a shared OpenShift cluster with only 2 schedulable workers (the other nodes are
+    # tainted for masters/infra), and this test starts 4 mongot pods.
+    # The operator default of 2 CPU per pod would ask for 8 CPU total, which doesn't fit, and the pods stay Pending.
+    # Real mongot CPU usage is well under 1 CPU, so we just ask for less. Other search tests run on Kind where
+    # scheduling isn't tight, so they keep the default.
+    # Memory stays at 2 Gi because mongot's JVM heap size (-Xmx) is derived from it in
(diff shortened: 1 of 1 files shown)


viveksinghggits (approved):

docker/mongodb-kubernetes-tests/tests/search/search_sharded_openshift_external.py
```diff
@@ -219,8 +219,15 @@ def mdbs(namespace: str, mdb: MongoDB, helper: SearchDeploymentHelper) -> MongoD
     # mongod connects via Route hostnames instead.
     if _route_hostnames:
         resource["spec"]["loadBalancer"]["managed"]["externalHostname"] = derive_lb_endpoint_template(_route_hostnames)
+    # We run on a shared OpenShift cluster with only 2 schedulable workers (the other nodes are
+    # tainted for masters/infra), and this test starts 4 mongot pods.
+    # The operator default of 2 CPU per pod would ask for 8 CPU total, which doesn't fit, and the pods stay Pending.
+    # Real mongot CPU usage is well under 1 CPU, so we just ask for less. Other search tests run on Kind where
+    # scheduling isn't tight, so they keep the default.
+    # Memory stays at 2 Gi because mongot's JVM heap size (-Xmx) is derived from it in
```
nammn: does this mean we need at least 2 gb mem?

nammn (approved):

docker/mongodb-kubernetes-tests/tests/search/search_sharded_openshift_external.py
Julien-Ben: Operator will default the JVM ram to half of the container memory, and I aim to keep about 1g for it.
The bottleneck was not the memory so I've left it that way.

It doesn't mean it's a strict requirement, I'll trim this comment

PR #1013, opened 2026-04-22
fix: honor klog -stderrthreshold even when -logtostderr is true

pierluigilenoci:
## What changed

klog v2 defaults `-logtostderr` to `true`, which silently ignores `-stderrthreshold` — all log levels go to stderr unconditionally. This has been an [open issue since 2020](https://github.com/kubernetes/klog/issues/212).

klog **v2.140.0** introduced a fix behind an opt-in flag (`legacy_stderr_threshold_behavior`). This PR enables the fix in `main.go` so that `-stderrthreshold` is honored, while preserving the current default behavior (`stderrthreshold=INFO` means all logs still go to stderr unless the user explicitly overrides it on the command line).

### Why this matters

In production Kubernetes clusters, log-aggregation systems (Fluentd, Fluent Bit, Loki, Datadog, …) typically ingest from container stderr. When every severity goes there indiscriminately:

- **Log storage costs increase** — INFO noise stored alongside real errors
- **Alerting becomes unreliable** — stderr contains everything, so severity-based alerts don't work
- **Signal-to-noise ratio degrades** — finding real errors in a stream of INFO is a needle-in-a-haystack problem

The `-stderrthreshold` flag exists precisely to solve this, but it has never worked under the default klog v2 configuration — until now.

### What the fix does

```go
klog.InitFlags(nil)
flag.Set("legacy_stderr_threshold_behavior", "false") // enable the klog fix
flag.Set("stderrthreshold", "INFO")                   // preserve current default
```

- **Placement**: after `klog.InitFlags(nil)` (registers the flags) but before the flags are used
- **`stderrthreshold=INFO`**: keeps current behavior — all logs still go to stderr by default
- **User override**: users can now pass `-stderrthreshold=WARNING` or `-stderrthreshold=ERROR` and it will actually take effect

### References

- Upstream issue: https://github.com/kubernetes/klog/issues/212
- Upstream fix: https://github.com/kubernetes/klog/pull/432

### DIFF
--- a/main.go
+++ b/main.go
@@ -134,6 +134,11 @@ func run() error {
 	operator.OmUpdateChannel = make(chan event.GenericEvent)
 
 	klog.InitFlags(nil)
+	// Opt into the new klog behavior so that -stderrthreshold is honored even
+	// when -logtostderr=true (the default).
+	// Ref: kubernetes/klog#212, kubernetes/klog#432
+	flag.Set("legacy_stderr_threshold_behavior", "false") //nolint:errcheck
+	flag.Set("stderrthreshold", "INFO")                   //nolint:errcheck
 	initializeEnvironment()
 
 	imageUrls := images.LoadImageUrlsFromEnv()


pierluigilenoci:
Hi — friendly ping. Is this PR still on the radar for review? Happy to rebase or make changes if needed. Thanks!

pierluigilenoci:
Hi — friendly ping. This PR is ready from our side. The evergreen CI check seems to be failing on something unrelated to the code change. Would you be able to take a look when you get a chance? Happy to address any feedback. Thank you!

pierluigilenoci:
Hi team — just checking in again. The Evergreen CI check appears to require manual authorization for external contributors. Would someone be able to trigger the CI run so the PR can be validated? The change itself is minimal (honoring klog stderrthreshold). Thanks for your time!

pierluigilenoci:
Hi team — friendly follow-up on this PR. The Evergreen CI check requires manual authorization for external contributors. Would someone be able to trigger the CI run when convenient? The change is minimal (honoring klog -stderrthreshold flag). Happy to address any feedback. Thank you!

PR #1021, opened 2026-04-27
KUBE-126 Voyage controller

fealebenpae:


api/voyageai/v1/vai/voyageai_validation.go
```diff
@@ -0,0 +1,16 @@
+package vai
+
+import (
+	"fmt"
+)
+
+// Validate performs basic validation of the VoyageAI resource.
+func (v *VoyageAI) Validate() error {
```
mircea-cosbuc, line 8: Is the validation going to be a validation webhook? If so, can we keep it in CEL, especially since it seems like we only have basic validation.

controllers/operator/voyageai_controller.go
```diff
@@ -0,0 +1,553 @@
...
+	"github.com/mongodb/mongodb-kubernetes/pkg/kube/service"
+	"github.com/mongodb/mongodb-kubernetes/pkg/statefulset"
+	"github.com/mongodb/mongodb-kubernetes/pkg/util"
+	"github.com/mongodb/mongodb-kubernetes/pkg/util/env"
+)
+
+const (
+	voyageAITLSCertPath = "/etc/voyageai/tls"
+
+	voyageAITLSCertFile = voyageAITLSCertPath + "/tls.crt"
```
mircea-cosbuc, line 45: Do we have any concerns around certificate rotation?

controllers/operator/voyageai_controller.go
```diff
@@ -0,0 +1,553 @@
...
+		return nil, fmt.Errorf("failed to ensure VoyageAI Deployment: %w", err)
+	}
+
+	log.Info("VoyageAI Deployment created/updated")
+	return dep, nil
+}
+
+func buildEnvVars(spec *vaiv1.VoyageAISpec, tlsEnabled bool) []corev1.EnvVar {
+	envs := []corev1.EnvVar{
+		// The default security context runs as UID 2000 which has no /etc/passwd
```
mircea-cosbuc, line 290: Are we compatible with managed security context where the UID and GID are managed by the k8s clusters (e.g. in Openshift)?

controllers/operator/voyageai_controller_test.go
```diff
@@ -0,0 +1,1231 @@
...
+}
+
+func TestVoyageAI_DeploymentSpec_SecurityContext(t *testing.T) {
+	tests := []struct {
+		name                   string
+		managedSecurityContext string
+		wantDefault            bool
+	}{
+		{name: "operator managed", managedSecurityContext: "false", wantDefault: true},
+		{name: "platform managed", managedSecurityContext: "true", wantDefault: false},
```
mircea-cosbuc, line 312: Nice, this answers my previous question around managed security context.

mircea-cosbuc (approved):
LGTM, some small nits and maybe a question for the future around TLS certificate rotation.

controllers/operator/voyageai_controller.go
fealebenpae: We don't need to perform a complex rotation like for mongod - but we would most likely want to follow up to make sure pods restart when the certificate in the mounted Secret changes, not just when a new TLS Secret is configured.

api/voyageai/v1/vai/voyageai_validation.go
fealebenpae: Everything is validated by the apiserver with the OpenAPI schema for now, we didn't find the need for a webhook just yet. We might get rid of this if the need doesn't present itself in the next iteration.

helderjs (approved):

controllers/operator/voyageai_controller.go
fealebenpae: We should be - the base CUDA image didn't even like the hard-coded 2000 UID/GID, so we had to adjust some behavior via environment variables and after that change we should work with any UID/GID.

PR #1035, opened 2026-04-29
KUBE-47: Apply per-cluster statefulSet override in multi-cluster AppDB

Julien-Ben:
# Summary

In multi-cluster AppDB topology, `spec.applicationDatabase.clusterSpecList[i].statefulSet` was silently dropped: only the global `spec.applicationDatabase.podSpec` was merged into the generated StatefulSet. Any per-member-cluster override (hostAliases, tolerations, nodeSelector, affinity, env, volumes) never reached the pods of that cluster.

The fix is a few lines in `AppDbStatefulSet`: after the existing global `podSpec` merge, also merge the per-cluster `StatefulSetConfiguration` when the topology is multi-cluster.

This is a quick, localized fix. The proper pattern, already used in other controllers like Sharded, is to resolve the override in the controller and pass it via `DatabaseStatefulSetOptions.StatefulSetSpecOverride` rather than reading the spec from inside the construct function. A follow-up to align AppDB with that pattern is worth filing separately.

## Proof of Work

- Unit test in `appdb_construction_test.go` builds an AppDB with two clusters, each with a different `hostAliases` block, and asserts each cluster's STS gets only its own. Also asserts the merge does not clobber `Replicas` and `ServiceName` from the base construction.
- e2e: `multicluster_appdb.py` fixture now sets a per-cluster `hostAliases` on each cluster spec item and `test_create_om` reads back each member cluster's AppDB STS to confirm the alias landed. Note: only verified at initial create; the later scale tests rebuild `clusterSpecList` from `cluster_spec_list(...)` and drop the override (called out in a comment).

## Checklist

### DIFF
--- a/changelog/20260429_fix_apply_per_cluster_statefulset_override_in_multi.md
+++ b/changelog/20260429_fix_apply_per_cluster_statefulset_override_in_multi.md
@@ -1,0 +1,6 @@
+---
+kind: fix
+date: 2026-04-29
+---
+
+* **MongoDBOpsManager**, **AppDB**: In multi-cluster AppDB topology, fixed a bug where the per-cluster `spec.applicationDatabase.clusterSpecList[i].statefulSet` override was silently ignored.
--- a/controllers/operator/construct/appdb_construction.go
+++ b/controllers/operator/construct/appdb_construction.go
@@ -454,6 +454,19 @@ func AppDbStatefulSet(opsManager om.MongoDBOpsManager, podVars *env.PodEnvVars, 
 	if podSpec != nil {
 		sts.Spec = merge.StatefulSetSpecs(sts.Spec, appsv1.StatefulSetSpec{Template: *podSpec})
 	}
+
+	// In multi-cluster topology, also apply the per-cluster statefulSet override
+	// from clusterSpecList[i].statefulSet to that cluster's STS.
+	// Quick fix. The proper pattern, used by OpsManager, MongoDBMultiCluster and
+	// Sharded, is to resolve the override in the controller and pass it via
+	// DatabaseStatefulSetOptions.StatefulSetSpecOverride, instead of reading the
+	// spec from inside the construct function.
+	if appDb.IsMultiCluster() {
+		clusterSpecItem := appDb.GetMemberClusterSpecByName(scaler.MemberClusterName())
+		if clusterSpecItem.StatefulSetConfiguration != nil {
+			sts.Spec = merge.StatefulSetSpecs(sts.Spec, clusterSpecItem.StatefulSetConfiguration.SpecWrapper.Spec)
+		}
+	}
 	return sts, nil
 }
 
--- a/controllers/operator/construct/appdb_construction.go
+++ b/controllers/operator/construct/appdb_construction.go
@@ -455,12 +455,8 @@ func AppDbStatefulSet(opsManager om.MongoDBOpsManager, podVars *env.PodEnvVars, 
 		sts.Spec = merge.StatefulSetSpecs(sts.Spec, appsv1.StatefulSetSpec{Template: *podSpec})
 	}
 
-	// In multi-cluster topology, also apply the per-cluster statefulSet override
-	// from clusterSpecList[i].statefulSet to that cluster's STS.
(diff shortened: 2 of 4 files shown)


changelog/20260429_fix_apply_per_cluster_statefulset_override_in_multi.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2026-04-29
+---
+
+* **MongoDBOpsManager**, **AppDB**: In multi-cluster AppDB topology, fixed a bug where the per-cluster `spec.applicationDatabase.clusterSpecList[i].statefulSet` override was silently ignored.
```
vinilage, line 6: LGTM!

vinilage (approved):
Awesome to have this fix so quickly!!! Thanks @Julien-Ben.
LGTM from my side.

Julien-Ben:
@MaciejKaras @mircea-cosbuc 
I pulled you in as reviewers because the ones picked by the bot are on long PTO.
Maciej you were chosen by a random throw of dice.

controllers/operator/construct/appdb_construction.go
```diff
@@ -454,6 +454,19 @@ func AppDbStatefulSet(opsManager om.MongoDBOpsManager, podVars *env.PodEnvVars,
 	if podSpec != nil {
 		sts.Spec = merge.StatefulSetSpecs(sts.Spec, appsv1.StatefulSetSpec{Template: *podSpec})
 	}
+
+	// In multi-cluster topology, also apply the per-cluster statefulSet override
```
MaciejKaras: nit: I would shorten this comment or get rid of it completely

controllers/operator/construct/appdb_construction_test.go
```diff
@@ -37,6 +37,69 @@ func TestAppDBAgentFlags(t *testing.T) {
 	assert.Contains(t, command[len(command)-1], "-Key1=Value1", "-Key2=Value2")
 }
 
+// TestAppDBMultiClusterPerClusterStatefulSetOverride verifies that a
```
MaciejKaras: nit: Instead of the long comment I would just mention the Jira ticket it fixes, because the test title already states what is actually tested,

docker/mongodb-kubernetes-tests/tests/multicluster_appdb/multicluster_appdb.py
```diff
@@ -37,9 +37,22 @@ def ops_manager_unmarshalled(
     resource.create_admin_secret(api_client=central_cluster_client)
 
     resource["spec"]["backup"] = {"enabled": False}
+    appdb_cluster_spec_list = cluster_spec_list(appdb_member_cluster_names, [2, 3])
+    # Per-cluster statefulSet override (hostAliases). Verifies that
+    # clusterSpecList[i].statefulSet is merged into the AppDB STS of that cluster.
+    # Only verified at initial create. Later tests rebuild clusterSpecList via
```
MaciejKaras: >Only verified at initial create. Later tests rebuild clusterSpecList via cluster_spec_list(...) and drop this override.

It worries me that we are reverting this configuration later on. Can we not just have additional function wrapper over `cluster_spec_list(...)` that adds missing override?

MaciejKaras (approved):
Looks good, added just couple small comments.

controllers/operator/construct/appdb_construction.go
Julien-Ben: Thanks for your review, addressed all 3 comments !

controllers/operator/construct/appdb_construction_test.go
Julien-Ben: True, I removed it completely

docker/mongodb-kubernetes-tests/tests/multicluster_appdb/multicluster_appdb.py
Julien-Ben: You are right, that is better, I used the wrapper approach

MaciejKaras (approved):
LGTM!

PR #1037, opened 2026-04-29
Add git worktree setup script and switch_context fixes

lsierant:
## Worktree scripts

Three scripts handle git worktree lifecycle:

**`scripts/dev/create_worktree.sh [-f] <branch>`** — creates and initializes a
worktree from the command line. Accepts a branch name (local or remote; creates
it from the current branch if it doesn't exist yet). The worktree is placed in
the **parent directory** of the main clone, with slashes in the branch name
converted to underscores:

```
branch: user/my-feature  →  ../user_my-feature/
```

**`scripts/dev/init_worktree.sh [-f] <worktree_root> <main_repo_root>`** — shared
initialization logic called by both `create_worktree.sh` and the `post-checkout`
hook. Idempotent (skips files that already exist; pass `-f` to overwrite):
- copies `scripts/dev/contexts/private-context`, `.generated`, and `bin` from the main clone
- runs `make switch` for the current context
- recreates the python venv

**`.githooks/post-checkout`** — git hook that calls `init_worktree.sh`
automatically whenever a linked worktree is checked out. Guards against
false triggers:
- skips file checkouts (`$3 == 0`)
- skips during rebase (detects `rebase-merge` / `rebase-apply` dirs)
- skips in the main clone (`.git` is a directory there, not a file)

## Handling python venv
By default create_worktree.sh creates new venv inside worktree dir. If you want to share venv across multiple worktrees (e.g. to share the venv in IntelliJ project config) specify in private-context a different path for venv: 

```
export PROJECT_VENV_PATH=~/.operator-dev/mck-venv
```

### 1. Hook (post-checkout) — for git UIs, external tools, and agents

Install once per clone (symlinks only this hook, leaving `pre-commit` opt-in):

```bash
ln -sf "$(pwd)/.githooks/post-checkout" .git/hooks/post-checkout
```

After that, any standard `git worktree add` — whether issued from the
command line, a git UI, or an external agent — automatically initializes
the new worktree. Best when the worktree is being created by something
other than you typing a command (IDE, git GUI, automation).

### 2. `create_worktree.sh` — for quick command-line use

```bash
scripts/dev/create_worktree.sh user/my-feature
scripts/dev/create_worktree.sh -f user/my-feature   # force re-init
```

Best when you want to spin up a worktree quickly from just a branch name,
without having to look up the full path or check out the branch first. Also
sources the context env vars into the current shell session after init so
you are immediately ready to work.

The two approaches are complementary: you can have the hook installed and
still use `create_worktree.sh` — the idempotent init logic means the second
pass is a no-op.

## Tests

`scripts/test/bash/worktree/test_worktree.bats` — bats-core integration tests, no mocking. Covers `init_worktree.sh` (idempotency, force flag), `create_worktree.sh` (path conversion, branch creation, force flag), and the `post-checkout` hook (rebase guard, main-clone guard, automatic init on worktree add).

Run with:
```
make test-bash                                  # all suites
scripts/test/bash/run.sh worktree               # only this suite
```

Wired into the `unit_tests` evergreen variant as a new `bash_tests` task (sibling of `lint_repo`, `unit_tests_golang`, etc.) — runs on every PR. The suite is intentionally NOT in pre-commit so the local hook stays fast. The runner self-installs `bats-core` into `~/.cache/bats-core` if it isn't already on `PATH`, so no separate provisioning is needed in the CI image.

## switch_context.sh fixes

Bundled here since worktree init depends on them:
- handle `export KEY=value` lines alongside `declare -x KEY=value`
- skip empty lines when generating `operator.export.env`
- do not overwrite an already-set workdir
- clarify log messages to reflect that the script generates context files

## Proof of work

Two evergreen patches against the new `bash_tests` task on the `unit_tests` variant. The first runs against a clean tree; the second runs against a deliberately-broken assertion in `test_worktree.bats` (`[[ ! -f .current_context ]]`). The first must pass and the second must fail in `bash_tests` — confirmation that the suite actually catches regressions in CI.

- ✅ Passing baseline (commit `41741e99e`): https://spruce.mongodb.com/version/69fc420775c38e0007291e7e — `bash_tests` SUCCESS
- ❌ Deliberate failure (commit `e231ce379`, since reverted): https://spruce.mongodb.com/version/69fc422e04602f00072eb83a — `not ok 1 init_worktree: copies private-context, .generated, bin and creates venv`, other 10 tests pass


fealebenpae (approved):
Bash tests are 🔥

lucian-tosa (approved):

PR #1048, opened 2026-05-03
Search/ga base

anandsyncs:
**Search GA — `search/ga-base` integration branch**

# Based on PR #1166

## Chain of upstream PRs as of 2026-06-28

* PR #1166:
  `master` ← `lsierant/devcontainer`

  * **PR #1048 (THIS ONE)**:
    `lsierant/devcontainer` ← `search/ga-base`

# Multi-cluster `MongoDBSearch` — Search GA integration branch

This is the long-lived integration branch for **MongoDB Search GA** (target **June 30, 2026**, shipping in **MCK 1.9.0**). All Search GA PRs target `search/ga-base`; the branch merges to `master` as a single change set at GA.

- **Current HEAD:** `f63bf7b32` — **0 behind / 42 ahead** of `master` (rebased onto master 2026-06-22).
- **Epic:** KUBE-4. Authoritative scope = the Search GA eng design doc + PD + ORR.
- **GA models (both external-mongod-only):** centralized multi-cluster (single operator fans out) **and** simulated / operator-per-cluster MC.

> Full team-review status (landed / pending / not-started / out-of-scope, with the GA-readiness checklist and ORR snapshot) is maintained separately and refreshed per audit. This description is the branch-level summary.

## Status at a glance (2026-06-22)

- ✅ **Engine complete & merged:** MC-first CRD, centralized MC (RS + sharded), simulated/per-cluster MC, managed Envoy L7 LB + hardening, TLS/x509/SCRAM, version gating, per-cluster mongot wiring.
- ✅ **Hard ORR blocker landed:** `routed_from_another_shard` (#1215, KUBE-124).
- 🔄 **Remaining:** a fleet of open feature/test PRs to land (several need a rebase after the 06-22 master rebase), a short unstarted-code list, observability, GA docs, and ORR/security sign-off.

## Landed on `search/ga-base` (by theme)

| Theme | Key PRs |
|-------|---------|
| MC-first CRD foundation (`spec.clusters[]`, `clusterIndex`, persisted `ClusterMapping`, placeholders, MC validators, state ConfigMap, sizing→clusters[]) | #1027, #1028, #1029, #1030, #1033, #1034, #1064, #1206 |
| Centralized MC reconcile — external RS / external sharded | #1059, #1117, e2e harness #1047 (KUBE-67) |
| Simulated / operator-per-cluster MC | #1126 (KUBE-32), tests #1177 |
[...]

PR #1059, opened 2026-05-04
Phase 2 (Q2-RS-MC): per-cluster MongoDBSearch over external MongoDBMulti RS

anandsyncs:
# Based on PR #1048

## Chain of upstream PRs as of 2026-05-13

* PR #1048:
  `master` ← `search/ga-base`

  * **PR #1059 (THIS ONE)**:
    `search/ga-base` ← `mc-search-phase2-q2-rs-mongothost-fix`

## Summary

MC search Phase 2 — operator support for `MongoDBSearch` deployed across multiple member clusters against an external `MongoDBMulti` ReplicaSet source. Per-cluster mongot StatefulSets / ConfigMaps / Services / Envoys, per-cluster reconcile fan-out, per-cluster Envoy SNI matching, MC admission validation, persisted `ClusterMapping`, cross-cluster owner-label scheme, and the strict `q2_mc_rs_steady` e2e test. Target: gate **G2** — real `$search` + `$vectorSearch` queries return rows from each cluster's local pod seed.

Tracks the active spec [docs/superpowers/specs/2026-04-30-rs-mc-mvp-to-green-design.md](https://github.com/mongodb/mongodb-kubernetes/blob/claude/vigilant-mahavira-456daf/docs/superpowers/specs/2026-04-30-rs-mc-mvp-to-green-design.md) and plan [docs/superpowers/plans/2026-05-03-mc-mvp-base-and-phase2.md](https://github.com/mongodb/mongodb-kubernetes/blob/claude/vigilant-mahavira-456daf/docs/superpowers/plans/2026-05-03-mc-mvp-base-and-phase2.md).

## Highlights

- **Multicluster-first reconcile** (`controllers/searchcontroller/`) — single-cluster is `N=1` of a uniform per-cluster work list, not a parallel branch. Each entry in `spec.clusters[]` becomes its own `rsWorkItem{ClusterName, ClusterIndex, Client}`, and `applyReconcileUnit` writes resources via the matching member-cluster client. Envoy controller follows the same shape with `clusterWorkItem`.
- **Persisted `ClusterMapping` is the source of truth for cluster indexes** — indexes come from the per-CR state ConfigMap (`<name>-state`, written via `AssignClusterIndices`), not from `range` position over `spec.clusters[]`. Indexes are stable across reorder and survive operator restarts. Envoy controller carries the canonical `ClusterIndex == -1` sentinel for the first-reconcile race ("not yet in mapping" → `workflow.Pending`); the primary RS controller writes the state CM before reading, so the sentinel is not needed there.
- **`EffectiveClusters()` cascade — REPLACE-if-nil for 5 spec fields** (`api/v1/search/mongodbsearch_types.go`). Cluster-level overrides win; nil/empty inherits the top-level default:
  - Pointer scalars + small structs (`Replicas`, `ResourceRequirements`, `Persistence`, `StatefulSetConfiguration`) — REPLACE-if-nil
  - `JVMFlags` — REPLACE-if-empty (no append, matches sharded MC semantics)
  - `EffectiveClusterFor(clusterName)` + `GetReplicasForCluster(clusterName)` accessors plumb the resolved spec through `CreateSearchStatefulSetFunc` so each cluster's overrides actually reach its own STS — not just `clusters[0]`.
- **Owner labels at every member-cluster write site** — `mongodb.com/search-name`, `mongodb.com/search-namespace`, `mongodb.com/cluster-name` stamped on STS + headless Service + proxy Service + mongot ConfigMap + Envoy Deployment + Envoy ConfigMap. Consumed by `MapMemberClusterObjectToSearch` for cross-cluster watch routing (`pkg/handler/labels_search.go`) — cross-cluster owner refs don't carry, so labels do the work.
- **Per-cluster naming helpers** (`api/v1/search/`) — `StatefulSetNamespacedNameForCluster(idx)`, `ProxyServiceNamespacedNameForCluster(idx)`, `MongotConfigConfigMapNameForCluster(idx)`, `SearchServiceNamespacedNameForCluster(idx)`, `LoadBalancerDeploymentNameForCluster(name)` — always index-suffixed in MC; legacy unindexed names reserved for the single-cluster degenerate case.
- **Per-cluster Envoy SNI** (`controllers/operator/mongodbsearchenvoy_controller.go`) — `server_names` resolves to `{search-name}-search-{idx}-proxy-svc.{ns}.svc.cluster.local`. Proxy Service selector matches per-cluster Envoy pods (fixes earlier `OperationFailure: Error connecting to Search Index Management service` at `test_create_search_index`).
- **MC admission**: rejects `len(spec.clusters) > 1` without `spec.source.external.hostAndPorts`; rejects MC without `spec.loadBalancer.managed`; requires `{clusterName}` or `{clusterIndex}` in `externalHostname` for MC.
- **Single status owner**: only `mongodbsearch_controller` writes the full CR status. The Envoy controller writes only the `/status/loadBalancer` substatus via scoped JSON patch — no two-writer race.

## Tests added

- **`TestMongoDBSearchReconcile_Success_MultiCluster`** (`controllers/operator/mongodbsearch_controller_test.go`) — drives the full MC reconcile end-to-end with 2 member-cluster clients. Asserts per-cluster STS / headless Svc / proxy Svc / mongot CM lands on the correct member client, carries all 3 owner labels, and does NOT leak onto other clients.
- **`TestEffectiveClustersCascade`** + **`TestEffectiveClusterFor`** + **`TestGetReplicasForCluster`** (`api/v1/search/mongodbsearch_types_test.go`) — locks the cascade semantics: each of the 5 fields' override path is exercised by its own table row.
- **`q2_mc_rs_steady.py`** strict assertions: `Phase=Running`, per-cluster Envoy Ready, `$search` + `$vectorSearch` data plane via per-cluster `mongotHost` patched into OM AC. Owner-label provenance verified on all per-cluster resources. Persisted `ClusterMapping` verified via state ConfigMap read. Mongot ConfigMap `hostAndPorts` content asserted against the seed list.
- **MC e2e test helpers** consolidated into canonical homes per ga-base review feedback (PR #1048) — `MCSearchDeploymentHelper` in `tests/common/search/search_deployment_helper.py`; `replicate_secret` + `assert_resource_in_cluster` in `tests/common/multicluster/multicluster_utils.py`. No `tests/common/multicluster_search/` subpackage; no test-helper unit tests.

## Out of scope (MVP scope acceptances)

- **No finalizer / no cross-cluster CR-delete cleanup**. Kubernetes GC does not cross cluster boundaries. Member-cluster Envoy Deployments + ConfigMaps and per-cluster mongot STS / Services / ConfigMaps are accepted to leak on CR delete and on managed→unmanaged LB transition. The parked finalizer commit (`search/jben/search-finalizer-parked`) is explicitly out of scope for MVP.
- **No per-cluster status surface**. `SearchClusterStatusItem`, `ClusterStatusList`, `status.loadBalancer.clusters[]`, `ShardLoadBalancerStatus`, `ObservedReplicas` were removed on ga-base (PR #1074) and are not reintroduced here. Per-cluster failures surface in the top-level `status.message` only. Reintroduction parked in draft PR #1095 for post-MVP.
- **Per-cluster error aggregation is fail-fast** — first cluster error stops the reconcile loop; no `multierror.Append` across clusters for MVP.
- **`StatefulSetConfiguration` cascade is REPLACE-if-nil for the whole struct** — canonical sharded MC does outer REPLACE-if-nil + deep merge of `SpecWrapper.Spec.Template` via `merge.PodTemplateSpecs`. The deep-merge is a follow-up; current behavior requires the user to spell out the full PodTemplate to override a single nested field.
- Phase 3 (Q2-Sh-MC), Phase 4 (Q1-RS-MC), Phase 5 (Q1-Sh-MC) — separate PRs.

## Verification

- ✅ Unit tests green: `go test ./controllers/searchcontroller/... ./controllers/operator/... ./api/v1/search/...`
- ✅ Surgical EVG patch on PR HEAD `bdb9770f8`: `e2e_search_q2_mc_rs_steady` green on both `e2e_multi_cluster_2_clusters` and `e2e_static_multi_cluster_2_clusters`.
- ✅ Lint autofix (gofumpt + black + isort) committed as `1c1dfd148`; `run_pre_commit` green on the auto-trigger patch.
- 🔍 Full PR Evergreen patch results: see commit-status checks linked above.


controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -130,17 +194,79 @@ func (r *MongoDBSearchReconcileHelper) buildReconcilePlan(log *zap.SugaredLogger
...
+	// Discriminate on the raw spec, not on EffectiveClusters() — the latter
+	// auto-promotes nil into a 1-element slice and would funnel the legacy
+	// path into the indexed-naming loop.
+	if r.mdbSearch.Spec.Clusters == nil || len(*r.mdbSearch.Spec.Clusters) == 0 {
+		return r.buildSingleClusterReplicaSetUnit(hostSeeds)
+	}
+	clusters := *r.mdbSearch.Spec.Clusters
+
+	units := make([]reconcileUnit, 0, len(clusters))
+	for idx, cluster := range clusters {
```
anandsyncs: [Claude] `idx` is the spec-array position — reordering `spec.clusters[]` renames every per-cluster STS / Service / ConfigMap.

- Envoy resolves via `state.ClusterMapping` at `mongodbsearchenvoy_controller.go:212-215` with `-1` sentinel for unmapped
- Controller loads the mapping at `mongodbsearch_controller.go:181` but doesn't pass it here

Plumb the mapping through `NewMongoDBSearchReconcileHelperWithMembers`; resolve `idx := mapping[c.ClusterName]`; emit `workflow.Pending` for `-1`.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -130,17 +194,79 @@ func (r *MongoDBSearchReconcileHelper) buildReconcilePlan(log *zap.SugaredLogger
...
+	hostSeeds, err := rsSource.HostSeeds("")
 	if err != nil {
 		return reconcilePlan{}, err
 	}
 
-	svcName := r.mdbSearch.SearchServiceNamespacedName().Name
+	// Discriminate on the raw spec, not on EffectiveClusters() — the latter
+	// auto-promotes nil into a 1-element slice and would funnel the legacy
+	// path into the indexed-naming loop.
+	if r.mdbSearch.Spec.Clusters == nil || len(*r.mdbSearch.Spec.Clusters) == 0 {
```
anandsyncs: [Claude] Forking single-cluster into a dedicated function is the anti-pattern the same PR's Envoy controller avoids.

- Envoy's `buildClusterWorkList` returns a 1-element work list with `ClusterName: ""` for the SC case (`mongodbsearchenvoy_controller.go:206-223`)
- The struct comment at lines 149-152 promises "downstream code never branches" — this fork falsifies it

Drop `buildSingleClusterReplicaSetUnit`; synthesize a 1-element unit with the legacy unindexed names + `clusterName=""` inside the same loop.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -79,22 +81,71 @@ type MongoDBSearchReconcileHelper struct {
 	mdbSearch            *searchv1.MongoDBSearch
 	db                   SearchSourceDBResource
 	operatorSearchConfig OperatorSearchConfig
+	// secretGaps records per-cluster customer-replicated-secret gaps reported
+	// by the parent controller's CheckSecretsPresence call. The helper folds
+	// gaps into the top-level status.message so warnings surface without a
+	// second writeback (per-cluster status types were dropped in ga-base #1074).
+	secretGaps []SecretCheckResult
```
anandsyncs: [Claude] `secretGaps` is set from `mongodbsearch_controller.go:190` but never read in this file. The field comment promises "the helper folds gaps into the top-level status.message" — no code does.

Controller surfaces gaps via `surfaceMissingSecrets` + `RequeueAfter` already, so the field is orphaned. Either wire it into the final `workflow.OK()/Pending(...)` message before returning, or delete the field + `SetSecretGaps` + parent call site.

api/v1/search/mongodbsearch_types.go
```diff
@@ -462,6 +478,42 @@ func (s *MongoDBSearch) MongotConfigConfigMapNamespacedName() types.NamespacedNa
 	return types.NamespacedName{Name: s.Name + "-search-config", Namespace: s.Namespace}
 }
 
+// MongotConfigConfigMapNameForCluster returns the per-cluster mongot ConfigMap
+// name. Index 0 matches the legacy single-cluster name for back-compat.
+func (s *MongoDBSearch) MongotConfigConfigMapNameForCluster(clusterIndex int) types.NamespacedName {
+	if clusterIndex == 0 {
```
anandsyncs: [Claude] Only this helper short-circuits on index 0. Siblings always emit the index:

- `StatefulSetNamespacedNameForCluster`
- `SearchServiceNamespacedNameForCluster`
- `ProxyServiceNamespacedNameForCluster`
- `LoadBalancerDeploymentNameForCluster`
- `LoadBalancerConfigMapNameForCluster`

Drop the `if clusterIndex == 0` branch; have `buildSingleClusterReplicaSetUnit` call `MongotConfigConfigMapNamespacedName()` directly.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -373,6 +467,140 @@ func (r *MongoDBSearchReconcileHelper) reconcile(ctx context.Context, log *zap.S
...
+	mods reconcileUnitMods,
+) (*appsv1.StatefulSet, error) {
+	mods = mods.withDefaults()
+
+	unitClient, err := r.clientForCluster(unit.clusterName)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := r.ensureSearchService(ctx, log, unitClient, unit.headlessSvc, buildHeadlessService(r.mdbSearch, unit)); err != nil {
```
anandsyncs, line 515: [Claude] STS / headless Svc / proxy Svc / mongot CM written to member clusters without `MongoDBSearchOwnerNameLabel` / `MongoDBSearchOwnerNamespaceLabel` / `MongoDBSearchClusterNameLabel`. Consequences:

- Member-cluster events on these resources don't enqueue the central CR (`MapMemberClusterObjectToSearch` reads exactly these labels)
- Label-based GC cleanup can't list-and-delete them

Envoy stamps them via `envoyLabelsForCluster` (`mongodbsearchenvoy_controller.go:776-784`). Introduce `searchLabelsForCluster(search, clusterName, clusterIndex)` and merge into:

- `unit.podLabels` (STS pod-template)
- `buildHeadlessService` Service `ObjectMeta.Labels`
- `buildProxyService` Service `ObjectMeta.Labels`
- mongot ConfigMap `ObjectMeta.Labels`

controllers/operator/mongodbsearch_controller.go
```diff
@@ -186,7 +186,8 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
 		}
 	}
 
-	reconcileHelper := searchcontroller.NewMongoDBSearchReconcileHelper(r.kubeClient, mdbSearch, searchSource, r.operatorSearchConfig)
+	reconcileHelper := searchcontroller.NewMongoDBSearchReconcileHelperWithMembers(r.kubeClient, mdbSearch, searchSource, r.operatorSearchConfig, r.memberClusterClientsMap)
```
anandsyncs: [Claude] PR adds no `Finalizer` / `DeletionTimestamp` handling. Per-cluster Envoy Deployment/ConfigMap + helper's mongot STS / Service / ConfigMap will leak on CR delete — cross-cluster owner refs don't GC.

- Parked commit `cdfc72108` (`search/jben/search-finalizer-parked`) defines `SearchFinalizer` + `deleteAllEnvoyResources` iterating `state.ClusterMapping`
- #1074's description hand-offs the work to this PR

Cherry-pick `cdfc72108`, then extend cleanup to iterate `state.ClusterMapping` and call per `(memberName, idx)`:

- `StatefulSetNamespacedNameForCluster(idx)`
- `SearchServiceNamespacedNameForCluster(idx)`
- `ProxyServiceNamespacedNameForCluster(idx)`
- `MongotConfigConfigMapNameForCluster(idx)`

Also folds in the pre-existing `(0)` hardcodes at `mongodbsearchenvoy_controller.go:304,311`.

api/v1/search/mongodbsearch_types.go
```diff
@@ -450,6 +450,22 @@ func (s *MongoDBSearch) ProxyServiceNamespacedName() types.NamespacedName {
 	return types.NamespacedName{Name: s.Name + "-search-0-" + ProxyServiceSuffix, Namespace: s.Namespace}
 }
 
+// ProxyServiceNamespacedNameForCluster returns the proxy Service name for one
```
anandsyncs: Too much text in the comment.

api/v1/search/mongodbsearch_types.go
anandsyncs: Apply this comment and resolve.

controllers/operator/mongodbsearch_controller.go
anandsyncs: Apply this comment and resolve.

controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -404,6 +406,55 @@ func TestNewMongoDBSearchReconciler_MultiCluster(t *testing.T) {
 	assert.NotNil(t, r.memberClusterSecretClientsMap["eu-west-k8s"].KubeClient)
 }
 
+// TestOperatorSchemeKnowsMongoDBSearch is a regression test for the v5 e2e
```
anandsyncs: Too much text in this comment.

controllers/operator/mongodbsearchenvoy_controller.go
```diff
@@ -372,52 +376,86 @@ func buildReplicaSetRoute(search *searchv1.MongoDBSearch) envoyRoute {
...
+	return []envoyRoute{buildReplicaSetRouteForCluster(search, clusterIndex, clusterName)}
 }
 
 // buildReplicaSetRouteForCluster builds a single RS-mode route for one cluster.
-func buildReplicaSetRouteForCluster(search *searchv1.MongoDBSearch, clusterName string) envoyRoute {
-	route := buildReplicaSetRoute(search)
-	route.ClusterID = clusterName
-	route.SNIHostname = applyClusterIDToSNI(route.SNIHostname, clusterName, search.GetManagedLBEndpoint() != "")
-	return route
+// The SNI hostname is the per-cluster proxy-svc FQDN derived from
```
anandsyncs: Too much text in this comment.

controllers/operator/mongodbsearchenvoy_controller.go
```diff
@@ -506,6 +542,18 @@ func (r *MongoDBSearchEnvoyReconciler) ensureDeployment(ctx context.Context, sea
...
+						// terminate it as Istio mTLS. mongod's gRPC connection to
+						// `mongotHost` / `searchIndexManagementHostAndPort` arrives as
+						// raw application TLS (using the LB server cert), so the
+						// Istio sidecar's mTLS handshake never completes — mongod
+						// sees a 20s connect timeout and returns code 125 ("Error
+						// connecting to Search Index Management service"). Excluding
+						// the proxy port from Istio inbound capture lets mongod's
+						// TLS reach Envoy directly. Single-cluster installs without
+						// Istio injection treat the annotation as a no-op.
+						"traffic.sidecar.istio.io/excludeInboundPorts": fmt.Sprintf("%d", searchv1.EnvoyDefaultProxyPort),
```
anandsyncs, line 506: Why do we have this istio reference here?

controllers/operator/mongodbsearchenvoy_controller_test.go
```diff
@@ -102,6 +102,93 @@ func TestBuildReplicaSetRoute(t *testing.T) {
 	}
 }
 
+// TestBuildReplicaSetRouteForCluster_PerClusterPlaceholders regresses the
```
anandsyncs: Too much comment.

controllers/operator/mongodbsearchenvoy_controller_test.go
```diff
@@ -462,7 +550,72 @@ func TestBuildRoutesForCluster_Sharded_PerClusterShardSNI(t *testing.T) {
 	assert.Len(t, seen, 4, "per-(cluster, shard) SNIs must all be distinct")
 }
 
-// --- reconciler constructor with member-cluster client maps -------------------
+// TestEnvoyFilterChain_PerClusterSNI verifies that for each member cluster, the
```
anandsyncs: Don't put big comments in tests

controllers/searchcontroller/external_search_source_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	searchv1 "github.com/mongodb/mongodb-kubernetes/api/v1/search"
+)
+
+// TestExternalSearchSource_HostSeeds_ReturnsTopLevelHostAndPorts is the
```
anandsyncs: separate test for this is unncessary

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -130,17 +174,79 @@ func (r *MongoDBSearchReconcileHelper) buildReconcilePlan(log *zap.SugaredLogger
 		log.Infof("Reconciling MongoDBSearch for sharded source deployment with %d shards", shardedSource.GetShardCount())
 		return r.buildShardedPlan(shardedSource)
 	}
-	return r.buildReplicaSetPlan()
+	return r.buildReplicaSetPlan(r.db)
 }
 
-func (r *MongoDBSearchReconcileHelper) buildReplicaSetPlan() (reconcilePlan, error) {
-	hostSeeds, err := r.db.HostSeeds("")
+// buildReplicaSetPlan produces a reconcilePlan for the replica-set topology.
```
anandsyncs: unnecessary big comment

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: apply this comment and resolve

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: apply this comment and resolve.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: apply this and resolve

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -79,20 +80,88 @@ type MongoDBSearchReconcileHelper struct {
...
 	db SearchSourceDBResource,
 	operatorSearchConfig OperatorSearchConfig,
+) *MongoDBSearchReconcileHelper {
+	return NewMongoDBSearchReconcileHelperWithMembers(client, mdbSearch, db, operatorSearchConfig, nil, nil)
+}
+
+// NewMongoDBSearchReconcileHelperWithMembers constructs a helper with the
+// per-member-cluster client map and persisted clusterMapping populated.
+// Pass nil/empty for single-cluster.
+func NewMongoDBSearchReconcileHelperWithMembers(
```
Julien-Ben: We don't need two functions for reconcile helper initialization.

controllers/operator/mongodbsearchenvoy_controller_test.go
```diff
@@ -102,6 +102,83 @@ func TestBuildReplicaSetRoute(t *testing.T) {
 	}
 }
 
+// Regression: externalHostname template must substitute both {clusterName}
+// and {clusterIndex}; an earlier path left {clusterIndex} literal in SNI.
```
Julien-Ben: I don't think this comment should be written as "we used have X bug so we wrote this test."
Just explaining the why/what briefly is enough

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -130,44 +209,94 @@ func (r *MongoDBSearchReconcileHelper) buildReconcilePlan(log *zap.SugaredLogger
...
 		return r.buildShardedPlan(shardedSource)
 	}
-	return r.buildReplicaSetPlan()
+	return r.buildReplicaSetPlan(r.db)
 }
 
-func (r *MongoDBSearchReconcileHelper) buildReplicaSetPlan() (reconcilePlan, error) {
-	hostSeeds, err := r.db.HostSeeds("")
+// buildReplicaSetPlan returns one reconcileUnit per cluster. Single-cluster is
+// represented as a 1-element work list with clusterName "" and unindexed names;
```
Julien-Ben: Could we index every single resources with 0- now ? To enable for easy migration later

api/v1/search/mongodbsearch_types.go
```diff
@@ -462,6 +472,33 @@ func (s *MongoDBSearch) MongotConfigConfigMapNamespacedName() types.NamespacedNa
...
+		Namespace: s.Namespace,
+	}
+}
+
+// StatefulSetNamespacedNameForCluster returns the index-suffixed StatefulSet
+// name (`<name>-search-<idx>`). The unindexed name from
+// StatefulSetNamespacedName is reserved for the single-cluster path.
+func (s *MongoDBSearch) StatefulSetNamespacedNameForCluster(clusterIndex int) types.NamespacedName {
+	return types.NamespacedName{
+		Name:      fmt.Sprintf("%s-search-%d", s.Name, clusterIndex),
```
Julien-Ben, line 484: TODO for later:
For all resources, ensure that even in single cluster case, we leave a -0 cluster index. There should not be any resource with no index. A resource should be easy to extend to multi cluster later, without breaking.

api/v1/search/mongodbsearch_types.go
```diff
@@ -699,42 +736,111 @@ func (s *MongoDBSearch) GetEndpointForShard(shardName string) string {
...
+	}
+	return out
+}
+
+// EffectiveClusterFor returns the cascaded ClusterSpec for a specific cluster
+// name. Returns the auto-promoted single-cluster spec when clusterName is empty
+// (legacy/single-cluster path). Returns a zero ClusterSpec if the named cluster
+// is not in spec.clusters[] (defensive — shouldn't happen in healthy reconcile
+// state but avoids panic).
+func (s *MongoDBSearch) EffectiveClusterFor(clusterName string) ClusterSpec {
```
Julien-Ben: We shouldn't return a "zero clusterSpec" but a proper error, that forces the reconciler to handle it.

api/v1/search/mongodbsearch_types_test.go
```diff
@@ -87,6 +89,143 @@ func TestEffectiveClusters(t *testing.T) {
...
+	assert.Equal(t, topJVMFlags, spec.JVMFlags)
+	assert.Same(t, clusterAPersistence, (*spec.Clusters)[0].Persistence)
+	assert.Nil(t, (*spec.Clusters)[0].ResourceRequirements)
+	assert.Nil(t, (*spec.Clusters)[2].Persistence)
+	assert.Nil(t, (*spec.Clusters)[2].Replicas)
+	assert.Empty(t, (*spec.Clusters)[2].JVMFlags)
+}
+
+// TestGetReplicasForCluster verifies the cascade-aware per-cluster replica accessor.
+func TestGetReplicasForCluster(t *testing.T) {
```
Julien-Ben: Can be removed, should be (or is already) asserted in the test above.

api/v1/search/mongodbsearch_types_test.go
```diff
@@ -334,3 +473,31 @@ func TestGetManagedLBEndpointForClusterShard_NotManaged(t *testing.T) {
...
+
+	got := s.ProxyServiceNamespacedNameForCluster(0)
+	assert.Equal(t, "mdb-search-search-0-proxy-svc", got.Name)
+	assert.Equal(t, "ns", got.Namespace)
+	assert.Equal(t, s.ProxyServiceNamespacedName(), got, "index 0 must match legacy ProxyServiceNamespacedName")
+
+	got1 := s.ProxyServiceNamespacedNameForCluster(1)
+	assert.Equal(t, "mdb-search-search-1-proxy-svc", got1.Name)
+
+	got2 := s.ProxyServiceNamespacedNameForCluster(2)
```
Julien-Ben: 2 is redundant with 1, remove it

controllers/operator/mongodbsearch_controller.go
```diff
@@ -186,7 +186,7 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
 		}
 	}
 
-	reconcileHelper := searchcontroller.NewMongoDBSearchReconcileHelper(r.kubeClient, mdbSearch, searchSource, r.operatorSearchConfig)
+	reconcileHelper := searchcontroller.NewMongoDBSearchReconcileHelperWithMembers(r.kubeClient, mdbSearch, searchSource, r.operatorSearchConfig, r.memberClusterClientsMap, state.ClusterMapping)
```
Julien-Ben: Unify the two reconciler (`withMembers` not needed)

controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -404,6 +407,35 @@ func TestNewMongoDBSearchReconciler_MultiCluster(t *testing.T) {
 	assert.NotNil(t, r.memberClusterSecretClientsMap["eu-west-k8s"].KubeClient)
 }
 
+// Regression: per-cluster Service writes failed with "no kind is registered
```
Julien-Ben: TODO: remove "regression:" comments. We don't need historical explanation of why we add a trivial unit test

controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -404,6 +407,35 @@ func TestNewMongoDBSearchReconciler_MultiCluster(t *testing.T) {
...
+// for type search.MongoDBSearch in scheme" when member-cluster clients used a
+// scheme without searchv1; apiv1.AddToScheme must register MongoDBSearch.
+func TestOperatorSchemeKnowsMongoDBSearch(t *testing.T) {
```
Julien-Ben: Remove this test, this should fail when the operator starts. When it is instructed to watch a resource that is not installed

controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -606,3 +638,162 @@ func TestMongoDBSearchControllerReconcile_StateConfigMap_ReAddCluster(t *testing
...
+	search.Spec.Clusters = &clusters
+
+	eastClient := mock.NewEmptyFakeClientBuilder().Build()
+	westClient := mock.NewEmptyFakeClientBuilder().Build()
+	memberClients := map[string]client.Client{
+		"us-east": eastClient,
+		"us-west": westClient,
+	}
+
+	// External source — no MongoDBCommunity required on the central client.
```
Julien-Ben: Useless comment

controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -606,3 +638,162 @@ func TestMongoDBSearchControllerReconcile_StateConfigMap_ReAddCluster(t *testing
...
+			sts := &appsv1.StatefulSet{}
+			require.NoError(t, tc.mc.Get(ctx, stsName, sts), "STS must exist on owning member client")
+			headlessSvc := &corev1.Service{}
+			require.NoError(t, tc.mc.Get(ctx, headlessName, headlessSvc), "headless Service must exist on owning member client")
+			proxySvc := &corev1.Service{}
+			require.NoError(t, tc.mc.Get(ctx, proxyName, proxySvc), "proxy Service must exist on owning member client")
+			cm := &corev1.ConfigMap{}
+			require.NoError(t, tc.mc.Get(ctx, cmName, cm), "mongot ConfigMap must exist on owning member client")
+
+			// Owner-labels at every member-cluster write site so cross-cluster
```
Julien-Ben: Do we do cross cluster watching ? If not adapt comment.
We can keep the owner labels

controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -606,3 +638,162 @@ func TestMongoDBSearchControllerReconcile_StateConfigMap_ReAddCluster(t *testing
...
+			// watch routing (MapMemberClusterObjectToSearch) + cleanup-by-label
+			// can find the resource — owner refs do not cross cluster boundaries.
+			for _, obj := range []client.Object{sts, headlessSvc, proxySvc, cm} {
+				labels := obj.GetLabels()
+				assert.Equal(t, search.Name, labels[khandler.MongoDBSearchOwnerNameLabel], "owner-name label on %T", obj)
+				assert.Equal(t, search.Namespace, labels[khandler.MongoDBSearchOwnerNamespaceLabel], "owner-namespace label on %T", obj)
+				assert.Equal(t, tc.clusterName, labels[khandler.MongoDBSearchClusterNameLabel], "cluster-name label on %T", obj)
+			}
+
+			// Per-cluster resources do NOT leak onto the other member client or
```
Julien-Ben: We can remove the leak assertions.

controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -606,3 +638,162 @@ func TestMongoDBSearchControllerReconcile_StateConfigMap_ReAddCluster(t *testing
...
 }
+
+// TestMongoDBSearchReconcile_Success_MultiCluster drives Reconcile() end-to-end
+// in MC mode with two member-cluster clients. It verifies the full fan-out:
+// each cluster gets its index-suffixed STS / headless Service / proxy Service /
+// mongot ConfigMap on its own member client (never on the central client or the
+// wrong member client), the owner-labels are stamped on every per-cluster write
+// so member-cluster watch routing + cross-cluster GC can find them, and the
+// top-level CR reaches PhaseRunning after STSes are marked ready.
+func TestMongoDBSearchReconcile_Success_MultiCluster(t *testing.T) {
```
Julien-Ben, line 611: TODO:

Use the "full reconcile loop test" pattern to address more complex cases. Including for example more complex field overrides

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
Julien-Ben: [AI] Done in #1101 — the two constructors are collapsed into one. Single-cluster call sites pass nil for the member-cluster wiring args.

controllers/operator/mongodbsearchenvoy_controller_test.go
Julien-Ben: [AI] Done in #1101 — the historical-bug preamble is gone.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
Julien-Ben: [AI] Deferred — switching the single-cluster path to indexed names is a breaking rename of every existing search resource on upgrade and needs migration handling. Left as a TODO for a follow-up.

api/v1/search/mongodbsearch_types.go
Julien-Ben: [AI] Deferred — same scope concern as the related thread on the helper: renaming the unindexed single-cluster resources to a -0 form is a breaking change for existing deployments. Tracking as a TODO for a follow-up.

api/v1/search/mongodbsearch_types.go
Julien-Ben: [AI] Done in #1101 — EffectiveClusterFor now returns (ClusterSpec, error). The reconciler-side callers handle the missing-cluster case explicitly, and the container builder takes the resolved per-cluster spec directly so the duplicate lookup is gone.

api/v1/search/mongodbsearch_types_test.go
Julien-Ben: [AI] Done in #1101 — TestGetReplicasForCluster removed; the cascade coverage lives in TestEffectiveClustersCascade.

api/v1/search/mongodbsearch_types_test.go
Julien-Ben: [AI] Done in #1101 — the third case (idx 2) was removed; the second case (idx 1) is enough to cover the index-suffix formatting.

controllers/operator/mongodbsearch_controller.go
Julien-Ben: [AI] Done in #1101 — single constructor now (the withMembers shim is gone).

controllers/operator/mongodbsearch_controller_test.go
Julien-Ben: [AI] Done in #1101 — the test was removed entirely (see the next thread), so the regression preamble is gone with it.

controllers/operator/mongodbsearch_controller_test.go
Julien-Ben: [AI] Done in #1101 — test removed. Operator startup already requires the scheme to register MongoDBSearch (apiv1.AddToScheme in main.go), so the assertion was duplicative.

controllers/operator/mongodbsearch_controller_test.go
Julien-Ben: [AI] Done in #1101 — comment removed.

controllers/operator/mongodbsearch_controller_test.go
Julien-Ben: [AI] Done in #1101 — comment rephrased to focus on the labels-as-cross-cluster-identity claim (owner refs don't carry between clusters). Whether the watch is wired is orthogonal to the label-stamping invariant, so the watch-routing mention is gone. Owner labels stay.

controllers/operator/mongodbsearch_controller_test.go
Julien-Ben: [AI] Done in #1101 — leak assertions removed. The positive per-cluster assertions already pin where each resource lands.

controllers/operator/mongodbsearch_controller_test.go
Julien-Ben: [AI] Noted for a follow-up. The cascade is currently exercised at the unit-test level via TestEffectiveClustersCascade; a full reconcile-loop test driving cluster-level overrides → per-cluster STS spec would catch broader end-to-end issues and is worth adding once the cascade surface stabilises.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
Julien-Ben: [AI] Already addressed at the current PR HEAD — no code change needed in #1101. The helper now reads the index from the persisted state mapping at `controllers/searchcontroller/mongodbsearch_reconcile_helper.go:280` (`r.clusterMapping[c.ClusterName]`), and the controller passes the mapping in at `controllers/operator/mongodbsearch_controller.go:189` after `AssignClusterIndices` has run. Indexes are no longer positional. The mapping is written before being read, so the -1 sentinel isn't needed on this path.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
Julien-Ben: [AI] Already addressed at the current PR HEAD — no code change needed in #1101. There's no `buildSingleClusterReplicaSetUnit` function; `buildRSWorkList` (`mongodbsearch_reconcile_helper.go:273-283`) returns a 1-element list with empty `ClusterName` for the single-cluster path and N items for MC. Same shape as Envoy's `buildClusterWorkList`.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
Julien-Ben: [AI] Already addressed at the current PR HEAD — the `secretGaps` field no longer exists on the helper struct (`mongodbsearch_reconcile_helper.go:78-92`). Gap surfacing happens entirely in the controller via `surfaceMissingSecrets` + RequeueAfter. No code change in #1101.

api/v1/search/mongodbsearch_types.go
Julien-Ben: [AI] Already addressed at the current PR HEAD — `MongotConfigConfigMapNameForCluster` (`api/v1/search/mongodbsearch_types.go:476-481`) always emits the index now; the `clusterIndex == 0` short-circuit was removed by an earlier commit on this branch. No code change in #1101.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
Julien-Ben: [AI] The owner labels are already stamped on the four member-cluster resources:

- STS ObjectMeta — `mongodbsearch_reconcile_helper.go:142-151` (`withSearchOwnerLabels`)
- Headless Service — `mongodbsearch_reconcile_helper.go:927-929`
- Proxy Service — `mongodbsearch_reconcile_helper.go:972-974`
- Mongot ConfigMap — `mongodbsearch_reconcile_helper.go:778-780`

These flow through the same `searchOwnerLabels(search, clusterName)` helper that Envoy uses for its Deployment / ConfigMap. The STS pod-template is intentionally untouched — `withSearchOwnerLabels` documents that pod-template labels would touch the immutable selector and force an STS recreate on upgrade. Watch routing reads labels from the resource itself, not from pods, so the routing path is intact. No code change in #1101 for this thread, but happy to revisit if pod-template labels are needed for a separate reason.

controllers/operator/mongodbsearch_controller.go
Julien-Ben: [AI] Out of scope for this batch of fixes. The parent PR description explicitly defers cross-cluster GC + finalizer to a follow-up, and WORK_NOTES marks the work as parked alongside this PR. The parked commit `cdfc72108` is intact on `search/jben/search-finalizer-parked` and will be brought back as a follow-up PR that also extends cleanup to per-cluster mongot STS / Services / mongot ConfigMap and folds the hardcoded `(0)` in `mongodbsearchenvoy_controller.go:285,292` into the same iteration. Not addressed in #1101.

api/v1/search/mongodbsearch_types.go
Julien-Ben: [AI] Done in #1101 — comment collapsed to a single line.

controllers/operator/mongodbsearch_controller_test.go
Julien-Ben: [AI] Done in #1101 — the test was removed entirely (operator startup already requires the scheme to register MongoDBSearch), so the comment is gone with it.

controllers/operator/mongodbsearchenvoy_controller.go
Julien-Ben: [AI] Done in #1101 — comments on `buildRoutesForCluster` and `buildReplicaSetRouteForCluster` are trimmed.

controllers/operator/mongodbsearchenvoy_controller.go
Julien-Ben: [AI] The annotation excludes Envoy's proxy port from Istio sidecar inbound capture: mongod talks to Envoy over raw app-TLS (not Istio mTLS), so REDIRECT-mode capture stalls the TLS handshake (20s timeout, fast-failed by mongod as code 125). It's a no-op when Istio is absent. The 3-line comment at the call site already documents this; no code change in #1101 but happy to expand or move the comment if a different shape is preferred.

controllers/operator/mongodbsearchenvoy_controller_test.go
Julien-Ben: [AI] Done in #1101 — regression preamble removed.

controllers/operator/mongodbsearchenvoy_controller_test.go
Julien-Ben: [AI] The function-level preamble at the current PR HEAD is already 2 lines (`mongodbsearchenvoy_controller_test.go:540-541`). The other strings inside the test are per-assertion failure messages that fire on real failure (e.g. "per-cluster SNIs must all be distinct; collisions break per-cluster TLS routing"). I'd lean toward keeping those — they make a failing test cheaper to triage. Happy to trim if you prefer; no code change in #1101 for this thread.

controllers/searchcontroller/external_search_source_test.go
Julien-Ben: [AI] Already addressed at the current PR HEAD — `controllers/searchcontroller/external_search_source_test.go` was removed by a prior commit on this branch. No code change in #1101.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
Julien-Ben: [AI] Done in #1101 — `buildReplicaSetPlan` docstring collapsed to 2 lines.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -111,6 +170,16 @@ type reconcileUnit struct {
 	logFields           []any                // k/v fields attached to the per-unit logger (nil for single-unit topologies)
 	tlsResource         tls.TLSConfigurableResource
 	mongotConfigFn      mongot.Modification
+
+	// Per-cluster routing fields; clusterName == "" means central client.
```
Julien-Ben: Formatting: don't separate these two fields from the others, and the comment should be on the side (and shorter)

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -130,44 +199,91 @@ func (r *MongoDBSearchReconcileHelper) buildReconcilePlan(log *zap.SugaredLogger
...
+type rsWorkItem struct {
+	ClusterName  string
+	ClusterIndex int
+}
+
+// buildRSWorkList returns a 1-element work list with clusterName "" for the
+// legacy single-cluster path, or one item per spec.clusters[i] with
+// clusterIndex resolved from the persisted state.ClusterMapping for MC.
+func (r *MongoDBSearchReconcileHelper) buildRSWorkList() []rsWorkItem {
+	if r.mdbSearch.Spec.Clusters == nil || len(*r.mdbSearch.Spec.Clusters) == 0 {
```
Julien-Ben, line 259: Double check if functions in mongodb_search_types file could help simplify this

controllers/searchcontroller/mongodbsearch_reconcile_helper_test.go
```diff
@@ -383,6 +386,49 @@ func TestBuildProxyServiceForShard_ManagedLB_Ready(t *testing.T) {
 	assert.Equal(t, search.LoadBalancerDeploymentNameForCluster(0), svc.Spec.Selector["app"])
 }
 
+// Regression: per-cluster proxy Service selector must match the per-cluster
```
Julien-Ben: Don't use "regression:" to comment test.
Just explain very briefly what they do if necessary

docker/mongodb-kubernetes-tests/tests/common/multicluster/multicluster_utils.py
```diff
@@ -10,3 +19,82 @@ def multi_cluster_pod_names(replica_set_name: str, cluster_index_with_members: l
 def multi_cluster_service_names(replica_set_name: str, cluster_index_with_members: list[tuple[int, int]]) -> list[str]:
     """List of multi-cluster service names for given replica set name and a list of member counts in member clusters."""
     return [f"{pod_name}-svc" for pod_name in multi_cluster_pod_names(replica_set_name, cluster_index_with_members)]
+
+
+_KIND_METHOD: dict[str, str] = {
```
Julien-Ben: What is this ? Do we need this design pattern ? What problem is it solving ?
It is over-engineered, make it clear what is invoked.

docker/mongodb-kubernetes-tests/tests/multicluster_search/fixtures/search-q2-mc-rs-search.yaml
```diff
@@ -0,0 +1,20 @@
+---
+# MongoDBSearch (multi-cluster) for Q2-MC-RS e2e.
```
Julien-Ben: This doesn't need a comment

docker/mongodb-kubernetes-tests/tests/multicluster_search/fixtures/search-q2-mc-rs.yaml
```diff
@@ -0,0 +1,21 @@
+---
+# MongoDBMultiCluster source RS for Q2-MC-RS e2e.
```
Julien-Ben: Remove comments in this fixture

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
+"""Q2-MC-RS e2e: 2-cluster MongoDBSearch with external MongoDBMulti source.
+
+Strict assertions across the full path:
```
Julien-Ben: Do not list assertions blindly, tell the reader about the why and what briefly. Use natural language

This is valid for the whole file.
Do not comment obvious things (typing, declaring a small variable...)

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+    )
+
+
+# =============================================================================
+# Test steps
+# =============================================================================
+
+
+@mark.e2e_search_q2_mc_rs_steady
+def test_create_kube_config_file(
```
Julien-Ben: Remove this test

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+        SOURCE_BUNDLE_SECRET,
+        member_cluster_clients,
+        central_cluster_client,
+        mdb,
+    )
+
+
+@mark.e2e_search_q2_mc_rs_steady
+def test_create_mdb_resource(mdb: MongoDBMulti):
+    """STRICT — Phase 2 ops code makes per-cluster MongoDBMulti reach Running."""
```
Julien-Ben: Remove this comment

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+@mark.e2e_search_q2_mc_rs_steady
+def test_deploy_lb_certificates(
+    namespace: str,
+    multi_cluster_issuer: str,
+    helper: MCSearchDeploymentHelper,
+):
+    """Managed LB server + client certs, with SAN list covering every
+    cluster's per-cluster proxy-svc FQDN.
+
+    Phase 2's LB cert SAN validator (commit dbdd35b0c) requires every
```
Julien-Ben: Remove this comment, don't mention commit IDs in comment

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+        f"{MDBS_RESOURCE_NAME}-{MONGOT_USER_NAME}-password",
+        # External CA Secret — operator at controllers/searchcontroller/external_search_source.go
+        # mounts spec.source.external.tls.ca as a Secret volume on per-cluster mongot pods.
+        # The CA is created as both ConfigMap+Secret with the same name in the central cluster
+        # (see create_issuer_ca); replicate the Secret half here.
+        CA_CONFIGMAP_NAME,
+    ]
+
+    for secret_name in secrets_to_replicate:
+        replicate_secret(
```
Julien-Ben: Use the existing helper for that

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+    source_cm = central_core.read_namespaced_config_map(name=CA_CONFIGMAP_NAME, namespace=namespace)
+    cm_data = dict(source_cm.data or {})
+    for mcc in member_cluster_clients:
+        create_or_update_configmap(namespace, CA_CONFIGMAP_NAME, cm_data, api_client=mcc.api_client)
+        logger.info(f"replicated CA ConfigMap {CA_CONFIGMAP_NAME} into cluster {mcc.cluster_name}")
+
+
+@mark.e2e_search_q2_mc_rs_steady
+def test_create_search_resource(mdbs: MongoDBSearch):
+    """STRICT — Phase 2 operator code makes Running real now.
```
Julien-Ben: Remove any comment that were tracking the implementation in progress.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+        sts = apps.read_namespaced_stateful_set(name=sts_name, namespace=namespace)
+        headless = core.read_namespaced_service(name=svc_name, namespace=namespace)
+        proxy = core.read_namespaced_service(name=proxy_svc_name, namespace=namespace)
+        cm = core.read_namespaced_config_map(name=cm_name, namespace=namespace)
+        _assert_search_owner_labels(sts.metadata.labels or {}, mcc.cluster_name, f"STS {sts_name}")
+        _assert_search_owner_labels(headless.metadata.labels or {}, mcc.cluster_name, f"headless Service {svc_name}")
+        _assert_search_owner_labels(proxy.metadata.labels or {}, mcc.cluster_name, f"proxy Service {proxy_svc_name}")
+        _assert_search_owner_labels(cm.metadata.labels or {}, mcc.cluster_name, f"mongot CM {cm_name}")
+
+        # mongot CM source-hosts content — the Phase 2 fan-out renders the
```
Julien-Ben: Remove this comment

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+        )
+
+
+@mark.e2e_search_q2_mc_rs_steady
+def test_verify_per_cluster_envoy_deployment(
+    namespace: str,
+    helper: MCSearchDeploymentHelper,
+    member_cluster_clients: List[MultiClusterClient],
+):
+    """STRICT — every cluster's Envoy Deployment + ConfigMap exist and the
```
Julien-Ben: Remove any "STRICT"-style comment

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+    mdbs.assert_lb_status()
+
+
+@mark.e2e_search_q2_mc_rs_steady
+def test_patch_per_cluster_mongot_host(
+    mdb: MongoDBMulti,
+    helper: MCSearchDeploymentHelper,
+    member_cluster_clients: List[MultiClusterClient],
+):
+    """Patch Ops Manager AC so each cluster's mongods point `mongotHost`
```
Julien-Ben: This comment is too long

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+
+@pytest.mark.skip(
+    reason=(
+        "validateLBCertSAN is implemented but not wired into the reconcile "
+        "loop yet (TODO: controllers/searchcontroller/mongodbsearch_reconcile_helper.go:1660). "
+        "When wired, add a negative test that submits a managed LB cert with a "
+        "missing per-cluster SAN and asserts the resource fails to reach Running."
+    )
+)
+@mark.e2e_search_q2_mc_rs_steady
```
Julien-Ben: This test can be removed

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -0,0 +1,1318 @@
...
+        )
+
+
+# =============================================================================
+# Admission / validator negative paths
+# =============================================================================
+
+
+@mark.e2e_search_q2_mc_rs_steady
+def test_admission_rejects_mc_without_external_hosts(
```
Julien-Ben: Remove this test

Julien-Ben: [AI] Done in #1102 — `test_admission_rejects_mc_without_external_hosts` removed.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — the skipped `test_lb_cert_san_validator_when_wired` placeholder is gone.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — trimmed from ~25 lines to a short why-only note (MongoDBMulti CRD doesn't yet expose per-cluster `additionalMongodConfig`, so the per-process AC keys are set directly).

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — the `STRICT —` prefix is stripped from every test docstring file-wide, and the module docstring is rewritten in natural language.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — the mongot-CM source-hosts comment is gone; the subsequent assertion already documents intent.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — implementation-in-progress wording (the `STRICT` prefixes and the manual AC-patch narrative) stripped file-wide.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — replaced the custom `replicate_secret` loop with the canonical `kubetester.create_or_update_secret(..., api_client=cluster.api_client)` pattern (see `docker/mongodb-kubernetes-tests/tests/multicluster/conftest.py`). The `replicate_secret` helper itself is dropped from `multicluster_utils.py`.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — comment removed, including the commit SHA. Same scrub applied to `fixtures/search-q2-mc-rs-search.yaml` (which referenced `98209bb9c`).

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — comment removed.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — `test_create_kube_config_file` removed.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
Julien-Ben: [AI] Done in #1102 — module docstring rewritten as a short natural-language summary, `STRICT —` prefix stripped from every test docstring, and individual verbose comments trimmed (including the 25-line AC-patch narrative and the commit-SHA reference). Keeping this bar going forward.

docker/mongodb-kubernetes-tests/tests/multicluster_search/fixtures/search-q2-mc-rs.yaml
Julien-Ben: [AI] Done in #1102 — fixture stripped to the YAML body.

docker/mongodb-kubernetes-tests/tests/multicluster_search/fixtures/search-q2-mc-rs-search.yaml
Julien-Ben: [AI] Done in #1102 — fixture stripped to the YAML body.

docker/mongodb-kubernetes-tests/tests/common/multicluster/multicluster_utils.py
Julien-Ben: [AI] Done in #1102 — `assert_resource_in_cluster` and the `_KIND_METHOD` lookup table are deleted. Every call site followed up with a typed read for label or readiness inspection that already trips on a missing resource, so the existence check was redundant. The orphaned `replicate_secret` helper is dropped at the same time.

controllers/searchcontroller/mongodbsearch_reconcile_helper_test.go
Julien-Ben: [AI] Done in #1102 — the `Regression:` prefix is dropped; the comment is now a one-line why.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
Julien-Ben: [AI] Audit — no code change in #1102. `buildRSWorkList` only assembles the (clusterName, clusterIndex) pairs by walking `spec.clusters[]` and looking up the persisted index. The four name helpers from `api/v1/search/mongodbsearch_types.go` (`StatefulSetNamespacedNameForCluster`, `SearchServiceNamespacedNameForCluster`, `ProxyServiceNamespacedNameForCluster`, `MongotConfigConfigMapNameForCluster`) are already consumed via `rsResourceNames` at `mongodbsearch_reconcile_helper.go:272-285`. The unindexed-vs-indexed branch in `rsResourceNames` is intentional pending the deferred `-0` naming alignment for single-cluster (tracked in WORK_NOTES). Happy to revisit if there's a specific simplification in mind.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
Julien-Ben: [AI] Done in #1102 — `clusterName` and `clusterIndex` are back in the main field block with a short inline comment on `clusterName`.

PR #1064, opened 2026-05-05
Store State in a config map, use cluster indexes for resource names

Julien-Ben:
## Summary

Two focused fixes on top of `search/ga-base`:

### 1. Replace cluster-index annotation with `StateStore[SearchDeploymentState]` (`9f8bf66`)
The `clusterName -> clusterIndex` mapping for `MongoDBSearch` was persisted in the `mongodb.com/v1.lastClusterNumMapping` annotation on the CR. Annotations conflate user intent with operator bookkeeping and are visible/strippable. Switch to a per-CR `<name>-state` ConfigMap managed via `StateStore[SearchDeploymentState]`, matching the pattern used by `ShardedClusterDeploymentState` and `AppDBDeploymentState`.

The reconciler runs `AssignClusterIndices` against current `spec.clusters` and writes the merged mapping back via `stateStore.WriteState` before the helper runs. Steady-state reconciles are no-ops (`reflect.DeepEqual` guard).

`validateClustersNoRename` is dropped — it relied on reading the annotation at admission time, the StateStore is not visible there, and the index machinery already handles remove+add safely.

### 2. Envoy resource names use `clusterIndex`, not `clusterName` (`3551187`)
`clusterName` is a user-supplied Kubernetes cluster identifier that can be long (e.g. `gke-us-east1-prod-main`) and routinely overruns the 63-char DNS label limit when appended to the LB name base. Switch the per-cluster Envoy `Deployment` and `ConfigMap` to the stable cluster index from the StateStore mapping:

- `LoadBalancerDeploymentNameForCluster(int)` -> `<name>-search-lb-0-<i>`
- `LoadBalancerConfigMapNameForCluster(int)` -> `<name>-search-lb-0-<i>-config`

Single-cluster installs use index `0` uniformly. The empty-string back-compat branch is dropped (pre-GA).

The Envoy controller reads `state.ClusterMapping` via `loadOrInitSearchState` (read-only); clusters not yet registered in the mapping surface a per-cluster `Pending` and skip `reconcileForCluster` until the main controller writes the state CM.

Adjacent fix: `buildProxyService` selector now references the suffixed Envoy Deployment name (`LoadBalancerDeploymentNameForCluster(0)`) — the legacy unsuffixed name no longer matches the Envoy pods' `app` label, so leaving the selector unchanged would have broken the managed-LB Ready flip.

## Test plan

- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] `go test ./api/... ./controllers/... ./pkg/...` (47 packages) — all pass
- [x] Full reconcile-loop unit tests exercise the state CM behaviour: initial create, no-op on stable mapping, operator-restart with pre-seeded CM, monotonic extension on cluster add, and index preservation on remove + re-add
- [x] Three Envoy reconcile-loop tests exercise: (1) no state CM -> per-cluster Pending, (2) reconcile uses indices from the seeded mapping verbatim, (3) stable index across cluster removals
- [x] Local kind smoke (`tests/search/search_replicaset_external_mongodb_proxy_service.py`): 12 passed, 1 skipped, 1 failed (the failure is a known local-dev DNS limitation — pymongo connecting to in-cluster DNS from the host process — not a code regression). Envoy index-suffixed naming verified end-to-end.
- [x] Evergreen patch [69fa0f2e](https://evergreen.mongodb.com/version/69fa0f2e8e31db0007047e4d) — all 32 tasks succeeded across 6 managed-LB e2e tasks x 5 variants

## Follow-ups

- [ ] Refactor the four `StateConfigMap` reconcile-loop tests (`TestMongoDBSearchControllerReconcile_StateConfigMap{,_NoOpOnStableMapping,_OperatorRestart,_ReAddCluster}`) into a single table-driven test (per @anandsyncs review). Tracked with a `// TODO(@anandsyncs PR #1064)` marker in `mongodbsearch_controller_test.go`.
- [ ] Add multi-cluster search e2e tests covering: state CM GC on CR deletion (owner ref), operator pod restart preserves mapping, monotonic-index invariant under random add/remove. The python helpers in `docker/mongodb-kubernetes-tests/tests/common/multicluster_search/` are already in place; the test files are not yet written.

### DIFF
--- a/controllers/operator/mongodbsearch_controller_test.go
+++ b/controllers/operator/mongodbsearch_controller_test.go
@@ -420,6 +420,9 @@ func TestMongoDBSearchReconcile_MissingSecret_Requeues(t *testing.T) {
 	require.True(t, res.RequeueAfter > 0, "must requeue when a customer-replicated secret is missing")
 }
 
+// TODO(@anandsyncs PR #1064): refactor StateConfigMap reconcile-loop tests
+// (this one + _NoOpOnStableMapping, _OperatorRestart, _ReAddCluster) into a
+// single table-driven test.
 func TestMongoDBSearchControllerReconcile_StateConfigMap(t *testing.T) {
 	ctx := context.Background()
 
--- a/controllers/operator/mongodbsearch_controller_test.go
+++ b/controllers/operator/mongodbsearch_controller_test.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"testing"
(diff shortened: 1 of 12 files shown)


controllers/operator/mongodbsearch_controller_test.go
```diff
@@ -418,3 +419,187 @@ func TestMongoDBSearchReconcile_MissingSecret_Requeues(t *testing.T) {
 	require.NoError(t, err, "missing secret must surface as RequeueAfter, not an error")
 	require.True(t, res.RequeueAfter > 0, "must requeue when a customer-replicated secret is missing")
 }
+
+func TestMongoDBSearchControllerReconcile_StateConfigMap(t *testing.T) {
```
anandsyncs, line 426: Make these test cases table driven

PR #1074, opened 2026-05-07
Search/jben/ga base fixes 2

Julien-Ben:
Round 2 of PR-1048 review-comment cleanup, on top of `search/ga-base`. Almost entirely simplification — no behavioural additions on this branch.

Most of it is type-surface tightening for the GA scope cut: drops `ShardOverride`, per-cluster status fields (`status.loadBalancer.clusters[]`, `clusterStatusList[]`), `ManagedLBConfig.Replicas`, and the `PerCluster*` wrappers — `ClusterSpec.LoadBalancer` is now plain `*LoadBalancerConfig`. CRDs regenerated.

Envoy controller goes multicluster-first — `selectEnvoyClient` is gone, the per-cluster client lives on `clusterWorkItem` and the lookup happens once during `buildClusterWorkList`. Single-cluster becomes "MC with a 1-element work list".

Smaller things: `CheckSecretsPresence` collapsed central+members into one map, bespoke `EnqueueRequestForSearchOwnerMultiCluster` replaced with `handler.EnqueueRequestsFromMapFunc`, AST-drift detector + `tests/common/multicluster_search/` deleted (`mc_proxy_svc_fqdn` pre-staged for #1059).

Deferred to #1059 by design: `EffectiveClusters()` cascade, helper-side per-cluster fan-out, `LoadBalancerDeploymentNameForCluster(0)` hardcode in the helper.

Parked on `search/jben/search-finalizer-parked` (single commit): cross-cluster GC for Envoy resources via `mongodb.com/search-finalizer`. Solves the AI-5 leak (cross-cluster OwnerRef GC doesn't work; today's pattern in sharded-cluster `OnDelete` is fire-and-forget). Kept out of this PR to keep it scoped to simplification — bring it back alongside #1059 so the same finalizer hook can clean up the helper's per-cluster mongot resources too.

Unit tests green; reconcile-loop coverage on the envoy controller. Targeted EVG patch (from the pre-parking branch state): https://evergreen.mongodb.com/version/69fca3b00627bf00078a8868.


api/v1/search/phase_test.go
```diff
@@ -0,0 +1,53 @@
+package search
```
anandsyncs, line 1: Why is this new file added here?

controllers/operator/mongodbsearch_controller.go
```diff
@@ -119,6 +121,28 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
 		return result, err
 	}
 
+	// Finalizer: cross-cluster Envoy GC.
```
anandsyncs: Is this in GA scope?

PR #1080, opened 2026-05-08
KUBE-17, KUBE-26, KUBE-27: search connectivity tool with cache-aware availability detection

lsierant:
# Based on PR #1048

* PR #1166:

  * PR #1048:

    * **PR #1080 (THIS ONE)**:
      `search/ga-base` ← `lsierant/KUBE-17-connectivity-tool`

## Summary

- `tests/common/search/connectivity.py` — `SearchConnectivityTool` for `$search` one-shot + paging, failure-class taxonomy (`cursor_lost` / `transient_network` / `other`), pod/STS disruption helpers (`delete_pods`, `wait_for_all_pods_replaced`, `wait_for_pods_by_label_replaced`, `wait_for_mongot_statefulset_drained`, `patch_mongot_readiness_probe_to_false`).
- `tests/common/search/bootstrap_test_mixins.py` — layered reusable mixins for deploy → search → data → smoke (RS, sharded, MC).
- `tests/common/search/background_availability_tester.py` — daemon-thread outage-window verdict accumulator.
- New e2e markers: `e2e_search_connectivity_tool`, `e2e_search_connectivity_tool_sharded`, `e2e_search_failure_modes`.
- `kubetester/__init__.py` — promoted `list_matching_pods` / `pod_is_ready` to public; added `wait_for_pods_ready` / `wait_for_no_pods_ready`.

## Local e2e (devcontainer against EVG-host kind)

| Marker | Result |
|---|---|
| `e2e_search_connectivity_tool` | 25 passed, 1 skipped (5:07) |
| `e2e_search_connectivity_tool_sharded` | 18 passed, 1 skipped (7:50) |
| `e2e_search_failure_modes` | 17 passed, 1 skipped (10:01) |

## Proof of Work

- PoW-Patch (full search suite — cloudqa + OM 8.0 + multi-cluster + openshift): https://spruce.mongodb.com/version/6a1f5d31f2a89a000729d810
- All `e2e_search*` tasks green, including `e2e_search_connectivity_tool_mc_sharded` — the new `test_cluster_tagged_search_routes_in_cluster`, validated on a fresh deploy (per-cluster member tags + `nearest` tagSet read preference pin each shard's `$search` to a same-cluster mongot; no cross-cluster fallback).
- Two failing tasks are base-branch, not introduced here:
  - `lint_repo` — `govulncheck` flags Go standard-library CVEs (go1.26.3, fixed in 1.26.4); a base toolchain bump, no repo code involved.
  - `unit_tests_python` — flaky `scripts/test/wt_ctl/test_network_registry.py::test_concurrent_allocate_serializes` (wt-ctl tooling, outside this diff).
- A later commit removes the fragile `test_single_cluster_mongot_outage_preserves_other_clusters` (superseded by the locality test); not re-patched, as removing an already-green test cannot regress the suite.


docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,671 @@
...
+   sentinel-injection path is not used.
+
+For the strongest guarantee that "this query reached upstream", callers should
+use one-shot mode with cache-busting *and* assert
+``cache_hit_hint is False``. Paging-mode results expose the same flag per
+page, so a long-running tester that wants a high-confidence verdict can simply
+filter to pages where ``cache_hit_hint is False`` and assert at least one such
+page succeeded over the observation window.
+
+Per the U1.1 plan: no unit tests with mocks — verified end-to-end on real
```
lsierant: do not refer to Ux.y plan at all.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,671 @@
...
+    # ------------------------------------------------------------------
+    # Helpers
+    # ------------------------------------------------------------------
+
+    @property
+    def _collection(self):
+        return self.search_tester.client[self.db_name][self.col_name]
+
+    @staticmethod
+    def _classify_error(exc: BaseException) -> tuple[str, Optional[int]]:
```
lsierant, line 156: explain what is this for

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,671 @@
...
+    def make_cache_buster_query(self, base_term: str = "movie") -> tuple[dict, str]:
+        """Build a ``$search`` pipeline that mongot cannot serve from cache.
+
+        The pipeline pairs a stable ``must`` clause (so the query reliably
+        matches some documents) with a unique random token in the ``should``
+        clause. mongot keys its query cache by query identity, so a fresh
+        token guarantees a real evaluation. Returns ``(pipeline_stage, token)``
+        — the caller is responsible for building the full aggregation pipeline.
+        """
+        token = f"kube17_{uuid.uuid4().hex[:12]}"
```
lsierant: do not refer to jira tickets in such identifiers

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,671 @@
...
+        stage = {
+            "$search": {
+                "compound": {
+                    "must": [{"text": {"query": base_term, "path": "plot"}}],
```
lsierant, line 174: explain the semantics between must and should. Is existence of should effectively ignored when must is present?

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,671 @@
...
+            except Exception:  # pragma: no cover
+                logger.debug("cursor.close() raised on cleanup")
+
+        return results
+
+    # ------------------------------------------------------------------
+    # Sentinel propagation — strongest "really upstream" check
+    # ------------------------------------------------------------------
+
+    def insert_sentinel(self, prefix: str = "kube17_sentinel") -> str:
```
lsierant: do not use kube17 str;

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,671 @@
...
+        """Insert a sentinel document and return its title.
+
+        The title is unique per call. After waiting for the search index to
+        re-tokenize (typically a few seconds), a ``$search`` for the title
```
lsierant: we don't tokenize in non-vector search queries. We build index.

lsierant: usages of it should use polling-style of wait for that, not fixed delay to avoid race conditions

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,671 @@
...
+        SHOULD return the document — if it doesn't, mongot is either stale
+        or unreachable. This is the strongest "the upstream search index is
+        actually current" check we can run from a client.
+        """
+        title = f"{prefix}_{uuid.uuid4().hex[:12]}"
```
lsierant: is the title fully random here?

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,671 @@
...
+        self._collection.insert_one({"title": title, "plot": "kube17 sentinel doc"})
+        return title
+
+    def search_for_sentinel(self, title: str, timeout_ms: int = 2000) -> QueryResult:
+        """Search for a sentinel document inserted via ``insert_sentinel``."""
+        stage = {"$search": {"text": {"query": title, "path": "title"}}}
+        return self.oneshot_search(query=stage, limit=5, timeout_ms=timeout_ms)
```
lsierant: do we poll for this doc or just blindly wait specified nr of seconds? We should poll to avoid races.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
```diff
@@ -0,0 +1,400 @@
+"""E2E test for the search connectivity tool (KUBE-17 / U1.1).
```
lsierant: don't mention U1.1, only jira ticket

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+@mark.e2e_search_connectivity_tool
+def test_create_search_index(mdb: MongoDB):
+    search_tester = get_rs_search_tester(mdb, USER_NAME, USER_PASSWORD, use_ssl=True)
+    search_tester.create_search_index("sample_mflix", "movies")
+    search_tester.wait_for_search_indexes_ready("sample_mflix", "movies", timeout=300)
+
+
+# ---------------------------------------------------------------------------
```
lsierant: Can we make the whole bootstrap above more reusable?

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
```diff
@@ -0,0 +1,400 @@
...
+        f"latency was {result.latency_ms:.1f}ms (threshold "
+        f"{tool.cache_latency_threshold_ms}ms)."
+    )
+
+    verdict = tool.verdict([result])
+    assert verdict.upstream_alive, f"verdict.upstream_alive should be True; got {verdict.as_dict()}"
+
+
+@mark.e2e_search_connectivity_tool
+def test_oneshot_vector_search_runs_or_skips(mdb: MongoDB):
```
lsierant: let's first focus on search queries, vector later.

lsierant: the test that's either skipped or passed is meaningless. let's remove it

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
```diff
@@ -0,0 +1,400 @@
...
+        # buffer-probe heuristic is the load-bearing signal here, not
+        # latency.
+        cache_latency_threshold_ms=10.0,
+    )
+
+    statefulset_name = search_resource_names.mongot_statefulset_name(MDBS_RESOURCE_NAME)
+    apps_v1 = client.AppsV1Api()
+
+    # Open the cursor while mongot is healthy. We deliberately use a
+    # large-ish total page count and small batch size so the local
```
lsierant: what is that large-ish total page count, I see pages=2

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
```diff
@@ -0,0 +1,400 @@
...
+    assert any(p.success and p.cache_hit_hint is False for p in pre_pages), (
+        "expected at least one upstream-confirmed page before scaling mongot down; "
+        "the cache-detection heuristic is broken before we even introduce a fault"
+    )
+
+    # Scale mongot StatefulSet to 0. The Service still resolves but has no
+    # endpoints, so subsequent getMore round-trips will fail (timeout or
+    # connection refused via envoy).
+    logger.info(f"scaling StatefulSet {statefulset_name} replicas -> 0")
+    apps_v1.patch_namespaced_stateful_set_scale(
```
lsierant: we should leverage the operator to downscale to zero using MongoDBSearch CR

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
```diff
@@ -0,0 +1,400 @@
...
+        ready = sts.status.ready_replicas or 0
+        return ready == 0, f"ready_replicas={ready}"
+
+    run_periodically(
+        mongot_pods_gone,
+        timeout=120,
+        sleep_time=5,
+        msg=f"mongot StatefulSet {statefulset_name} to scale to 0",
+    )
+    # Give kube-proxy a moment to update endpoints; envoy circuit breakers
```
lsierant: we don't need that, remove

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
```diff
@@ -0,0 +1,400 @@
...
+    # also need a beat to register the upstream as unhealthy.
+    time.sleep(5)
```
lsierant: don't use sleeps - it's race'y

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
```diff
@@ -0,0 +1,400 @@
...
+    # claim upstream is alive.
+    post_pages = tool.paging_search(pages=8, interval_seconds=0.5, batch_size=10)
+    logger.info("post-outage pages: %s", "; ".join(str(p) for p in post_pages))
+
+    post_verdict = tool.verdict(post_pages)
+    logger.info(f"post-outage verdict: {post_verdict.as_dict()}")
+
+    # The actual deliverable assertion: with mongot scaled to 0, no page
+    # should be reported as a fresh upstream success. The tool should
+    # surface either failures or cache-only successes.
```
lsierant: we must surface error here. Cache only success is not telling us anything. We need to achieve the connectivity or lost cursor error with the cursor. We'll introduce restarting mongot/envoy/mongods at some point to simulate losing long living cursors.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
```diff
@@ -0,0 +1,400 @@
...
+    # to 0.
+    assert (post_verdict.failed + post_verdict.cache_only_succeeded) > 0, (
+        f"post-outage verdict shows no failures and no cache-only successes — "
+        f"either mongot is still alive (test setup bug) or the connectivity "
+        f"tool isn't classifying results correctly. Verdict: {post_verdict.as_dict()}"
+    )
+
+    # Restore mongot for cleanup so subsequent tests in the same module can
+    # still run if anyone adds them.
+    apps_v1.patch_namespaced_stateful_set_scale(
```
lsierant: scale via the operator and wait for it to become ready normally.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Removed the U1.1 reference from the docstring at line 48 in commit bac540eb5; the "no unit tests with mocks" guidance now stands on its own without referring to the plan. Also dropped a lingering U1.3 follow-up mention from the test file in commit 38b6e273e.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Added a docstring on `_classify_error` (now connectivity.py lines 185-192) explaining the role: "We use the class name as a coarse, stringly-typed bucket for verdict aggregation (`ConnectivityVerdict.error_breakdown` keys on this) so callers can do `error_breakdown[\"OperationFailure\"] > 0` style checks without importing pymongo. The server error code is only present for `OperationFailure` (network errors do not have codes)." (commit bac540eb5)

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Renamed the cache-buster token prefix from `kube17_` to `cb_` (commit bac540eb5; see `make_cache_buster_query` line 214). No KUBE-* identifier remains in any variable or class name in the file.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Added a docstring on `make_cache_buster_query` (connectivity.py lines 200-213) that spells out the semantics: "`must` filters the result set — documents that do not match are excluded; `should` only contributes to the relevance score (a document still appears in the result even if it does not match `should`, just with a lower score). So the random `should` token does not change *which* documents come back; what it changes is the **query identity** — mongot keys its per-query cache on the full query text, and a fresh token forces a real evaluation rather than a cache lookup." (commit bac540eb5)

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Renamed the sentinel default prefix from `kube17_sentinel` to `sentinel` in `insert_sentinel` (commit bac540eb5; current connectivity.py line 600). The string `kube17` no longer appears anywhere in the connectivity-tool source.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Updated the `insert_sentinel` docstring (commit bac540eb5): replaced the misleading "tokenize" wording with "rebuild the search index against the underlying collection". The docstring also explicitly warns "never sleep a fixed amount and assume the document is queryable, that is racy".

lsierant: [AI] `search_for_sentinel` was rewritten as a real polling helper (commit bac540eb5) with three knobs: `overall_timeout_seconds=60`, `poll_interval_seconds=1.0`, `per_query_timeout_ms=2000`. It polls until `returned_count > 0` or the overall deadline elapses, no fixed-duration sleeps. The `insert_sentinel` docstring also points at this helper as the canonical caller pattern.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Yes — the `insert_sentinel` docstring now opens with "Insert a sentinel document and return its (fully random) title" and the implementation is `title = f"{prefix}_{uuid.uuid4().hex[:12]}"` (commit bac540eb5). Every call produces a unique, server-unseen value so mongot cannot serve the search query from cache.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Polled, not fixed-sleeped. `search_for_sentinel` now drives a polling loop with `overall_timeout_seconds`, `poll_interval_seconds`, and `per_query_timeout_ms` knobs (commit bac540eb5). The single-shot 2-second timeout is gone; it now keeps polling until the sentinel surfaces or the overall deadline elapses.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Module docstring now reads "E2E test for the search connectivity tool (KUBE-17)." — only the JIRA ticket, no U1.1 reference (commit bac540eb5). Also dropped a lingering "follow-up under U1.3" line from the test docstring in commit 38b6e273e, so no Ux.y references remain.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Extracted the 14-stage RS-managed-LB bootstrap into a new module `tests/common/search/connectivity_bootstrap.py` (commit bac540eb5). The test file now reads as thin pytest shells: `bootstrap.install_operator(namespace, operator_installation_config)`, `bootstrap.install_tls_certificates(helper, issuer, RS_MEMBERS)`, `bootstrap.create_database_resource(mdb)`, `bootstrap.create_search_resource(mdbs)`, etc. Sibling search e2es can adopt the same helpers without duplicating the bootstrap.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Vector-search functionality removed from this PR scope. The standalone vector test was deleted in commit bac540eb5; the test file now exercises only `$search` queries and paging.

lsierant: [AI] `test_oneshot_vector_search_runs_or_skips` removed entirely in commit bac540eb5. The always-skipped-on-cloud-qa, never-meaningful test is gone.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] The misleading "large-ish total page count" comment was removed in commit bac540eb5; the page counts in the test now scope correctly to their use sites (`pages=2` for the pre-outage probe across at least one getMore boundary, `pages=8` for the post-outage observation window).

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Switched to operator-driven scale-down via the MongoDBSearch CR. Required relaxing the CRD `+kubebuilder:validation:Minimum=1` to `Minimum=0` on three mongot replicas fields (top-level `spec.replicas`, `spec.clusters[].replicas`, `ShardOverride.Replicas`); `loadBalancer.managed.replicas` kept at `Min=1` (different concept). Also fixed `GetReplicas()` in `mongodbsearch_types.go` to honor an explicit 0 (was returning 1 silently because of `*Replicas > 0`). Added unit test `TestGetReplicasHonorsExplicitZero` and updated `TestHasMultipleReplicas`. Test now does `mdbs["spec"]["replicas"] = 0; mdbs.update()` and waits for `ready_replicas==0` (commit 38b6e273e).

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] The `time.sleep(5)` "give kube-proxy a moment to update endpoints" line is gone (commit bac540eb5). Endpoint readiness is now polled via `run_periodically(mongot_pods_gone, ...)` instead of a fixed delay.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] No fixed-duration sleeps remain in the test file. All wait-for-condition loops use `kubetester.run_periodically(..., timeout=N, sleep_time=K)` which polls until the condition is satisfied or the timeout elapses (commits bac540eb5 + 38b6e273e).

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Two changes address this:

1. The existing outage test asserts strictly: `post_verdict.failed > 0` plus a check that the observed error class is one of `{OperationFailure, ServerSelectionTimeoutError, NetworkTimeout, AutoReconnect, ConnectionFailure}` (commit bac540eb5). Cache-only success on its own is no longer accepted.

2. Added a new test `test_paging_through_mongot_pod_restart_surfaces_lost_cursor` (commit d566c0efd) that exercises the long-living-cursor scenario you flagged. It opens a paging cursor against healthy mongot, deletes the mongot pod via `core_v1.delete_namespaced_pod`, waits for the StatefulSet to bring back a fresh pod (watching by uid change), then continues paging on the SAME cursor. Asserts `post_verdict.cursor_lost > 0`.

The actual surface error turns out to be `OperationFailure(code=1, codeName=InternalError)` with message `Remote error from mongot :: caused by :: Received RST_STREAM with error code 2` rather than the canonical `CursorNotFound(code=43)`. The gRPC stream between mongod and mongot tears down when the mongot pod dies, and the new mongot pod has no record of the cursor session-side state. The new failure-class taxonomy in `connectivity.py` recognises both signal patterns and maps them to the `cursor_lost` bucket.

Local verdict from the new test: `cursor_lost=1, cursor_lost_observed=True, transient_network=0` — the cursor-lost signal surfaces deterministically.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Cleanup now does `mdbs["spec"]["replicas"] = 1; mdbs.update(); mdbs.assert_reaches_phase(Phase.Running, timeout=300)` — operator-driven scale-back, exercises the operator recovery path (StatefulSet recreation + readiness) instead of bypassing it with a direct StatefulSet patch (commit 38b6e273e).

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Vector-search functionality removed from this PR's scope (commit bac540eb5). Specifically: the vector_search / make_vector_search_query paths in connectivity.py were dropped; the test_vector_search_* test that was previously gated on AI_MONGODB_EMBEDDING_QUERY_KEY was removed (so the test no longer either-skips-or-passes — addresses the threaded follow-up too); only text/compound search queries remain. Vector-search will land in a follow-up once the embedding-API contract is locked down.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] The "large-ish" phrasing was a stale comment relative to the actual call. The current paging_search invocations are: pages=3 in test_paging_search_first_page_is_upstream, pages=2 (pre-outage warm-up) and pages=8 (post-outage probing) in the scale-to-0 test, and pages=2 (pre-restart warm-up) and pages=10 (post-restart probing) in the new lost-cursor test (search_connectivity_tool.py lines 230, 268, 303, 402, 467). Numbers match what the assertions actually need; misleading "large-ish" comment is gone (commits bac540eb5 / d566c0efd).

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Switched to operator-driven scale-down via the MongoDBSearch CR (commit 38b6e273e). The test now sets mdbs["spec"]["replicas"] = 0 and calls mdbs.update() (search_connectivity_tool.py lines 282-285), then waits for the operator to reconcile via polling on ready_replicas. The CRD minimum:1 constraint was relaxed to allow 0 (per the comment at lines 277-279 of the test). Direct StatefulSet patching is gone.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Removed the voyage_query_key_available import and its consumer (the auto-embed gating block in the vector-search assertion) when vector-search was dropped from PR scope (commit bac540eb5). That eliminated the unused-on-cloud-qa code path. If a different removal was intended, point at the file/line and I'll address it directly.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] All time.sleep() and bare sleep() calls removed from search_connectivity_tool.py (commits bac540eb5 + d566c0efd). The test now uses kubetester.kubetester.run_periodically for every wait condition: mongot_pods_gone polling for the scale-to-0 path (line 291), mdbs.assert_reaches_phase(Phase.Running) for cleanup wait (line 349), and a pod_recreated poll loop in the new lost-cursor test (line 453). `grep -nE "time\.sleep|\bsleep\(" tests/search/search_connectivity_tool.py` returns 0 matches.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Both signals are now surfaced explicitly. The existing scale-to-0 test (test_paging_through_mongot_outage_surfaces_connectivity_error) asserts failed > 0 AND that at least one error matches the connectivity-class taxonomy (OperationFailure "no healthy upstream", NetworkTimeout, AutoReconnect, ConnectionFailure, ServerSelectionTimeoutError) — cache-only success cannot satisfy the assertion. The new test_paging_through_mongot_pod_restart_surfaces_lost_cursor (commit d566c0efd) restarts the mongot pod mid-cursor and asserts cursor_lost > 0; the cursor-not-found error surfaces as "Executor error during getMore :: Remote error from mongot :: Received RST_STREAM with error code 2", classified as cursor_lost in the failure-class taxonomy. The Verdict dataclass exposes cursor_lost, transient_network, and other_failed buckets so future scenarios can target a specific failure class. Latest local e2e: 17 passed / 1 expected skip / 0 fail.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool.py
lsierant: [AI] Cleanup now uses operator-driven scale-up. The test sets mdbs["spec"]["replicas"] = 1, calls mdbs.update(), then waits via mdbs.assert_reaches_phase(Phase.Running, timeout=300) (search_connectivity_tool.py line 349). Direct StatefulSet patch in cleanup is gone (commit 38b6e273e).

api/v1/search/mongodbsearch_types.go
```diff
@@ -90,8 +90,11 @@ type MongoDBSearchSpec struct {
 	// For Sharded source: the number mongot pods per shard.
 	// When Replicas > 1, a load balancer configuration (spec.loadBalancer)
 	// is required to distribute traffic across mongot instances.
+	// 0 is allowed so operators (and operator-driven test harnesses) can
```
lsierant, line 98: don't mention test harnesses; 0 is allowed simply to allow to scale down to zero.

api/v1/search/mongodbsearch_types.go
```diff
@@ -928,9 +932,10 @@ func EffectiveClusters(s *MongoDBSearch) []ClusterSpec {
 func (s *MongoDBSearch) GetReplicas() int {
 	// Single legitimate read of the deprecated top-level field — this is the
 	// operator-side default ("1 when unset") for the legacy single-cluster path.
+	// An explicit 0 is honored (callers can take mongot offline via the CR).
```
lsierant, line 838: not take offline but scale to zero (use different wording)

controllers/operator/envoy_config_builder.go
```diff
@@ -186,6 +193,7 @@ func buildFilterChain(route envoyRoute, tlsEnabled bool, caKeyName string) (*lis
 		},
 		StreamIdleTimeout: durationpb.New(300 * time.Second),
 		RequestTimeout:    durationpb.New(300 * time.Second),
+		AccessLog:         accessLog,
```
lsierant, line 251: we need to configure unified logging outputted from the pod. Now the access log will be json, but the rest will be in a different (non-json format)

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool_sharded.py
```diff
@@ -0,0 +1,734 @@
...
+        core_v1 = client.CoreV1Api()
+
+        search_tester = get_search_tester(mdb, cfg.user_name, cfg.user_password, use_ssl=True)
+        tool = SearchConnectivityTool(search_tester)
+
+        # 1. Sanity: fanout works.
+        pre = tool.oneshot_search()
+        assert pre.success, f"pre-disrupt search failed: {pre.error_class} {pre.error_message}"
+
+        original_pod_count = sum(
```
lsierant: let's use a common function for that, we already have functions like _list_matching_pods;

docker/mongodb-kubernetes-tests/tests/common/search/testdata/fixtures/sharded-paging-clean/golden/cursor_trees.txt
```diff
@@ -0,0 +1 @@
+# sharded topology — see sharded_cursor_trees.txt
```
lsierant: why this is empty?

docker/mongodb-kubernetes-tests/tests/common/search/testdata/fixtures/sharded-paging-clean/golden/sharded_cursor_trees.txt
```diff
@@ -0,0 +1,10 @@
+
+=== per-sharded-cursor tree view — 1 cursor(s) ===
+
+  cursor 6216190576695441574  lsid=628a6b198d7142848bef97dc15a362e0  mongos=mdb-sh-conn-tool-mongos-0  num_shards=2  branches=0  wire_ops=6
+    mongos.aggregate  ts=2026-05-19T14:18:52.782000+00:00  dur_ms=2422  nShards=2
+    mongos.getMore  ts=2026-05-19T14:18:53.343000+00:00  dur_ms=0  nShards=2
```
lsierant: why do we have ony mongos here?

docker/mongodb-kubernetes-tests/tests/common/search/testdata/capture_client_wire_ops.py
```diff
@@ -0,0 +1,403 @@
...
+ADMIN_USER = os.environ.get("MCK_ADMIN_USER", "mdb-admin-user")
+ADMIN_PASS = os.environ.get("MCK_ADMIN_PASS", "mdb-admin-user-pass")
+
+
+# ---------------------------------------------------------------------------
+# kubectl helpers (subprocess — devc has kubectl on PATH)
+# ---------------------------------------------------------------------------
+
+
+def kubectl(args: list[str], *, check: bool = True) -> str:
```
lsierant: all those operations should rely on python kube api

docker/mongodb-kubernetes-tests/tests/common/search/testdata/capture_client_wire_ops.py
```diff
@@ -0,0 +1,403 @@
...
+        st = get_rs_search_tester(mdb, ADMIN_USER, ADMIN_PASS, use_ssl=True)
+    return SearchConnectivityTool(st)
+
+
+# ---------------------------------------------------------------------------
+# Fault injectors
+# ---------------------------------------------------------------------------
+
+
+def apply_fault(
```
lsierant: is this fault injection actually used anywhere? If not useful for log analysis then let's get rid of it.

docker/mongodb-kubernetes-tests/tests/common/search/testdata/capture_client_wire_ops.py
```diff
@@ -0,0 +1,403 @@
...
+
+class _MDB:
+    """Stand-in for an MCK MongoDB CR with .name/.namespace."""
+    def __init__(self, name: str, namespace: str):
+        self.name = name
+        self.namespace = namespace
+
+
+def make_tool(namespace: str, mdb_name: str, topology: str) -> SearchConnectivityTool:
+    mdb = _MDB(mdb_name, namespace)
```
lsierant: use proper MongoDB types

docker/mongodb-kubernetes-tests/tests/common/search/testdata/capture_client_wire_ops.py
```diff
@@ -0,0 +1,403 @@
...
+    tool = make_tool(namespace, mdb_name, topology)
+    pages_read = 0
+    fault_info: Optional[dict] = None
+    verbosity_was_bumped = False
+
+    try:
+        # Bump server-side verbosity so the analyzer's COMMAND/NETWORK
+        # joins have something to chew on. RS: hits the primary mongod.
+        # Sharded: hits mongos (per-shard mongod bump requires kubectl
+        # exec, which the e2e test does — out of scope here).
```
lsierant: why cannot we do it here? 
how e2e test is doing it, which one?

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,1554 @@
...
+        self.search_tester = search_tester
+        self.db_name = db_name
+        self.col_name = col_name
+        # pymongo only accepts event_listeners at MongoClient construction
+        # time — rebuild the client with the listener attached.
+        self._listener = _RecordingCommandListener()
+        self._install_listener_on_search_tester(search_tester, self._listener)
+
+    @staticmethod
+    def _install_listener_on_search_tester(
```
lsierant: let's reorder all _ methods in classes at the end of the class having the normal methods visible first in the class.

docker/mongodb-kubernetes-tests/tests/common/search/log_store.py
```diff
@@ -0,0 +1,1510 @@
...
+            ))
+        self.conn.executemany(
+            """INSERT INTO mongot_cmds (pod, timestamp, command, client_id, cursor_id)
+               VALUES (?, ?, ?, ?, ?)""",
+            rows,
+        )
+
+
+# ----------------------------------------------------------------------
+# SQL-backed builders — PR2: replaces build_cursor_trees
```
lsierant: remove mentions like PR2 and similar

docker/mongodb-kubernetes-tests/tests/common/search/log_store.py
```diff
@@ -0,0 +1,1510 @@
...
+    return final
+
+
+def build_sharded_cursor_trees_sql(
+    store: "LogStore",
+    *,
+    shard_pod_prefixes: dict,
+    shard_mongot_pod_prefixes: Optional[dict] = None,
+):
+    """SQL-backed equivalent of ``build_sharded_cursor_trees``.
```
lsierant: remove mentions of previous state, there is sql only now

docker/mongodb-kubernetes-tests/tests/common/search/log_store.py
```diff
@@ -0,0 +1,1510 @@
...
+
+    Returns the same ``list[ShardedCursorTree]`` shape so the existing
```
lsierant: state objectively without mentioning previous state

tmp/orchestration/reports/log_store_migration_plan.md
```diff
@@ -0,0 +1,1388 @@
+# SQLite analyzer migration — implementation plan
```
lsierant: remove the whole file and move it into tmp/ reports

docker/mongodb-kubernetes-tests/tests/common/search/all_nodes_availability_tool.py
```diff
@@ -0,0 +1,382 @@
...
+Spatial fan-out tool that drives ``SearchConnectivityTool`` against
+EVERY connectable mongod in a single-cluster MongoDB deployment — every
+RS member, or every mongod across every shard. Returns per-node
+``NodeAvailabilityResult`` objects so availability tests can express
+"is search reachable on every node?" in one call instead of hand-rolled
+``for member_index in range(N)`` loops.
+
+This module is **single-cluster only**. The multi-cluster sibling
+(fan-out across clusters AND members within each cluster) lives in the
+MC worktree and is built on the same ``NodeAvailabilityResult``
```
lsierant: there is no MC worktree anymore, we have everything now here; adjust similar mentions

docker/mongodb-kubernetes-tests/tests/common/search/all_nodes_availability_tool.py
```diff
@@ -0,0 +1,382 @@
...
+@dataclass
+class NodeAvailabilityResult:
+    """Per-node verdict from ``AllNodesAvailabilityTool`` fan-out.
+
+    ``oneshot_result`` is populated by ``oneshot_all`` / ``run``;
+    ``paging_results`` + ``paging_verdict`` are populated by ``paging_all``
+    / ``run``. A run-mode-specific call leaves the other side ``None`` /
+    empty so callers can branch on what was actually exercised.
+
+    ``served_by_mongot_pod`` is reserved for the multi-cluster sibling
```
lsierant: verify if this is actually needed

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -0,0 +1,1554 @@
...
+    server_connection_id: Optional[int] = None
+    lsid: Optional[str] = None  # hex form of the lsid UUID, or None
+    cursor_id: Optional[int] = None  # set on getMore (cursor in cmd) / aggregate reply
+    duration_micros: Optional[int] = None  # CommandSucceededEvent / CommandFailedEvent
+    n_returned: Optional[int] = None  # extracted from reply.cursor.firstBatch/nextBatch
+    database_name: Optional[str] = None
+    failure: Optional[str] = None  # error description for "failed"
+
+
+class _RecordingCommandListener(pymongo.monitoring.CommandListener):
```
lsierant: verify if we actually need the CommandListeners

docker/mongodb-kubernetes-tests/tests/common/search/background_availability_tester.py
```diff
@@ -0,0 +1,303 @@
+"""Background availability tester for MongoDBSearch (KUBE-26).
```
lsierant: the filename should be background_availability_tester

docker/mongodb-kubernetes-tests/tests/common/search/bootstrap_test_mixins.py
```diff
@@ -0,0 +1,692 @@
+"""Reusable pytest test-class mixins for the RS managed-LB search bootstrap.
+
+The 15-step "deploy operator → deploy MongoDB → deploy MongoDBSearch →
+load sample data → build search index" sequence was duplicated almost
+verbatim across the KUBE-17 connectivity-tool e2e, the KUBE-26
+background-tester e2e, and the KUBE-27 failure-modes e2e. This module
```
lsierant: don't mention previous tickets; state objectively what has been done here and how the dedupication works;

docker/mongodb-kubernetes-tests/tests/search/_probe_cache_layers.py
```diff
@@ -0,0 +1,345 @@
+"""Empirical probe: where (if anywhere) does paging-search cache?
```
lsierant: verify if we still need those _probe files.

controllers/operator/envoy_config_builder.go
lsierant: [AI] Added `--log-format '{"time":"%Y","level":"%l","logger":"%n","thread":%t,"source":"%g:%#","message":"%j"}'` to the envoy container args in `mongodbsearchenvoy_controller.go`, so the component log (incl. the `http2:debug` per-frame output) is emitted as JSON in the same shape as the existing HCM access log. The `%j` escape keeps the bracketed `[Tags: "ConnectionId":...,"StreamId":...]` body well-formed inside the surrounding JSON. The test-side parser was updated: `parse_envoy_debug_log` now calls a new `_envoy_line_message_and_ts` helper that does `json.loads(line)` for JSON-format lines and extracts the bracketed body from `message` before feeding it through the existing connection/stream regex set. The plain-text bracketed-timestamp fallback path is preserved so older fixtures keep working.

docker/mongodb-kubernetes-tests/tests/search/search_connectivity_tool_sharded.py
lsierant: [AI] Added a shared `list_pods_by_prefix(core_v1, namespace, prefix)` helper in `tests/common/search/connectivity.py` (single source of truth for the "list pods whose `metadata.name.startswith(prefix)`" pattern) and switched both the `search_connectivity_tool_sharded.py:682` pod-count check and the existing `hard_kill_pods_by_prefix` / `delete_mongot_pod_for_cursor` call sites in `connectivity.py` to use it instead of inlining the comprehension.

docker/mongodb-kubernetes-tests/tests/common/search/testdata/fixtures/sharded-paging-clean/golden/cursor_trees.txt
lsierant: [AI] DEFER. The empty `cursor_trees.txt` is the placeholder that `generate_golden.py` writes for sharded fixtures because the RS-shaped per-cursor tree builder (`build_cursor_trees_sql`) has no useful output once mongos sits in front of the mongods — the meaningful view for sharded is in the sibling `sharded_cursor_trees.txt`. Two options for a real fix and both belong in a separate change rather than this PR: (a) stop emitting the placeholder for sharded fixtures (drop the `cursor_trees.txt.write_text("# sharded topology ...")` block in `generate_golden.py`) and teach `test_analyzer_golden.py` not to assert its presence for sharded; (b) make `build_cursor_trees_sql` join mongos-anchored ops as their own per-cursor tree alongside the per-shard view. I left the file as-is in this PR since this is the connectivity-tool PR, not the analyzer refactor — happy to file a follow-up KUBE ticket.

docker/mongodb-kubernetes-tests/tests/common/search/testdata/fixtures/sharded-paging-clean/golden/sharded_cursor_trees.txt
lsierant: [AI] DEFER. The mongos-only fanout in `sharded_cursor_trees.txt` is the symptom of `build_sharded_cursor_trees_sql` returning `branches=0` for this fixture even though the `unified_timeline.txt` clearly shows mongot interceptor / batch traffic on both shard mongots (`mdb-sh-conn-tool-search-0-mdb-sh-conn-tool-0-0` and `-1-1`). The join from a top-level mongos `aggregate` to per-shard sub-cursors goes through a per-shard time-window match against the mongod `aggregate` / `getMore` records, and on this capture the shard mongod COMMAND/NETWORK lines are missing because the capture script only bumped verbosity on mongos. I addressed the upstream cause in the same PR: `capture_client_wire_ops.py` now iterates the shards via `get_shard_mongod_tester` and bumps each shard primarys `command`/`network` levels to 2 (matching what the e2e test path does). Regenerating the fixture against an updated capture run is the right way to surface the per-shard branches in the golden — that regeneration is out of scope here and should land alongside the fixture refresh; left a TODO in the commit message.

docker/mongodb-kubernetes-tests/tests/common/search/testdata/capture_client_wire_ops.py
lsierant: [AI] Rewrote the kubectl-subprocess shims at the top of `capture_client_wire_ops.py` to use the python kubernetes client. `list_pods` now goes through `CoreV1Api().list_namespaced_pod(namespace=...)`; `get_shard_count` reads the MongoDB CR via `CustomObjectsApi().get_namespaced_custom_object("mongodb.com", "v1", ...)`; `fetch_pod_log` calls `read_namespaced_pod_log(name=..., namespace=..., since_seconds=...)`; `delete_pod` issues `delete_namespaced_pod(..., body=V1DeleteOptions(grace_period_seconds=0))`. Kubeconfig is loaded lazily via `config.load_kube_config()` with `load_incluster_config()` fallback so importing the module from a non-cluster context (e.g. unit tests) does not error. The `subprocess` import and the unused `restart_deployment` helper are removed.

docker/mongodb-kubernetes-tests/tests/common/search/testdata/capture_client_wire_ops.py
lsierant: [AI] Confirmed by grep that nothing in `tests/` or `scripts/` invokes this script with `--fault`; the only references were in the `MIGRATION_PLAN.md` doc (which is being removed in this PR per another comment) and the docstring of the script itself. Removed the entire `apply_fault` function, the `--fault` / `--fault-after-pages` / `--fault-settle-seconds` CLI args, the fault-aware branch in `run_capture` (now just calls `tool.paging_search(...)` directly), the `fault_info` slot in the metadata dict, and the fault example block in the module docstring.

docker/mongodb-kubernetes-tests/tests/common/search/testdata/capture_client_wire_ops.py
lsierant: [AI] Replaced the local `_MDB` stand-in class with `kubetester.mongodb.MongoDB(name=mdb_name, namespace=namespace).load()`. `make_tool` now passes a real `MongoDB` instance into `get_search_tester` / `get_rs_search_tester`, matching what the e2e helpers do. The per-shard verbosity bump (separate comment) also benefits from this — it reads `mdb["spec"]["shardCount"]` (with a `get_shard_count` API fallback) off the same loaded object.

docker/mongodb-kubernetes-tests/tests/common/search/testdata/capture_client_wire_ops.py
lsierant: [AI] Done — added the per-shard mongod bump in `run_capture`. For `topology=="sharded"` the script now (a) bumps mongos COMMAND/NETWORK to 2 via `set_mongos_debug_logs` (as before), and (b) for each `shard_index in range(shard_count)` builds a `SearchTester` via `get_shard_mongod_tester(mdb, shard_index, member_index=0, ...)` (the directConnection mongod factory the sharded e2e `search_connectivity_tool_sharded` uses — see `tests/common/search/sharded_search_helper.py`) and calls `set_mongod_debug_logs(shard_tester.client, command_level=2, network_level=2)` on it. The `finally` block restores levels to 0 on every shard client and closes them; `shard_count` is read off the loaded MongoDB CR spec with the API fallback.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Reordered `SearchConnectivityTool` so the public surface comes first and every underscore-prefixed helper is grouped under a single "Private helpers" banner at the very end of the class. Order is now: `__init__` → `listener` property → `make_cache_buster_query` → `oneshot_search` / `oneshot_vector_search` → `paging_search` / `paging_cursor_open` / `paging_cursor_read_pages` → `insert_sentinel` / `search_for_sentinel` → `verdict` → private block (`_install_listener_on_search_tester`, `_collection`, `_classify_error`, `_cursor_id`, `_begin_capture`, `_end_capture`, `_count_wire_ops`, `_default_paging_stage`, `_paging_open_and_first_error`). The duplicate `listener` property has been collapsed. Other classes in the file already follow this layout; only `SearchConnectivityTool` was interleaved.

docker/mongodb-kubernetes-tests/tests/common/search/log_store.py
lsierant: [AI] Replaced the `# SQL-backed builders — PR2: replaces build_cursor_trees` section banner in `log_store.py` with a neutral `# SQL-backed cursor-tree builders` heading, and rewrote the module top docstring so it describes the store + its builders directly instead of framing the file as "Phase 1 of the SQLite migration (KUBE-17), purely additive, parity-test against the Python builder it is replacing".

docker/mongodb-kubernetes-tests/tests/common/search/log_store.py
lsierant: [AI] Rewrote the `build_sharded_cursor_trees_sql` docstring. It no longer says "SQL-backed equivalent of `build_sharded_cursor_trees`" (which implied a parallel Python builder still exists); it now reads "Build per-cursor sharded fanout trees from the log store" and just states the join semantics directly (per-shard time-window for the mongos fanout, ±5s session match, ±2s mongot fallback). Same edit on `test_analyzer_golden.py` to drop the "Post-PR5 the only backend …" framing.

docker/mongodb-kubernetes-tests/tests/common/search/log_store.py
lsierant: [AI] Same rewrite as the comment one line up: the `build_sharded_cursor_trees_sql` docstring now states what it does and the join shape ("per-shard time-window for the mongos fanout, ±5s session match, ±2s mongot fallback") instead of describing itself as a parity-targeted rewrite of a prior `build_sharded_cursor_trees` Python builder. The "Mirrors the Python builders join semantics" / "regression test sees byte-identical golden output" sentences are gone.

tmp/orchestration/reports/log_store_migration_plan.md
lsierant: [AI] `git mv` of `docker/mongodb-kubernetes-tests/tests/common/search/testdata/MIGRATION_PLAN.md` to `tmp/orchestration/reports/log_store_migration_plan.md` — the file is now part of the orchestration reports tree, not the test data. Also cleaned up the only in-tree pointer to it (a `# scaffolding for PR1` reference in the modules section banner was already removed by the docstring edits in another comment).

docker/mongodb-kubernetes-tests/tests/common/search/all_nodes_availability_tool.py
lsierant: [AI] Rewrote the module docstring of `all_nodes_availability_tool.py` to drop the "single-cluster only / MC sibling lives in the MC worktree" framing. The new lead-in just describes the spatial fan-out primitive over every RS member or every shard mongod and points at the two factory entrypoints; no claim about a separate multi-cluster file. The `served_by_mongot_pod` field docstring, which also referenced "the multi-cluster sibling", is gone too (the field itself is removed per the next comment).

docker/mongodb-kubernetes-tests/tests/common/search/all_nodes_availability_tool.py
lsierant: [AI] Verified — `served_by_mongot_pod` is declared on `NodeAvailabilityResult` but no production / test code anywhere in `docker/mongodb-kubernetes-tests/` reads or writes it (only references are the field declaration itself and its docstring). Removed the field from the dataclass and the surrounding "reserved for the multi-cluster sibling" docstring paragraph; the contract now matches what is actually populated by `oneshot_all` / `paging_all`.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
lsierant: [AI] Verified — CommandListener is the load-bearing instrumentation in this PR, not optional. `tool.listener.snapshot_since(...)` feeds `parse_client_wire_ops` in the sharded e2e (`search_connectivity_tool_sharded.py:206-207`), and the listener-derived `mongod_wire_ops` / `wire_ops` fields on `QueryResult` are asserted in `search_connectivity_tool.py:98,113,126,138,362,807,886,895`, `multicluster_search/search_connectivity_tool_mc_rs.py:128,143,181,225,285`, `search_connectivity_tool_sharded.py:172,457`, `search_availability_background_tester.py:179`, and `all_nodes_availability_tool.py:93,284` — every connectivity-tool e2e relies on it to distinguish "real wire op" from "buffer-only" pages. Keeping the `_RecordingCommandListener` as-is; the alternative (parsing mongos/mongod NETWORK logs after the fact) loses the per-call scoping that `_begin_capture` / `_end_capture` provides.

docker/mongodb-kubernetes-tests/tests/common/search/background_availability_tester.py
lsierant: [AI] `git mv tests/common/search/availability_tester.py tests/common/search/background_availability_tester.py`. Updated the two importers — `tests/search/search_failure_modes.py:66` and `tests/search/search_availability_background_tester.py:43` now both read `from tests.common.search.background_availability_tester import SearchAvailabilityBackgroundTester`. No other callers exist.

docker/mongodb-kubernetes-tests/tests/common/search/bootstrap_test_mixins.py
lsierant: [AI] Rewrote the `bootstrap_test_mixins.py` top docstring. The opening paragraph no longer names previous tickets (was: "duplicated almost verbatim across the KUBE-17 connectivity-tool e2e, the KUBE-26 background-tester e2e, and the KUBE-27 failure-modes e2e"); it now states the current behaviour objectively — three sibling test-class mixins parented at `SearchE2EFixtures`, each covering one layer of the deploy → search → data bootstrap; deduplication works because all three search e2es pull in only the layers they need and stay in sync by sharing the mixin code rather than copying it.

docker/mongodb-kubernetes-tests/tests/search/_probe_cache_layers.py
lsierant: [AI] Verified — none of the `_probe_*.py` scripts under `tests/search/` are invoked by pytest (they are standalone scripts with `if __name__ == "__main__"`), and none are imported from any production / test code. The only in-tree reference was a docstring `_probe_envoy_debug_h2.py` pointer in `mongot_log_analyzer.py`. Deleted all seven probe files (`_probe_cache_layers.py`, `_probe_envoy_debug_h2.py`, `_probe_envoy_stats_signal.py`, `_probe_midcursor_kill.py`, `_probe_mongod_verbosity.py`, `_probe_pymongo_client_ids.py`, `_probe_stream_correlate.py`) via `git rm`, and rewrote the two docstring references in `mongot_log_analyzer.py` so the envoy debug-logging instructions stand on their own (POST `/logging?paths=http2:debug,http:debug,router:debug` on the admin endpoint) without naming a probe file.

controllers/operator/envoy_config_builder.go
```diff
@@ -342,6 +350,70 @@ func buildDownstreamTLSTransportSocket(caKeyName string) (*corev3.TransportSocke
...
+//   - %UPSTREAM_HOST% — which mongot endpoint envoy actually picked.
+//   - %BYTES_RECEIVED% / %BYTES_SENT% / %DURATION% — basic per-stream
+//     traffic stats for the post-hoc test report.
+//
+// JSON formatting keeps the per-line schema stable across envoy releases
+// and is easier for the analyzer's parser to consume than the default
+// text format.
+func buildHCMAccessLog() ([]*accesslogv3.AccessLog, error) {
+	jsonFields, err := structpb.NewStruct(map[string]interface{}{
+		"ts":             "%START_TIME%",
```
lsierant: we need to align the format here with the envoy log:
"--log-format", `{"time":"%Y","level":"%l","logger":"%n","thread":%t,"source":"%g:%#","message":"%j"}`,
	
both should be compatible. Also adjust the log collector/analyser accordingly

lsierant: both logs will be interleaved - how to make it obvious that those are access logs and not envoy logs? What's the common practice here with envoy?

controllers/operator/mongodbsearchenvoy_controller.go
```diff
@@ -611,7 +611,16 @@ func buildEnvoyPodSpec(search *searchv1.MongoDBSearch, clusterIndex int, tlsCfg
...
 				Command: []string{"/usr/local/bin/envoy"},
-				Args:    []string{"-c", "/etc/envoy/envoy.json", "--log-level", "info"},
+				// --log-format emits component logs as JSON so they share a
+				// uniform shape with the JSON access log we configure in the
+				// HCM (see buildHCMAccessLog). %j escapes the message body
+				// (which can carry quotes from envoy "[Tags: ... ]" frames)
+				// so each line stays as a single well-formed JSON object.
+				Args: []string{
+					"-c", "/etc/envoy/envoy.json",
+					"--log-level", "info",
```
lsierant, line 636: let's create a new jira ticket under KUBE-4 epic, label envoy with making envoy's log level configurable. We should have this configurable under spec.cluster.loadBalancer.logLevel: <level>

docker/mongodb-kubernetes-tests/kubetester/__init__.py
```diff
@@ -405,6 +405,119 @@ def get_pod_when_ready(
...
+            name_prefix=name_prefix,
+            api_client=api_client,
+        )
+        if not pods:
+            return False, "no pods"
+        ready = [p for p in pods if _pod_is_ready(p)]
+        want = expected_count if expected_count is not None else len(pods)
+        return len(ready) >= want, f"ready={len(ready)}/{len(pods)} want={want}"
+
+    selector = label_selector or f"name~{name_prefix}*"
```
lsierant, line 464: f"name~{name_prefix}*" - is ambiguous; make it explicit log as name={name_prefix}

controllers/operator/envoy_config_builder.go
```diff
@@ -93,12 +95,15 @@ func buildEnvoyBootstrapConfig(routes []envoyRoute, tlsEnabled bool, caKeyName s
...
-			// enable just some endpoints because it's recommended to not enable al the admin endpoints by default
+			Address: socketAddress("0.0.0.0", uint32(EnvoyAdminPort)),
+			// enable just some endpoints because it's recommended to not enable all the admin endpoints by default.
+			// /logging is whitelisted so e2e log-gather tooling can bump per-frame http2/http verbosity at runtime
+			// (POST /logging?paths=http2:debug,...) without rebuilding the envoy image.
 			AllowPaths: []*matcherv3.StringMatcher{
 				{MatchPattern: &matcherv3.StringMatcher_Exact{Exact: "/ready"}},
 				{MatchPattern: &matcherv3.StringMatcher_Prefix{Prefix: "/stats"}},
 				{MatchPattern: &matcherv3.StringMatcher_Exact{Exact: "/drain_listeners"}},
+				{MatchPattern: &matcherv3.StringMatcher_Prefix{Prefix: "/logging"}},
```
lsierant: logging is useful for enabling debug log at runtime to investigate issues.

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -48,6 +48,12 @@ const (
 
 	ForceWireprotoAnnotation = "mongodb.com/v1.force-search-wireproto"
 
+	// ResourceDisabledAnnotation, when set to "true" on a MongoDBSearch CR,
+	// short-circuits the reconciler: it returns Result{} + nil without
+	// mutating any owned objects. Useful for tests that need to mutate
+	// owned StatefulSets directly without the operator reverting them.
+	ResourceDisabledAnnotation = "mongodb.com/resourceDisabled"
```
viveksinghggits: nit:
```suggestion
	// DisableReconciliationAnnotation, when set to "true" on a MongoDBSearch CR,
	// short-circuits the reconciler: it returns Result{} + nil without
	// mutating any owned objects. Useful for tests that need to mutate
	// owned StatefulSets directly without the operator reverting them.
	DisableReconciliationAnnotation = "mongodb.com/disable-reconciliation"
```

docker/mongodb-kubernetes-tests/kubetester/__init__.py
```diff
@@ -405,6 +405,119 @@ def get_pod_when_ready(
...
+    """
+    if not label_selector and not name_prefix:
+        raise ValueError("must provide label_selector or name_prefix")
+    kwargs: Dict[str, Any] = {}
+    if label_selector:
+        kwargs["label_selector"] = label_selector
+    pods = client.CoreV1Api(api_client=api_client).list_namespaced_pod(namespace, **kwargs).items
+    if name_prefix:
+        pods = [p for p in pods if p.metadata.name.startswith(name_prefix)]
+    return pods
```
viveksinghggits, line 427: I think we should have two functions, one to filter using labels and other to filter using names, considering `name_prefix` and `label_selector` are mutually exclusive.

docker/mongodb-kubernetes-tests/kubetester/__init__.py
```diff
@@ -405,6 +405,119 @@ def get_pod_when_ready(
...
+
+
+def pod_is_ready(pod: client.V1Pod) -> bool:
+    conds = pod.status.conditions or []
+    return any(c.type == "Ready" and c.status == "True" for c in conds)
```
viveksinghggits, line 432: do we not have these utilities written already? I think we should have.

docker/mongodb-kubernetes-tests/tests/common/search/background_availability_tester.py
```diff
@@ -0,0 +1,282 @@
...
+    def wait_for_operations(
+        self,
+        count: int,
+        *,
+        since: Optional[int] = None,
+        timeout: float = 120.0,
+        stop_on_fault: bool = False,
+    ) -> int:
+        """Block until ``count`` more operations (success or failure) are
+        recorded beyond ``since`` (defaults to the current count). With
```
viveksinghggits, line 137: what exactly is current count here? is it `self.operations_count`?

docker/mongodb-kubernetes-tests/tests/common/search/bootstrap_test_mixins_mc.py
```diff
@@ -0,0 +1,566 @@
...
+logger = test_logger.get_test_logger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Per-layer configuration dataclasses.
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class MongoDBMultiRsDeploymentConfig:
```
viveksinghggits: very nit:  instead of calling this `DeploymentConfig` can we not call it `InstallationConfig`, because DeploymentConfig is an OCP resource I got confused for a second whether it has to do anything with OCP deployment config.

viveksinghggits (commented):
skimmed through the PR. looks good overall, left some nits.

docker/mongodb-kubernetes-tests/tests/common/search/background_availability_tester.py
lsierant: yes, operation, one getMore or one query.

viveksinghggits:
evergreen retry

ashortt-mdb:
LGTM overall — nice work on the connectivity harness. 👍

One heads-up for whoever picks up the `private_kind_code_snippets` failure (`createSearchIndex` → *"Error connecting to Search Index Management service"*): **the mongot image version is a red herring** — not worth chasing down that path.

The final commit (`f154c98cc`) reverts `MDB_SEARCH_VERSION` from `9872ab5089` (the temp clientId/cursorId debug-log build, v1.65.1 base) back to `1.64.0-66-g7b0f85e89c`. But that's just restoring the baseline — `search/ga-base` already pins the same build, and so does the PR tip:

```
ga-base:  1.64.0-66-g7b0f85e89c
PR tip:   1.64.0-66-g7b0f85e89c   ← identical
```

So PR and base run the same mongot, which rules the version out as the differentiator for the snippet failure. Whatever's breaking `createSearchIndex` is in the diff itself, not the image.

viveksinghggits:
> LGTM overall — nice work on the connectivity harness. 👍
> 
> One heads-up for whoever picks up the `private_kind_code_snippets` failure (`createSearchIndex` → _"Error connecting to Search Index Management service"_): **the mongot image version is a red herring** — not worth chasing down that path.
> 
> The final commit (`f154c98cc`) reverts `MDB_SEARCH_VERSION` from `9872ab5089` (the temp clientId/cursorId debug-log build, v1.65.1 base) back to `1.64.0-66-g7b0f85e89c`. But that's just restoring the baseline — `search/ga-base` already pins the same build, and so does the PR tip:
> 
> ```
> ga-base:  1.64.0-66-g7b0f85e89c
> PR tip:   1.64.0-66-g7b0f85e89c   ← identical
> ```
> 
> So PR and base run the same mongot, which rules the version out as the differentiator for the snippet failure. Whatever's breaking `createSearchIndex` is in the diff itself, not the image.

I am looking into the code snippets failure and it's about the connectivity between loadbalancer and mongot. It looks like mongots have updated/new labels now and LB service is not changed for those new labels.

PR #1081, opened 2026-05-08
KUBE-72: per-worktree NAMESPACE for cloud-qa OM project isolation

lsierant:
[skip-ci] Stacked dev-tooling fixup off `lsierant/devcontainer` — base is **not** master, so this PR stays draft permanently per the team's stacked-PR convention. Lukasz merges via local git push; GH auto-closes.

## Problem

Every parallel devcontainer worktree on the same host shared one `NAMESPACE`. Cloud-qa Ops Manager projects created during e2e are named `${NAMESPACE}` and `${NAMESPACE}-{slug}`, so two parallel worktrees both manipulated the **same shared cloud-qa namespace** of OM projects. The destructive cleanup at the start of `make prepare-local-e2e` (`scripts/dev/delete_om_projects.sh`) deleted everything matching `${NAMESPACE}` / `${NAMESPACE}-*`, including the peer worktree's live projects mid-run. Symptoms looked like flake (random "project not found", agent registration errors, cert-hash mismatches) but were structural.

## Fix

Reuse the existing per-worktree `MCK_DEVC_NET_PREFIX` (numeric 16..31, picked by `dc_select_network.sh` and persisted in `.devcontainer/.env`) to suffix `NAMESPACE`. Whatever the user set in their personal `private-context` becomes `${NAMESPACE}-${MCK_DEVC_NET_PREFIX}` — every downstream OM-project-name composition site (`configure_operator.sh` x2, `prepare-multi-cluster/main.go`, ~15 pytest fixtures, `delete_om_projects.sh`'s filter) automatically scoped — **zero changes** to those sites. Idempotent guard skips re-suffixing when `NAMESPACE` already ends in `-${MCK_DEVC_NET_PREFIX}`.

## Deliverables

- **`scripts/dev/contexts/root-context`** — append `-${MCK_DEVC_NET_PREFIX}` to whatever `NAMESPACE` the user's `private-context` chose. Loads the prefix from `.devcontainer/.env` if not already in env, so host-side `make switch` (run before the container is up) sees the same value as the container.
- **`.devcontainer/compose.yml`** — forward `MCK_DEVC_NET_PREFIX` into the devcontainer's environment block so root-context (sourced inside the container during `make switch`) sees it.
- **`scripts/dev/delete_om_projects.sh`** — mirror of the same derivation, so host-side cleanup callers (`wt_teardown.sh`) delete the right scope without first having to refresh the worktree's context via `make switch`.
- **`scripts/dev/wt_teardown.sh`** — adds Step 1: run `delete_om_projects.sh` against the worktree's `${NAMESPACE}-${prefix}-*` scope before tearing down compose / removing the worktree, so torn-down worktrees don't orphan their OM projects. New `--keep-om-projects` flag to skip.
- **`scripts/test/bash/worktree/test_om_project_namespace.bats`** — 6 new bats covering: appends-to-NAMESPACE behavior, idempotency on re-sourcing, no-prefix fallback, `.devcontainer/.env` loading, env-set wins over file, and the `delete_om_projects.sh` mirror. `setup()` snapshots `private-context` + `.devcontainer/.env` and stubs `private-context` with a known `NAMESPACE`; `teardown()` restores both. Bats guarantees teardown runs even on crash, so the worktree is never left dirty.

## Fixup-commit shape

This PR is a **stack of `fixup!` commits** all targeting `141a0a6c8 Devcontainers in isolated worktrees with evg hosts`. They are intended to be `git rebase --autosquash`'d into the original Devcontainers commit before the parent branch eventually merges. Do not expect a separate squash-merge.

## EVG CI back-compat

- EVG tasks set `NAMESPACE` via `evg-private-context` (random per-task value via `generate_random_namespace` written to `${NAMESPACE_FILE}`) and don't set `MCK_DEVC_NET_PREFIX`, so the suffix branch never fires and behavior is unchanged.
- GHA's `gha-private-context` exports `NAMESPACE=""` (empty) and likewise no prefix, no suffixing happens.
- No pipeline (`.evergreen*.yml`) edits needed.

## Stale-project cleanup

Use the existing `scripts/evergreen/e2e/setup_cloud_qa.py delete_all` — it age-cleans every project older than 70 minutes in the org via `clean_unused_projects(org)`. Pre-fix orphans (any project not actively in use) get swept by the next run. No standalone bash cleanup script needed.

## Proof of Work

- **EVG patch (round 3):** https://evergreen.mongodb.com/version/69fe04efe6c2a60007ca5d0f — green-equivalent.
  - `unit_tests_golang` / `unit_tests_helm` / `unit_tests_python` / `bash_tests` / `go_bump_policy_tests` — all passed.
  - `lint_repo` — `govulncheck` only (KUBE-71).
- **Local validation:**
  - 17/17 worktree bats green (`bash scripts/test/bash/run.sh worktree`).
  - `pre-commit run --files <changed>` clean (shellcheck, etc.).
- **e2e:** N/A (pure dev-tooling change — no controllers/, api/, helm_chart/, public/crds.yaml, pkg/ touched).
- **JIRA:** https://jira.mongodb.org/browse/KUBE-72 (linked under epic [KUBE-4](https://jira.mongodb.org/browse/KUBE-4)).

### Patch history

- **Round 1** (`69fdf2295882250007b10af0`) — initial submission. green-equivalent (unit_tests_python flake on unrelated `release_info_test`, recovered on retry).
- **Round 2** (`69fe01e383e34600079e643c`) — applied 8 review comments; bash_tests broke because the regex used to extract the user's NAMESPACE from `private-context` didn't match EVG's `NAMESPACE_FILE`-driven shape.
- **Round 3** (`69fe04efe6c2a60007ca5d0f`) — bats now stub `private-context` with a known `NAMESPACE` for determinism. Green-equivalent.

## Known unrelated failures

- **`lint_repo` / `govulncheck`** — 4 GO stdlib CVEs flagged (GO-2026-4982, GO-2026-4980, GO-2026-4971, GO-2026-4918). Pre-existing master-level issue; tracked under **KUBE-71**. Not introduced by this PR (no Go files touched).


scripts/dev/contexts/root-context
```diff
@@ -9,6 +9,34 @@ source "${script_dir}/private-context"
...
+# prefix as the in-container flow.
+if [[ -z "${MCK_DEVC_NET_PREFIX:-}" && -f "${PROJECT_DIR}/.devcontainer/.env" ]]; then
+  devc_prefix_line="$(grep '^MCK_DEVC_NET_PREFIX=' "${PROJECT_DIR}/.devcontainer/.env" 2>/dev/null | tail -n1 || true)"
+  if [[ -n "${devc_prefix_line}" ]]; then
+    export "${devc_prefix_line?}"
+  fi
+  unset devc_prefix_line
+fi
+
+# Only applies when NAMESPACE is the default 'ls' — explicit user overrides
```
lsierant: ls is not a default of any sort; it's just Lukasz's namespace from private-context. 'ls' should not be mentioned anywhere.

scripts/dev/cleanup_legacy_ls_om_projects.sh
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env bash
+#
+# One-shot cleanup of legacy `ls` / `ls-*` cloud-qa OM projects.
```
lsierant: repurpose the cleanup for a namespace prefix - other could have different ones. How this compares to existing cloudqa  delete_all python script? isn't it redundant?

scripts/dev/cleanup_legacy_ls_om_projects.sh
```diff
@@ -0,0 +1,111 @@
...
+
+echo "==> Listing all cloud-qa OM projects (this may take a moment)..."
+all_projects="$(curl -s -u "${OM_USER}:${OM_API_KEY}" --digest \
+  "${OM_HOST}/api/public/v1.0/groups?itemsPerPage=500" \
+  | jq -r '.results[]?.name' | sort -u)"
+
+# Legacy pattern: name == "ls" OR (name starts with "ls-" AND second segment is
+# NOT a 2-digit number in [16,31]). The new per-worktree pattern is
+# `ls-<digit>...` where the digit is one of 16..31 (MCK_DEVC_NET_PREFIX range).
+legacy="$(echo "${all_projects}" \
```
lsierant: this is too much of awk scripting!

lsierant: let's figure out if this script is redundant and perhaps expand on the python one for cloudqa

scripts/dev/delete_om_projects.sh
```diff
@@ -11,6 +11,28 @@ test "${MDB_BASH_DEBUG:-0}" -eq 1 && set -x
...
+# made before the worktree context is refreshed) delete the right scope.
+if [[ -z "${MCK_DEVC_NET_PREFIX:-}" && -f "${PROJECT_DIR:-.}/.devcontainer/.env" ]]; then
+  devc_prefix_line="$(grep '^MCK_DEVC_NET_PREFIX=' "${PROJECT_DIR:-.}/.devcontainer/.env" 2>/dev/null | tail -n1 || true)"
+  if [[ -n "${devc_prefix_line}" ]]; then
+    export "${devc_prefix_line?}"
+  fi
+  unset devc_prefix_line
+fi
+if [[ -n "${MCK_DEVC_NET_PREFIX:-}" && "${NAMESPACE:-}" == "ls" ]]; then
+  NAMESPACE="ls-${MCK_DEVC_NET_PREFIX}"
```
lsierant: ls again to be removed

scripts/test/bash/worktree/test_om_project_namespace.bats
```diff
@@ -0,0 +1,218 @@
...
+# We test the derivation by sourcing each script in isolation with a stubbed
+# environment and asserting the resulting NAMESPACE.
+
+setup() {
+    REAL_PROJECT_DIR="$(git rev-parse --show-toplevel)"
+
+    # Build a sandbox dir that mimics a worktree's relevant tree:
+    #   <sandbox>/.devcontainer/.env         (carries MCK_DEVC_NET_PREFIX)
+    #   <sandbox>/scripts/dev/contexts/...   (real scripts, copied)
+    #   <sandbox>/release.json               (root-context jq's this)
```
lsierant: do we need sandboxing? Couldn't we just rely on real project files to test real thing?

lsierant: of course running the tests shouldn't leave dirty/broken state of the current worktree.

scripts/test/bash/worktree/test_om_project_namespace.bats
```diff
@@ -0,0 +1,218 @@
...
+# cleanup_legacy_ls_om_projects.sh
+# ---------------------------------------------------------------------------
+#
+# Verify the awk filter that picks legacy `ls` / `ls-{nondigit}-...` names
+# while preserving in-use `ls-<16..31>-...` names. This is the only piece of
+# logic we can test without hitting the OM API.
+
+@test "cleanup_legacy_ls_om_projects: awk filter selects the right projects" {
+    legacy="$(printf '%s\n' \
+        ls \
```
lsierant: too much awk! do we need that?

scripts/dev/contexts/root-context
lsierant: [AI] Fixed in 8cc47a52e. The derivation now appends `-${MCK_DEVC_NET_PREFIX}` to whatever NAMESPACE the user already has (no `"ls"` literal anywhere). Idempotency is via `[[ "${NAMESPACE}" != *"-${MCK_DEVC_NET_PREFIX}" ]]` so re-sourcing the file does not pile suffixes.

scripts/dev/cleanup_legacy_ls_om_projects.sh
lsierant: [AI] Confirmed redundant — `scripts/evergreen/e2e/setup_cloud_qa.py delete_all` already age-cleans every project >70 min old in the org via `clean_unused_projects(org)`. Deleted the bash script in 8cc47a52e. Updated the PR description to point users at `setup_cloud_qa.py delete_all` instead.

scripts/dev/cleanup_legacy_ls_om_projects.sh
lsierant: [AI] Script deleted in 8cc47a52e (redundant with `setup_cloud_qa.py delete_all` — see reply on the file-level comment).

lsierant: [AI] Done — script removed; `setup_cloud_qa.py delete_all` handles legacy cleanup.

scripts/dev/delete_om_projects.sh
lsierant: [AI] Fixed in 8cc47a52e — same change as the root-context block: appends `-${MCK_DEVC_NET_PREFIX}` to whatever NAMESPACE was set, no `"ls"` literal.

scripts/test/bash/worktree/test_om_project_namespace.bats
lsierant: [AI] Sandbox dropped in 8cc47a52e. Tests now run against the real `scripts/dev/contexts/root-context` and the real `scripts/dev/contexts/private-context`, deriving the expected NAMESPACE from whatever the user has set there.

lsierant: [AI] In 8cc47a52e: `setup()` snapshots `.devcontainer/.env` to `${BATS_TEST_TMPDIR}` and `teardown()` restores it (or removes it if it did not exist before). Bats guarantees teardown runs even if the test crashes, so the worktree is always restored.

scripts/test/bash/worktree/test_om_project_namespace.bats
lsierant: [AI] Removed in 8cc47a52e — the awk filter test went away with the legacy-cleanup script it was testing.

.devcontainer/compose.yml
```diff
@@ -9,6 +9,12 @@ services:
       K8S_FWD_PROXY: *proxy_address
       EVG_HOST_PROXY: "http://gost-proxy:8080"
       GOTOOLCHAIN: local
+      # Forward the per-worktree network prefix so root-context (sourced inside
+      # the container during `make switch`) can derive a per-worktree NAMESPACE
+      # (ls-${MCK_DEVC_NET_PREFIX}). That makes cloud-qa OM project names
```
lsierant, line 14: is still see ls mentioned here. ls is just $NAMESPACE set in private-context

PR #1085, opened 2026-05-10
KUBE-26: background availability tester pytest harness

lsierant:
# Based on PR #1080

## Chain of upstream PRs as of 2026-05-15

* PR #1080:
  `search/lb-improvements` ← `lsierant/KUBE-17-connectivity-tool`

  * **PR #1085 (THIS ONE)**:
    `lsierant/KUBE-17-connectivity-tool` ← `lsierant/KUBE-26-background-tester`

[skip-ci] KUBE-26 — background availability tester (stacked on PR #1080 / KUBE-17)

## Motivation

Layered on top of the connectivity tool from KUBE-17, this PR adds the
background availability tester — a pytest harness that drives the
connectivity tool over a configurable observation window from a daemon
thread and produces a `ConnectivityVerdict` snapshot on demand. This is
the foundation downstream availability scenarios (KUBE-27 failure modes,
KUBE-30 envoy lifecycle, KUBE-37 pod lifecycle) build on: without a
harness that **proves** it can detect availability loss, every
fault-injection test downstream is a no-op.

## What's new

A reusable Python module
`docker/mongodb-kubernetes-tests/tests/common/search/availability_tester.py`
exposing:

- **`SearchAvailabilityBackgroundTester`** — extends
  `kubetester.mongotester.BackgroundHealthChecker` (mirroring the
  existing `MongoDBBackgroundTester` pattern), drives a
  `SearchConnectivityTool` in either `oneshot` or `paging` mode, and
  captures every iteration's `QueryResult` for verdict snapshots.
  - `oneshot` mode: each iteration runs one cache-busted `$search` —
    reaches mongot every time, so any mongot/envoy outage surfaces
    immediately as a failure.
  - `paging` mode: each iteration reads one page from a long-living
    cursor; the cursor is rotated periodically (`paging_reset_every`,
    default 50) and reopened automatically after any failure so the
    tester keeps probing rather than spamming the same dead-cursor
    error. Use this when the test wants to exercise both upstream
    connectivity AND cursor longevity together.
- **`get_verdict()` / `get_results()`** — thread-safe snapshots of the
  current state. Tests can call these mid-run to log progress without
  stopping the tester.
- **`assert_steady_state(min_iterations, require_upstream_succeeded,
  max_failed)`** — used in fault-free observation windows; raises
  `AssertionError` with a verdict-shaped message if the snapshot fails
  the criteria.
- **`assert_outage_detected(accept_classes)`** — the deliverable signal
  for "the harness actually detects upstream loss"; verifies the verdict
  surfaces at least one failure in the configured classes
  (`cursor_lost`, `transient_network`, or `other`).

## Tests in the deliverable e2e

`tests/search/search_availability_background_tester.py` deploys a
single-cluster managed-LB MongoDBSearch (RS+Search+TLS, identical
bootstrap to the KUBE-17 connectivity-tool e2e via the shared
`tests.common.search.connectivity_bootstrap` module) and runs two
scenarios after the bootstrap:

1. **`test_steady_state_window_reports_alive`** — paging mode for 12s
   against a healthy cluster; asserts `upstream_alive` and
   `failed == 0` over ≥ 8 recorded iterations. Smoke test for the
   harness; without this baseline, the outage scenario could trivially
   pass on a tester that's not actually probing.
2. **`test_outage_window_detects_availability_loss`** — oneshot mode;
   ~6s healthy probe, then `core_v1.delete_namespaced_pod` deletes the
   mongot pod (StatefulSet recreates it within ~10-20s), ~25s of
   probing through the outage. Asserts `cursor_lost > 0` OR
   `transient_network > 0` — the deliverable per KUBE-26 acceptance:
   "demonstrates on a deliberately broken cluster that the tester
   detects availability loss." Cleanup waits for `Phase.Running`.

## Local verification

Latest run (single-cluster managed-LB on EVG-host kind):

```
=== 15 passed, 1 skipped, 2387 deselected, 24 warnings in 105.74s (0:01:45) ===
```

Skip is `test_create_ops_manager` (`@skip_if_cloud_manager`, expected
on cloud-qa).

Outage verdict from the deliverable test:

```
{'total': 48, 'succeeded': 9, 'succeeded_with_retry': 0, 'failed': 39,
 'cursor_lost': 0, 'transient_network': 39, 'other_failed': 0,
 'upstream_succeeded': 9, 'cache_only_succeeded': 0,
 'unknown_succeeded': 0, 'upstream_alive': True,
 'cursor_lost_observed': False,
 'error_breakdown': {'OperationFailure': 39},
 'first_error': "upstream connect error or disconnect/reset before
                 headers. reset reason: remote connection failure,
                 transport failure reason: delayed connect error:
                 Connection refused, ..."}
```

39 of 48 oneshot probes failed during the window mongot was being
recreated, all classified as `transient_network` (envoy returns
"no healthy upstream" / "Connection refused" while the upstream is
gone). 9 successes are pre-fault healthy probes. `upstream_alive: True`
because the recovered post-fault probes again confirm upstream contact.
This is exactly the signal we need: the tester catches the outage AND
confirms it's not just stuck in cache-only mode.

## Why oneshot for the outage scenario

The fault-detection scenario uses `oneshot` mode rather than `paging`
because mongod's server-side cache can keep serving a long-living
cursor's `getMore` calls during a brief mongot outage — so paging mode
saw zero failures even when mongot was actually down for ~15 seconds.
This is exactly the "cache caveat" the KUBE-17 connectivity tool was
designed to expose. Oneshot mode with `cache_buster=True` issues a
fresh aggregation each iteration, forcing a real upstream evaluation,
which surfaces the outage immediately.

The steady-state scenario keeps `paging` because that mode is what
real downstream tests (KUBE-27) will use, and it needs an explicit
proof point.

## Branch / rebase note

This PR is **stacked on PR #1080 / KUBE-17 connectivity-tool**, NOT on
`search/lb-improvements`. The branch (`lsierant/KUBE-26-background-tester`)
sits on top of `lsierant/KUBE-17-connectivity-tool` HEAD because it
imports the connectivity tool's `SearchConnectivityTool`, `QueryResult`,
`ConnectivityVerdict`, and `connectivity_bootstrap` directly. The
stacked PR stays draft permanently; merge happens via local git when
PR #1080 lands and this PR is rebased onto its new tip.

## Out of scope (deferred)

- The 5 enumerated failure-mode scenarios (single mongot pod restart,
  envoy pod restart, query before search index built, mongod missing
  search params, new shard added mid-flight) — KUBE-27, builds on this
  harness.
- `paging` mode for fault scenarios — needs a stronger cache-bust
  strategy (e.g. close-and-reopen cursor on every getMore boundary)
  before it's usable; tracked under the cache caveat investigation.
- Multi-cluster scaffold integration (KUBE-32) — this PR runs
  single-cluster only; multi-cluster work is filed separately.


docker/mongodb-kubernetes-tests/tests/search/search_availability_background_tester.py
```diff
@@ -0,0 +1,390 @@
...
+        # objects: ``mdbs["spec"]["replicas"] = 0; mdbs.update()`` does
+        # NOT actually propagate to the API. Pod deletion via
+        # core_v1.delete_namespaced_pod sidesteps that.
+        core_v1 = client.CoreV1Api()
+        pods = core_v1.list_namespaced_pod(
+            namespace=mdb.namespace,
+            label_selector=f"statefulset.kubernetes.io/pod-name={statefulset_name}-0",
+        ).items
+        if not pods:
+            # Older k8s versions don't set that label; fall back to listing
```
lsierant: what version? how old? does it matter?

docker/mongodb-kubernetes-tests/tests/search/search_availability_background_tester.py
```diff
@@ -0,0 +1,390 @@
...
+        # The StatefulSet's controller will recreate it within ~10-30s,
+        # but during the recreation window the cursor's gRPC stream to
+        # mongot is dead — any post-fault getMore must fail (cursor_lost
+        # because the new pod has no record of the cursor's session, or
+        # transient_network because envoy briefly has no healthy
+        # upstream).
+        #
+        # We use pod-delete rather than CR-driven scale-to-0 here
+        # because the python kubernetes-client's openapi serializer
+        # silently drops zero-valued int fields when patching custom
```
lsierant: so how we could scale to zero in other places? This should work here.

lsierant: it's OK to just delete pod in order to induce availability loss.

docker/mongodb-kubernetes-tests/tests/search/search_availability_background_tester.py
```diff
@@ -0,0 +1,390 @@
...
+
+        # Cleanup — wait for the StatefulSet's controller to recreate
+        # mongot pods so sibling tests see a healthy cluster.
+        try:
+            mdbs.assert_reaches_phase(Phase.Running, timeout=180)
+        except Exception as e:
+            logger.warning(f"cleanup mdbs.assert_reaches_phase(Running) timed out: {e}")
+
+    verdict = tester.assert_outage_detected(
+        accept_classes=("cursor_lost", "transient_network"),
```
lsierant: how can we simulate actual cursor loss? Would more than one mongot instance work in this case, when cursor is opened against one and we delete it, the second one will be used by mongod and envoy and there will not be network error?

PR #1112, opened 2026-05-14
SKUNK-209: refactor(appdb): collapse AppDB monitoring into single automation agent

MaciejKaras:
## Summary

- Removes the `mongodb-agent-monitoring` sidecar container from AppDB StatefulSets; both automation and monitoring now run as a single process in the `mongodb-agent` container
- Monitoring is activated by embedding `mmsGroupId`/`mmsApiKey` (plus optional TLS params) into `monitoringVersions[].additionalParams` in the main automation config — no pod restart required when Ops Manager credentials become available
- `spec.appDB.monitoringAgent` is deprecated: field marked as deprecated in CRD schema, admission webhook emits a non-blocking warning when set, and a runtime warning is logged

## Consequences & follow-ups

**Behavioral consequences (no code change, documented):**

- Upgrading to this version triggers a **one-time rolling restart** of all AppDB pods (sidecar removal); a downgrade triggers a second roll (re-adding it). No data-loss risk.
- Existing AppDBs get a **one-time automation-config version bump** as monitoring folds into the main automation config.
- The agent API key now resides in the main AppDB automation-config secret (`<name>-config`) instead of the removed `<name>-monitoring-config` secret — a relocation alongside the SCRAM creds/keyfile already there, not a new exposure.
- The old `*-monitoring-config` Secret/ConfigMap are **left orphaned** on upgrade. This is consistent with how the operator already treats the primary AC and agent-key secrets (no finalizer / scale-in cleanup); they may be deleted manually.
- **Observability change:** the dedicated `mongodb-agent-monitoring` container and its status port are removed. The monitoring component still writes its own `monitoring-agent.log`, now on the single `mongodb-agent` container's log volume — `kubectl logs -c mongodb-agent-monitoring` breaks; read the file in the `mongodb-agent` container instead.

**Out of scope (follow-ups):**

- Wiring `spec.appDB.agent.monitoringAgent.logRotate` / `backupAgent.logRotate` into the automation config (pre-existing MDB-parity gap).
- Cleanup of orphaned / operator-managed AppDB secrets on scale-in or resource deletion (pre-existing operator behavior).

**Product impact:** lower per-pod resource footprint, monitoring toggle without a pod restart, and architecture aligned with standard Ops Manager monitoring and the `MongoDB` resource (a step toward a single controller for both).

> [!NOTE]
> The additional benefit is that the EVG patch timespan was reduced by ~10 minutes (from 1h30m to 1h19m) and the overall cost of PR patch by 10% (from 80$ to 73$).

## Proof of Work

- `controllers/operator/appdbreplicaset_controller_test.go` — `TestConfigureMonitoring_*`, `TestBuildAppDbAutomationConfig_*`, `TestTryConfigureMonitoring_*`, `TestReadExistingPodVars_*` (incl. `TestReadExistingPodVars_MissingAgentKeyIsNonFatal`), `TestMonitoringToggled*`, `TestMonitoringAgentStartupParametersIgnored`, `TestStartupParametersDeprecationWarning`
- `controllers/operator/construct/appdb_construction_test.go` — `TestAppDbStatefulSet_SingleAgentContainer*`
- `api/mongodb/v1/om/opsmanager_validation_test.go` — `TestWarnMonitoringAgentStartupParameters*`

## Checklist

### DIFF
--- a/controllers/operator/appdbreplicaset_controller.go
+++ b/controllers/operator/appdbreplicaset_controller.go
@@ -1806,19 +1806,31 @@ func (r *ReconcileAppDbReplicaSet) readExistingPodVars(ctx context.Context, om *
 		return env.PodEnvVars{}, xerrors.Errorf("error reading credentials: %w", err)
 	}
 
-	agentAPIKey, err := secret.ReadKey(ctx, memberClient, util.OmAgentApiKey, kube.ObjectKey(om.Namespace, agents.ApiKeySecretName(projectId)))
-	if err != nil {
-		return env.PodEnvVars{}, xerrors.Errorf("error reading agent API key for project %s: %w", projectId, err)
-	}
-
-	return env.PodEnvVars{
-		User:        cred.PublicAPIKey,
-		ProjectID:   projectId,
-		AgentAPIKey: agentAPIKey,
+	podVars := env.PodEnvVars{
+		User:      cred.PublicAPIKey,
+		ProjectID: projectId,
 		SSLProjectConfig: env.SSLProjectConfig{
 			SSLMMSCAConfigMap: om.Spec.GetOpsManagerCA(),
 		},
-	}, nil
+	}
+
+	var appdbSecretPath string
+	if r.VaultClient != nil {
+		appdbSecretPath = r.VaultClient.AppDBSecretPath()
+	}
+
+	// Read through the Vault-aware SecretClient so the key resolves on Vault backends
+	// (where it has no plain Kubernetes Secret). A missing key is non-fatal: monitoring
+	// is simply left unconfigured until the key exists (configureMonitoring clears
+	// monitoringVersions when the key is empty).
+	agentAPIKey, err := r.getMemberCluster(r.getNameOfFirstMemberCluster()).SecretClient.ReadSecretKey(
+		ctx, kube.ObjectKey(om.Namespace, agents.ApiKeySecretName(projectId)), appdbSecretPath, util.OmAgentApiKey)
+	if err != nil {
+		log.Warnf("Agent API key for project %s not readable yet (%v); AppDB monitoring will be configured once it is available", projectId, err)
+		return podVars, nil
+	}
+	podVars.AgentAPIKey = agentAPIKey
+	return podVars, nil
 }
 
 func (r *ReconcileAppDbReplicaSet) publishACVersionAsConfigMap(ctx context.Context, cmName string, opsManager *omv1.MongoDBOpsManager, version int, memberCluster multicluster.MemberCluster) workflow.Status {
--- a/controllers/operator/appdbreplicaset_controller.go
+++ b/controllers/operator/appdbreplicaset_controller.go
@@ -43,7 +43,6 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/controllers/operator/watch"
 	"github.com/mongodb/mongodb-kubernetes/controllers/operator/workflow"
 	"github.com/mongodb/mongodb-kubernetes/pkg/agent"
-	"github.com/mongodb/mongodb-kubernetes/pkg/agentVersionManagement"
 	"github.com/mongodb/mongodb-kubernetes/pkg/authentication/scram"
 	"github.com/mongodb/mongodb-kubernetes/pkg/automationconfig"
 	"github.com/mongodb/mongodb-kubernetes/pkg/dns"
@@ -68,14 +67,9 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/pkg/vault"
 )
 
-type agentType string
-
 const (
 	appdbCAFilePath              = "/var/lib/mongodb-automation/secrets/ca/ca-pem"
 	appDBACConfigMapVersionField = "version"
-
-	monitoring agentType = "MONITORING"
-	automation agentType = "AUTOMATION"
 
 	// Used to note that for this particular case it is not necessary to pass
 	// the hash of the Prometheus certificate. This is to avoid having to
@@ -485,7 +479,7 @@ func (r *ReconcileAppDbReplicaSet) shouldReconcileAppDB(ctx context.Context, ops
 		return true, nil
 	}
 
-	desiredAc, err := r.buildAppDbAutomationConfig(ctx, opsManager, automation, UnusedPrometheusConfiguration, memberCluster.Name, log)
+	desiredAc, err := r.buildAppDbAutomationConfig(ctx, opsManager, nil, UnusedPrometheusConfiguration, memberCluster.Name, log)
 	if err != nil {
 		return false, xerrors.Errorf("error building AppDB Automation Config: %w", err)
 	}
@@ -605,15 +599,6 @@ func (r *ReconcileAppDbReplicaSet) ReconcileAppDB(ctx context.Context, opsManage
 			appdbOpts.AgentImage = images.ContainerImage(r.imageUrls, util.AgentImageUrlEnv, agentVersion)
 		}
 	} else {
-		// instead of using a hard-coded monitoring version, we use the "newest" one based on the release.json.
-		// This ensures we need to care less about CVEs compared to the prior older hardcoded versions.
-		legacyMonitoringAgentVersion, err := r.getLegacyMonitoringAgentVersion(ctx, opsManager, log)
-		if err != nil {
-			return r.updateStatus(ctx, opsManager, workflow.Failed(xerrors.Errorf("Error reading monitoring agent version: %w", err)), log, appDbStatusOption)
-		}
-
-		appdbOpts.LegacyMonitoringAgentImage = images.ContainerImage(r.imageUrls, util.AgentImageEnv, legacyMonitoringAgentVersion)
-
 		// AgentImageEnv contains the full container image uri e.g. quay.io/mongodb/mongodb-agent:107.0.0.8502-1
 		// In non-static containers we don't ask OM for the correct version, therefore we just rely on the provided
 		// environment variable.
@@ -660,7 +645,7 @@ func (r *ReconcileAppDbReplicaSet) ReconcileAppDB(ctx context.Context, opsManage
 
 	workflowStatus = workflow.RunInGivenOrder(publishAutomationConfigFirst,
 		func() workflow.Status {
-			return r.deployAutomationConfigAndWaitForAgentsReachGoalState(ctx, log, opsManager, allStatefulSetsExist, appdbOpts)
+			return r.deployAutomationConfigAndWaitForAgentsReachGoalState(ctx, log, opsManager, &podVars, allStatefulSetsExist, appdbOpts)
 		},
 		func() workflow.Status {
 			return r.deployStatefulSet(ctx, opsManager, log, podVars, appdbOpts)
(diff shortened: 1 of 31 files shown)


api/v1/mdb/mongodb_types.go
```diff
@@ -634,8 +634,13 @@ type AgentConfig struct {
 	SystemLog *automationconfig.SystemLog `json:"systemLog,omitempty"`
 }
 
+// Deprecated: MonitoringAgentConfig has no effect. The monitoring agent has been merged into
```
MaciejKaras: We don't need this comment as long as we have deprecation comment on api/v1/om/appdb_types.go

api/v1/mdb/mongodb_types.go
```diff
@@ -634,8 +634,13 @@ type AgentConfig struct {
...
+// the automation agent and no longer runs as a separate process.
 type MonitoringAgentConfig struct {
-	StartupParameters StartupParameters `json:"startupOptions"`
+	// Deprecated: startupOptions has no effect. The monitoring agent has been merged into the
```
MaciejKaras: We don't need this comment as long as we have deprecation comment on api/v1/om/appdb_types.go. Also marking it as optional is unnecessary, because the whole struct should be excluded

api/v1/om/opsmanager_validation.go
```diff
@@ -313,13 +321,17 @@ func (om *MongoDBOpsManager) RunValidations() []v1.ValidationResult {
 	return validationResults
 }
 
-func (om *MongoDBOpsManager) ProcessValidationsWebhook() error {
+func (om *MongoDBOpsManager) ProcessValidationsWebhook() (admission.Warnings, error) {
```
MaciejKaras: we only care about ProcessValidationsOnReconcile(), we can skip webhook warnings and do not change this method at all.

controllers/operator/construct/appdb_construction.go
```diff
@@ -40,30 +37,19 @@ const (
...
-	LegacyMonitoringAgentImage string
+	InitAppDBImage string
+	MongodbImage   string
+	AgentImage     string
 
 	PrometheusTLSCertHash string
 }
 
-func getMonitoringAgentLogOptions(spec om.AppDBSpec) string {
-	return fmt.Sprintf(" -logFile=/var/log/mongodb-mms-automation/monitoring-agent.log -maxLogFileDurationHrs=%d -logLevel=%s", spec.GetAgentMaxLogFileDurationHours(), spec.GetAgentLogLevel())
```
MaciejKaras, line 66: Instead of removing them, we should add those settings to AutomationConfig monitoringVersions if possible. The options available in monitoringVersions -> https://www.mongodb.com/docs/ops-manager/current/reference/cluster-configuration/#monitoring

controllers/operator/construct/appdb_construction.go
```diff
@@ -375,27 +345,23 @@ func ShouldMountSSLMMSCAConfigMap(podVars *env.PodEnvVars) bool {
...
 
-	if ShouldEnableMonitoring(podVars) {
-		monitoringModification = addMonitoringContainer(*appDb, *podVars, opts, externalDomain, architectures.IsRunningStaticArchitecture(opsManager.Annotations), log)
-	} else {
-		// Otherwise, let's remove for now every podTemplateSpec related to monitoring
-		// We will apply them when enabling monitoring
-		if podSpec != nil {
-			podSpec.Spec.Containers = removeContainerByName(podSpec.Spec.Containers, monitoringAgentContainerName)
+	if podSpec != nil {
+		originalLen := len(podSpec.Spec.Containers)
```
MaciejKaras, line 358: I think we should also emit warning in the spec validation when users specify monitoring-agent container

controllers/operator/construct/appdb_construction.go
```diff
@@ -375,27 +345,23 @@ func ShouldMountSSLMMSCAConfigMap(podVars *env.PodEnvVars) bool {
...
+		podSpec.Spec.Containers = removeContainerByName(podSpec.Spec.Containers, monitoringAgentContainerName)
+		if len(podSpec.Spec.Containers) < originalLen {
+			log.Warnf("Removed deprecated container %q from user-supplied podTemplateSpec; monitoring is now managed via the automation config additionalParams", monitoringAgentContainerName)
 		}
 	}
 
 	// We copy the Automation Agent command from community and add the agent startup parameters
-	automationAgentCommand := AutomationAgentCommand(architectures.IsRunningStaticArchitecture(opsManager.Annotations), true, opsManager.Spec.AppDB.GetAgentLogLevel(), opsManager.Spec.AppDB.GetAgentLogFile(), opsManager.Spec.AppDB.GetAgentMaxLogFileDurationHours())
+	automationAgentCommand := AutomationAgentCommand(architectures.IsRunningStaticArchitecture(opsManager.Annotations), false, opsManager.Spec.AppDB.GetAgentLogLevel(), opsManager.Spec.AppDB.GetAgentLogFile(), opsManager.Spec.AppDB.GetAgentMaxLogFileDurationHours())
```
MaciejKaras: We always pass false to this AutomationAgentCommand function I believe. Maybe we can simplify it?

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -485,7 +479,8 @@ func (r *ReconcileAppDbReplicaSet) shouldReconcileAppDB(ctx context.Context, ops
 		return true, nil
 	}
 
-	desiredAc, err := r.buildAppDbAutomationConfig(ctx, opsManager, automation, UnusedPrometheusConfiguration, memberCluster.Name, log)
+	// nil podVars: monitoring state is not compared for drift detection, only processes/replica sets
```
MaciejKaras: remove comment

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -1453,26 +1421,34 @@ func configureMonitoring(ac *automationconfig.AutomationConfig, log *zap.Sugared
...
+				Name:             om.MonitoringAgentDefaultVersion,
+				AdditionalParams: params,
 			}
 			log.Debugw("Added monitoring agent configuration", "host", hostname, "tls", tls)
 			monitoringVersions = append(monitoringVersions, mv)
 		} else {
-			if tls {
-				monitoringVersions[foundIdx].AdditionalParams = om.NewTLSParams(appdbCAFilePath, pemKeyFile)
-			} else {
-				om.ClearTLSParams(monitoringVersions[foundIdx].AdditionalParams)
```
MaciejKaras, line 1474: I don't see om.ClearTLSParams being called on the right side. Do we actually clean those tls params when tls is disabled?

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -1824,9 +1805,17 @@ func (r *ReconcileAppDbReplicaSet) readExistingPodVars(ctx context.Context, om *
 		return env.PodEnvVars{}, xerrors.Errorf("error reading credentials: %w", err)
 	}
 
+	agentAPIKey, err := secret.ReadKey(ctx, memberClient, util.OmAgentApiKey, kube.ObjectKey(om.Namespace, agents.ApiKeySecretName(projectId)))
+	if err != nil {
+		// Non-fatal: log and continue with empty key. Monitoring will self-correct on next reconcile.
```
MaciejKaras: I would rather return error and reconcile again

docs/superpowers/plans/2026-05-13-appdb-single-agent-monitoring.md
MaciejKaras, line 1: remove this file

api/v1/mdb/mongodb_types.go
MaciejKaras: Done — removed the struct-level deprecated comment from `MonitoringAgentConfig`.

api/v1/mdb/mongodb_types.go
MaciejKaras: Done — removed `// +optional` and `omitempty` from the field. The CRD now marks `startupOptions` as required when `monitoringAgent` is present (consistent with `AgentConfig`'s non-deprecated counterpart).

api/v1/om/opsmanager_validation.go
MaciejKaras: Done — reverted `ProcessValidationsWebhook()` to return `error` only; `ValidateCreate`/`ValidateUpdate` return `nil, err`. Warnings remain in `RunValidations()` and are picked up exclusively by `ProcessValidationsOnReconcile()`.

controllers/operator/construct/appdb_construction.go
MaciejKaras: Done — `configureMonitoring` now accepts an `extraParams map[string]string` argument. `buildAppDbAutomationConfig` passes `map[string]string(opsManager.Spec.AppDB.MonitoringAgent.StartupParameters)` so those params are merged into each host's `additionalParams`. Added a unit test `TestMonitoringAgentStartupParametersForwardedToAdditionalParams` to cover this.

controllers/operator/construct/appdb_construction.go
MaciejKaras: Done — added `warnMonitoringAgentContainer` validator in `opsmanager_validation.go` that checks `spec.appDB.podSpec.podTemplate.spec.containers` for a container named `mongodb-agent-monitoring` and emits a warning. Registered in `RunValidations()`.

controllers/operator/construct/appdb_construction.go
MaciejKaras: Done — removed the `withAgentAPIKeyExport bool` parameter entirely (it was always `false` at both call sites). Also removed the now-dead `GetMongodbUserCommandWithAPIKeyExport` helper.

controllers/operator/appdbreplicaset_controller.go
MaciejKaras: Done — comment removed.

controllers/operator/appdbreplicaset_controller.go
MaciejKaras: Good catch. The `params` map is rebuilt fresh each reconcile so stale TLS keys wouldn't survive, but it's clearer to be explicit. Added an `else { om.ClearTLSParams(params) }` branch so the intent is unambiguous.

controllers/operator/appdbreplicaset_controller.go
MaciejKaras: Done — now returns an error which causes a requeue rather than continuing with an empty key.

docs/superpowers/plans/2026-05-13-appdb-single-agent-monitoring.md
MaciejKaras: Done — file deleted.

MaciejKaras:

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_256_connectivity.py
```diff
@@ -172,58 +161,20 @@
...
-    assert "connectionString.standard" in non_admin_standard_secret
-    assert "connectionString.standardSrv" in non_admin_standard_secret
-    # authSource in the connection string must match the user's spec.db (non-admin database)
-    assert f"authSource={NON_ADMIN_USER_DATABASE}" in non_admin_standard_secret["connectionString.standard"]
-    assert f"authSource={NON_ADMIN_USER_DATABASE}" in non_admin_standard_secret["connectionString.standardSrv"]
 
 
 @mark.e2e_replica_set_scram_sha_256_user_connectivity
 def test_credentials_can_connect_to_db(replica_set: MongoDB, standard_secret: Dict[str, str]):
+    print("Connecting with {}".format(standard_secret["connectionString.standard"]))
```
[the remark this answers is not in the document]
MaciejKaras: This is test data

josvazg (approved):
Approved, but please seek approval from someone with more context here.

> Verdict: Pass with comments — no blocking bugs found
> 
>   The architectural direction is sound and well-executed. Eliminating the second agent container removes ~300 lines of net code and kills an
>   entire class of sync problems (two secrets, two agent versions, file-mounted API keys). Good simplification.
> 
>   ---
>   1. Dead code — should be cleaned up before merge
> 
>   containerImageModification (appdb_construction.go:518) is defined but never called anywhere in the codebase. It was clearly left over from a
>   refactor iteration. Either wire it up or delete it — dead private functions are noise and confuse future readers.
> 
>   // defined here, called nowhere:
>   func containerImageModification(containerName string, image string, args []string) statefulset.Modification {
> 
>   ---
>   2. shouldReconcileAppDB passes nil podVars — add a comment
> 
>   desiredAc, err := r.buildAppDbAutomationConfig(ctx, opsManager, nil, ...)
> 
>   This is actually fine — the function only compares Processes[*].Disabled flags and never looks at MonitoringVersions. But nil here looks like
>    a bug to anyone reading it cold. One line comment explaining why nil is intentional would save the next person from investigating it.
> 
>   ---
>   3. SSLRequireValidMMSServerCertificates is always false in the AppDB path
> 
>   tryConfigureMonitoringInOpsManager and readExistingPodVars both return a PodEnvVars that never sets SSLRequireValidMMSServerCertificates —
>   it's always the zero value false. So configureMonitoring will always write sslRequireValidMMSServerCertificates=false into the AC when TLS is
>    enabled.
> 
>   The old monitoring container had the same behavior (that field was also never populated for AppDB). So this is not a regression introduced
>   here, but the new explicit "false" in the AC is visible to anyone diffing configs. Worth calling out: is this intentional? If customers ever
>   set requireValidCert: true via OM project config, it would be silently overridden to false in the monitoring params. A follow-up ticket is
>   appropriate if this isn't the right steady state.
> 
>   ---
>   4. MonitoringAgent.StartupParameters silently dropped — warning message needs a migration story
> 
>   The validation warning says "remove this field", but gives no guidance on what to do if the user's custom params had a real purpose (e.g.
>   custom log paths, custom SSL certs). The message could say: "If you configured TLS or log options, use spec.appDB.agent instead." Right now a
>    user who gets this warning has no idea whether they're losing functionality silently.
> 
>   ---
>   5. Minor: upgrade causes a brief monitoring gap — fine, but changelog-worthy
> 
>   On upgrade: the monitoring container is removed from the pod (restart), then the operator pushes the new AC with monitoringVersions
>   populated. Between those two events there's a window where no monitoring data is sent. This is expected and acceptable for this kind of
>   architectural change, but worth a sentence in the release notes so customers aren't surprised by a monitoring gap during rolling restarts.

MaciejKaras:
Abandoned in favour of https://github.com/mongodb/mongodb-kubernetes/pull/1186

PR #1123, opened 2026-05-19
Suppress Code QL complaints on tests

josvazg:
# Summary

Avoid noisy Code QL issues irrelevant on test code.

## Proof of Work

Should reduce reports to legit only, or none.

## Checklist


josvazg:
I do not think this makes sense. I assume for this project we probably want to supres most of the complaints, specially those in python code, which is building or testing infra.

PR #1126, opened 2026-05-20
KUBE-32: simulated multi-cluster mode for MongoDBSearch (RS + sharded)

anandsyncs:
# Based on PR #1048

## Chain of upstream PRs as of 2026-06-16

* PR #1166:

  * PR #1048:

    * **PR #1126 (THIS ONE)**:
      `search/ga-base` ← `search/simulated-mc`

# Summary

Adds a third MongoDBSearch deployment mode — **simulated multi-cluster**: each Kubernetes cluster runs its own MCK operator, and every operator reconciles the same `MongoDBSearch` CR but acts only on the `spec.clusters[i]` entry whose `clusterName` matches its own identity (`OPERATOR_CLUSTER_NAME`). There is no cross-cluster fan-out; an operator whose cluster isn't listed silently no-ops. This unblocks customers who can't grant one operator cross-cluster kubeconfig access but still want a single logical CR spanning clusters. Both **replica-set** and **sharded** external sources are supported. Legacy single-cluster and central-multi-cluster modes are unchanged.

Supersedes #1152 and the former base simulated-MC PR — both folded in here and rebased onto current `search/ga-base`.

## Architecture: hub-and-spoke vs simulated multi-cluster

Side-by-side reconcile flow for both topologies. Each lane runs the shared `prepareSearch` gate (built once at construction by `newPrepareSearch`: `ValidateSpec` → `ValidateSimulatedMCClusterIndices` → `LocalizeToCluster`; the reconcile bodies never branch on mode) and `resolveClusterMapping`/`AssignClusterIndices` once, then splits into the two reconcilers — `MongoDBSearchReconciler` (mongot) and `MongoDBSearchEnvoyReconciler` (Envoy) — across CONFIG BUILD / DEPLOY / STATUS. Hub-and-spoke fans out per-cluster writes via `memberClusterClientsMap`; simulated-MC narrows to the local entry (silent no-op if unlisted) and writes to `r.kubeClient`. Status ownership is split per Rule 6 (mongot → `Status.Phase`; Envoy → `Status.LoadBalancer`).

Hub-and-spoke vs simulated multi-cluster reconcile flow

## Lines of code by area

| Area | Files | +Added | −Deleted | Net |
|---|---:|---:|---:|---:|
| Production Go | 12 | 283 | 150 | **+133** |
| Go unit tests | 8 | 1,480 | 455 | +1,025 |
| Python e2e (incl. shared helpers in `tests/common/search/`) | 14 | 2,251 | 179 | +2,072 |
| Generated (CRD YAML ×3 + deepcopy) | 4 | 50 | 0 | +50 |
| Helm + Evergreen wiring | 6 | 51 | 0 | +51 |
| **Total** | **44** | **4,115** | **784** | **+3,331** |

**~91% of the added lines are tests** (Go full-reconcile + Python e2e). The "generated" row is purely additive: the new `clusterIndex` schema rendered into the three CRD files (16 lines each) plus the deepcopy — no regeneration churn. Rebased onto current `search/ga-base` (`35439898d`) with zero conflicts; the feature delta is byte-identical (`git range-diff` `=` on every commit). The diff was scope-audited hunk-by-hunk: pre-existing ga-base comments and test shapes are byte-identical to base wherever the feature didn't force a change.

## Review guide

The branch is 6 logical commits (5 feature + 1 pre-existing-ga-base test-compile fix), structured for commit-by-commit review (oldest first):

1. `5069f0fb4` **search: simulated multi-cluster mode for MongoDBSearch (API + reconcile)** — the only production commit. Focus on the constructor-built gate (`newPrepareSearch`: `ValidateSpec` runs on the *un-narrowed* spec → `ValidateSimulatedMCClusterIndices` → `LocalizeToCluster`; both reconcilers store the resulting `prepareSearch` func, so the reconcile bodies never branch on mode) and the Envoy per-cluster work list: a constructor-chosen `clientForCluster` resolves each work item's client (central client only when no member clients are configured; an unregistered cluster name surfaces per-cluster Pending instead of writing centrally). Secondary: pin application in `AssignClusterIndices` and the `clusterIndex` CEL rules (uniqueness + required for multi-cluster).
2. `a41547abf` **test(search): unit coverage for simulated-MC reconcile** — full-reconcile tables for RS + sharded projection. Focus on the pinned-index-7 regression tables (the *resolved* cluster index, not array position, must drive the SNI hostname), no-match-leaves-status-untouched, re-pin updates mapping + resource names, and unregistered-cluster Pending with zero central writes. Also includes the Envoy LB-cleanup-across-mapping / state-load-error-central-fallback / registered-but-missing-from-mapping-Pending tables, a resolve-cluster-mapping-failure path, and a DNS-label-overflow row keyed on the *pinned* index. Pre-existing ga-base tests keep their original shapes/names wherever the feature didn't force a change.
3. `59ed48e2a` **test(search): shared simulated-MC e2e helpers** — the package `conftest.py` (fixtures: per-cluster operators, users, CR apply; plus the shared test bodies both e2e files call — per-file `test_*` wrappers keep the pytest ids) and **extensions to the canonical helpers** in `tests/common/search/` rather than parallel copies: `mc_search_helper.py` gains the resolver-callback AC patcher (`patch_mongot_host_via_ac` — the two existing `*_via_om` entry points now delegate to it), an explicit-mode on-disk mongotHost core, and opt-in cds.json upstream anchors on `verify_per_cluster_envoy_sni`; all new params are default-preserving, so existing consumers (`bootstrap_test_mixins*`, q2/q3) are byte-compatible and gain the stronger vacuous-pass guards. Focus on those guards and the cross-cluster tools-pod support. Note: `tools_pod.run_command` now raises on non-zero exit and returns stdout-only (previously returned merged output and never raised) — all four pre-existing callers audited; this also fixes a previously vacuous negative assertion in `search_enterprise_tls.py`.
4. `f284b6b0c` **test(search): RS + sharded simulated-MC e2e** — `simulated_mc_rs.py` + `simulated_mc_sharded.py` on the three-operator topology. Focus on the data-plane assertions: restore → `$search` index READY → deterministic per-cluster row counts; sharded flip routing verified-applied before querying; `$vectorSearch` per cluster and through each local mongos; cross-cluster isolation + local-only status. The sharded flip assertion also reads back the four search setParameters (mTLS/gRPC) on the source mongod+mongos before querying, type-coerced so the on-disk string booleans compare correctly.
5. `d90aba8d2` **ci(search): wire simulated-MC e2e tasks into multi-cluster kind variants** — small; confirm `e2e_search_simulated_mc_rs` / `e2e_search_simulated_mc_sharded` line up with the existing `e2e_multi_cluster_kind` task group.
6. `f9d4273f2` **fix: pass architectures arg to defaultReplicaSetReconciler call** — a one-line fix for a *pre-existing* `search/ga-base` test-compile bug (`mongodbreplicaset_controller_test.go:893` was the only call site missing the trailing `architectures.NonStatic` arg the helper requires on ga-base), which otherwise leaves the `controllers/operator` test package uncompilable (`unit_tests_golang` + `lint_repo` red). Not feature code; drops cleanly once ga-base fixes it upstream.

Five files exceed ~400 changed lines: `tests/multicluster_search/simulated_mc_sharded.py` (809), `controllers/operator/mongodbsearch_controller_test.go` (765), `controllers/operator/mongodbsearchenvoy_controller_test.go` (674), `tests/multicluster_search/simulated_mc_rs.py` (580), `tests/multicluster_search/conftest.py` (492). Suggested reading order: `api/mongodb/v1/search/mongodbsearch_types.go` before the two controllers; `conftest.py` and the `mc_search_helper.py` extensions before the two e2e test files.

## What's included

- **API** — `ClusterIndex *int32` on `ClusterSpec`: **manually provided, never auto-assigned**. Required on every `spec.clusters[]` entry whenever more than one cluster is specified (CRD CEL at admission + a `ValidateSpec` mirror for stale-CRD installs; uniqueness enforced at both layers); optional for single-entry specs (unpinned resolves to the legacy index 0). The index is customer-changeable — re-pinning renames that cluster's per-cluster resources and orphans those at the old index, as documented on the field. `LocalizeToCluster` method; `AssignClusterIndices` applies pins verbatim (no monotonic assignment exists).
- **Controllers** — simulated-MC projection shared across both reconcilers via a single `prepareSearch` gate (a `prepareSearchFunc` chosen once at construction by `newPrepareSearch(operatorClusterName)`, so the reconcile body never branches on mode), plus shared `resolveClusterMapping` and `EnqueueMemberClusterObjectToSearch`.
- **Helm / wiring** — `operator.clusterIdentity.clusterName` injects the `OPERATOR_CLUSTER_NAME` env var, plumbed through `main.go` into both controller registrations.
- **Tests** — Go full-reconcile + API unit tests; RS and sharded e2e (`simulated_mc_rs.py`, `simulated_mc_sharded.py`) with a real per-cluster `$search` data plane; shared bodies live in `conftest.py`; wired into the existing `e2e_multi_cluster_kind` task group as `e2e_search_simulated_mc_rs` / `e2e_search_simulated_mc_sharded`.

## Key design notes

- **Identity from `OPERATOR_CLUSTER_NAME`** (Helm-injected). Unset ⇒ simulated-MC disabled, both legacy paths unchanged. No new CRD/ConfigMap/RBAC.
- **`LocalizeToCluster`** narrows `spec.clusters[]` to the local entry, then the reconciler runs unchanged over a 1-element slice. Unlisted cluster ⇒ `Result{}, nil` (status untouched).
- **`clusterIndex` is pinned-only** — there is no automatic index assignment. Multi-cluster specs must pin every entry (rejected at admission otherwise, in *all* modes); per-cluster-operator mode additionally requires the pin even on a single-entry spec, validated on the un-narrowed spec *before* localize. The persisted mapping still retains removed clusters' entries (orphan bookkeeping) and is what the Envoy reconciler reads.
- **No `if simulatedMC {} else {}` in reconcile** — the per-CR state ConfigMap is used identically in both modes; the managed-LB endpoint helpers resolve `{clusterIndex}`/`{clusterName}` from the persisted mapping index — the same value the per-cluster resource names use (fixes a prior SNI-hostname divergence on unpinned remove/re-add).
- **RS + sharded** — the sharded reconcile helpers collapse correctly to "1 cluster × N shards" after projection; the sharded e2e places a mongos per cluster so `$search` routes through that cluster's local Envoy + mongot path.

## Testing

- **Go unit** (full-reconcile + API): `LocalizeToCluster`, `AssignClusterIndices`, `ValidateSimulatedMCClusterIndices`, `ProjectedReconcilesLocalOnly` + `NoMatchSilentNoOp` (RS + sharded), and a regression guard asserting the **resolved** cluster index (not array position) drives the SNI hostname (pinned index 7 ≠ array position 0).
- **E2E** (three-operator topology — central operator + a simulated MCK operator scoped to `mongodbsearch` per member cluster): `e2e_search_simulated_mc_rs` (RS, TLS everywhere, restore → `$search` index → aggregation returns rows per cluster) and `e2e_search_simulated_mc_sharded` (sharded, per-cluster `$search` **and** `$vectorSearch` through each local mongos). Both assert zero cross-cluster resource leak + local-only status. Wired into `e2e_multi_cluster_kind` (+ static variant).
- **Rebased onto current `search/ga-base` (`35439898d`)** with zero conflicts; feature delta byte-identical. Local gates re-run green on the new base: `go build ./...`, `go vet`, `gofmt`, Go unit (`api/search` + `searchcontroller` + operator Search/Envoy/ClusterIndex/SimulatedMC), `black`/`isort`/`py_compile`, and `make manifests` = **zero CRD drift** (so `lint_repo`'s CRD check passes; the commit-6 fix makes the operator test package compile). E2E re-validation on the rebased tip is in flight; the feature-identical pre-rebase heads were Evergreen-green: `e2e_search_simulated_mc_sharded` (`6a2fef6c`, exec 0) and `e2e_search_simulated_mc_topology` (`6a2f487a`) on both kind variants, plus an earlier all-green run (`6a2c26db`) covering `e2e_search_simulated_mc_rs` × both kind variants and the q2/q3 + connectivity-tool suites that exercise every existing consumer of the extended `tests/common/search` helpers and the `tools_pod` change.

### Out of scope (v0)

- Finalizer-based cross-cluster CR cleanup (member-cluster resource leaks on CR delete are an accepted limitation).
- Per-cluster status aggregation — status is local-only by design; a global view is read per-cluster.

## Tickets

KUBE-32


api/mongodb/v1/search/cluster_index.go
```diff
@@ -2,28 +2,44 @@ package search
 
 // AssignClusterIndices returns a new clusterName -> clusterIndex mapping that:
 //   - preserves every entry in existing (never deletes a name, even if it is
-//     not present in currentClusterNames — that's how indices are not reused
-//     on remove/re-add);
-//   - appends every name in currentClusterNames that is not already in the
-//     mapping, assigning indices monotonically starting at max(existing)+1 (or
-//     0 when existing is empty).
+//     not present in current — that's how indices are not reused on
```
anandsyncs: make the comment more concise

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -916,42 +909,40 @@ func (s *MongoDBSearch) GetManagedLBEndpointForShard(shardName string) string {
 }
 
 // GetManagedLBEndpointForCluster resolves {clusterName} and {clusterIndex} in
-// the externalHostname template for spec.clusters[i]. Returns "" when managed
+// the externalHostname template, substituting the caller-supplied resolved
+// cluster index and name directly. The resolved index is the user-pinned
+// ClusterIndex in simulated-MC (or the mapping index in legacy MC); callers
```
anandsyncs: There is no legacy MC, there are 2 viable paths. 
Operator per cluster(a.k.a simulated) and Hub and spoke MC.

api/mongodb/v1/search/cluster_index_test.go
```diff
@@ -22,45 +30,72 @@ func TestAssignClusterIndices(t *testing.T) {
...
+			want:     map[string]int{"us-east": 7, "us-west": 3},
+		},
+		{
+			name:     "pinned ClusterIndex overrides existing",
+			existing: map[string]int{"us-east": 0, "us-west": 1},
+			current:  []ClusterSpec{{ClusterName: "us-east", ClusterIndex: pin(9)}, {ClusterName: "us-west"}},
+			want:     map[string]int{"us-east": 9, "us-west": 1},
+		},
+		{
+			name:     "mixed pinned + unpinned: pin honored, rest monotonic from max",
```
anandsyncs: there should be no mixed pinned or unpinned, it's either pinned or unpinned.

api/mongodb/v1/search/cluster_index_test.go
```diff
@@ -22,45 +30,72 @@ func TestAssignClusterIndices(t *testing.T) {
...
+			existing: map[string]int{},
+			current:  []ClusterSpec{{ClusterName: "us-east", ClusterIndex: pin(5)}, {ClusterName: "us-west"}},
+			want:     map[string]int{"us-east": 5, "us-west": 6},
+		},
+		{
+			// Two distinct names pinning the same index: AssignClusterIndices keeps
+			// both at that index (it keys on name, not index). Uniqueness across
+			// clusterIndex values is a CRD CEL rule enforced at admission, not here.
+			name:     "duplicate pinned index across distinct names: both kept (CEL enforces uniqueness at admission)",
```
anandsyncs: two clusters cannot have same index.
if one clusterIndex is assigned explicitly then all of them need to be assigned explicitly.

api/mongodb/v1/search/mongodbsearch_types_test.go
```diff
@@ -522,3 +585,68 @@ func TestMongotConfigConfigMapNameForCluster(t *testing.T) {
 	got1 := s.MongotConfigConfigMapNameForCluster(1)
 	assert.Equal(t, "mdb-search-search-1-config", got1.Name)
 }
+
+func TestMongoDBSearch_LocalizeToCluster(t *testing.T) {
+	twoClusters := func() *[]ClusterSpec {
+		return &[]ClusterSpec{{ClusterName: "us-east"}, {ClusterName: "us-west"}}
```
anandsyncs: if this is for simulated clusters then index is mandatory.

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -1075,3 +1059,33 @@ func (s *MongoDBSearch) GetOwnerLabels() map[string]string {
...
+
+// ValidateSimulatedMCClusterIndices enforces the simulated-multi-cluster contract:
+// spec.clusters must be non-empty and every entry must pin a ClusterIndex.
+// Called only by operators running in simulated-MC mode (operatorClusterName set).
+func (s *MongoDBSearch) ValidateSimulatedMCClusterIndices() error {
+	if s.Spec.Clusters == nil || len(*s.Spec.Clusters) == 0 {
+		return fmt.Errorf("simulated multi-cluster mode requires spec.clusters to be set")
+	}
+	for _, c := range *s.Spec.Clusters {
+		if c.ClusterIndex == nil {
```
anandsyncs, line 1058: They also need to be unique.

api/mongodb/v1/search/mongodbsearch_types_test.go
```diff
@@ -453,49 +442,123 @@ func TestGetManagedLBEndpointForClusterShard_ClusterIndex(t *testing.T) {
 			Clusters:     &clusters,
 		},
 	}
-	assert.Equal(t, "search-0.shard-a.lb.example.com:443", s.GetManagedLBEndpointForClusterShard(0, "shard-a"))
-	assert.Equal(t, "search-1.shard-b.lb.example.com:443", s.GetManagedLBEndpointForClusterShard(1, "shard-b"))
+	assert.Equal(t, "search-0.shard-a.lb.example.com:443", s.GetManagedLBEndpointForClusterShard(0, "us-east-k8s", "shard-a"))
+	assert.Equal(t, "search-1.shard-b.lb.example.com:443", s.GetManagedLBEndpointForClusterShard(1, "eu-west-k8s", "shard-b"))
+}
+
+// Regression guard: {clusterIndex} substitution uses exactly the resolved index
```
anandsyncs: make these comments more concise.

api/mongodb/v1/search/mongodbsearch_validation_test.go
```diff
@@ -772,6 +772,76 @@ func TestValidateMCMatchTagsNonEmpty(t *testing.T) {
 	}
 }
 
+// TestValidateSpec_MCShardedSource covers the un-narrowed MC sharded path that the
+// simulated-MC reconcilers MUST validate before LocalizeToCluster narrows
+// spec.clusters[] to a 1-element slice. After narrowing, the MC validators
+// short-circuit via their `len(spec.clusters) <= 1` guards and silently accept
+// misconfigured templates / duplicate shardNames. These cases lock in the
+// pre-localization contract.
+func TestValidateSpec_MCShardedSource(t *testing.T) {
```
anandsyncs: remove this guard if the relevant details are already asserted in another test.

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -152,6 +154,15 @@ type ClusterSpec struct {
 	// +optional
 	// +kubebuilder:validation:MaxLength=253
 	ClusterName string `json:"clusterName,omitempty"`
+	// ClusterIndex is the stable integer in per-cluster resource names. Required in
+	// simulated multi-cluster mode. Changing it renames this cluster's per-cluster
```
lsierant: we should probably avoid mentioning "simulated multi-cluster" in public docs. Let's figure out a better name for that.

api/mongodb/v1/search/cluster_index.go
```diff
@@ -1,29 +1,39 @@
...
+// AssignClusterIndices returns a new clusterName -> clusterIndex mapping that
+// preserves every entry in existing (so indices are not reused on remove/re-add),
+// honors any user-pinned spec.clusters[].ClusterIndex, and assigns the remaining
+// unpinned names monotonically from max(existing/pinned)+1.
 //
 // existing is not mutated; callers may safely pass in a map from any persistence
 // layer. The returned map is always non-nil.
-func AssignClusterIndices(existing map[string]int, currentClusterNames []string) map[string]int {
-	result := make(map[string]int, len(existing)+len(currentClusterNames))
+func AssignClusterIndices(existing map[string]int, current []ClusterSpec) map[string]int {
```
lsierant, line 7: the complexity of automatically assigned indices is not worth it in general

anandsyncs: We can resolve this in the slack discussion and then fix in a followup PR

anandsyncs: We decided to go with manual indexes for now.

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -897,39 +906,40 @@ func (s *MongoDBSearch) GetManagedLBEndpointForShard(shardName string) string {
...
 
 // GetManagedLBEndpointForClusterShard returns the externalHostname template with
 // {clusterName}, {clusterIndex}, and {shardName} all resolved for the
-// (spec.clusters[i], shardName) pair. Used for sharded multi-cluster
-// MongoDBSearch deployments. Returns "" when managed LB is not configured.
-func (s *MongoDBSearch) GetManagedLBEndpointForClusterShard(i int, shardName string) string {
-	out := s.GetManagedLBEndpointForCluster(i)
+// (cluster, shard) pair. Used for sharded multi-cluster MongoDBSearch
+// deployments. Returns "" when managed LB is not configured.
+func (s *MongoDBSearch) GetManagedLBEndpointForClusterShard(clusterIndex int, clusterName, shardName string) string {
```
anandsyncs, line 925: Managed LB is mandatory for all multi cluster scenarios. Is this being followed?

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -897,39 +906,40 @@ func (s *MongoDBSearch) GetManagedLBEndpointForShard(shardName string) string {
...
 		return ""
 	}
 	return strings.ReplaceAll(out, ShardNamePlaceholder, shardName)
 }
 
 // GetManagedLBEndpointForClusterLevel derives the mongos-facing endpoint by stripping
-// the leading "{shardName}." from externalHostname. Returns "" when not derivable;
-// callers fall back to the cluster-level proxy Service FQDN.
-func (s *MongoDBSearch) GetManagedLBEndpointForClusterLevel(i int) string {
+// the leading "{shardName}." from externalHostname, then substituting the
```
anandsyncs: Mach es kürzer!

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -938,17 +948,16 @@ func (s *MongoDBSearch) GetManagedLBEndpointForClusterLevel(i int) string {
...
 	}
-	hasClusterPlaceholder := strings.Contains(trimmed, ClusterNamePlaceholder) ||
-		strings.Contains(trimmed, ClusterIndexPlaceholder)
-	clusters := s.Spec.Clusters
-	if i < 0 || i >= len(clusters) {
-		if hasClusterPlaceholder {
-			return ""
-		}
-		return trimmed
+	// A {clusterName} placeholder with no name (the operator-managed sharded
```
anandsyncs: Mach es kürzer!

api/mongodb/v1/search/mongodbsearch_validation.go
```diff
@@ -281,11 +283,11 @@ func generateShardResourceNames(s *MongoDBSearch, shardName string, clusterIndex
...
 }
 
-// maxValidationClusterIndex returns the largest persisted index admission can foresee
-// (= len(spec.clusters)-1 for a fresh assignment; 0 for single-cluster).
+// maxValidationClusterIndex returns the cluster index used for worst-case DNS-length
+// checks — len(spec.clusters)-1, a heuristic pinned indices can exceed (see TODO).
 //
-// TODO: ClusterMapping is monotonic-append-only, so persisted indices can exceed
-// len-1 after remove→re-add cycles. Admission underestimates the real max; read
+// TODO: pinned indices may exceed len-1 (Maximum=999) and removed clusters' mapping
```
anandsyncs, line 288: validate this and check if it needs to updated.

api/mongodb/v1/search/mongodbsearch_validation.go
```diff
@@ -599,8 +621,18 @@ func validateExternalHostnameDNSLength(s *MongoDBSearch) v1.ValidationResult {
 
 	// Iterate the cross-product. len(shardNames) == 0 runs a single pass per
 	// cluster with no shard substitution.
-	for i := range s.Spec.Clusters {
-		base := s.GetManagedLBEndpointForCluster(i)
+	for i, c := range s.Spec.Clusters {
+		// Resolve the cluster index for clusters[i]: the pinned ClusterIndex when
```
anandsyncs: Is this sensible? Why is there a 5 line comment explaining this simple concept.

controllers/operator/mongodbsearch_controller.go
```diff
@@ -74,18 +74,82 @@ func loadOrInitSearchState(
 	return state, store, nil
 }
 
+// resolveClusterMapping is called only by the main reconciler — the single writer of the
```
anandsyncs: clarify that main reconciler is the search reconciler

PR #1042, opened 2026-04-30
Replicaset external members

lucian-tosa:
# Based on PR #800

## Chain of upstream PRs as of 2026-06-03

* PR #800:
  `master` ← `vm-migration-feature-branch`

  * **PR #1042 (THIS ONE)**:
    `vm-migration-feature-branch` ← `lucian/rs-external-members`

# Summary

## Checklist

- [ ] Have you linked a jira ticket and/or is the ticket in the title?
- [ ] Have you checked whether your jira ticket required DOCSP changes?
- [ ] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

PR #1133, opened 2026-05-22
feat: sharded cluster external members support for VM-to-Kubernetes migration

filipcirtog:
# Based on PR #1042

* PR #800:

  * PR #1042:
    `vm-migration-feature-branch` ← `lucian/rs-external-members`

    * **PR #1133 (THIS ONE)**:
      `lucian/rs-external-members` ← `filip/sharded-external-members`

## Why
The RS external members feature (on the base branch) lets VM-managed replica set members coexist in Ops Manager alongside K8s-managed members during a VM-to-Kubernetes migration. Sharded clusters need the same capability, but have three distinct process types (mongos, config server, shard mongods) each with their own replica set names and shard identities. These concepts have no equivalent in a plain replica set.

Without this PR, the operator would remove VM-managed processes from Ops Manager on first reconcile, breaking the running cluster before the migration can begin.

## What

**New API fields**
- `spec.externalMembers`: a list of VM-managed processes (processName, hostname, type, replicaSetName) that the operator must not remove or overwrite in Ops Manager during reconciliation.
- `spec.shardNameOverrides`: an immutable slice that maps K8s shard StatefulSet names to their AC identities. Entries are matched by `shardName` so order is not significant. Two forms are supported.

  Brevity form (`shardName` only): all three identities are equal. Use this when the VM deployment named the shard the same as the K8s default (`{resourceName}-{index}`).

  Full form (`shardName` + `shardId` + `replicaSetName`): use this when the VM deployment used different names. All three fields must be set together or omitted together.

  - `shardName`: the K8s StatefulSet name for the shard, always computed as `{resourceName}-{index}`. Acts as the key for matching entries and for `shardOverrides` references. Cannot be changed after creation.
  - `shardId`: the shard `_id` in the AC sharding section. The merge matches shards by `_id`, so this must match the value already present in the AC. 
  - `replicaSetName`: the `replicaSetName` of the shard's processes in the AC. Used to look up the correct process list when applying internal-auth cert paths and when cleaning up after scale-down.

- `spec.configServerNameOverride` and `spec.shardedClusterNameOverride`: fields that override the AC replicaSetName for the config server and the AC cluster name for the sharded cluster, so that the operator can manage a sharded cluster whose AC names were assigned by a VM deployment.

**OM deployment layer**
- Extends `mergeMongosProcesses`, `mergeConfigReplicaSet`, and `mergeShards` to skip external members during the removal pass, mirroring the existing RS behaviour.
- Adds `ShardNamePrefix` to `DeploymentShardedClusterMergeOptions` so that `findReplicaSetsRemovedFromShardedCluster` uses the K8s StatefulSet name prefix for orphaned-shard detection even when `shardedClusterNameOverride` changes the AC cluster name.

**Process naming during migration**
- Applies `IsLegacyDeployment` consistently across `createDesiredMongosProcesses`, `createDesiredConfigSrvProcessesAndMemberOptions`, `createDesiredShardProcessesAndMemberOptions`, and `getHealthyProcessNamesToWaitForReadyState` so that the desired process list and the healthy-wait list always use the same naming scheme (bare pod names for legacy AC, `k8s/<namespace>/<pod>` for new deployments). Without this, `WaitForReadyState` would wait for names that never appear in the AC during migration.

**Validation**
- `externalMembers` entries are immutable once set.
- `shardNameOverrides` entries are immutable once set, must have unique `shardName` values validated against the `{resourceName}-{index}` format, and must be either brevity form (shardName only) or full form (shardName + shardId + replicaSetName).
- `shardName` is enforced as a required field in the CRD schema.
- `validateShardedACIdentity` runs before the merge during migration and verifies that the AC cluster name, config server RS name, and all shard RS names and shard IDs match the values the resource resolves to. This prevents a mistyped override from corrupting the AC.

**E2E tests**
- Adds plain TLS sharded migration (`e2e_vm_migration_sharded`) and X509 sharded migration (`e2e_vm_migration_sharded_x509`), covering the full promote-and-prune cycle.


api/v1/mdb/sharded_cluster_validation.go
```diff
@@ -195,6 +196,28 @@ func shardOverridesClusterSpecListsCorrect(m MongoDB) v1.ValidationResult {
...
+// ShardName values must be unique across all entries.
+func shardNameOverridesValidForm(m MongoDB) v1.ValidationResult {
+	seen := make(map[string]bool, len(m.Spec.ShardNameOverrides))
+	for i, o := range m.Spec.ShardNameOverrides {
+		if o.ShardName == "" {
+			return v1.ValidationError("spec.shardNameOverrides[%d]: shardName is required", i)
+		}
+		if seen[o.ShardName] {
+			return v1.ValidationError("spec.shardNameOverrides[%d]: shardName %q is a duplicate", i, o.ShardName)
+		}
```
lucian-tosa: we might as well do the same for shardId and replicasetname

api/v1/mdb/shardedcluster.go
```diff
@@ -7,6 +7,27 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/mongodb-community-operator/pkg/automationconfig"
 )
 
+// ShardNameOverride maps one Kubernetes shard to its automation config identity during a VM-to-Kubernetes migration.
+// The slice is position-significant: index 0 targets shard 0, index 1 targets shard 1, and so on.
```
lucian-tosa: The order should not be significant due to the k8s name mapping.

api/v1/mdb/shardedcluster.go
```diff
@@ -7,6 +7,27 @@ import (
...
+// Shards beyond the slice length keep their K8s default names.
+// Once set, entries are immutable and must not be reordered or extended.
+// ShardName is the override name used to reference this shard in shardOverrides.shardNames.
+// ShardId and ReplicaSetName are the corresponding automation config values for this shard.
```
lucian-tosa: nit: duplicate comments

api/v1/mdb/shardedcluster.go
```diff
@@ -7,6 +7,27 @@ import (
...
+type ShardNameOverride struct {
```
lucian-tosa, line 19: we should add a comment for which field corresponds with the statefulset/pod name

api/v1/mdb/mongodb_validation.go
```diff
@@ -431,6 +431,21 @@ func noExternalMembersAdditionOrChanges(newObj, oldObj MongoDbSpec) v1.Validatio
 	return v1.ValidationSuccess()
 }
 
+func noShardNameOverridesChanges(newObj, oldObj MongoDbSpec) v1.ValidationResult {
+	if len(newObj.ShardNameOverrides) < len(oldObj.ShardNameOverrides) {
+		return v1.ValidationError("Cannot remove shardNameOverrides entries from an existing MongoDB resource.")
```
lucian-tosa: in case you scale down, this should be allowed

docker/mongodb-kubernetes-tests/tests/tls/vm_migration_x509.py
```diff
@@ -417,3 +432,14 @@ def test_promote_and_prune(mdb_migration: MongoDB, vm_sts):
...
+    mdb_health_checker.assert_healthiness()
+
+
+@mark.e2e_vm_migration_x509
+def test_process_names(om_tester: OMTester, namespace: str, mdb_migration: MongoDB):
+    ac_tester = om_tester.get_automation_config_tester()
+    process_names = [process["name"] for process in ac_tester.get_all_processes()]
+    assert f"k8s/{namespace}/{mdb_migration.name}-0" in process_names
+    assert f"k8s/{namespace}/{mdb_migration.name}-1" in process_names
+    assert f"k8s/{namespace}/{mdb_migration.name}-2" in process_names
```
lucian-tosa, line 479: these are not in line with what we are doing for replicasets. We should use the same naming scheme

api/v1/mdb/sharded_cluster_validation.go
filipcirtog: Done! Thanks!

api/v1/mdb/shardedcluster.go
filipcirtog: Fixed, entries are now matched by ShardName (the K8s StatefulSet name) rather than by slice index, so order is not significant.

api/v1/mdb/shardedcluster.go
filipcirtog: Fixed!

api/v1/mdb/shardedcluster.go
filipcirtog: Done!

api/v1/mdb/mongodb_validation.go
filipcirtog: Fixed. Split into two validators: `noShardNameOverridesAddedOrModified` (blocks adds and changes) and `noShardNameOverridesRemovedForActiveShards` (blocks removal only for shards still within shardCount, allowing removal of scaled-away shards).

docker/mongodb-kubernetes-tests/tests/tls/vm_migration_x509.py
filipcirtog: Could you please follow-up o this? This a replicaset test.

filipcirtog: Solved. Thank you!

api/v1/mdb/mongodb_types.go
```diff
@@ -1360,10 +1399,24 @@ func (m *MongoDB) MongosRsName() string {
...
 }
 
+func (m *MongoDB) ConfigACRsName() string {
+	if m.Spec.ConfigServerNameOverride != "" {
+		return m.Spec.ConfigServerNameOverride
+	}
+	return m.ConfigRsName()
+}
+
 func (m *MongoDB) ShardRsName(i int) string {
```
lucian-tosa: We should rethink the naming of the methods. The one you added below should be the only one used

api/v1/mdb/mongodb_validation.go
```diff
@@ -431,6 +431,73 @@ func noExternalMembersAdditionOrChanges(newObj, oldObj MongoDbSpec) v1.Validatio
...
+	for _, o := range old.Spec.ShardNameOverrides {
+		if _, exists := newByShardName[o.ShardName]; exists {
+			continue
+		}
+		// Entry was removed. It is only allowed if the shard no longer exists.
+		for i := 0; i < m.Spec.ShardCount; i++ {
+			if m.ShardRsName(i) == o.ShardName {
+				return v1.ValidationError("Cannot remove shardNameOverrides entry for active shard %q.", o.ShardName)
+			}
+		}
```
lucian-tosa: nit: you can use another set to avoid another loop

api/v1/mdb/sharded_cluster_validation.go
```diff
@@ -195,6 +196,71 @@ func shardOverridesClusterSpecListsCorrect(m MongoDB) v1.ValidationResult {
...
+			return v1.ValidationError("shards %d and %d both resolve to AC shard _id %q", prev, i, shardId)
+		}
+		shardIdToShardIdx[shardId] = i
+	}
+	if i, ok := rsNameToShardIdx[m.ConfigACRsName()]; ok {
+		return v1.ValidationError("shard %d resolves to the same AC replicaSetName %q as the config server", i, m.ConfigACRsName())
+	}
+	if i, ok := rsNameToShardIdx[m.GetShardedClusterName()]; ok {
+		return v1.ValidationError("shard %d resolves to AC replicaSetName %q which equals the sharded cluster name", i, m.GetShardedClusterName())
+	}
```
lucian-tosa: Is it a problem if the name of the sharded cluster collides with a replica set name? They should be independent

api/v1/mdb/sharded_cluster_validation.go
```diff
@@ -195,6 +196,71 @@ func shardOverridesClusterSpecListsCorrect(m MongoDB) v1.ValidationResult {
...
+		if prev, ok := rsNameToShardIdx[rsName]; ok {
+			return v1.ValidationError("shards %d and %d both resolve to AC replicaSetName %q", prev, i, rsName)
+		}
+		rsNameToShardIdx[rsName] = i
+		shardId := m.ShardACShardId(i)
+		if prev, ok := shardIdToShardIdx[shardId]; ok {
+			return v1.ValidationError("shards %d and %d both resolve to AC shard _id %q", prev, i, shardId)
+		}
+		shardIdToShardIdx[shardId] = i
+	}
```
lucian-tosa, line 257: This validates the same thing as above, doesn't it?

controllers/om/replicaset/om_replicaset.go
```diff
@@ -126,6 +126,43 @@ func PrepareScaleDownFromMongoDB(omClient om.Connection, rs *mdbv1.MongoDB, log
 		return xerrors.Errorf("dev error: the number of members being scaled down was > 1, scaling more than one member at a time is not possible! %s", podNames)
 	}
 
-	log.Debugw("Setting votes to 0 for members", "members", podNames)
-	return PrepareScaleDownFromMap(omClient, map[string][]string{rs.Name: podNames}, podNames, log)
+	processNames := podNamesToProcessNames(omClient, rs.Name, rs.Namespace, rs.Spec.GetExternalMemberProcessNames(), podNames, log)
+	log.Debugw("Setting votes to 0 for members", "members", processNames)
+	return PrepareScaleDownFromMap(omClient, map[string][]string{rs.Name: processNames}, processNames, log)
```
lucian-tosa: This is not needed anymore. Agents can remove members even if they are voting or are primaries.

api/v1/mdb/sharded_cluster_validation.go
lucian-tosa, line 1: we should also validate that shardCount is not less than the overrides (i.e. if shards are getting removed, the overrides for that shard should also be removed)

controllers/om/deployment.go
```diff
@@ -849,9 +894,16 @@ func (d Deployment) mergeShards(opts DeploymentShardedClusterMergeOptions, log *
...
+func (d Deployment) handleShardsRemoval(finalizing bool, s ShardedCluster, removedShardRsNames []string, log *zap.SugaredLogger) bool {
 	junkReplicaSets := d.findReplicaSetsRemovedFromShardedCluster(s.Name())
 
+	// Shards removed by this merge are not in the draining list yet, add them when their replica set exists.
+	for _, rsName := range removedShardRsNames {
+		if !stringutil.Contains(junkReplicaSets, rsName) && d.GetReplicaSetByName(rsName) != nil {
+			junkReplicaSets = append(junkReplicaSets, rsName)
+		}
+	}
+
```
lucian-tosa: can you explain how this works?

controllers/operator/common_controller.go
```diff
@@ -451,52 +451,204 @@ const MaxVotingMembers = 7
...
+		total := externalVoting + len(k8sVotingPositions)
+		if total <= MaxVotingMembers {
+			continue
+		}
+
+		acVoting := collectACVotingMembers(rs, externalSet)
+		excess := total - MaxVotingMembers
+		// Treat all K8s voting positions as "newly voting" since K8s members may not yet exist in the AC.
+		return workflow.Failed(xerrors.Errorf("%s", formatTooManyVotingMembersError(
+			rsName, total, acVoting, k8sVotingPositions, excess,
```
lucian-tosa: it would be cleaner to reuse the validation for the replicaset

controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -304,6 +305,14 @@ func (r *ShardedClusterReconcileHelper) getShardNameToShardIdxMap() map[string]i
...
 	for shardIdx := 0; shardIdx < max(r.sc.Spec.ShardCount, r.deploymentState.Status.ShardCount); shardIdx++ {
 		mapping[r.sc.ShardRsName(shardIdx)] = shardIdx
+		if o := r.sc.ShardNameOverrideForShard(shardIdx); o != nil {
+			if o.ShardId != "" {
+				mapping[o.ShardId] = shardIdx
+			}
+			if o.ReplicaSetName != "" {
+				mapping[o.ReplicaSetName] = shardIdx
+			}
+		}
```
lucian-tosa: if shardName, shardId, and replicasetname are all different, we get 3 entried in the mapping with the same value. Is that desired?

api/v1/mdb/mongodb_types.go
filipcirtog: updated! thank you!

api/v1/mdb/mongodb_validation.go
filipcirtog: updated! thank you!

api/v1/mdb/sharded_cluster_validation.go
filipcirtog: updated! thank you!

controllers/om/replicaset/om_replicaset.go
filipcirtog: as discussed, the whole function should get removed in a separate pr.

controllers/operator/common_controller.go
filipcirtog: updated! thank you!

api/v1/mdb/sharded_cluster_validation.go
filipcirtog: No, these check two different AC fields. ShardACRsName(i) resolves the shard's replicaSetName, while ShardACShardId(i) resolves its shard _id. In full-override form these are independent values and can differ.

api/v1/mdb/sharded_cluster_validation.go
filipcirtog: updated. thanks!

controllers/operator/mongodbshardedcluster_controller.go
filipcirtog: it does not make any difference, does it?

filipcirtog: or do you suggest having a separate condition for that?

lucian-tosa (approved):

api/v1/mdb/sharded_cluster_validation.go
lucian-tosa: where was this added? It should be enforced that overrides are less than or equal to shardcount

controllers/operator/mongodbshardedcluster_controller.go
lucian-tosa: I'm just not sure what the reasoning behind it is. This is supposed to be a map from shard name to index. Now it also contains a key for the replica set name if it is overridden and also for the shardid. But are these keys actually used?

filipcirtog: removed. thank you!

api/v1/mdb/sharded_cluster_validation.go
filipcirtog:
```
if !validateShardName(o.ShardName, m.Spec.ShardCount, m.Name) {
			return v1.ValidationError("spec.shardNameOverrides[%d]: shardName %q must be %s-{index} with index < %d", i, o.ShardName, m.Name, m.Spec.ShardCount)
		}
```

```
// If the MDB resource name is foo, and we have n shards, we verify that shard names ∈ {foo-0 , foo-1 ..., foo-(n-1)}
func validateShardName(shardName string, shardCount int, resourceName string) bool {
	// The shard number should not have leading zeros except for 0 itself
	pattern := fmt.Sprintf(`^%s-(0|[1-9][0-9]*)$`, resourceName)

	re := regexp.MustCompile(pattern)
	if !re.MatchString(shardName) {
		return false
	}

	// Extract the shard number from the matched part
	parts := re.FindStringSubmatch(shardName)
	shardNumber, err := strconv.Atoi(parts[1])
	if err != nil {
		return false
	}

	if shardNumber < 0 || shardNumber >= shardCount {
		return false
	}
	return true
}
```

PR #1134, opened 2026-05-22
feat: connectivity dry-run validation for sharded cluster VM migration

filipcirtog:
# Based on PR #1133

* PR #800:

  * PR #1042:
    `vm-migration-feature-branch` ← `lucian/rs-external-members`

    * PR #1133:
      `lucian/rs-external-members` ← `filip/sharded-external-members`

      * **PR #1134 (THIS ONE)**:
        `filip/sharded-external-members` ← `filip/sharded-connectivity-validation`

## Why
Before committing to a full VM-to-Kubernetes migration, operators need to verify that all external VM members are reachable from inside the cluster — without touching Ops Manager config or StatefulSets. The RS migration already has this via the `migration-dry-run` annotation. Sharded clusters need the same safety net: a sharded cluster has more process types (mongos, config server, multiple shards) and more failure modes, so a reachability check before migration is even more valuable.

## What
- Adds `runConnectivityValidationDryRun` to `ShardedClusterReconcileHelper`: setting the annotation `mongodb.com/migration-dry-run: "true"` launches a validator Job (built from the mongos StatefulSet spec) that pings all external members. No OM config or StatefulSets are modified.
- The resource stays in `PhaseConnectivityValidation` while the Job runs; transitions to passed (condition `NetworkConnectivityVerification=True`) when all members are reachable. Removing the annotation resumes normal reconciliation.
- X509 path: the validator reads the automation agent certificate subject DN from the agent cert secret and passes it to the Job for mutual TLS validation.
- Makes `workflow.failedStatus.WithAdditionalOptions` variadic to align with the other status types.
- Adds unit tests for all dry-run outcomes (running, passed, failed, no healthy cluster, no external mongos, operator image unset, x509 cert subject read error).
- Adds dry-run step to both sharded E2E test suites before the first full reconcile.


controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -908,7 +910,7 @@ func (r *ShardedClusterReconcileHelper) Reconcile(ctx context.Context, log *zap.
 	r.automationAgentVersion = automationAgentVersion
 
 	workflowStatus := r.doShardedClusterProcessing(ctx, sc, conn, projectConfig, log)
```
lucian-tosa, line 962: This line already makes a lot of changes before we even check if we are doing a dry-run.
All the dryrun steps should be moved above this line

controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -2460,6 +2483,97 @@ func (r *ShardedClusterReconcileHelper) migrateToNewDeploymentState(sc *mdbv1.Mo
...
+// runConnectivityValidationDryRun launches (or polls) a connectivity-validator Kubernetes Job
+// that checks whether all external MongoDB members in the current Ops Manager deployment are
+// reachable from within the cluster. No StatefulSets or Ops Manager config are modified.
+// The Job is built from the mongos StatefulSet spec so it uses the same credentials volumes and
+// mounts as the mongos STS. Sharded cluster clients connect through mongos, so the connection
+// string targets external mongos members and the validator pings each external member directly.
+//
+// The MongoDB resource phase stays PhaseConnectivityValidation for both in-progress and passed
+// outcomes; the migration condition carries ConnectivityCheckRunning or ConnectivityCheckPassed.
+func (r *ShardedClusterReconcileHelper) runConnectivityValidationDryRun(ctx context.Context, sc *mdbv1.MongoDB, opts deploymentOptions, operatorImage string, log *zap.SugaredLogger) workflow.Status {
```
lucian-tosa, line 2745: There is a leftover TODO on the original method in the replica set controller to move it to the common controller.
These 2 should be combined into only one function reusable by both replicasets and sharded in the common controller

docker/mongodb-kubernetes-tests/kubetester/omtester.py
```diff
@@ -272,6 +272,23 @@ def assert_cluster_available(self, name: str):
...
+    def assert_cluster_available(self, name: str, timeout: int = 120):
+        deadline = time.time() + timeout
+        availability = None
+        while time.time() < deadline:
+            availability = self.get_cluster_availability(name)
+            if availability == "available":
+                return
+            time.sleep(5)
+        assert False, f"Cluster {name} is not available after {timeout}s, last availability: {availability}"
+
```
lucian-tosa: are these used?

lucian-tosa (approved):

docker/mongodb-kubernetes-tests/kubetester/omtester.py
filipcirtog: sorry I was testing something for a different patch and forgot to remove.

filipcirtog: removed! thanks!

PR #1145, opened 2026-05-27
KUBE-2: fix JVM params accumulating in mms.conf across pod restarts

filipcirtog:
## Problem

The `mmsconfiguration` binary appended the `Xmx`/`Xms` JVM parameter block to `mms.conf` on every pod cycle without checking for or removing existing entries. Over time this caused the file to accumulate thousands of duplicate lines, eventually causing the JVM to fail at startup with:

```
bash: Argument list too long
```

The bug has existed since the operator's inception but only recently surfaced in a customer environment where the file had grown to tens of thousands of duplicate entries. The immediate workaround was to truncate `mms.conf` manually.

Root cause: [edit_mms_configuration.go](docker/mongodb-kubernetes-init-ops-manager/mmsconfiguration/edit_mms_configuration.go) called `appendLinesToFile` unconditionally on every reconciliation.

## Solution

Replace the blind append with a read-strip-write approach:

1. Read the existing `mms.conf`.
2. Strip all previously written operator JVM blocks (identified by the existing banner comment used as a marker).
3. Write the cleaned content back with exactly one fresh block.

This makes the operation idempotent. Files that have already accumulated thousands of duplicate blocks are cleaned up on the first reconciliation after this change is deployed.

## Changes

- **`writeJvmParamsToFile`** (new): reads the file, strips existing operator blocks, writes cleaned content + one fresh block.
- **`removeOperatorJvmBlocks`** (new): strips every occurrence of the banner comment + param line pair using `strings.Cut`.
- **`appendLinesToFile`** (removed): dead code, no longer needed.
- **Tests** (4 new):
  - `UpdateConfFile_Idempotent`: 10 consecutive reconciliations produce exactly one param line.
  - `UpdateConfFile_ReplacesExistingBlock`: changing params removes the old value.
  - `UpdateConfFile_PreservesBaseContent`: a file pre-seeded with 3 duplicate operator blocks is cleaned up, base OM content is fully preserved, and exactly one fresh block is written.
  - `UpdateConfFile_Mms` (updated): switched from fragile line-index assertion to `Contains` check.

## Testing

```
ok  github.com/mongodb/mongodb-kubernetes/docker/mongodb-kubernetes-init-ops-manager/mmsconfiguration  (all 7 tests pass)
```

### DIFF
--- a/changelog/20260527_fix_jvm_params_accumulating_in_mms_conf.md
+++ b/changelog/20260527_fix_jvm_params_accumulating_in_mms_conf.md
@@ -3,4 +3,4 @@ kind: fix
 date: 2026-05-27
 ---
 
-* **MongoDBOpsManager JVM parameters**: Fixed an issue in the `MongoDBOpsManager` resource where JVM parameter blocks were appended to `mms.conf` on every pod restart without removing previous entries, causing duplicate configuration entries to accumulate.
+* **MongoDBOpsManager**: Fixed an issue in the `MongoDBOpsManager` resource where JVM parameter blocks were appended to `mms.conf` on every pod restart without removing previous entries, causing duplicate configuration entries to accumulate.
--- a/changelog/20260527_fix_jvm_params_accumulating_in_mms_conf.md
+++ b/changelog/20260527_fix_jvm_params_accumulating_in_mms_conf.md
@@ -3,4 +3,4 @@ kind: fix
 date: 2026-05-27
 ---
 
-* Fixed an issue where JVM parameter blocks were appended to `mms.conf` on every pod restart without removing previous entries.
+* **MongoDBOpsManager JVM parameters**: Fixed an issue in the `MongoDBOpsManager` resource where JVM parameter blocks were appended to `mms.conf` on every pod restart without removing previous entries, causing duplicate configuration entries to accumulate.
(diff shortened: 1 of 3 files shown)


MaciejKaras (approved):
Nice fix!

mircea-cosbuc (approved):
❤️ LGTM

changelog/20260527_fix_jvm_params_accumulating_in_mms_conf.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2026-05-27
+---
+
+* **MongoDBOpsManager JVM parameters**: Fixed an issue in the `MongoDBOpsManager` resource where JVM parameter blocks were appended to `mms.conf` on every pod restart without removing previous entries, causing duplicate configuration entries to accumulate.
```
viveksinghggits:
```suggestion
* **MongoDBOpsManager**: Fixed an issue in the `MongoDBOpsManager` resource where JVM parameter blocks were appended to `mms.conf` on every pod restart without removing previous entries, causing duplicate configuration entries to accumulate.
```

Just MongoDBOpsManager would be better I think. But change it only if you also agree.

viveksinghggits (approved):

PR #1146, opened 2026-05-27
[KUBE-8] Add liveness probe to envoy load balancer

viveksinghggits:
# Summary

The envoy loadbalancer already had readiness probe which configures when envoy is ready/ready to accept traffic, but it lacked the liveness probe. Liveness probes are used to configure when a pod should be restarted, if in case it's stuck or something like that.
Figuring out the endpoint that can be used in liveness probe was a challenge because apart from `/ready` there is another endpoint `/server_info` that can be used to figure out if if envoy's admin interface is responsive. But after spending some time and looking at this PR https://github.com/envoyproxy/gateway/pull/5535 it looks like `/ready` would be a good idea because `/ready` is used in that upstream PR. 

## Proof of Work

NA

## Checklist

### DIFF
--- a/controllers/operator/mongodbsearchenvoy_controller.go
+++ b/controllers/operator/mongodbsearchenvoy_controller.go
@@ -628,6 +628,17 @@ func buildEnvoyPodSpec(search *searchv1.MongoDBSearch, clusterIndex int, tlsCfg 
 					InitialDelaySeconds: 5,
 					PeriodSeconds:       5,
 				},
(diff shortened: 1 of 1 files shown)


Julien-Ben (approved):
LGTM
When I created the envoy controller I indeed couldn't find a proper liveness endpoint. But from your PR description I see that it is fine to consider liveness == readiness in this case.

Don't forget to lint again before merging ;)

PR #1150, opened 2026-05-27
[KUBE-12] Make some circuit brekaers configurable via search CR

viveksinghggits:
# Based on PR #1048

* PR #1166:

  * PR #1048:

    * **PR #1150 (THIS ONE)**:
      `search/ga-base` ← `search/envoy-circuitbreaker-config`

# Summary

In the current implementation we just has the circuit breaker related config, like `MaxConnections`, `MaxRequests` and `MaxPendingRequests` hardcoded to the default value (1024). It makes more sense for some of these configs to be exposed via search CR so that users can configure those fields.
This PR does that exposing the `MaxConnections`, `MaxRequests` and `MaxPendingConnection` using the search CR.

This is how these fields can be configured using the search CR

```
  spec:
    loadBalancer:
      managed:
        circuitBreakers:
          maxConnections: 2048
          maxRequests: 4096
          maxPendingRequests: 512
```

## Proof of Work

NA

## Checklist


controllers/operator/envoy_config_builder.go
```diff
@@ -274,9 +303,9 @@ func buildCluster(route envoyRoute, tlsEnabled bool, caKeyName string) (*cluster
 			Thresholds: []*clusterv3.CircuitBreakers_Thresholds{
 				{
 					Priority:           corev3.RoutingPriority_DEFAULT,
-					MaxConnections:     wrapperspb.UInt32(1024),
-					MaxPendingRequests: wrapperspb.UInt32(1024),
-					MaxRequests:        wrapperspb.UInt32(1024),
+					MaxConnections:     wrapperspb.UInt32(circuitBreakerValue(cbMaxConnections(cb), 1024)),
```
Julien-Ben, line 315: Nit: let's replace the magic numbers with constant(s)

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -231,6 +231,28 @@ type ManagedLBConfig struct {
...
 	Deployment *common.DeploymentConfiguration `json:"deployment,omitempty"`
+	// CircuitBreakers configures Envoy circuit breaker thresholds for upstream clusters.
+	// When not set, defaults to 1024 for maxConnections, maxRequests, and maxPendingRequests.
+	// +optional
+	CircuitBreakers *EnvoyCircuitBreakers `json:"circuitBreakers,omitempty"`
+}
+
+type EnvoyCircuitBreakers struct {
+	// MaxConnections is the maximum number of TCP connections to the upstream cluster.
+	// Defaults to 1024 if not specified.
```
Julien-Ben, line 245: Could we use `kubebuilder:default=` here ?

I'm mostly concerned by the risk of this documentation drifting with the actual behavior, if we later change the value in the code

Julien-Ben (approved):

controllers/operator/envoy_config_builder.go
viveksinghggits: Even though these values seem duplicated right now, but they are not really. These values are for different fields so they can have different values tomorrow that's why I think it's ok not move them to consts. WDYT?

api/mongodb/v1/search/mongodbsearch_types.go
viveksinghggits: mmm, I somewhat agree with the fact that if we change the default value we will have to change the comment here but that's true for any documentation right? if we change something we will have to update the document to reflect that. And these API/field definitions directly go in our docs.

For the default using kubebuilder, I don't think it's the correct use case of that. the diff is if we use kubebuilder default the default value will be stored in the CR after resource is created. And because of that later if we decide to change the default it won't be straight forward to go ahead and update the values in already created CRs.

lsierant (commented):
I'm generally against merging this PR. Let's specify high enough defaults for those settings and not expose API for those. I don't see any reasonable situation where someone might need to change this and explaining in docs what exactly this would change will be very difficult.

PR #1155, opened 2026-05-28
KUBE-75: fix cross-cluster ownerReferences causing resource GC in remote member clusters

filipcirtog:
### Problem

In multi-cluster mode, Kubernetes resources (StatefulSets, Services, Secrets) were being written to member cluster clients with `ownerReferences` pointing to Custom Resources that only exist in the central cluster. The Kubernetes garbage collector follows those references, finds the owner absent in the member cluster, and deletes the resource as an orphan.

The root cause is that `kube.BaseOwnerReference(cr)` makes a resource owned *by* the CR. Calling it before writing to a member cluster client silently introduces a cross-cluster GC dependency.

### Fix

Apply the following guard wherever resources are written to member clusters:

```go
ownerRefs := kube.BaseOwnerReference(cr)
if cr.Spec.IsMultiCluster() {
    ownerRefs = nil
}
resource.OwnerReferences = ownerRefs
```

In multi-cluster mode, cleanup relies on label-based deletion via `GetOwnerLabels()` instead of GC-based ownerReference cleanup. The fix covers AppDB and BackupDaemon StatefulSets, OpsManager and BackupDaemon HTTPS cert Secrets, OpsManager and AppDB Services, sharded cluster StatefulSets and headless Services, MultiReplicaSet headless Services, and MongoDBUser connection string Secrets.

### Multi-cluster StatefulSet watch and `MultiClusterStatefulSetAnnotations`

Because ownerReferences are removed from member-cluster StatefulSets, watch callbacks can no longer use the standard `EnqueueRequestForOwner`. All three multi-cluster controllers now register a per-member-cluster `source.Kind` watch backed by `EnqueueRequestForOwnerMultiCluster`, which reads `MongoDBMultiResourceAnnotation` on each StatefulSet to map it back to the parent CR:

- `mongodbmultireplicaset_controller.go` — pre-existing watch, variable naming and comment improved.
- `mongodbopsmanager_controller.go` — **new**: AppDB StatefulSet changes in member clusters now trigger OpsManager reconciliation.
- `mongodbshardedcluster_controller.go` — **new**: sharded StatefulSet changes in member clusters now trigger reconciliation.

`MultiClusterStatefulSetAnnotations(crName)` is introduced in `pkg/handler/enqueue_owner_multi.go` as a single named helper for building that annotation map, replacing scattered inline `map[string]string{MongoDBMultiResourceAnnotation: crName}` literals across AppDB, sharded, and MongoDBMultiCluster construction.

### Tests

Unit tests added:
- `TestOpsManagerStatefulSet_MultiClusterOwnerRef`
- `TestBackupDaemonInKubernetes_ServiceOwnerReference`
- `TestOpsManagerInKubernetes_ServiceOwnerReference`
- `TestAppDBStatefulSet_MultiClusterOwnerRef`

E2E tests added (`test_statefulsets_multi_cluster_identity`):
- `multi_cluster_replica_set.py`
- `multicluster_appdb.py`
- `multi_cluster_sharded_simplest.py`


filipcirtog:
closed in favour of #1171 and #1172

PR #1156, opened 2026-05-28
KUBE-74 - Continue reconciliation with failed cluster without auto-failover

lucian-tosa:
# Summary

This PR fixes 2 issues that appear when automatic failover is disabled:
1. Reconciliation would stop if there is any failed cluster in the annotation. This means that even healthy clusters are not reconciled
2. Since automatic failover is disabled, we are not moving workloads between clusters when one cluster goes down. Therefore, the flickering issue (healthy <-> failed) is not an issue here. It is safe to remove the annotation when a cluster is back up and retry reconciliation on that cluster again.

This is far from a fundamental change to the cluster health checking. It is not in line with the MC sharded approach (where we check for health on each reconcile instead of background). But it does align it the sense that reconciliation continues on the healthy clusters.

## Proof of Work

New unit tests with mocked cluster health checker will add and remove the annotation when needed.
Updated `multi_cluster_network_partition` test to check if the failed cluster annotation is removed after recovery.

## Checklist


docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_recover_network_partition.py
```diff
@@ -112,13 +163,14 @@ def test_delete_database_statefulset_in_failed_cluster(
...
-def test_mongodb_multi_enters_failed_state(
+def test_mongodb_multi_is_stable(
     mongodb_multi: MongoDBMulti,
     namespace: str,
     central_cluster_client: client.ApiClient,
 ):
+    # Even if a cluster is failed, MongoDBMultiCluster should be stable and reconcile the rest of the healthy clusters
     mongodb_multi.load()
-    mongodb_multi.assert_reaches_phase(Phase.Failed, timeout=100)
+    mongodb_multi.assert_reaches_phase(Phase.Running, timeout=100)
```
mircea-cosbuc, line 173: Is it correct to assume that even in the case of losing majority the `MongoDBMultiCluster` resource would stay in running state?

pkg/multicluster/memberwatch/memberwatch.go
```diff
@@ -109,20 +113,39 @@ func (m *MemberClusterHealthChecker) WatchMemberClusterHealth(ctx context.Contex
 		for k, v := range m.Cache {
 			if v.IsClusterHealthy(log) {
 				log.Infof("Cluster %s reported healthy", k)
+				if multicluster.ShouldPerformFailover() {
+					continue
+				}
+
+				// If failover is disabled we should remove the cluster from the annotation after a number of health checks have succeeded
+				m.HealthyStreak[k] = min(m.HealthyStreak[k]+1, requiredHealthyStreak)
```
mircea-cosbuc: Do we need to protect this from concurrent updates?

mircea-cosbuc (commented):
Looks mostly good to me, I just have 2 clarifying questions.

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_recover_network_partition.py
lucian-tosa: Yes that is correct. I agree we should communicate this better across the board. Right now, the only way to determine that a cluster is unhealthy is by looking at logs or at the annotation.
As for the other resources, there is no indication.
We would need a new status condition to indicate unhealthy clusters, and maybe even a new phase. But we should implement this consistently for all resources.

pkg/multicluster/memberwatch/memberwatch.go
lucian-tosa: not needed, there is no parallelization here

changelog/20260528_fix_reconciliation_of_clusters_without_auto.md
```diff
@@ -0,0 +1,9 @@
+---
+kind: fix
```
anandsyncs, line 2: file name has two "fix" in name

anandsyncs (approved):

mircea-cosbuc (approved):

mircea-cosbuc (approved):

changelog/20260528_fix_reconciliation_of_clusters_without_auto.md
```diff
@@ -0,0 +1,7 @@
...
+date: 2026-05-28
+---
+
+* **MongoDBMulticluster**: Fixed an issue where reconciliation of an entire resource could be blocked if automatic failover was disabled and a member cluster was marked as failed. Now, the operator will reconcile healthy clusters and skip the failed clusters.
+* **MongoDBMulticluster**: When automatic failover is disabled, the operator will now remove the failed clusters annotation once the clusters respond successfully to health checks a consecutive number of times, allowing for recovery without manual intervention. The number of consecutive successful health checks required to remove the failed annotation is configurable via the `MDB_REQUIRED_HEALTHY_STREAK` environment variable, or the `multiCluster.requiredHealthyStreak` helm value. The default is 5.
```
m1kola: `multiCluster.requiredHealthyStreak` and especially `MDB_REQUIRED_HEALTHY_STREAK` might not be a clear name to indicate it is about member cluster healthy streak.

Maybe include "member cluster" in name?

pkg/multicluster/memberwatch/memberwatch.go
```diff
@@ -227,6 +248,27 @@ func AddFailoverAnnotation(ctx context.Context, mrs mdbmulti.MongoDBMultiCluster
...
 
+func removeClusterFromFailedAnnotation(ctx context.Context, mrs mdbmulti.MongoDBMultiCluster, clustername string, client kubernetesClient.Client) error {
+	failedClusters := readFailedClusterAnnotation(mrs.Annotations)
+
+	remaining := make([]failedcluster.FailedCluster, 0, len(failedClusters))
+	for _, failed := range failedClusters {
+		if failed.ClusterName != clustername {
+			remaining = append(remaining, failed)
+		}
+	}
```
m1kola: Nit: [`slices.DeleteFunc`](https://pkg.go.dev/slices#DeleteFunc) might be useful here.

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_recover_network_partition.py
m1kola: > We would need a new status condition to indicate unhealthy clusters, and maybe even a new phase. But we should implement this consistently for all resources.

In MCK 2.x we will have `MemberCluster` CRs for member clusters with status and conditions. We could eventually add health checks for member clusters there.

helm_chart/values.yaml
```diff
@@ -187,6 +187,7 @@ multiCluster:
   kubeConfigSecretName: mongodb-enterprise-operator-multi-cluster-kubeconfig
   performFailOver: true
   clusterClientTimeout: 10
+  requiredHealthyStreak: 5
```
m1kola: Q: Since we already have implicit defaulting in the operator, does it still make sense to have a value in `values.yaml`?

m1kola (commented):
I didn't review this thoroughly, but from 2.x/installation UX perspective this looks good. It will be another setting we would need to migrate from helm/env vars to OperatorConfig, but it is ok.

The only slight concern that requires attention is naming of the settings. Perhaps it could be clearer that it is about health of member clusters and not about workloads.

helm_chart/values.yaml
lucian-tosa: Having it in values.yaml makes the variable more visible IMO. If it's hidden away in the templates, or in docs, users might not realize it is configurable. Wdyt?

m1kola: It makes sense. The only thing is that we duplicate the default in multiple places at the moment: in both values files and in a go constant. But it is not a big deal.

changelog/20260528_fix_reconciliation_of_clusters_without_auto.md
lucian-tosa: `MDB_MEMBER_CLUSTER_REQUIRED_HEALTHY_STREAK` and `multiCluster.memberClusterRequiredHealthyStreak` ?

pkg/multicluster/memberwatch/memberwatch.go
lucian-tosa: fixed

helm_chart/values.yaml
lucian-tosa: Yes, that is not ideal, but having this visible is worth the risk of managing 2 duplicates. Added a comment on the go constant to help.

m1kola (approved):
Approving with the caveat that I haven't reviewed the whole PR. I mostly looked at it from the perspective of new operator settings and how it fits with MCK 2.x/OperatorConfig.

mircea-cosbuc (approved):

PR #1158, opened 2026-05-28
KUBE-78: bound automation-agent-stderr.log size with in-place rotation

filipcirtog:
## Problem

`automation-agent-stderr.log` is written via shell redirection (`2>>`) in `agent-launcher.sh` and is never rotated. Under normal operation the file stays small, but if the automation agent fails to initialise its own log (for example, due to corrupt `.slogger-state-*` files), it falls back to writing all trace and debug output to stderr. With nothing bounding the file, it can exhaust all available PVC disk space and crash the pod.

This has already caused a production incident: a customer's pod ran out of disk space, requiring a full recovery from S3 with oplog replay and manual PVC deletion.

## Solution

Add a background watchdog (`watch_and_rotate_log`) that runs alongside the agent and keeps `automation-agent-stderr.log` under a configurable size limit.

When the file exceeds the limit the watchdog trims it in-place — writing the last `limit / 2` bytes to a temp file and copying it back over the original — so the agent's open file descriptor stays valid and no data is lost from the agent's perspective.

The default limit is **50 MB**, injected by the operator as `MDB_LOG_FILE_AUTOMATION_AGENT_STDERR_MAX_SIZE_MB=50`. Users who need a tighter bound can override this via `podTemplate` env vars.

## Changes

- **`agent-launcher-lib.sh`**
  - `rotate_log_if_needed` — single-shot check: reads file size with portable `wc -c`, trims in-place via temp file + `cat` (avoids `$()` newline-stripping and preserves the inode).
  - `watch_and_rotate_log` — thin loop that calls `rotate_log_if_needed` every 30 seconds.
- **`agent-launcher.sh`** — starts `watch_and_rotate_log` in the background immediately after the agent is launched.
- **`database_construction.go`** — adds `MDB_LOG_FILE_AUTOMATION_AGENT_STDERR_MAX_SIZE_MB` constant and injects it with value `"50"` into the container env vars.
- **`agent-launcher-lib_test.sh`** — 7 bash tests covering: below/at/above threshold, content preservation, trailing newline correctness, non-existent file safety, temp file cleanup.
- **`database_construction_test.go`** — updated env var count assertions and added `stderrMaxSizeEnvVar` checks to all `TestGetAutomationLogEnvVars` subtests.
- **`changelog/`** — entry added.

## Testing

```
# Go unit tests
ok  github.com/mongodb/mongodb-kubernetes/controllers/operator/construct  (all pass)

# Shell unit tests
PASS: file below threshold is not rotated
PASS: file at threshold is not rotated
PASS: file above threshold is trimmed to keep_bytes
PASS: rotation keeps newest content
PASS: trailing newline is preserved after rotation
PASS: non-existent file does not cause an error
PASS: temp file is removed after rotation
Results: 7 passed, 0 failed
```

### DIFF
--- a/changelog/20260528_fix_automation_agent_stderr_log_rotation.md
+++ b/changelog/20260528_fix_automation_agent_stderr_log_rotation.md
@@ -1,0 +1,6 @@
+---
+kind: fix
+date: 2026-05-28
+---
+
+* Fixed an issue where `automation-agent-stderr.log` was never rotated and could grow unboundedly, exhausting all available PVC disk space. The operator now sets a default maximum size of 50 MB for this file (configurable via the `MDB_LOG_FILE_AUTOMATION_AGENT_STDERR_MAX_SIZE_MB` environment variable on the database pod).
--- a/controllers/operator/construct/database_construction.go
+++ b/controllers/operator/construct/database_construction.go
@@ -60,13 +60,14 @@ const (
 	AgentAPIKeySecretPath = "/mongodb-automation/agent-api-key" //nolint
 	AgentAPIKeyVolumeName = "agent-api-key"                     //nolint
 
-	LogFileAutomationAgentEnv        = "MDB_LOG_FILE_AUTOMATION_AGENT"
-	LogFileAutomationAgentVerboseEnv = "MDB_LOG_FILE_AUTOMATION_AGENT_VERBOSE"
-	LogFileAutomationAgentStderrEnv  = "MDB_LOG_FILE_AUTOMATION_AGENT_STDERR"
-	LogFileMongoDBAuditEnv           = "MDB_LOG_FILE_MONGODB_AUDIT"
-	LogFileMongoDBEnv                = "MDB_LOG_FILE_MONGODB"
-	LogFileAgentMonitoringEnv        = "MDB_LOG_FILE_MONITORING_AGENT"
-	LogFileAgentBackupEnv            = "MDB_LOG_FILE_BACKUP_AGENT"
+	LogFileAutomationAgentEnv               = "MDB_LOG_FILE_AUTOMATION_AGENT"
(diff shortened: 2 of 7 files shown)


nammn (commented):
Can we discuss in slack to just write to stderr/stdout and remove the whole file in the first place?

MaciejKaras:
> Can we discuss in slack to just write to stderr/stdout and remove the whole file in the first place?
@nammn you mean to pipe the logs directly to stderr? Current situation:
```shell
"${AGENT_BINARY_PATH}" "${agentOpts[@]}" "${splittedAgentFlags[@]}" 2>> "${MDB_LOG_FILE_AUTOMATION_AGENT_STDERR}" >> >(json_log "automation-agent-stdout") &
```

After the change:
```shell
"${AGENT_BINARY_PATH}" "${agentOpts[@]}" "${splittedAgentFlags[@]}" 2>> (json_log "automation-agent-stderr")  >> >(json_log "automation-agent-stdout") &
```

The difference would be that stderr logs will not be persisted on the disk, but it already happens with stdout I guess.

nammn:
Yes!  
  
Its an anti-pattern in the first place to write to a file. Commonly customers are not manually reading that file but rather use a platform (loki, datadog, om) which "scrapes" from stdout. 

Long-term we shouldn't write any logs to disks and remove all of that manual file handling. We discussed on standup that @filipcirtog will talk to @vinilage whether there are customers with **strong** **opinions** **on** **that**. Otherwise I would just remove it**.**  



I see a future where:

- we don't write to files
- we don't redirect and tail 
- we have the agent and mongod write json logs by default to stdout

filipcirtog:
closed in favour of #1321

filipcirtog:
closed in favour of PR #1321

PR #1167, opened 2026-06-01
Search: API doc & code-quality nits

Julien-Ben:
Comment/doc-only cleanup of the MongoDBSearch API types. No behavior change.

- Trim deprecated top-level field docs to redirects
- Simplify field descriptions; drop "degenerate" jargon
- Unify const declarations

### DIFF
--- a/api/mongodb/v1/search/mongodbsearch_types.go
+++ b/api/mongodb/v1/search/mongodbsearch_types.go
@@ -75,10 +75,10 @@ func (p *Prometheus) GetPort() int32 {
 }
 
 type MongoDBSearchSpec struct {
-	// Optional version of MongoDB Search component (mongot). If not set, then the operator will set the most appropriate version of MongoDB Search.
+	// Version of MongoDB Search (mongot) to run. If unset, the operator picks the most appropriate version.
 	// +optional
 	Version string `json:"version"`
-	// MongoDB database connection details from which MongoDB Search will synchronize data to build indexes.
+	// Source is the MongoDB database that MongoDB Search syncs from to build its indexes.
 	// +optional
 	Source *MongoDBSource `json:"source"`
 	// Deprecated: use spec.clusters[].replicas instead; this top-level form will be removed.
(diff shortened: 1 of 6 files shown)


api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -75,46 +75,26 @@ func (p *Prometheus) GetPort() int32 {
...
-	// Setting both spec.replicas and spec.clusters at the same time is rejected by admission.
-	// Replicas is the number of mongot pods to deploy.
-	// For ReplicaSet source: the number of mongot pods in total.
-	// For Sharded source: the number mongot pods per shard.
-	// When Replicas > 1, a load balancer configuration (spec.loadBalancer)
-	// is required to distribute traffic across mongot instances.
-	// 0 is allowed so operators (and operator-driven test harnesses) can
-	// take mongot offline cleanly via the CR — the StatefulSet is scaled
-	// to 0 by the reconciler; the MongoDBSearch CR itself stays around.
+	// Deprecated: use spec.clusters[].replicas instead; this top-level form will be removed.
```
lsierant: what's form?

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -167,27 +149,33 @@ type SyncSourceSelector struct {
...
+// the matching top-level default.
 type ClusterSpec struct {
 	// ClusterName is the Kubernetes cluster name. Required and immutable
-	// when len(spec.clusters) > 1; optional in the single-cluster degenerate case.
-	// MaxLength bounds the per-element cost contributed to the parent
-	// clusters[] uniqueness CEL rule. 253 matches the DNS subdomain limit
-	// that K8s cluster names obey.
+	// when len(spec.clusters) > 1; optional in the single-cluster case.
+	// MaxLength is 253 to match the DNS subdomain limit Kubernetes cluster
+	// names obey (and to bound the clusters[] uniqueness CEL rule's per-element cost).
```
lsierant: > and to bound the clusters[] uniqueness CEL rule's per-element cost

do we need that comment? I don't uderstand it

lsierant (approved):

api/mongodb/v1/search/mongodbsearch_types.go
Julien-Ben: In that case it is the meaning "Shape or Structure" of the word form

api/mongodb/v1/search/mongodbsearch_types.go
Julien-Ben: It was meant to say that the Common Expression Language rules, that check if clusters are unique would be guaranteed to be cheap if we validate max length.
I dropped and simplified this part.

api/mongodb/v1/search/mongodbsearch_types.go
Julien-Ben: I used "field" instead

PR #1170, opened 2026-06-01
KUBE-75: treat null in additionalMongodConfig as field removal

MaciejKaras:
> [!NOTE]
> This PR is split out from the work in #1155, which bundled several independent multi-cluster and MongoDB fixes. That work is being broken into granular, separately reviewable PRs; this one carries the `additionalMongodConfig` field-removal fix only.

# Summary

Setting a field to `null` in `additionalMongodConfig` to remove it from a deployment did not take effect — the field was either ignored or reappeared on the next reconciliation. The root cause is that `nil` map values were not treated as "remove this field":

- `maputil.traverse` now treats `nil` as absent, so `RemoveFieldsBasedOnDesiredAndPrevious` includes the field in `itemsToRemove`.
- `MergeMaps` now deletes keys whose source value is `nil` instead of forwarding `"field": null` to Ops Manager.

Previously a `null` override was dropped or re-introduced; now it reliably removes the field from the deployment.

## Proof of Work

- `pkg/util/maputil/maputil_test.go` — `TestRemoveFieldsBasedOnDesiredAndPrevious_NilValueMeansRemove`
- `pkg/util/maputil/mapmerge_test.go` — `TestMergeMaps` nil-deletes-key and nil-is-no-op-when-absent cases

## Checklist

### DIFF
--- a/pkg/util/maputil/mapmerge.go
+++ b/pkg/util/maputil/mapmerge.go
@@ -31,6 +31,11 @@ func MergeMaps(dst, src map[string]interface{}) {
 				dstMap := dst[key].(map[string]interface{})
 				MergeMaps(dstMap, t)
 			}
+		case nil:
+			// nil means the caller wants this field removed; delete it from the
+			// destination so it is not sent to OM as "verbosity": null, which
+			// would either reintroduce the field or cause type confusion.
+			delete(dst, key)
 		default:
 			{
 				dst[key] = castValue(dst[key], t)
--- a/pkg/util/maputil/mapmerge.go
+++ b/pkg/util/maputil/mapmerge.go
@@ -32,9 +32,6 @@ func MergeMaps(dst, src map[string]interface{}) {
 				MergeMaps(dstMap, t)
 			}
 		case nil:
-			// nil means the caller wants this field removed; delete it from the
-			// destination so it is not sent to OM as "verbosity": null, which
-			// would either reintroduce the field or cause type confusion.
 			delete(dst, key)
 		default:
 			{
--- a/pkg/util/maputil/maputil.go
+++ b/pkg/util/maputil/maputil.go
@@ -100,8 +100,6 @@ func traverse(currentValue interface{}, currentPath []string) []string {
 			return allPaths
 		}
 	case nil:
-		// nil means the caller wants this field removed; treat it as absent so
-		// RemoveFieldsBasedOnDesiredAndPrevious will include it in itemsToRemove.
 		return []string{}
 	default:
 		{
--- a/pkg/util/maputil/maputil.go
+++ b/pkg/util/maputil/maputil.go
@@ -99,6 +99,10 @@ func traverse(currentValue interface{}, currentPath []string) []string {
 			}
 			return allPaths
 		}
+	case nil:
+		// nil means the caller wants this field removed; treat it as absent so
+		// RemoveFieldsBasedOnDesiredAndPrevious will include it in itemsToRemove.
+		return []string{}
 	default:
 		{
 			// We found the "terminal" node in the map - need to dump the current path
--- a/changelog/20260601_fix_additional_mongod_config_null_field_removal.md
+++ b/changelog/20260601_fix_additional_mongod_config_null_field_removal.md
@@ -1,0 +1,6 @@
+---
+kind: fix
+date: 2026-06-01
+---
+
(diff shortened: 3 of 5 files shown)


pkg/util/maputil/mapmerge.go
```diff
@@ -31,6 +31,11 @@ func MergeMaps(dst, src map[string]interface{}) {
 				dstMap := dst[key].(map[string]interface{})
 				MergeMaps(dstMap, t)
 			}
+		case nil:
+			// nil means the caller wants this field removed; delete it from the
```
MaciejKaras: remove comment

pkg/util/maputil/maputil.go
```diff
@@ -99,6 +99,10 @@ func traverse(currentValue interface{}, currentPath []string) []string {
 			}
 			return allPaths
 		}
+	case nil:
+		// nil means the caller wants this field removed; treat it as absent so
```
MaciejKaras: remove comment

mircea-cosbuc (approved):

viveksinghggits (approved):
What happens if we just remove the field instead of setting it to `null`, does that have the same behavior as setting the field to null? that means it's removed from deployment?

MaciejKaras:
> What happens if we just remove the field instead of setting it to `null`, does that have the same behavior as setting the field to null? that means it's removed from deployment?

@viveksinghggits the behaviour for missing field is preserved. If the field does not exist in src map it will not be removed in the dst map. As I understand this is the expected behaviour for `MergeMaps` func

viveksinghggits:
> > What happens if we just remove the field instead of setting it to `null`, does that have the same behavior as setting the field to null? that means it's removed from deployment?
> 
> @viveksinghggits the behaviour for missing field is preserved. If the field does not exist in src map it will not be removed in the dst map. As I understand this is the expected behaviour for `MergeMaps` func

yea, I was not really worried about `MergeMaps`, I was more keep towards the user experience customers are going to have. If there was a field in CR under `additionalMongodConfig` initially and later customer removed, would that actually be removed from deployment's mognod config? I am assuming this handled but just wanted to confirm.

MaciejKaras:
> > > What happens if we just remove the field instead of setting it to `null`, does that have the same behavior as setting the field to null? that means it's removed from deployment?
> > 
> > 
> > @viveksinghggits the behaviour for missing field is preserved. If the field does not exist in src map it will not be removed in the dst map. As I understand this is the expected behaviour for `MergeMaps` func
> 
> yea, I was not really worried about `MergeMaps`, I was more keep towards the user experience customers are going to have. If there was a field in CR under `additionalMongodConfig` initially and later customer removed, would that actually be removed from deployment's mognod config? I am assuming this handled but just wanted to confirm.

@viveksinghggits I've double checked with Claude and it will work in both situations :) Thanks for asking.

```
Yes — in both removal styles, and that's exactly what this is about. The removal is driven by maputil.RemoveFieldsBasedOnDesiredAndPrevious(...) in controllers/om/process.go (mergeFrom):

maputil.MergeMaps(p, operatorProcess)
p["args2_6"] = maputil.RemoveFieldsBasedOnDesiredAndPrevious(p.Args(), specArgs26, prevArgs26)

where for a replica set it's wired as (common_controller.go):

d.MergeReplicaSet(rs, spec.GetAdditionalMongodConfig().ToMap(), lastMongodConfig, log)
//                     desired = current CR config            previous = last achieved spec

previous comes from GetLastAdditionalMongodConfigByType(r.deploymentState.LastAchievedSpec, …). RemoveFieldsBasedOnDesiredAndPrevious removes from the deployment any path that was in previous but is no longer in desired. So:

1. Customer deletes the key from the CR (drops the YAML line) → it's in previous but not desired → removed from the deployment's mongod config. This already worked before this PR (pre-existing previous-vs-desired diff).
2. Customer sets the key to null → this is the case this PR fixes. Before, traverse treated a nil-valued leaf as a present path, so it stayed in the desired flat-list and was not removed (the field lingered or got re-pushed as "field": null). Now traverse treats nil as absent and MergeMaps deletes nil-valued keys, so a null is handled identically to deleting the key — the field is removed from the deployment.

So the customer experience is consistent now: removing a previously-set field — whether by deleting the key or setting it to null — removes it from the deployment's mongod config. The e2e_sharded_cluster_mongod_options_and_log_rotation e2e test exercises this end-to-end (sets systemLog.verbosity, storage.journal.commitIntervalMs, operationProfiling.mode, then removes them via null and asserts they're gone from the mongos/configsrv/shard processes).
```

PR #1171, opened 2026-06-01
KUBE-75: correct MongoDBUser connection string secret ownership

MaciejKaras:
> [!NOTE]
> This PR is split out from the work in #1155, which bundled several independent multi-cluster and MongoDB fixes. That work is being broken into granular, separately reviewable PRs; this one carries the MongoDBUser connection-string Secret ownership fixes only.

# Summary

Two related bugs are fixed:

- The connection string `Secret` created in the central cluster did not carry a controller owner reference to the `MongoDBUser` CR, so Kubernetes garbage collection never cleaned it up when the user was deleted. The central-cluster `Secret` now gets a controller reference via `SetControllerReference`. Member-cluster `Secret`s intentionally carry **no** owner references (a cross-cluster reference would be GC'd as an orphan) and are instead deleted explicitly in `preDeletionCleanup`.
- The ownership guard on the connection string `Secret` was always satisfied regardless of the actual owner, letting the operator silently overwrite a `Secret` controlled by a different resource. The guard now compares the existing `Secret`'s controller UID against the user's UID.

## Proof of Work

- `controllers/operator/mongodbuser_controller_test.go` — `TestConnectionStringSecret_HasControllerRef_AfterReconciliation`, `TestConnectionStringSecret_NotExplicitlyDeleted_OnUserDeletion`
- e2e: `multi_cluster_scram.py`, `multi_cluster_tls_with_scram.py`, `multi_cluster_sharded_simplest.py`, `users/users_finalizer.py` — controller-ref presence on the central cluster and secret deletion on user removal

## Checklist

### DIFF
--- a/controllers/operator/mongodbuser_controller.go
+++ b/controllers/operator/mongodbuser_controller.go
@@ -18,6 +18,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	mdbv1 "github.com/mongodb/mongodb-kubernetes/api/mongodb/v1/mdb"
 	"github.com/mongodb/mongodb-kubernetes/api/mongodb/v1/mdbmulti"
@@ -276,30 +277,41 @@ func (r *MongoDBUserReconciler) updateConnectionStringSecret(ctx context.Context
 	if err != nil && !apiErrors.IsNotFound(err) {
 		return err
 	}
-	if err == nil && !secret.HasOwnerReferences(existingSecret, user.GetOwnerReferences()) {
-		return xerrors.Errorf("connection string secret %s already exists and is not managed by the operator", secretName)
+	if err == nil {
+		existingController := metav1.GetControllerOf(&existingSecret)
+		if existingController != nil && existingController.UID != user.UID {
+			return xerrors.Errorf("connection string secret %s already exists and is not managed by the operator", secretName)
+		}
 	}
 
 	mongoAuthUserURI := connectionBuilder.BuildConnectionString(user.Spec.Username, password, connectionstring.SchemeMongoDB, map[string]string{"authSource": user.Spec.Database})
 	mongoAuthUserSRVURI := connectionBuilder.BuildConnectionString(user.Spec.Username, password, connectionstring.SchemeMongoDBSRV, map[string]string{"authSource": user.Spec.Database})
 
-	connectionStringSecret := secret.Builder().
+	// Member cluster secrets must not carry ownerReferences pointing to the MongoDBUser CR
+	// in the central cluster. A cross-cluster ownerReference causes the Kubernetes garbage
+	// collector to delete the secret as an orphan. Cleanup is handled by explicitly deleting
+	// the secret from each member cluster in preDeletionCleanup.
+	memberClusterSecret := secret.Builder().
 		SetName(secretName).
 		SetNamespace(user.Namespace).
 		SetField("connectionString.standard", mongoAuthUserURI).
 		SetField("connectionString.standardSrv", mongoAuthUserSRVURI).
 		SetField("username", user.Spec.Username).
 		SetField("password", password).
-		SetOwnerReferences(user.GetOwnerReferences()).
 		Build()
 
 	for _, c := range r.memberClusterSecretClientsMap {
-		err = secret.CreateOrUpdate(ctx, c, connectionStringSecret)
+		err = secret.CreateOrUpdate(ctx, c, memberClusterSecret)
 		if err != nil {
 			return err
 		}
 	}
-	return secret.CreateOrUpdate(ctx, r.SecretClient, connectionStringSecret)
+
+	centralClusterSecret := memberClusterSecret
+	if err := controllerutil.SetControllerReference(&user, &centralClusterSecret, r.client.Scheme()); err != nil {
+		return err
+	}
+	return secret.CreateOrUpdate(ctx, r.SecretClient, centralClusterSecret)
 }
 
 func AddMongoDBUserController(ctx context.Context, mgr manager.Manager, memberClustersMap map[string]cluster.Cluster, backupEnableDelay time.Duration) error {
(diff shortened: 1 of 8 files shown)


controllers/operator/mongodbuser_controller.go
```diff
@@ -276,30 +277,41 @@ func (r *MongoDBUserReconciler) updateConnectionStringSecret(ctx context.Context
...
+		if existingController != nil && existingController.UID != user.UID {
+			return xerrors.Errorf("connection string secret %s already exists and is not managed by the operator", secretName)
+		}
 	}
 
 	mongoAuthUserURI := connectionBuilder.BuildConnectionString(user.Spec.Username, password, connectionstring.SchemeMongoDB, map[string]string{"authSource": user.Spec.Database})
 	mongoAuthUserSRVURI := connectionBuilder.BuildConnectionString(user.Spec.Username, password, connectionstring.SchemeMongoDBSRV, map[string]string{"authSource": user.Spec.Database})
 
-	connectionStringSecret := secret.Builder().
+	// Member cluster secrets must not carry ownerReferences pointing to the MongoDBUser CR
```
MaciejKaras: remove the comment

controllers/operator/mongodbuser_controller.go
```diff
@@ -511,6 +523,17 @@ func (r *MongoDBUserReconciler) preDeletionCleanup(ctx context.Context, user *us
 		return r.updateStatus(ctx, user, workflow.Failed(xerrors.Errorf("Failed to perform AutomationConfig cleanup: %w", err)), log)
 	}
 
+	// Delete the connection string Secret from every member cluster. In multi-cluster mode
```
MaciejKaras: remove comment

filipcirtog (approved):

controllers/operator/mongodbuser_controller.go
```diff
@@ -276,30 +277,37 @@ func (r *MongoDBUserReconciler) updateConnectionStringSecret(ctx context.Context
...
 		Build()
 
 	for _, c := range r.memberClusterSecretClientsMap {
-		err = secret.CreateOrUpdate(ctx, c, connectionStringSecret)
+		err = secret.CreateOrUpdate(ctx, c, memberClusterSecret)
 		if err != nil {
 			return err
 		}
 	}
-	return secret.CreateOrUpdate(ctx, r.SecretClient, connectionStringSecret)
```
josvazg, line 302: > // mongodbuser_controller.go:276
> existingSecret, err := r.client.GetSecret(ctx, types.NamespacedName{...})
> 
>   r.client is the central cluster client. The guard will prevent overwriting a central cluster Secret owned by another controller — correct. However, member cluster Secrets are written unconditionally in the loop at line 299–304 without any ownership check. If a member cluster already has a same-named Secret owned by a different resource, the operator silently overwrites it. This was the pre-existing behaviour, but it's now asymmetric (central = guarded, members = unguarded). Probably acceptable given member cluster Secrets intentionally carry no controller refs, but worth a comment.

controllers/operator/mongodbuser_controller.go
```diff
@@ -276,30 +277,37 @@ func (r *MongoDBUserReconciler) updateConnectionStringSecret(ctx context.Context
...
+
+	centralClusterSecret := memberClusterSecret
+	if err := controllerutil.SetControllerReference(&user, &centralClusterSecret, r.client.Scheme()); err != nil {
+		return err
+	}
+	return secret.CreateOrUpdate(ctx, r.SecretClient, centralClusterSecret)
```
josvazg, line 310: >  // line 306
>  centralClusterSecret := memberClusterSecret  // corev1.Secret value copy

> Build() returns corev1.Secret (value), so the struct fields are copied. However the Data map[string][]byte is a reference type — both memberClusterSecret and centralClusterSecret share the same underlying map. SetControllerReference only touches ObjectMeta.OwnerReferences, so in practice this is harmless, but it's a trap if Data were mutated between the two uses. Not a current bug.

I guess fine if it was conscious.

josvazg (approved):
LGTM with some comments

>   Missing unit test for the ownership guard rejection path
> 
>   There's a test that a controller ref is set, and a test that the central Secret isn't explicitly deleted — but no test covering the
>   new guard rejecting a secret whose controller UID doesn't match the user. If someone breaks the UID comparison it goes unnoticed.
> 
>   Flaky-prone GC wait in e2e
> 
>   wait_until(central_secret_deleted, timeout=30)  # multi_cluster_scram, tls_with_scram, users_finalizer
> 
>   30 s for Kubernetes GC is tight in loaded or slow environments. Consider bumping to 60–120 s.
> 
>   Silent error swallowing in multi_cluster_tls_with_scram.py
> 
>   def wait_for_user_deleted() -> bool:
>       try:
>           mongodb_user.load()
>           return False
>       except kubernetes.client.ApiException as e:
>           return e.status == 404  # non-404 errors → returns False, keeps polling silently
> 
>   The multi_cluster_scram.py version logs non-404 errors before returning False. The tls_with_scram version doesn't — non-404 API errors will
>   be silently retried until timeout, making failures hard to diagnose. Minor, but worth aligning.

controllers/operator/mongodbuser_controller.go
MaciejKaras, line 310: @filipcirtog How do we know this is central cluster client?

controllers/operator/mongodbuser_controller.go
```diff
@@ -276,30 +277,37 @@ func (r *MongoDBUserReconciler) updateConnectionStringSecret(ctx context.Context
...
 	for _, c := range r.memberClusterSecretClientsMap {
-		err = secret.CreateOrUpdate(ctx, c, connectionStringSecret)
+		err = secret.CreateOrUpdate(ctx, c, memberClusterSecret)
 		if err != nil {
 			return err
 		}
 	}
-	return secret.CreateOrUpdate(ctx, r.SecretClient, connectionStringSecret)
+
+	centralClusterSecret := memberClusterSecret
```
nammn, line 306: as discussed with @MaciejKaras:
- who gces that one if we don't have ownerrefs
- do we even need it in the first place?

controllers/operator/mongodbuser_controller.go
filipcirtog: Discussed offline but added here for further context. Will add a comment to provide more context.

We first declare the secret.
```
memberClusterSecret := secret.Builder().
        SetName(secretName).
        SetNamespace(user.Namespace).
        SetField("connectionString.standard", mongoAuthUserURI).
        SetField("connectionString.standardSrv", mongoAuthUserSRVURI).
        SetField("username", user.Spec.Username).
        SetField("password", password).
        Build()
```
        
We first do this for the memberClusters, in which the central one does not take part.
```
for _, c := range r.memberClusterSecretClientsMap {
        err = secret.CreateOrUpdate(ctx, c, memberClusterSecret)
        if err != nil {
            return err
        }
    }
```

We then set ownerRef on the secret and rename it for better visibility that this should be on the central one.
```
centralClusterSecret := memberClusterSecret
    if err := controllerutil.SetControllerReference(&user, &centralClusterSecret, r.client.Scheme()); err != nil {
        return err
    }
```
Then we do this, by using `r.SecretClient` which is always the central cluster client and it comes from the operator's own cluster.

filipcirtog: Thanks for raising this. In my opinion this should be ok, it should be harmless.

controllers/operator/mongodbuser_controller.go
filipcirtog: 1. The central cluster secret has an ownerReference set via `SetControllerReference`, so Kubernetes GC handles its cleanup automatically when the MongoDBUser CR is deleted.
2. Should we remove this? If so, should we do this in the scope of this PR.

Julien-Ben:
HI @MaciejKarasn, is this one good to merge ?

MaciejKaras:
> HI @MaciejKarasn, is this one good to merge ?

@Julien-Ben yes, but I wanted to wait for one more reviewer to approve, since @filipcirtog that is a second approver is actually the author. I've only created a separate PR for his work.

Julien-Ben (approved):

Julien-Ben:
@MaciejKaras 
I see, done ✅

PR #1175, opened 2026-06-01
test: support custom kind images and detect compliance

josvazg:
# Summary

The official kindest/node image for Kubernetes 1.36.1 is not yet published, so we build and host it at:

`ghcr.io/mongodb/kindest-node`

We also add `kind.images.min/max` to `kubernetes-versions.json` to customize the kind test images.

A workflow to check the policy was respected has been added.

The current Kubernetes versions were out of policy so they were bumped.

## Checklist

### DIFF
--- a/.github/workflows/check-kubernetes-versions.yml
+++ b/.github/workflows/check-kubernetes-versions.yml
@@ -1,0 +1,54 @@
+name: Check Kubernetes Versions
+
+on:
+  schedule:
+    - cron: '0 0 * * 1'  # Every Monday at 00:00 UTC
+  workflow_dispatch:
+    inputs:
+      operator_release_date:
+        description: 'Override OPERATOR_RELEASE_DATE (YYYY-MM-DD). Leave blank to use today.'
+        required: false
+        type: string
+        default: ''
+
+permissions:
+  contents: read
+
+jobs:
+  check-kubernetes-versions:
+    name: Check Kubernetes Versions
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Check Kubernetes versions
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          OPERATOR_RELEASE_DATE: ${{ inputs.operator_release_date }}
+        run: scripts/check-kube-versions.sh
+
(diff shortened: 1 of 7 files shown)


.github/workflows/check-kubernetes-versions.yml
```diff
@@ -0,0 +1,54 @@
+name: Check Kubernetes Versions
```
MaciejKaras, line 1: q: Maybe instead of only checking compliance the workflow could create a PR that bumps the respective versions?

josvazg: I already have this almost ready to integrate here and in the Atlas Operator:
https://github.com/josvazg/gobump

I hesitate to invest more in scripts and maybe do a `kubepolicy` program that manages our Kubernetes versions as that gobump would do for Go. That `gobump` does interface with git and creates PRs, kubepolicy would do the same.

MaciejKaras (approved):

kubernetes-versions.json
```diff
@@ -1,7 +1,13 @@
...
+    "min": "1.34.3",
+    "max": "1.36.1"
   },
-  "openshift": "4.21"
+  "openshift": "4.21",
+  "kind": {
+    "images": {
+      "min": "268558157000.dkr.ecr.eu-west-1.amazonaws.com/docker-hub-mirrors/kindest/node",
+      "max": "268558157000.dkr.ecr.eu-west-1.amazonaws.com/docker-hub-mirrors/kindest/node"
+    }
```
lucian-tosa, line 11: nit: Is this needed? Would we use different registries for min and max? Or would we even use a different registry from our mirror in ECR?

.github/workflows/check-kubernetes-versions.yml
lucian-tosa, line 1: I'm not against github actions, but why not use evergreen periodic jobs? We already use periodic tasks to run cleanups.

josvazg: Mimics [Atlas Operator more closely](https://github.com/mongodb/mongodb-atlas-kubernetes/actions/workflows/check-kubernetes-versions.yml), this policy is the same in both, it only changes how it is applied at the low level details.
For maintenance, I would rather use the same high level code in both.

lucian-tosa (approved):

PR #1180, opened 2026-06-02
CLOUDP-407997: Add release promote

josvazg:
# Summary

Adds `mckctl release promote` the first real release automation step.

This command copies the tested staging image to the release candidate registry, marking it as promotable for release day. This action **must** only be run after all merge tests passed.

Dry-run mode prints what would be promoted without making any changes. Full verification always runs regardless of dry-run.

## Proof of Work

`internal/ci/release/promote_test.go` covers the promote logic with unit tests. 

Manual testing:
```shell
mcktli release promote --dry-run --version # <version> prints the would-be promote commands without touching the registry.
```

## Checklist

### DIFF
--- a/internal/ci/cli/release_promote.go
+++ b/internal/ci/cli/release_promote.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
+	"log"
 
 	"github.com/spf13/cobra"
 
@@ -54,9 +55,15 @@ func newReleasePromoteCmd() *cobra.Command {
 
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "print what would happen without copying any images")
 
-	_ = cmd.MarkFlagRequired("image")
-	_ = cmd.MarkFlagRequired("commit")
-	_ = cmd.MarkFlagRequired("version")
+	MustNotErr(cmd.MarkFlagRequired("image"))
+	MustNotErr(cmd.MarkFlagRequired("commit"))
+	MustNotErr(cmd.MarkFlagRequired("version"))
 
 	return cmd
+}
+
+func MustNotErr(err error) {
+	if err != nil {
+		log.Fatalf("fatal error: %v", err)
+	}
 }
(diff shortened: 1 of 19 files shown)


MaciejKaras (approved):

internal/ci/cli/release_promote.go
```diff
@@ -0,0 +1,62 @@
...
+	cmd.Flags().StringVar(&commit, "commit", "", "commit SHA to encode in the promoted tag (required)")
+	cmd.Flags().StringVar(&version, "version", "", "version to encode in the promoted tag (required)")
+	cmd.Flags().StringVar(&registryURL, "registry", "https://quay.io", "target OCI registry base URL")
+	cmd.Flags().StringVar(&repo, "repo", "mongodb/mongodb-kubernetes-operator", "target image repository")
+
+	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "print what would happen without copying any images")
+
+	_ = cmd.MarkFlagRequired("image")
+	_ = cmd.MarkFlagRequired("commit")
+	_ = cmd.MarkFlagRequired("version")
```
m1kola: q: Why ignore error?

internal/ci/release/promote.go
```diff
@@ -0,0 +1,106 @@
...
+// gcrPromoter implements Promoter via google/go-containerregistry.
+type gcrPromoter struct {
+	host     string
+	repo     string
+	insecure bool
+}
+
+// NewOCIPromoter returns a Promoter that copies images into the given registry+repo.
+func NewOCIPromoter(baseURL, repo string) Promoter {
+	insecure := strings.HasPrefix(baseURL, "http://")
```
m1kola: q: Do we need to handle http? Aren't all of our registries use https?

m1kola (approved):
Some questions (non-blocking). Looks good overall.

internal/ci/cli/release_promote.go
josvazg: You are right to flag these. But as these are programming time errors I would use panic so we do not let it slip further. The only possible error here is that any of those flags did not exist or we had a typo on thhose names.

internal/ci/release/promote.go
josvazg: This allows us to test with local registries if needed. I would leave it as it is more flexible.

PR #1182, opened 2026-06-02
CLOUDP-403367: Add OperatorConfig CRD

m1kola:
# Based on PR #1011

## Chain of upstream PRs as of 2026-06-03

* PR #1011:
  `master` ← `feature/mc-installation-ux`

  * **PR #1182 (THIS ONE)**:
    `feature/mc-installation-ux` ← `iux-operator-config-crd`

# Summary

This pull request introduces new `OperatorConfig` CRD for configuring the operator.

The most important changes are:

**OperatorConfig CRD Definition and API Types**
- Adds a new `OperatorConfig` CRD in `api/operator/v1/operatorconfig_types.go`.
- Introduces the `groupversion_info.go` file to register the new API group (`operator.mongodb.com/v1`) and set up the scheme for the CRD.

## Proof of Work

N/A - just adds a new CRD. It doesn't affect anything yet.

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes? (**Docs tracked separately**)
- [x] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

PR #1198, opened 2026-06-05
CLOUDP-411215: Wire OperatorConfig into operator startup

m1kola:
# Summary

Infrastructure PR to begin consuming the `OperatorConfig` CRD that was added in #1182.

- `pkg/operatorconfig` — new package with `Load()` (reads the CR, falls back to defaults when absent) (CR name is `operator-config`, overridable via `MDB_OPERATOR_CONFIG_NAME` for non-standard deployments).
- `controllers/operatorconfig` — new reconciler that watches `OperatorConfig` and triggers a graceful manager shutdown when the CR changes or gets deleted, allowing the pod to restart and pick up the new config/defaults.
- `main.go` — loads `OperatorConfig` early in startup (before the manager is created) and registers the reconciler unconditionally (not gated by `--watch-resource` like other CRDs).

What's next

Individual settings will be migrated from env vars to OperatorConfig in follow-up PRs.

## Proof of Work

Tests from the base branch should still pass.

## Checklist

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
@@ -52,6 +53,7 @@ import (
 	mcoConstruct "github.com/mongodb/mongodb-kubernetes/mongodb-community-operator/controllers/construct" //nolint:depguard
 	"github.com/mongodb/mongodb-kubernetes/pkg/images"
 	"github.com/mongodb/mongodb-kubernetes/pkg/multicluster"
+	"github.com/mongodb/mongodb-kubernetes/pkg/operatorconfig"
 	"github.com/mongodb/mongodb-kubernetes/pkg/pprof"
 	"github.com/mongodb/mongodb-kubernetes/pkg/telemetry"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util"
@@ -131,6 +133,8 @@ func run() error {
 	}
 
 	ctx := signals.SetupSignalHandler()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	operator.OmUpdateChannel = make(chan event.GenericEvent)
 
 	klog.InitFlags(nil)
@@ -158,6 +162,8 @@ func run() error {
 
 	// Get a config to talk to the apiserver
 	cfg := ctrl.GetConfigOrDie()
+
+	operatorConfigName := env.ReadOrDefault(util.OperatorConfigNameEnv, util.DefaultOperatorConfigName)
 
 	log.Debugf("Setting up tracing with ID: %s, Parent ID: %s, Endpoint: %s", traceIDHex, spanIDHex, endpoint)
 	ctx, tp, err := telemetry.SetupTracingFromParent(ctx, traceIDHex, spanIDHex, endpoint)
@@ -194,6 +200,19 @@ func run() error {
 		}
 	}
 
+	// Restrict the OperatorConfig informer to the single named instance in the operator's own
+	// namespace. Without this, a change to any OperatorConfig in any watched namespace would
+	// trigger a restart.
+	managerOptions.Cache.ByObject = map[client.Object]cache.ByObject{
+		&operatorv1.OperatorConfig{}: {
+			Namespaces: map[string]cache.Config{
+				currentNamespace: {
+					FieldSelector: fields.OneTermEqualSelector("metadata.name", operatorConfigName),
+				},
+			},
+		},
+	}
+
 	if isInLocalMode() {
 		// managerOptions.MetricsBindAddress = "127.0.0.1:8180"
 		managerOptions.Metrics = metricsServer.Options{
@@ -209,7 +228,17 @@ func run() error {
 	if err != nil {
 		return err
 	}
+	// TODO(m1kola): use operatorCfg to configure controllers as settings are migrated from env vars
+	_, err = operatorconfig.Load(ctx, mgr.GetAPIReader(), currentNamespace, operatorConfigName)
+	if err != nil {
+		return fmt.Errorf("loading OperatorConfig: %w", err)
+	}
+
 	log.Info("Registering Components.")
+
+	if err := mgr.Add(operatorconfig.NewWatcher(mgr.GetCache(), cancel)); err != nil {
+		return err
+	}
 
 	// Setup Scheme for all resources
 	if err := apiv1.AddToScheme(scheme); err != nil {
--- a/pkg/operatorconfig/operatorconfig.go
+++ b/pkg/operatorconfig/operatorconfig.go
@@ -1,0 +1,39 @@
+package operatorconfig
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	operatorv1 "github.com/mongodb/mongodb-kubernetes/api/operator/v1"
+)
+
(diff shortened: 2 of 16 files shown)


main.go
```diff
@@ -159,6 +164,16 @@ func run() error {
...
 
+	bootstrapClient, err := client.New(cfg, client.Options{Scheme: scheme})
+	if err != nil {
+		return fmt.Errorf("creating bootstrap client: %w", err)
+	}
+	operatorConfigName := env.ReadOrDefault(util.OperatorConfigNameEnv, util.DefaultOperatorConfigName)
+	operatorCfg, err := operatorconfig.Load(ctx, bootstrapClient, currentNamespace, operatorConfigName)
+	if err != nil {
+		return fmt.Errorf("loading OperatorConfig: %w", err)
+	}
```
josvazg: **q**: Could this use mgr.GetAPIReader() after ctrl.NewManager instead of a separate bootstrap client? It would close the window where the cache syncs a newer RV than what was read here, and avoids creating a second client.

pkg/operatorconfig/operatorconfig.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	if err != nil {
+		return operatorv1.OperatorConfig{}, fmt.Errorf("getting OperatorConfig %q in namespace %q: %w", name, namespace, err)
+	}
+	return withDefaults(cfg), nil
+}
+
+// withDefaults fills in Go-side defaults that mirror the CRD schema markers.
+// Required when constructing a spec without the API server's defaulting webhook.
+func withDefaults(cfg operatorv1.OperatorConfig) operatorv1.OperatorConfig {
```
josvazg, line 31: q: Should we clarify further?

```suggestion
// withDefaults fills in Go-side defaults that mirror the CRD schema markers.
// Required when constructing a spec without the API server's defaulting webhook.
// Only fields consumed at startup are defaulted here; add entries as each Spec field is wired in.
func withDefaults(cfg operatorv1.OperatorConfig) operatorv1.OperatorConfig {
```

controllers/operatorconfig/reconciler.go
```diff
@@ -0,0 +1,55 @@
...
+	var cfg operatorv1.OperatorConfig
+	if err := r.client.Get(ctx, req.NamespacedName, &cfg); err != nil {
+		if apierrors.IsNotFound(err) && r.loadedResourceVersion != "" {
+			zap.S().Info("OperatorConfig deleted — initiating graceful shutdown to revert to defaults")
+			r.shutdownOnce.Do(r.cancel)
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	if cfg.ResourceVersion == r.loadedResourceVersion {
```
josvazg: **q**: This compares ResourceVersion rather than Spec, so any metadata write (label, annotation) to the CR will also trigger a restart. Is that intentional?
If only spec changes should cause a reload, comparing cfg.Spec directly would be more targeted.

main.go
```diff
@@ -211,6 +236,10 @@ func run() error {
 	}
 	log.Info("Registering Components.")
 
+	if err := operatorconfigcontroller.AddOperatorConfigController(mgr, cancel, operatorCfg.ResourceVersion); err != nil {
```
josvazg: **q**: The loaded **operatorCfg** is only used for its ResourceVersion — none of the spec fields (DefaultArchitecture, MaxConcurrentReconciles, WatchedResources, Proxy, etc.) are passed to anything that acts on them. Is consuming those values expected to land in a follow-up, or is that wiring still missing here?

m1kola: Yes, integration of settings will be done one-by-one in follow up PRs. I already started work on this (https://github.com/mongodb/mongodb-kubernetes/pull/1210 is the first one).

pkg/operatorconfig/operatorconfig.go
m1kola: I think it is probably not an accurate clarification. I think we will probably have to default everything here for the case of absent CR.

main.go
m1kola: No, we can't use `mgr.GetAPIReader()`. Because we will need some of the values from the config to pass to reconcilers. So we read before we create the manager.

I'll check if I can the things the other way around:
1. Initiate a client
2. Pass an existing client to manager.

controllers/operatorconfig/reconciler.go
roothorp: I can see how we might want this to happen (and checking resourceversion is easier than comparing specs), but I agree this is probably worth trying to avoid when we are restarting the manager on changes.

m1kola: Good point! We probably don't need to restart on metadata changes. I'll explore the following options:
1. What you suggested - comparing the spec.
2. Not reconciling when metadata changes by adding predicates. That way we should still be able to rely on a simple version comparison.

roothorp (approved):
LGTM, but worth considering whether we want to just compare resource versions. This looks like a really solid base to work from on subsequent features as we migrate.

m1kola:
@josvazg @roothorp your reviews (especially the [thread about resource version](https://github.com/mongodb/mongodb-kubernetes/pull/1198#discussion_r3381761023)) made me rethink the approach.

Could you please take another look? The idea is largely the same, but 

* Initial operator config read in `main` now uses the client from manager and there is no `bootstrapClient`.
* Instead of having a `Reconciler` watching for the `OperatorConfig` changes I now have `operatorconfig.Watcher` which is a `Runnable`. Reason: the reconciler was not reconciling anything and it was just triggering a shutdown of the process.
* The `Watcher` relies on comparing generations of the old and new objects instead of resource version. Generation only gets updated when the spec changes.

controllers/operatorconfig/reconciler.go
m1kola: Made some changes which cover this as well. See https://github.com/mongodb/mongodb-kubernetes/pull/1198#issuecomment-4671659350.

Resolving.

main.go
m1kola: Made some changes which cover this as well. See https://github.com/mongodb/mongodb-kubernetes/pull/1198#issuecomment-4671659350.

Resolving.

pkg/operatorconfig/watcher.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// Wait until this registration's initial object replay is complete before marking
+	// the watcher as synced. Until then, AddFunc calls are suppressed (initial sync
+	// replay). informer.HasSynced() is not suitable here — it reflects the informer's
+	// global state, which is already true when Start() runs.
+	if !toolscache.WaitForCacheSync(ctx.Done(), reg.HasSynced) {
+		return nil
+	}
+	w.synced.Store(true)
```
josvazg, line 53: Nice catch using `reg.HasSynced` here instead of `informer.HasSynced()`

pkg/operatorconfig/watcher.go
```diff
@@ -0,0 +1,85 @@
...
+		return err
+	}
+
+	// Wait until this registration's initial object replay is complete before marking
+	// the watcher as synced. Until then, AddFunc calls are suppressed (initial sync
+	// replay). informer.HasSynced() is not suitable here — it reflects the informer's
+	// global state, which is already true when Start() runs.
+	if !toolscache.WaitForCacheSync(ctx.Done(), reg.HasSynced) {
+		return nil
+	}
```
josvazg, line 52: **q**: If the context is cancelled before the cache syncs, this returns nil silently and the watcher exits without ever becoming active — the operator keeps running but will never react to OperatorConfig changes. Should we at least a log line so the condition is observable?

josvazg (approved):

pkg/operatorconfig/watcher.go
m1kola: > and the watcher exits without ever becoming active — the operator keeps running but will never react to OperatorConfig changes

@josvazg cancellation of context here basically means that the `Runnable` must stop. This happens when the operator is shutting down (when manager exists from `mgr.Start(ctx)`). I believe there will be no scenario when operator is still running, but runnable is not reacting to changes.

roothorp (approved):

m1kola:
Merging. Failing tests are known issues in CI configuration/api key. Not related to this PR.

PR #1199, opened 2026-06-08
HELP-81729: Enforce restricted Pod Security Standard

MaciejKaras:
## Summary

Builds on the `warn`-mode PSS work from PR #473 by switching test namespaces to `pod-security.kubernetes.io/enforce: restricted`. Where PR #473 only surfaced warnings, this PR validates that our workloads actually comply with the restricted profile — non-compliant pods are now rejected at admission.

The changes fall into several categories, each required to make enforcement viable without breaking tests.

---

### 1. Core namespace enforcement

`scripts/funcs/kubernetes` and `mongodb-community-operator/scripts/dev/e2e.py` flip the namespace label from `warn` to `enforce`. Every test namespace created by CI now rejects pods that violate the restricted profile.

---

### 2. Test infrastructure pods — PSS security context added

Several test-only pods were created without any `securityContext` and were immediately rejected under enforcement. Each needed `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]` at the container level, and `runAsNonRoot: true` + `seccompProfile: RuntimeDefault` at the pod level:

- **`mongodb-tools-pod`** (`mongodb_tools_pod.py`) — runs `mongorestore` inside the cluster for search restore tests; also needed `runAsUser: 2000` to satisfy `runAsNonRoot`
- **MCO e2e test pod** (`replica_set_mount_connection_string_test.go`) — runs a Python app to verify connection strings
- **Envoy proxy deployment** (`envoy_helpers.py`) — test-side proxy for multi-mongot search tests
- **Nginx deployment** (`remote_fixtures/nginx.yaml`) — serves MongoDB binaries locally for remotemode tests; needed `runAsUser: 1337` and `capabilities.add: [NET_BIND_SERVICE]` to bind port 80 as non-root

---

### 3. User-defined sidecar containers in test fixtures

PSS enforcement applies to every container in a pod, including user-defined sidecars. Several test YAML fixtures defined sidecar containers without `securityContext`, causing `FailedCreate` events:

- `om_ops_manager_pod_spec.yaml` — `appdb-sidecar` container
- `om_localmode-single-pv.yaml` and `om_https_enabled.yaml` — curl init containers that download MongoDB binaries; these already had `runAsUser: 1337` (Istio CNI workaround), the missing fields were `allowPrivilegeEscalation: false` and `capabilities.drop: [ALL]`
- `sharded-cluster-custom-podspec.yaml` and `standalone-custom-podspec.yaml` — busybox sidecars used for custom podspec validation

---

### 4. Third-party components in dedicated namespaces (`warn` mode)

Some test dependencies use third-party images whose runtime behaviour we cannot change. Running them under `enforce` would either reject the pods or silently break the component. These are now deployed into dedicated namespaces labelled `pod-security.kubernetes.io/warn: restricted`:

- **OpenLDAP** — moved from the test namespace to a dedicated `openldap` namespace. The `ldap_namespace` pytest fixture (module-scoped, depends on `cert_manager`) creates/ensures that namespace before any LDAP test runs. TLS certificates are issued via a dedicated cert-manager `Issuer` created directly in the `openldap` namespace so the openldap pod can mount the secret.
- **Squid proxy** — moved to a dedicated `squid` namespace. The proxy URL uses the fully-qualified `squid-service.squid.svc.cluster.local` service name so the operator can still reach it across namespaces.
- **PyKMIP server** (`kubetester/kmip.py`) — moved to a dedicated `kmip` namespace. The KMIP service, configmap, bundle secret and StatefulSet are all created in `kmip`; the TLS cert `Certificate` CR and CA configmap remain in the test namespace (where the cert-manager `Issuer` lives). The FQDN `kmip-svc.kmip.svc.cluster.local` is added as a SAN so TLS validation succeeds across namespaces. Three fixture YAMLs updated accordingly (`om_ops_manager_backup_kmip.yaml`, `replica-set-kmip.yaml`, `replica-set-kmip-static.yaml`).

---

### 5. Tests downgraded to `warn`

Several categories of tests cannot run under `enforce` and have their namespace downgraded to `warn` via conftest or in-test patching:

**Operator upgrade tests** — tests that install released operator versions (e.g. v1.27, v1.32, v1.33) predate PSS compliance. Their Helm charts do not set the required `securityContext` fields, so the operator pod would be rejected under `enforce`. A `namespace` fixture override in `tests/upgrades/conftest.py`, `tests/multicluster_appdb/conftest.py`, and `tests/olm/conftest.py` patches the label from `enforce` to `warn` before any upgrade test runs. The OLM conftest includes a TODO noting it can be removed once the 1.9.0 release ships with the required fields.

**MCO operator upgrade Go test** (`replica_set_operator_upgrade_test.go`) — the test installs the old community operator via Helm before upgrading to MCK. The old MCO Helm chart lacks PSS security context. The namespace is patched to `warn` in-test via `CoreV1Client.Namespaces().Patch()` before `InstallCommunityOperatorViaHelm`.

**No-mesh multi-cluster OM tests** (`tests/multicluster_om/conftest.py`) — these tests use `macbre/nginx-http3` as an OM interconnect between kind clusters. The image writes to root-owned paths (`/var/run/nginx`, `/var/log/nginx`) at startup and cannot run as non-root without significant fixture changes. Moving the nginx to a dedicated namespace is not practical here because OM connects to it via a short service name. The namespace is downgraded to `warn` so PSS violations are still surfaced as warnings.

---

### 6. MCO e2e test pod — writable cache directories

The MCO e2e test pod runs `go test` and `helm install` inside the container as `runAsUser: 2000`. Both tools try to write cache files to `$HOME/.cache` and `~/.helm` respectively. With no `HOME` env var, they default to `/.cache` which is not writable by a non-root user even without `readOnlyRootFilesystem`.

Added to the test pod's env:
- `GOCACHE=/tmp/go-cache`
- `HELM_CACHE_HOME`, `HELM_CONFIG_HOME`, `HELM_DATA_HOME` → `/tmp/helm-*`

---

### 7. Agent writable cache directory

`DefaultSecurityContext()` sets `readOnlyRootFilesystem: true` (pre-existing, not introduced by this PR). Under enforcement the automation agent tries to create `/var/cache/mongodb-mms-automation` at startup and fails with `read-only file system`, delaying pod readiness by ~30 s.

Added an emptyDir volume mount for `/var/cache/mongodb-mms-automation` (as a subpath of the existing `agent` emptyDir) in both:
- Enterprise database pods (`GetNonPersistentAgentVolumeMounts` in `database_construction.go`)
- AppDB pods (`appdb_construction.go`)

---

### 8. Test assertion fixes

Several unit and e2e tests asserted exact security context values or volume mount counts that changed as a consequence of the above:

- **`construction_test.go`** — volume mount counts incremented by 1 (new cache mount); expected `SecurityContext` structs updated
- **`replica_set_custom_podspec.py`** — test was asserting `securityContext` using camelCase keys, but `V1Container.to_dict()` returns snake_case; removed the `security_context` key from the expected dict (the operator injects the full context; the custom-podspec test is not the right place to verify it)
- **`podspec_template_test.go`** — `TestMergeContainer` was missing a `SecurityContext` on the default side-car container in the test setup

---

### 9. Scram auth test timeout

`test_authentication_was_disabled` polled for OM config propagation with `timeout=10s` after CR deletion. The CR deletion itself takes ~35 s (finalizer cleanup), making the 10 s window unreachable. This was pre-existing flakiness that enforcement surfaced by slightly slowing pod startups. Increased to `timeout=120s, sleep_time=5s`.

## Proof of Work

Passing CI across all variants. Remaining known non-issues: operator-upgrade tests on `warn` by design (old released operators); a handful of Docker Hub `registry:2` pull timeouts (infra flake unrelated to this PR).

## Checklist


docker/mongodb-kubernetes-tests/tests/authentication/conftest.py
```diff
@@ -29,6 +30,14 @@
 AUTOMATION_AGENT_NAME = "mms-automation-agent"
 
 
+def _ensure_ldap_namespace() -> str:
```
MaciejKaras: nit: no underscore please

MaciejKaras: also make this a fixture named ldap_namespace and use in the openldap_tls function. How are multicluster tests doing this? Please align with them

MaciejKaras: Done — converted to a `ldap_namespace` fixture (`scope="session"`) and wired it up via pytest injection into `openldap`, `openldap_tls`, `secondary_openldap`, and `openldap_cert`. The `namespace` parameter was removed from the three install fixtures since it's no longer used there.

controllers/operator/construct/database_construction.go
```diff
@@ -1110,5 +1110,8 @@ func GetNonPersistentAgentVolumeMounts(volumes []corev1.Volume, volumeMounts []c
 	volumeMounts = append(volumeMounts, statefulset.CreateVolumeMount(util.PvMms, util.PvcMmsHomeMountPath, statefulset.WithSubPath(util.PvcMmsHome)))
 
 	volumeMounts = append(volumeMounts, statefulset.CreateVolumeMount(util.PvMms, util.PvcMountPathTmp, statefulset.WithSubPath(util.PvcNameTmp)))
+
+	// Agent cache directory — must be writable even with readOnlyRootFilesystem: true
```
MaciejKaras: remove comment

controllers/operator/construct/appdb_construction.go
```diff
@@ -417,11 +417,14 @@ func AppDbStatefulSet(opsManager om.MongoDBOpsManager, podVars *env.PodEnvVars,
...
 
+	agentCacheVolume := statefulset.CreateVolumeFromEmptyDir(util.PvcAgentCache)
+	agentCacheVolumeMount := statefulset.CreateVolumeMount(agentCacheVolume.Name, util.PvcMountPathAgentCache)
+
 	keyFileNsName := appDb.GetAgentKeyfileSecretNamespacedName()
 	keyFileVolume := statefulset.CreateVolumeFromEmptyDir(keyFileNsName.Name)
 	keyFileVolumeMount := statefulset.CreateVolumeMount(keyFileVolume.Name, "/var/lib/mongodb-mms-automation/authentication", statefulset.WithReadOnly(false))
 
-	mongodbAgentVolumeMounts := []corev1.VolumeMount{agentHealthStatusVolumeMount, keyFileVolumeMount, tmpVolumeMount}
+	mongodbAgentVolumeMounts := []corev1.VolumeMount{agentHealthStatusVolumeMount, keyFileVolumeMount, tmpVolumeMount, agentCacheVolumeMount}
```
MaciejKaras: The cache directory is optional for the agent — when mkdir /var/cache/mongodb-mms-automation fails with read-only file system, the agent logs the error and continues without it. The agent doesn't need that directory to reach goal state; it's used as a scratch/cache space, not for core operation.

  So the startup does succeed — but with two consequences:

  1. ~30s extra delay per pod — the agent's retry logic runs several times before giving up on the directory, burning ~30s before continuing with the actual work (downloading MongoDB, starting mongod, etc.)
  2. Error noise in logs — repeated read-only file system log lines that obscure real errors

docker/mongodb-kubernetes-tests/kubetester/kmip.py
```diff
@@ -188,14 +201,24 @@ class KMIPDeploymentStatus:
...
         """
         Waits and assert if the KMIP server is running.
         :return: raises an error if the server is not running within the timeout.
         """
-        KubernetesTester.wait_for_condition_stateful_set(
-            self.deployment.namespace,
-            self.deployment.statefulset_name,
-            "status.current_replicas",
-            1,
+        appsv1 = client.AppsV1Api()
```
MaciejKaras: do we need here a complete rewrite of `KubernetesTester.wait_for_condition_stateful_set()`?

docker/mongodb-kubernetes-tests/tests/multicluster_appdb/conftest.py
```diff
@@ -1,12 +1,30 @@
...
 import kubernetes
+from kubetester import label_namespace
 from kubetester.awss3client import AwsS3Client
 from pytest import fixture
 from tests.common.constants import S3_BLOCKSTORE_NAME, S3_OPLOG_NAME
 from tests.opsmanager.om_ops_manager_backup import create_aws_secret, create_s3_bucket
 
 
+@fixture(scope="module")
+def namespace(namespace: str) -> str:
```
MaciejKaras, line 12: maybe think about encapsulating the logic in some convienent function in kubetester so its reusable?

docker/mongodb-kubernetes-tests/kubetester/kmip.py
MaciejKaras: No, no full rewrite needed — the function had a one-line bug: `namespace = KubernetesTester.get_namespace()` on line 811 silently overwrote the `namespace` parameter, making it always look in the test namespace regardless of what was passed. Removing that line fixes the root issue and allows `assert_is_running` to go back to using `wait_for_condition_stateful_set` as originally intended.

docker/mongodb-kubernetes-tests/tests/multicluster_appdb/conftest.py
MaciejKaras: Good point — extracted as `downgrade_pss_to_warn(namespace)` in `kubetester/__init__.py`. All four conftest files (`tests/upgrades/`, `tests/multicluster_appdb/`, `tests/multicluster_om/`, `tests/olm/`) now call it instead of repeating the `label_namespace` dict inline.

pkg/util/constants.go
```diff
@@ -116,6 +116,8 @@ const (
 	PvcMms                         = "mongodb-mms-automation"
 	PvcMmsMountPath                = "/var/lib/mongodb-mms-automation"
 	PvMms                          = "agent"
+	PvcAgentCache                  = "mms-automation-cache"
+	PvcMountPathAgentCache         = "/var/cache/mongodb-mms-automation"
```
lucian-tosa: where does this path come from? Can it be changed?

docker/mongodb-kubernetes-tests/tests/multicluster_om/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+
+@fixture(scope="module")
+def namespace(namespace: str) -> str:
+    """Downgrade PSS from enforce to warn for no-mesh multicluster OM tests.
+
+    These tests deploy a third-party nginx interconnect (macbre/nginx-http3) that
+    writes to root-owned paths at startup and cannot run as non-root without
+    significant image changes. The namespace is downgraded to warn so the nginx
+    pod can start while PSS violations are still surfaced as warnings.
+    """
```
lucian-tosa, line 13: Couldn't the nginx deployment be moved to a different namespace similarly to what you did for the ldap and kmip tests?

MaciejKaras: It's possible, I would just need to rewrite the endpoints the OM exposes. Reason from Claude:
>Moving the nginx to a dedicated namespace is not practical here because OM connects to it via a short service name. 

I don't really agree with this, but since we already have many tests that are enforced I don't think it's a big deal.

pkg/util/constants.go
MaciejKaras: The path is hardcoded in the Automation Agent binary. But it's only used when FTDC export explicitly enabled in OM settings. I will drop this change.

MaciejKaras: Done

lucian-tosa (approved):

mircea-cosbuc (approved):

Issue #1008, opened 2026-04-21
MDBC - Default agent version upgrade


TravesPigg:
This is not a bug report, but rather a question regarding the mongodb-agent version for the MongoDB Community Operator chart.

The question is, should we continue to use the default version (`108.0.2.8729-1`) in the chart, or should we pass in an override for a newer release version? Are the newer versions safe to consume for community version of the operator?

I ask because the Wiz scan against the version currently in the chart returns a number of CVE vulnerabilities.

KrisJohnstone:
its pretty horrid:
```
grype quay.io/mongodb/mongodb-agent:108.0.2.8729-1
 ✔ Vulnerability DB                [updated]
 ✔ Pulled image
 ✔ Loaded image                                                                                                                                                        quay.io/mongodb/mongodb-agent:108.0.2.8729-1
 ✔ Parsed image                                                                                                                             sha256:38d05831c69fa0eb6d3d8a0677034d6f0fbc84ca57d94670020c09d93895b179
 ✔ Cataloged contents                                                                                                                              2575cab4728114a7814ed3ab9fb09cc32bbef4d3f49087de2920c5d1cf7ae8bc
   ├── ✔ Packages                        [488 packages]
   ├── ✔ Executables                     [629 executables]
   ├── ✔ File digests                    [4,318 files]
   └── ✔ File metadata                   [4,318 locations]
 ✔ Scanned for vulnerabilities     [797 vulnerability matches]
   ├── by severity: 37 critical, 252 high, 379 medium, 129 low, 0 negligible
   └── by status:   607 fixed, 190 not-fixed, 0 ignored
[0000]  WARN current database is invalid error=the vulnerability database was built 1 week ago (max allowed age is 5 days)
NAME                           INSTALLED                FIXED IN                      TYPE       VULNERABILITY        SEVERITY  EPSS           RISK
golang.org/x/crypto            v0.25.0                  0.31.0                        go-module  GHSA-v778-237x-gjrc  Critical  30.3% (96th)   27.4
golang.org/x/crypto            v0.28.0                  0.31.0                        go-module  GHSA-v778-237x-gjrc  Critical  30.3% (96th)   27.4
gopkg.in/square/go-jose.v2     v2.6.0                                                 go-module  GHSA-c5q2-7r4c-mv6g  Medium    4.9% (89th)    2.3
[...]

PR #1205, opened 2026-06-08
KUBE-105: Update agent for MCO with pct bot changes

josvazg:
# Summary

Update MongoDB Community Operator Agent image reference and allow automation to automate it with the MCK references automatically going forward.

## Proof of Work

CI must pass

## Checklist

### DIFF
--- a/scripts/evergreen/release/update_helm_values_files.py
+++ b/scripts/evergreen/release/update_helm_values_files.py
@@ -17,6 +17,7 @@ from helm_files_handler import (
     set_value_in_yaml_file,
     update_all_helm_values_files,
     update_community_agent_image_in_file,
+    update_community_agent_image_in_go_file,
 )
 
 RELEASE_JSON_TO_HELM_KEY = {
@@ -71,6 +72,7 @@ def update_community_manifests(agent_version: str):
     update_community_agent_image_in_file(
         "mongodb-community-operator/deploy/openshift/operator_openshift.yaml", agent_version
     )
+    update_community_agent_image_in_go_file("mongodb-community-operator/test/e2e/setup/test_config.go", agent_version)
 
 
 def update_helm_charts(operator_version, release, agent_version):
--- a/scripts/evergreen/release/update_helm_values_files.py
+++ b/scripts/evergreen/release/update_helm_values_files.py
@@ -12,7 +12,12 @@ import sys
 from typing import List
 
 from agent_matrix import get_supported_version_for_image
-from helm_files_handler import get_value_in_yaml_file, set_value_in_yaml_file, update_all_helm_values_files
+from helm_files_handler import (
+    get_value_in_yaml_file,
+    set_value_in_yaml_file,
+    update_all_helm_values_files,
+    update_community_agent_image_in_file,
+)
 
 RELEASE_JSON_TO_HELM_KEY = {
     "mongodbOperator": "operator",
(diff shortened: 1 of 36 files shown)


josvazg:
Related to KUBE-105 and issue https://github.com/mongodb/mongodb-kubernetes/issues/1008

scripts/evergreen/release/update_helm_values_files.py
```diff
@@ -44,13 +49,30 @@ def main() -> int:
...
+    update_community_manifests(agent_version)
     update_cluster_service_version(operator_version)
 
     return 0
 
 
-def update_helm_charts(operator_version, release):
+def get_latest_community_agent_version(release) -> str:
+    """Returns the agent version for the highest OM major from latestOpsManagerAgentMapping."""
+    latest_mapping = release["latestOpsManagerAgentMapping"]
```
m1kola, line 64: Is it the right way to get an agent image for community? Note that we already have `filterNonReleaseOut(get_supported_version_for_image("mongodb-agent"))`. Is it not an equivalent?

m1kola (commented):
I see this wired into pre-commit, but PR also mentions PCT bot. Are you referring to the fact that PCT bot also runs pre-commit on agent bumps like [this one](https://github.com/mongodb/mongodb-kubernetes/pull/1056)? Or something else?

josvazg:
> I see this wired into pre-commit, but PR also mentions PCT bot. Are you referring to the fact that PCT bot also runs pre-commit on agent bumps like [this one](https://github.com/mongodb/mongodb-kubernetes/pull/1056)? Or something else?

Yes, the idea is to rely on precommit for PCT bot updates to propagate any extra changes needed.

scripts/evergreen/release/update_helm_values_files.py
josvazg: These are not equivalent. `get_supported_version_for_image("mongodb-agent")` returns a list of all supported agent versions from the full OM compatibility matrix (`supportedImages.mongodb-agent.opsManagerMapping`), it's used for the OpenShift relatedImages.agent list (line 89). On the other hand, 
`get_latest_community_agent_version()` returns a single version string from `latestOpsManagerAgentMapping`, which is what we need to stamp a concrete version into the community manifests and values.yaml.

scripts/evergreen/release/update_helm_values_files.py
```diff
@@ -44,13 +49,30 @@ def main() -> int:
...
+
+    update_helm_charts(operator_version, release, agent_version)
+    update_community_manifests(agent_version)
     update_cluster_service_version(operator_version)
 
     return 0
 
 
-def update_helm_charts(operator_version, release):
+def get_latest_community_agent_version(release) -> str:
```
m1kola, line 62: Nit: Re regarding name. Is it specific to community agent? Doesn't seem to be so.

scripts/evergreen/release/update_helm_values_files.py
m1kola: Ah, I just thought `get_supported_version_for_image` would also contain the latest at the end or beginning of the list.

m1kola (approved):

josvazg:
We need to merge the 8.0.24 update first #1230, then rebase this on top.

josvazg:
rebased

nammn (approved):
i think we should auto-generate `operator_openshift.yaml` via helm template in our pre-commit e.g. via a dedicated helm chart file. But I won't block here

PR #1206, opened 2026-06-08
KUBE-48, KUBE-51, KUBE-52: Move MongoDBSearch sizing fields into clusters[]

Julien-Ben:
# Summary

Until now a MongoDBSearch could set `replicas` and `resourceRequirements` at the top level of the spec *and* under `spec.clusters[]`. Having sizing live in two places was confusing and got in the way of the multi-cluster work. This PR moves those fields so they only live under `spec.clusters[]`, and makes `clusters` required. So there's now a single place to configure the deployment parameters.

This is a breaking change to the MongoDBSearch CRD. That's on purpose -> we're tidying up the API before GA. No other CRDs are touched.

Also included:
- Validations regrouped by topology so they're easier to follow.
- Docs and code snippets updated to the new shape.
- The operator-upgrade search test is temporarily disabled (KUBE-102) since it relies on the old spec shape; to be re-enabled post-GA.

> Related Jira tickets: KUBE-48, KUBE-51, KUBE-52

## Proof of Work

E2E tests pass

## Checklist


api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -760,112 +741,41 @@ func (s *MongoDBSearch) GetEndpointForShard(shardName string) string {
...
 func (s *MongoDBSearch) EffectiveClusterFor(clusterName string) (ClusterSpec, error) {
-	effective := s.EffectiveClusters()
+	clusters := s.Spec.Clusters
 	if clusterName == "" {
-		if len(effective) > 0 {
-			return effective[0], nil
+		if len(clusters) > 0 {
+			return clusters[0], nil
 		}
 		return ClusterSpec{}, fmt.Errorf("cluster %q not found in spec.clusters", clusterName)
```
viveksinghggits: nit: unrelated to this PR but maybe we should fix this.


```suggestion
		return ClusterSpec{}, fmt.Error("No clusters are configured in search resource's Spec.Clusters")
```

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -198,6 +179,15 @@ type ClusterSpec struct {
 	JVMFlags []string `json:"jvmFlags,omitempty"`
 }
 
+// ReplicasOrDefault returns the cluster's mongot replica count, defaulting to 1
```
Julien-Ben, line 183: One thing I want to flag: we default replicas to 1 when unset here.
Should we do 0 instead ?

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -780,125 +770,65 @@ func (s *MongoDBSearch) GetEndpointForShard(shardName string) string {
...
-	return 1
-}
-
-// GetReplicasForCluster returns the per-cluster mongot replica count after
-// applying the EffectiveClusters cascade (cluster-set wins, top-level is the
-// default, "1" if neither is set). clusterName="" returns the single-cluster
-// auto-promoted value (equivalent to GetReplicas).
+// GetReplicasForCluster returns the per-cluster mongot replica count: the
+// cluster's Replicas, or the default "1" when unset. clusterName="" reads the
+// first entry (single-cluster case).
```
viveksinghggits, line 802: q: should we have a validation (if we don't have already), that if there are multiple clusters none of them can have `clusterName==""`, I am recommending this because if someone is transitioning from single cluster to multi cluster, they might forget to add the clusterName to the already added entry in clusters slice.

api/mongodb/v1/search/mongodbsearch_validation.go
```diff
@@ -590,52 +616,33 @@ func validateExternalHostnameDNSLength(s *MongoDBSearch) v1.ValidationResult {
...
-// "MC at GA = Q1 or Q2 = managed LB" rule symbolically without inspecting
-// per-cluster replicas.
-func validateMCRequiresLoadBalancerManaged(s *MongoDBSearch) v1.ValidationResult {
-	if s.Spec.Clusters == nil || len(*s.Spec.Clusters) <= 1 {
-		return v1.ValidationSuccess()
+// validateMCRequiresManagedLB enforces that a multi-cluster MongoDBSearch uses
+// spec.loadBalancer.managed. Both no-LB (Q5-MC / Q6-MC) and unmanaged-LB (Q3-MC /
+// Q4-MC) topologies are rejected for multi-cluster per spec §4.4 / §B0.2:
+// multi-cluster at GA requires Envoy (managed LB). The dispatch scopes this to
+// multi-cluster (len > 1); single-cluster keeps using no-LB / unmanaged LB.
```
viveksinghggits: nit: I don't think we should use these terminologies `Q5`, `Q3`, `B0` etc here in the comments. If someone looks into this comment later, they would have not idea what these chars are.

api/mongodb/v1/search/mongodbsearch_types.go
viveksinghggits: it looks like it's already handled in validations.

viveksinghggits (approved):
Looks pretty good.

api/mongodb/v1/search/mongodbsearch_validation.go
Julien-Ben: Well spotted, thanks !
Fixed

api/mongodb/v1/search/mongodbsearch_types.go
Julien-Ben: We have `validateClustersClusterNameNonEmpty` which runs as part of `multiClusterValidators`, if there are more than 1 cluster.

api/mongodb/v1/search/mongodbsearch_types.go
Julien-Ben: Fixed, it's better like this indeed !

api/mongodb/v1/search/mongodbsearch_types.go
lsierant: 1 is fine.

docker/mongodb-kubernetes-tests/tests/search/search_sharded_internal_mongodb_multi_mongot_managed_lb.py
```diff
@@ -52,7 +52,9 @@
 
 # Resource names
 MDB_RESOURCE_NAME = "mdb-sh-managed-lb"
-MDBS_RESOURCE_NAME = MDB_RESOURCE_NAME
+# Distinct from MDB_RESOURCE_NAME so the search and sharded controllers don't share the <name>-state ConfigMap.
+# Shortened (no "-lb"): derived per-shard proxy Service names would otherwise hit the 63-char DNS label limit.
+MDBS_RESOURCE_NAME = "mdb-sh-managed-search"
```
lsierant, line 57: weren't we supposed to fix this by automatically adding -search to the search's state cm?

lsierant (approved):

docker/mongodb-kubernetes-tests/tests/search/search_sharded_internal_mongodb_multi_mongot_managed_lb.py
Julien-Ben: Yes, but it is out of the scope of this PR I think. This is a pre-existing issue.
I tracked it in our TODO list.

Also, the real problem is not really to add "-search" or not. You could have a customer who for some reason name their sharded resource "XXX-search" and their search resource "XXX". They would still have a collision.
We need to properly detect the conflict.

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -120,11 +108,6 @@ type MongoDBSearchSpec struct {
 	// spec.loadBalancer.unmanaged is top-level only — there is no per-cluster form.
 	// +optional
 	LoadBalancer *LoadBalancerConfig `json:"loadBalancer,omitempty"`
```
anandsyncs, line 110: Do you plan to handle the cluster level wiring for this?

Julien-Ben: Yes, but it will come as a separate PR, as it involves more changes than I expected.

anandsyncs (approved):

PR #1212, opened 2026-06-09
KUBE-103: Make Search autoEmbedding API-key secret optional for internal VoyageAI

helderjs:
# Summary

The embeddingModelAPIKeySecret is now optional when providerEndpoint resolves to an in-cluster VoyageAI Service (verified by operator labels); external
providers still require it. mongot mandates the key files at bootstrap, so the operator fabricates placeholder files for the internal case. Regenerated CRDs.

## Proof of Work

Unit Test were addded

## Checklist

### DIFF
--- a/api/mongodb/v1/search/mongodbsearch_types.go
+++ b/api/mongodb/v1/search/mongodbsearch_types.go
@@ -145,8 +145,12 @@ type EmbeddingConfig struct {
 	ProviderEndpoint string `json:"providerEndpoint,omitempty"`
 	// EmbeddingModelAPIKeySecret would have the name of the secret that has two keys
 	// query-key and indexing-key for embedding model's API keys.
-	// +kubebuilder:validation:Required
-	EmbeddingModelAPIKeySecret corev1.LocalObjectReference `json:"embeddingModelAPIKeySecret"`
+	// It may be omitted only when ProviderEndpoint points to the Kubernetes Service
+	// exposing the self-hosted Voyage AI embedding model managed by this operator. For any
+	// other endpoint the secret remains required; the operator validates this at
+	// reconcile time.
+	// +optional
+	EmbeddingModelAPIKeySecret corev1.LocalObjectReference `json:"embeddingModelAPIKeySecret,omitempty"`
 }
 
 type MongoDBSource struct {
--- a/controllers/searchcontroller/mongodbsearch_reconcile_helper.go
+++ b/controllers/searchcontroller/mongodbsearch_reconcile_helper.go
@@ -681,6 +681,11 @@ const (
 	appLabelKey           = "app"
 	shardLabelKey         = "shard"
 	proxyServiceComponent = "search-proxy"
+
+	nameLabelKey       = "app.kubernetes.io/name"
+	managedByLabelKey  = "app.kubernetes.io/managed-by"
+	voyageAILabelValue = "voyageai"
+	operatorLabelValue = "mongodb-kubernetes-operator"
 )
 
 // buildHeadlessService builds a headless Service for a reconcile unit. All topology-specific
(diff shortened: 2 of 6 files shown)


fealebenpae (approved):
💯

Julien-Ben (approved):
Thanks for the fix ! 🙏

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -145,8 +145,12 @@ type EmbeddingConfig struct {
 	ProviderEndpoint string `json:"providerEndpoint,omitempty"`
 	// EmbeddingModelAPIKeySecret would have the name of the secret that has two keys
 	// query-key and indexing-key for embedding model's API keys.
-	// +kubebuilder:validation:Required
-	EmbeddingModelAPIKeySecret corev1.LocalObjectReference `json:"embeddingModelAPIKeySecret"`
+	// It is required for external embedding providers. It may be omitted only when
+	// ProviderEndpoint refers to an in-cluster VoyageAI service (ai.mongodb.com),
```
viveksinghggits: Q: I am assuming when we are configuring/setting up internal embedding provider we provide the K8s service name that is used to expose that embedding provider. If that is the case what is the significance of mentioning `ai.mongodb.com` here? 
I am trying to confirm this because we use `ai.mongodb.com/v1/embeddings` in our E2E tests as embedding provider, does that mean authn would be on network layer there as well?

viveksinghggits (commented):
Looks pretty good.

api/mongodb/v1/search/mongodbsearch_types.go
helderjs: > If that is the case what is the significance of mentioning ai.mongodb.com here?

Not sure if I got the question correctly, but the upcoming VoyageAI CRD is comming under the `ai.mongodb.com` group. However, the `Service` is a native kubernetes resource. Reading it again I see that looks like there's a `Service` resource under the `ai` group, which is not the case. I'm going to rephrase the comment.

> does that mean authn would be on network layer there as well?

Yes, there's no dedicated authn/authz in the service running, as of now, we rely on the netowrk layer only.

viveksinghggits: > Not sure if I got the question correctly, but the upcoming VoyageAI CRD is comming under the ai.mongodb.com group

Aaah, I understand it now. The question is do we really have to mention the Voyage CR K8s API Group in the comments here? If we have to, we should explicitly mention that it's CR's API Group other wise it's confusing with the actual external embedding model provider https://ai.mongodb.com/v1/embeddings. 
I hope it makes sense, I can also get in a quick call to work on this.

helderjs: I have changed the field comment, I hope it's more clear now.

viveksinghggits (approved):

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -145,8 +145,12 @@ type EmbeddingConfig struct {
 	ProviderEndpoint string `json:"providerEndpoint,omitempty"`
 	// EmbeddingModelAPIKeySecret would have the name of the secret that has two keys
 	// query-key and indexing-key for embedding model's API keys.
-	// +kubebuilder:validation:Required
-	EmbeddingModelAPIKeySecret corev1.LocalObjectReference `json:"embeddingModelAPIKeySecret"`
+	// It may be omitted only when ProviderEndpoint points to the Kubernetes Service
+	// exposing the workload of a VoyageAI resource managed by this operator. For any
+	// other endpoint the secret remains required; the operator validates this at
+	// reconcile time.
```
viveksinghggits, line 151: nit:
```suggestion
	// It may be omitted only when ProviderEndpoint points to the Kubernetes Service
	// exposing the self hosted Voyage AI embedding model managed by this operator. For any
	// other endpoint the secret remains required; the operator validates this at
	// reconcile time.
```

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -843,6 +879,63 @@ func (r *MongoDBSearchReconcileHelper) ensureEmbeddingConfig(ctx context.Context
...
+	svc := &corev1.Service{}
+	if err := r.client.Get(ctx, client.ObjectKey{Name: svcName, Namespace: svcNamespace}, svc); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, xerrors.Errorf("failed to resolve embedding provider Service %s/%s: %w", svcNamespace, svcName, err)
+	}
+
+	return svc.Labels["app.kubernetes.io/name"] == "voyageai" &&
+		svc.Labels["app.kubernetes.io/managed-by"] == "mongodb-kubernetes-operator", nil
```
lucian-tosa: Should we reuse a constant instead of inline string?

lucian-tosa (approved):

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
helderjs: Done

roothorp (approved):

PR #1215, opened 2026-06-09
[KUBE-124] Handle newly added mongodb shard and mongot group using `routed_from_another_shard` header

viveksinghggits:
# Summary

When a new shard is added to a MongoDB sharded deployment, that has search already setup, the respective mongot group comes up and would not be ready to to handle the search queries for some time because of the time it takes to build indexes etc. If there is search query that lands on the newly added shard it would be forwarded to the respective nonready mongot group which would return error and result into failing the query.

To fix this, we change envoy config for newly added mongot groups (envoy clusters) in such a way that if the mongot group (`shard-b`) is NEWLY added and is not ready the filter chain would redirect the query that is directed at newly added mongot group to any of the already running mongot group (`shard-a`) with specific header (routed_from_another_shard). When `shard-a` receives that query it would know that this query was not meant for it, because of the presence of that header, and would not do index lookup or anything like that and would just return empty cursor. This would result into partial search result but the query would not fail. Once the newly added shard's (`shard-b`) mongot group becomes ready, the envoy config (filter chain) is changed again to forward the request from shard to its own mongot group.

To achieve this we are adding a field `PendingMongotGroups` in the `MongoDBSearch.status` that would have the mongot groups that are newly created and are not ready. Once search controller updates this information, envoy controller would be triggered and by reading this `PendingMongotGroups` it would know which are the shards that are just added and are not ready, so it will create the envoy config respectively, i.e. have filter to redirect to already running mongot group with routed_from_another_shard. As soon as the newly added mongot group becomes ready, the shardname/mongot group will be removed from `PendingMongotGroups` by search controller and that would trigger envoy controller again which would result into the envoy getting changed back to normal behavior.

## Proof of Work

Full search-suite Evergreen patch — unit + lint + the complete `e2e_mdb_kind_search_task_group` and `e2e_mdb_kind_search_large_task_group` on both `e2e_mdb_kind_ubi_cloudqa_large` and `e2e_static_mdb_kind_ubi_cloudqa_large` — **71/71 tasks green**: https://spruce.mongodb.com/version/6a2d7af560a7ec0008ba6416

https://github.com/mongodb/mongodb-kubernetes/pull/1216

## Checklist


controllers/operator/mongodbsearchenvoy_controller.go
```diff
@@ -356,15 +371,32 @@ func buildShardRoutes(search *searchv1.MongoDBSearch, shardNames []string, clust
 			sniHostname = endpoint
 		}
 
+		_, isReady := readySet[shardName]
+		isPending := !isReady
```
viveksinghggits:
```suggestion
		_, ok := readySet[shardName]
		isPending := !ok
```

viveksinghggits: nit:
I think having ok make it more readable because it's `_, ok` is an idiomatic way to check if a value is in a map or not. Otherwise people might get confused as to how are we just negating ready shards (`isReady`) to get pending shards.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -545,6 +558,19 @@ func (r *MongoDBSearchReconcileHelper) reconcile(ctx context.Context, log *zap.S
 
 	plan.cleanup(ctx, log)
 
+	// Latch routing-ready shards across ALL units before the worst-of readiness
```
viveksinghggits: I was wondering if `Latch` really is a good verb to explain what's happening here. Because when I read this I am not able to understand what exactly Latch means in this context.
Should we change it to something else, or is it just me?

viveksinghggits: I am not able to suggest a good alternative because it's being used at quite a few places and I am not really sure what we mean at those places.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -698,6 +763,44 @@ def scale_mongot_statefulset(
...
+) -> list[str]:
+    """Delete the PVCs backing a mongot StatefulSet so a recreated mongot starts
+    on a clean volume.
+
+    A reused volume carries the old index catalog; on restart mongot can latch its
+    one-way ``/ready`` state on stale indexes (or loop initial-sync on a dead
+    collectionUUID). Call this only when the STS is gone or scaled to 0 — otherwise
+    the STS controller immediately recreates the claim. PVCs from a
+    volumeClaimTemplate are named ``<template>-<sts>-<ordinal>``, so we match on the
+    ``-<sts>-`` infix. Returns the deleted PVC names; missing PVCs are a no-op.
```
viveksinghggits, line 780: do we also expect users to perform this if they are planning to remove/scale down their sts resource?

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
lsierant: I'll refactor it to more friendly "routing ready switch" - thanks.

lsierant (approved):

PR #1218, opened 2026-06-10
KUBE-111: search: wire syncSourceSelector.matchTagSets to mongot replicationReader.tagSets [skip-ci]

anandsyncs:
Jira: [KUBE-111](https://jira.mongodb.org/browse/KUBE-111)

## What

Wires `MongoDBSearch.spec.clusters[].syncSourceSelector` tag selection (previously an inert API surface) into the generated mongot config as `syncSource.replicationReader.tagSets`, and covers it in the existing `e2e_search_q2_mc_rs_steady` e2e — both config render and runtime sync-source selection.

The selector is reduced to a single field, `matchTagSets`:
- The pre-existing single-map `matchTags` is replaced by `matchTagSets` (an ordered list of tag sets — the general form mongot accepts; `matchTags: {region: x}` was just `matchTagSets: [{region: x}]`).
- The never-wired `hosts` field is **removed** — the sync-source host list is already provided by `spec.source` (external `hostAndPorts` for RS, the sharded source config for sharded), so a second host list on the selector was redundant.

Both fields were inert before this PR, so neither change is a behavior break.

## How it builds on ga-base

ga-base already renders the `replicationReader` block on every mongot config (KUBE-125 / #1226) with `readPreference: secondaryPreferred` and no `tagSets`. This PR is purely additive: it populates `tagSets` from the per-cluster selector. mongot rejects non-empty `tagSets` under `readPreference: "primary"`, which is why ga-base already emits `secondaryPreferred`.

## Rendering semantics

- `matchTagSets`: an ordered list of tag sets (AND across keys within a set, key-sorted for a stable config hash) — mongot syncs from members matching the first set with matches, so a trailing empty set `{}` acts as a match-any fallback under the user's control.
- No `matchTagSets` input: the base config's match-any default is left untouched (replicationReader with `secondaryPreferred`, no `tagSets`). mongot treats absent/empty `tagSets` as match-any, so this is unchanged behavior for non-tag users.

> Note on mongot semantics: `tagSets` select a sync source over mongot's **full discovered replica set**, not over a host subset. The seed/host list (from `spec.source`) is used only for topology discovery; target selection is governed entirely by `tagSets` + read preference. This is why a separate `hosts` selector field added nothing.

## Changes

- `controllers/searchcontroller`: `replicationReaderTagSetsMod`, threaded per work item (RS + sharded) through the existing `mongot.Apply` chains; no-op when the cluster has no `matchTagSets`.
- `api`: `syncSourceSelector.matchTags` (single map) replaced by `matchTagSets` (ordered list of tag sets); the unused `hosts` field and the now-moot `validateMCMatchTagsNonEmpty` / `validateClustersSyncSourceSelector` validators removed; CRDs regenerated.
- Unit tests: mod mapping (defaults, single/multi-key sorting, ordered tag sets with fallback), per-cluster plan threading.

## E2E (folded into the existing `q2_mc_rs_steady`, no new test)

- Source RS members are tagged `nodeLocation=<clusterName>` via `memberConfig[].tags`; each `MongoDBSearch` cluster sets `syncSourceSelector.matchTagSets: [{nodeLocation: <clusterName>}]` — so each cluster's mongot syncs from its local members.
- The existing per-cluster mongot ConfigMap check now also asserts `replicationReader.tagSets` matches that cluster's tag set under a non-primary read preference.
- A new runtime assertion reads each cluster's mongot pod log and verifies every `Selected initial sync host` entry points at a member of the same cluster — proving `tagSets` constrained sync-source selection at runtime, not just that the config rendered. Fails closed: zero selection lines or an unrecognized host format counts as failure.
- The match-any default path is exercised live by the search e2es that set no tags (e.g. `q3_mc_sharded_external_mtls` and the connectivity-tool tests) — their mongot configs keep ga-base's default `replicationReader` with no `tagSets`.
- `spec.version` stays unpinned in tests: the operator falls back to the pipeline's dev image (a mongot-master build whose `replicationReader` understands `tagSets`); pinning a public release tag would ImagePullBackOff since the e2e ECR repo only carries dev tags.


controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -65,6 +67,11 @@ const (
 	// is the minimum search image version that is required to enable the auto embeddings for vector search
 	minSearchImageVersionForEmbedding = "0.60.0"
 
+	// minSearchImageVersionForTagSets is the minimum search image version exposing mongot's
+	// syncSource.replicationReader.tagSets, which syncSourceSelector.matchTags maps to. Added in
+	// mongot PR #5842 (CLOUDP-390504); first released after community-search 0.64.0.
+	minSearchImageVersionForTagSets = "0.65.0"
```
anandsyncs: remove this.

anandsyncs: Done — removed the version gate entirely (constant, validateSearchVersionForTagSets, reconcile call site, and its unit tests). The CRD field doc keeps the one-line note that mongot needs >= 0.65.0, since mongot's strict syncSource parser crash-loops on the unknown replicationReader block below that.

PR #1219, opened 2026-06-10
Sharded query availability when adding new shard

lsierant:
[skip-ci] Sharded $search shard-onboarding availability characterisation.

# Summary

Adding a shard to a sharded `$search` deployment cannot preserve query availability with today's mongot+operator: mongot's `/ready` is a sticky one-way latch with no per-index readiness, so there is an unavoidable window between data landing on a new shard and that shard's mongot being queryable. This PR adds an e2e that characterises that window precisely and guarantees it surfaces only as a clean "no mongot reachable" failure (SearchNotEnabled / Envoy no-healthy-upstream / MaxTimeMSExpired stall) and **never** as a mongot index-state rejection (`while in state INITIAL_SYNC`) — the bad mode where a Ready-latched mongot serves a still-syncing range.

What's included:
- `search_sharded_routed_from_another_shard.py`: data-first onboarding e2e — migrate data onto the new shard before its mongot exists, wire `mongotHost` before creating the mongot, then assert clean recovery. A staged class walks one shard through no-mongotHost -> mongotHost-without-Envoy -> mongot-not-ready -> recovered, asserting the failure *class* at each stage.
- `connectivity.py`: failure-class taxonomy (`search_not_enabled` 31082, `transient_network` "no healthy upstream", `mongot_unreachable` MaxTimeMSExpired 50) plus the `index_unavailable` tripwire (matched by message `while in state <X>`, checked ahead of the cursor-lost heuristic so a mongod-wrapped mongot rejection can't slip past). Helpers `assert_clean_no_mongot_gap` / `assert_no_index_unavailable`; `delete_mongot_pvcs` to avoid stale-index latch on recreated mongots.
- Operator: guard a nil `deploymentState.Status` in `initializeStateStore` (a status-less deployment-state configmap caused a nil deref via the read-only reconciler helper); gated the status->CR copy on a non-empty Phase so the synthesized status isn't copied on the write path.
- Test-harness: `OMTester.clear_feature_controls()` clears the operator's `EXTERNALLY_MANAGED_LOCK` (with a 401 retry) so the per-shard `mongotHost` `automationConfig` PUT is accepted.
- A dormant (module-skipped) inverse-contract test for the future transparent-reroute operator feature, plus investigation docs under `docs/investigations/`.
- Registers `e2e_search_sharded_routed_from_another_shard` as an Evergreen task in the cloudqa-large search group.

## Proof of Work

- Full suite green locally against the EVG-host kind cluster: **22 passed, 1 skipped, 0 failed** (~34 min). `index_unavailable == 0` in every probe batch; per-stage verdicts as expected (stage1 search_not_enabled, stage2 mongot_unreachable, stage3 transient_network, stage4 recovery).
- PoW-Patch: running — https://spruce.mongodb.com/version/6a2aa99a81efd10007620c0d (unit + lint + e2e_search_sharded_routed_from_another_shard)

## Checklist


controllers/operator/mongodbsearch_controller.go
```diff
@@ -154,23 +275,24 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
...
 		}
+		s.ClusterMapping = newMapping
+		return true
+	})
+	if err != nil {
+		return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Failed(err), log)
 	}
 
-	reconcileHelper := searchcontroller.NewMongoDBSearchReconcileHelper(r.kubeClient, mdbSearch, searchSource, r.operatorSearchConfig, r.memberClusterClientsMap, state.ClusterMapping)
+	routingLatch := &searchRoutingLatch{client: r.kubeClient, search: mdbSearch, state: state}
```
lsierant: routingLatch should be internal implementation details of reconciler helper, do we actually need it outside? We could just initialize it inside. If we need to pass state, then it's better to pass it instead of routingLatch

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -400,13 +397,9 @@ type MongoDBSearchStatus struct {
...
-	// mongot group we don't have at least MinMongotReadyReplicas replias ready.
-	// The envoy reconciler uses this to build config in such a way that quries that are directed
-	// to these shards/mongot groups are routed to a healthy mongot group with routed_from_anothere_shard header.
-	// Once a group reaches the threshold (at least MinMongotReadyReplicas replicas are ready from that mongot group),
-	// the search controller removes it from this list.
+	// PendingMongotGroups is a read-only mirror derived from the routing-readiness
+	// latch in the <name>-search-state ConfigMap: sharded managed-LB shards whose
+	// mongot group has never been routing-ready. Rebuilt on every reconcile.
 	// +optional
 	PendingMongotGroups []string `json:"pendingMongotGroups,omitempty"`
```
lsierant: we don't need that field at all probably considering we have it in the state config map

controllers/operator/mongodbsearchenvoy_controller.go
```diff
@@ -144,15 +147,17 @@ func (r *MongoDBSearchEnvoyReconciler) Reconcile(ctx context.Context, request re
 		return r.updateLBStatus(ctx, mdbSearch, workflow.Pending("Waiting for search source: %s", err), log)
 	}
 
-	// Load the per-CR state to get the stable clusterName → clusterIndex mapping.
-	state, _, err := loadOrInitSearchState(ctx, r.kubeClient, mdbSearch)
+	// Load the per-CR state to get the stable clusterName → clusterIndex mapping
+	// and the routing-readiness latch.
+	state, err := loadOrInitSearchState(ctx, r.kubeClient, mdbSearch)
```
lsierant: is this save to be not locked with mutex considering it's used both in search and searchenvoy controllers?

lsierant: ok, it appears it's only used in search controller and it's not actually writing the state - ensure it.

controllers/operator/mongodbsearch_controller.go
lsierant: [AI] Done: the latch is no longer part of the helper constructor. NewMongoDBSearchReconcileHelper now takes the *SearchDeploymentState (which also replaces the separate clusterMapping param) and builds the latch internally. The state machinery (SearchDeploymentState, ReadSearchState/MutateSearchState, searchRoutingLatch) moved to controllers/searchcontroller/search_state.go; the interface is unexported and kept only as an in-package test seam.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -688,33 +735,31 @@ func (r *MongoDBSearchReconcileHelper) applyReconcileUnit(
...
-		log.Infof("Removed shard %q from PendingMongotGroups (routing-ready: %d/%d replicas ready)",
-			unit.shardName, sts.Status.ReadyReplicas, r.mdbSearch.GetMinMongotReadyReplicasForRouting())
+	minReady := r.mdbSearch.GetMinMongotReadyReplicasForRouting()
+	if sts.Status.ReadyReplicas < minReady {
+		return nil
 	}
+	if err := r.routingLatch.MarkRoutingReady(ctx, unit.shardName); err != nil {
+		return err
+	}
+	log.Infof("Marked shard %q routing-ready (%d/%d replicas ready)", unit.shardName, sts.Status.ReadyReplicas, minReady)
```
lsierant: this is confusing, we should add also what is the total number of replicas and what is the minimum replicas to be explicit. If minReady == 1, then we'll have 1/1 replicas ready, where it could be more than 1 replicas in total

controllers/searchcontroller/mongodbsearch_reconcile_helper_test.go
```diff
@@ -95,13 +96,46 @@ func newTestFakeClient(objects ...client.Object) kubernetesClient.Client {
 	return kubernetesClient.NewClient(clientBuilder.Build())
 }
 
+// fakeRoutingLatch is an in-memory RoutingReadyLatch for tests.
+type fakeRoutingLatch struct {
```
lsierant: do not use fake routing here. Reconciler should use normal routing latch impl as we're using just fake k8s clients. Verify if we need routingLatch interface in the first place. It's better to simplify as we won't have more than one impl in practice.

PR #1223, opened 2026-06-10
CLOUDP-411215: Replace MDB_MAX_CONCURRENT_RECONCILES with OperatorConfig

m1kola:
## Summary

Part of the Installation UX epic (CLOUDP-260547). `maxConcurrentReconciles`
is the first operator-level setting moved from env-var / Helm-value
configuration to the `OperatorConfig` CRD.

The `MDB_MAX_CONCURRENT_RECONCILES` environment variable and the
`operator.maxConcurrentReconciles` Helm value are removed. The operator now
reads `.spec.maxConcurrentReconciles` from the `OperatorConfig` CR at
startup. When no `OperatorConfig` CR is present the default value of `1` is
applied automatically, preserving existing behaviour for users who do not
create an `OperatorConfig` resource.

A migration guide for users upgrading from MCK 1.x will be provided
separately as part of the MCK 2.0 release.

## Proof of Work

Existing E2Es are unaffected (no test references to removed Helm value or env var)

## Checklist

      (Migration guide will be prepared separately as part of MCK 2.0 docs.)

### DIFF
--- a/changelog/20260609_breaking_maxconcurrentreconciles_moved_to_operatorconfig.md
+++ b/changelog/20260609_breaking_maxconcurrentreconciles_moved_to_operatorconfig.md
@@ -1,0 +1,6 @@
+---
+kind: breaking
+date: 2026-06-09
+---
+
+* **Operator**: The `operator.maxConcurrentReconciles` Helm value and the `MDB_MAX_CONCURRENT_RECONCILES` environment variable have been removed. Configure the maximum number of concurrent reconciliations per controller using `.spec.maxConcurrentReconciles` in the `OperatorConfig` CR instead. The default value remains `1`. Users who previously set this value must create or update their `OperatorConfig` resource before upgrading to MCK 2.0.
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -24,7 +24,7 @@ spec:
       serviceAccountName: mongodb-kubernetes-operator
       containers:
         - name: mongodb-kubernetes-operator
-          image: "quay.io/mongodb/mongodb-kubernetes:1.8.2"
+          image: "quay.io/mongodb/mongodb-kubernetes:2.0.0"
           imagePullPolicy: Always
           args:
             - -watch-resource=mongodb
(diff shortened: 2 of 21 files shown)


changelog/20260609_breaking_maxconcurrentreconciles_moved_to_operatorconfig.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: breaking
```
m1kola, line 2: I recommend reviewing commit by commit. There is a operator version bump commit which is generated changes because it is a first change with "breaking" changelog in the feature branch and it adds a lot of files.

changelog/20260609_breaking_maxconcurrentreconciles_moved_to_operatorconfig.md
```diff
@@ -0,0 +1,6 @@
...
+date: 2026-06-09
+---
+
+* **Operator**: The `operator.maxConcurrentReconciles` Helm value and the `MDB_MAX_CONCURRENT_RECONCILES` environment variable have been removed. Configure the maximum number of concurrent reconciliations per controller using `.spec.maxConcurrentReconciles` in the `OperatorConfig` CR instead. The default value remains `1`. Users who previously set this value must create or update their `OperatorConfig` resource before upgrading to MCK 2.0.
```
josvazg, line 6: **q**: do we plan to add failsafes or guardrails for this? for instance, fail the controller, or log a warning, if the is still sees the ignored MDB_MAX_CONCURRENT_RECONCILES env var being used?

josvazg (approved):
LGTM

changelog/20260609_breaking_maxconcurrentreconciles_moved_to_operatorconfig.md
m1kola: No, there is no plan like this. The plan is to make clean up of old settings to be optional. This allows an easy rollback in case it is needed.

roothorp (approved):
LGTM (always good to get rid of nolint tags too!)

PR #1225, opened 2026-06-10
KUBE-108: Fix(appdb): clear stale PVC resize status after successful resize

josvazg:
# Summary

After a PVC resize completes on an AppDB replica set, the `status.applicationDatabase.pvc` field in the MongoDBOpsManager CRD was never cleared. The AppDB controller's final success updateStatus call was missing `NewPVCsStatusOptionEmptyStatus()`, unlike the regular `MongoDBReplicaSet` and sharded cluster controllers which already handle this correctly. As a result, the PVC Resize - STS has been orphaned phase persisted indefinitely in the CRD status even after the resize was fully complete and the AppDB was healthy.

The fix adds `status.NewPVCsStatusOptionEmptyStatus()` to the final `updateStatus` call in `ReconcileAppDB`, so the PVC status field is explicitly cleared on every successful reconcile.

## Proof of Work

A regression test `TestAppDB_PVCStatusClearedAfterSuccessfulResize` walks the full resize state machine — resize detected, PVCs resized, STS orphaned and recreated — and asserts the PVC status is nil after completion.

## Checklist

### DIFF
--- a/changelog/20260610_fix_appdb_stale_pvc_resize_status_after_successful_resize.md
+++ b/changelog/20260610_fix_appdb_stale_pvc_resize_status_after_successful_resize.md
@@ -1,0 +1,6 @@
+---
+kind: fix
+date: 2026-06-10
+---
+
+* **OpsManager AppDB**: Fixed an issue where the `status.applicationDatabase.pvc` field in the `MongoDBOpsManager` CRD retained a stale `PVC Resize - STS has been orphaned` phase indefinitely after a PVC resize completed successfully. The AppDB was healthy and fully operational, but the misleading status entry required manual intervention to clear.
--- a/changelog/20260610_fix_appdb_stale_pvc_resize_status_after_successful_resize.md
+++ b/changelog/20260610_fix_appdb_stale_pvc_resize_status_after_successful_resize.md
@@ -3,4 +3,4 @@ kind: fix
 date: 2026-06-10
 ---
 
-* **OpsManager AppDB**: Fixed an issue where the `status.applicationDatabase.pvc` field in the `MongoDBOpsManager` CRD retained a stale `PVC Resize - STS has been orphaned` phase indefinitely after a PVC resize completed successfully. The AppDB was healthy and fully operational, but the misleading status entry required manual intervention to clear.
+* **OpsManager AppDB**: Fixed an issue where the `status.applicationDatabase.pvc` field in the `MongoDBOpsManager` CRD retained a stale `PVC Resize - STS has been orphaned` phase indefinitely after a PVC resize completed successfully.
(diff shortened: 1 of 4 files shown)


changelog/20260610_fix_appdb_stale_pvc_resize_status_after_successful_resize.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2026-06-10
+---
+
+* **OpsManager AppDB**: Fixed an issue where the `status.applicationDatabase.pvc` field in the `MongoDBOpsManager` CRD retained a stale `PVC Resize - STS has been orphaned` phase indefinitely after a PVC resize completed successfully. The AppDB was healthy and fully operational, but the misleading status entry required manual intervention to clear.
```
nammn: nit: isn't "The AppDB was healthy and fully operational, but the misleading status entry required manual intervention to clear." too descriptive, wdyt about removing it?

controllers/operator/appdbreplicaset_controller_test.go
```diff
@@ -1595,3 +1597,77 @@ func TestClearTLSParams(t *testing.T) {
 		})
 	}
 }
+
+// TestAppDB_PVCStatusClearedAfterSuccessfulResize is a regression test for KUBE-108.
+// After a PVC resize completes, the AppDB controller must clear the stale PVC status
+// from the CRD. Without the fix, the PhaseSTSOrphaned entry persists indefinitely.
+func TestAppDB_PVCStatusClearedAfterSuccessfulResize(t *testing.T) {
```
nammn, line 1604: Q: Do we have the same test for the other resources?

nammn (commented):
LGTM! Just minor things

nammn (approved):

controllers/operator/appdbreplicaset_controller_test.go
josvazg: Good catch. The `MultiCluster` and `ShardedCluster` controllers already had full reconciler-level tests asserting nil PVCs after a resize (`TestReconcilePVCResizeMultiCluster` and `TestReconcilePVCResizeShardedCluster`).

The `MongoDBReplicaSet` controller only had `TestHandlePVCResize`, which unit-tests `HandlePVCResize` directly without going through the full Reconcile path — so it never exercised the final updateStatus clearing.

Added `TestReplicaSetReconciler_PVCStatusClearedAfterSuccessfulResize` to close that gap. Now all four controllers have full reconciler-level coverage for PVC status cleanup.

Coming up in next update

changelog/20260610_fix_appdb_stale_pvc_resize_status_after_successful_resize.md
josvazg: done

ashortt-mdb (approved):

PR #1227, opened 2026-06-10
[skip-ci] KUBE-110: Add spec.clusters[].advancedMongotConfigs passthrough to MongoDBSearch

anandsyncs:
## Summary

Adds an optional `spec.clusters[].advancedMongotConfigs` field to the MongoDBSearch CRD: an opaque block of mongot settings rendered **verbatim under the `advancedConfigs` key** of that cluster's operator-generated mongot config file. This lets customers supply the mongot tuning knobs being exposed by the Search team's "Expose MMS Configs in Community Mongot" design (CLOUDP-403140) without waiting for first-class CRD fields.

## Design

- **Per-cluster field**: lives in `spec.clusters[]`, the single home for per-cluster settings. Each cluster's mongot ConfigMap is rendered from its own entry via `EffectiveClusterFor(...)`.
- **Opaque block under its own key**: the operator neither reads nor modifies the block — it lands in `config.yml` as a nested `advancedConfigs:` section, exactly as written in the CR. Operator-generated sections (`syncSource`, `storage`, listeners, ...) are never touched; no merge or collision semantics exist.
- Mirrors `additionalMongodConfig` mechanically: wraps `v1.MapWrapper`; CRD renders `type: object` + `x-kubernetes-preserve-unknown-fields: true`.
- No per-key allowlisting or value validation in MCK — mongot owns interpretation of the block, ignoring unknown keys.
- Applies in both single-config and per-pod (auto-embedding leader/follower) modes; config changes roll mongot pods via the existing config-hash annotation.
- When the field is absent the rendered config is byte-identical to before, so operator upgrade causes no pod restarts.

## Testing

- **Unit**: block-verbatim-under-key + operator-sections-untouched, per-cluster independence, byte-identical-when-absent, and full-reconcile. All green.
- **e2e (operator renders + mongot consumes)**: `e2e_search_community_basic` sets the block on `clusters[0]` and asserts `config.yml`'s `advancedConfigs` section equals it verbatim (`test_advanced_mongot_configs_rendered`). Verified **end-to-end** against a mongot build (`1.70.0-26-gb5c7482bdd`) that contains the consuming change ([mongot PR 6518](https://github.com/10gen/mongot/pull/6518), "Move tuning configs under `advancedConfigs`") — the search pod starts and mongot honors the block.

## Notes

Rebased onto current `search/ga-base` (incl. KUBE-125/126/127 scram-auth work and #1234 `clusterName`→`name`), squashed to a single feature commit. Diff is the 8 feature files only. Title carries `[skip-ci]`; verification is via targeted Evergreen patches (unit + the search e2e) rather than the full auto-pipeline.

JIRA: [KUBE-110](https://jira.mongodb.org/browse/KUBE-110)


api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -113,6 +113,12 @@ type MongoDBSearchSpec struct {
 	// included in mongot config and mongot uses its built-in defaults.
 	// +optional
 	FeatureFlags *FeatureFlags `json:"featureFlags,omitempty"`
+	// AdditionalMongotConfig holds extra mongot settings merged into the operator-generated
+	// mongot config file, using mongot's own YAML key structure. Settings the operator
+	// manages (e.g. syncSource, storage, server listeners) take precedence on conflict.
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +optional
+	AdditionalMongotConfig *AdditionalMongotConfig `json:"additionalMongotConfig,omitempty"`
```
anandsyncs: make this per cluster

PR #1231, opened 2026-06-11
KUBE-116: fix connection string credential encoding for spaces and plus signs

filipcirtog:
# Summary

Generated connection string secrets had broken credentials for two cases:

- **Password with a space**: `url.QueryEscape` encoded space as `+`. The Go
  MongoDB driver decodes userinfo with `url.PathUnescape` (which does not treat
  `+` as a space), so the driver sent a literal `+` to SCRAM instead of a space.
  See: https://github.com/mongodb/mongo-go-driver/blob/master/x/mongo/driver/connstring/connstring.go#L934
- **Password with a `+`**: a raw `+` in userinfo is decoded as a space by
  pymongo's `unquote_plus`, corrupting the password.
  See: https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/uri_parser_shared.py#L146

The fix uses `url.QueryEscape` as the base encoder (correctly encodes `+` as
`%2B`) and replaces the space encoding from `+` to `%20`, which both the Go
driver and pymongo decode correctly. Extracted into
`stringutil.EncodeUserinfoComponent` and applied to both
`connectionstring.Builder` and `authtypes.User.GetLoginString`.

## Proof of Work

Unit tests covering space→`%20`, `+`→`%2B`, structural separators, and no-auth
cases. Community operator tests asserting correct encoding in both standard and
SRV URIs. e2e tests with a space-password user and a `+`-password user that
assert encoding and verify actual connectivity via pymongo.

## Checklist


pkg/util/stringutil/stringutil.go
```diff
@@ -84,6 +87,15 @@ func Remove(slice []string, s string) (result []string) {
...
 }
 
+// EncodeUserinfoComponent percent-encodes a username or password for use in a
+// MongoDB connection string. Space is encoded as %20 (not +) so that pymongo's
+// unquote_plus and the Go driver both decode it correctly, and + is encoded as
+// %2B so pymongo does not decode it as a space.
+// https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/uri_parser_shared.py#L146
+func EncodeUserinfoComponent(s string) string {
+	return strings.ReplaceAll(url.QueryEscape(s), "+", "%20")
+}
```
josvazg, line 97: nit: would not mind a unit test for this function at `stringutil_test.go`, e.g:

```go
  func TestEncodeUserinfoComponent(t *testing.T) {
      cases := []struct{ in, want string }{
          {"pass word", "pass%20word"},
          {"pass+word", "pass%2Bword"},
          {"p@ss", "p%40ss"},
          {"p%25x", "p%2525x"},  // literal % is double-encoded
          {"my:p@ss/w?rd# %[+]!$&'()*,;=~-._", "my%3Ap%40ss%2Fw%3Frd%23+%25%5B%2B%5D%21%24%26%27%28%29%2A%2C%3B%3D~-._"},
          {"plain", "plain"},
      }
      for _, c := range cases {
          assert.Equal(t, c.want, EncodeUserinfoComponent(c.in), c.in)
      }
  }
```

pkg/util/stringutil/stringutil.go
```diff
@@ -84,6 +87,15 @@ func Remove(slice []string, s string) (result []string) {
 	return result
 }
 
+// EncodeUserinfoComponent percent-encodes a username or password for use in a
+// MongoDB connection string. Space is encoded as %20 (not +) so that pymongo's
+// unquote_plus and the Go driver both decode it correctly, and + is encoded as
+// %2B so pymongo does not decode it as a space.
+// https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/uri_parser_shared.py#L146
```
josvazg, line 94: q: should we mention why `url.PathEscape` was not the right choice here because it doesn't encode `:` in
  usernames?

pkg/authentication/authtypes/authtypes.go
```diff
@@ -85,9 +83,7 @@ type User struct {
 
 func (u User) GetLoginString(password string) string {
 	if u.Database != constants.ExternalDB {
-		return fmt.Sprintf("%s:%s@",
-			url.QueryEscape(u.Username),
-			url.QueryEscape(password))
+		return stringutil.EncodeUserinfoComponent(u.Username) + ":" + stringutil.EncodeUserinfoComponent(password) + "@"
 	}
 	return ""
 }
```
josvazg, line 89: nit: would be great another unit test on this one.

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_256_connectivity.py
```diff
@@ -28,6 +28,14 @@
 NON_ADMIN_USER_PASSWORD = "my-password-2"
 NON_ADMIN_USER_DATABASE = "testdb"
 
+SPACE_PASSWORD_USER_NAME = "mms-user-3"
+SPACE_PASSWORD_SECRET_NAME = "mms-user-3-password"
+SPACE_PASSWORD_USER_PASSWORD = "my pass word"
+
+PLUS_PASSWORD_USER_NAME = "mms-user-4"
+PLUS_PASSWORD_SECRET_NAME = "mms-user-4-password"
+PLUS_PASSWORD_USER_PASSWORD = "my:p@ss/w?rd# %[+]!$&'()*,;=~-._"
```
josvazg, line 37: nit: This could be covered in a unit test more cheaply.

josvazg (approved):
LGTM, although we could use some more unit tests

pkg/authentication/authtypes/authtypes.go
filipcirtog: updated! thank you!

pkg/util/stringutil/stringutil.go
filipcirtog: updated! thank you!

pkg/util/stringutil/stringutil.go
filipcirtog: Documenting why each alternative wasn't picked would be noisy we could equally call out url.UserPassword, url.PathEscape, or others, but each of them has drawbacks since none of these are identical with what the mongodb drivers expect.

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_256_connectivity.py
filipcirtog: thanks! added in the unit test as well. in here the actual connection is tested so that we can make sure that the mongodb driver decodes it well. this is where it would have failed before. since we already spin up the deployment, it's not very costly to test this as well.

fealebenpae (approved):

fealebenpae:
@filipcirtog I think you should revert the switch_project test changes here, though - aggressive timeouts like that might obfuscate sources of flakiness that we'd want to address, and it's generally a good idea to make changes to e2e tests in separate PRs so we can more easily assess their impact. I don't think the rest of your PR touches anything that would require the switch_project changes.

PR #1233, opened 2026-06-11
KUBE-117: Skip broken tests at PR, still run at merge time

josvazg:
# Summary

This change will avoid the following broken tests and all its variants at PR time:
- `e2e_search_community_auto_embedding`
- `e2e_search_community_auto_embedding_multi_mongot`

Such **tests still run at merge time in master**, so releases will not happen until those are fixed, because the master branch is still broken.

Additionally, slack notifications:
- Alert us if more unexpected errors creeped through to master.
- OR when we should remove this skipping code as the broken tests recover.

The aim of this PR is to safely unblock development work without allowing broken releases not letting people get "use to red pipelines".

## Proof of Work

CI must pass clean at PR time, fail at merge time with only the expected failures.

## Checklist

### DIFF
--- a/.evergreen-functions.yml
+++ b/.evergreen-functions.yml
@@ -958,3 +958,17 @@ functions:
           - SLACK_WEBHOOK_URL
           - version_id
         binary: scripts/dev/run_python.sh scripts/evergreen/notify_master_failures.py
+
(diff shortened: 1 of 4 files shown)


josvazg:
Seems master is fixed, we do not need this anymore

PR #1234, opened 2026-06-11
MongoDBSearch: rename spec.clusters[].clusterName to name [skip-ci]

Julien-Ben:
> **Rebased & taken over — updated 2026-06-22.** The original author is out; I took over this stack. Both #1224 (shardOverrides) and #1238 (per-cluster loadBalancer) have now **merged into `search/ga-base`**, so this is the final PR of the stack and targets `search/ga-base` directly. It carries the `clusterName`→`name` rename plus the simulated-MC e2e fixture migration to per-cluster `loadBalancer`. Content is unchanged from the previously-green full multi-cluster + load-balancer e2e validation (byte-identical tree, just reparented). **The `name` rename is breaking post-GA — please confirm the `name` vs `clusterName` decision before merge.**

**Update (rebased onto rewritten ga-base):** `search/ga-base` was force-rewritten (per-cluster LB + shardOverrides re-landed as a squashed "KUBE-120: finalize API" + the #1262 CI fix). This PR was rebased onto the new ga-base; the rename commits applied clean, and the sim-MC fixture migration was reconciled with #1262 (kept only the non-redundant deltas — the `name` rename + the per-cluster callable). Rename verified complete (Go field + JSON tag + CRD all `name`); build + search unit suites green; zero CRD drift.

**Update (review follow-up):** addressed @viveksinghggits's nits — the Go field `ClusterSpec.ClusterName` is renamed to `Name` too (not just the JSON tag), and its doc comment fixed; the generated CRD is unchanged (JSON tag was already `name`). Also fixed two pre-existing lint failures surfaced on CI (a gofmt reformat and two Python `ty` `int | None` call-site narrowings) — neither related to the rename. Unit-covered; no e2e behavior change.

# Summary

Renames the per-cluster identifier field `spec.clusters[].clusterName` → `spec.clusters[].name`. This is the user-facing key on each `clusters[]` entry; the rename flows through the Go API types, the generated CRD/CEL validation, and all unit + e2e tests. No behavior change — purely the API surface.

Note: this makes MongoDBSearch the only CRD using `name` for the per-cluster key (AppDB / sharded / OM use `clusterName`). This is intentional for the MongoDBSearch `spec.clusters[]` API but is breaking post-GA, so it should be confirmed before merge.

## Proof of Work

- Unit suites for `api/mongodb/v1/search`, `controllers/searchcontroller`, and `controllers/operator` pass; `make manifests` produces zero CRD drift.
- The full multi-cluster + LB e2e suite (q2/q3 MC, simulated-MC RS + sharded, managed/unmanaged per-cluster-LB) is green on the rebased stack.

## Checklist

### DIFF
--- a/api/mongodb/v1/search/mongodbsearch_types.go
+++ b/api/mongodb/v1/search/mongodbsearch_types.go
@@ -136,7 +136,7 @@ type ClusterSpec struct {
 	// MaxLength is 253 — the DNS subdomain limit Kubernetes cluster names follow.
 	// +optional
 	// +kubebuilder:validation:MaxLength=253
-	ClusterName string `json:"name,omitempty"`
+	Name string `json:"name,omitempty"`
 	// ClusterIndex is the stable integer in per-cluster resource names. Required on every entry
 	// of a multi-cluster spec, and even on a single entry when each member cluster runs its own
 	// operator. Changing it renames this cluster's resources, orphaning those at the old index.
@@ -880,7 +880,7 @@ func (s *MongoDBSearch) EffectiveClusterFor(clusterName string) (ClusterSpec, er
 		return ClusterSpec{}, errors.New("no clusters are configured in spec.clusters")
 	}
 	for _, c := range clusters {
-		if c.ClusterName == clusterName {
+		if c.Name == clusterName {
 			return c, nil
 		}
 	}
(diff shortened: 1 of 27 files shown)


api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -126,17 +126,17 @@ type SyncSourceSelector struct {
...
 // per-field default applies.
 type ClusterSpec struct {
-	// ClusterName is the Kubernetes cluster name. Required and immutable
+	// Name is the Kubernetes cluster name. Required and immutable
 	// when len(spec.clusters) > 1; optional in the single-cluster case.
 	// MaxLength is 253 — the DNS subdomain limit Kubernetes cluster names follow.
 	// +optional
 	// +kubebuilder:validation:MaxLength=253
-	ClusterName string `json:"clusterName,omitempty"`
+	ClusterName string `json:"name,omitempty"`
```
viveksinghggits: nit: Should we rename `ClusterName` to `Name` too? I think we should do that.

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -126,17 +126,17 @@ type SyncSourceSelector struct {
...
 
-// ClusterSpec is one entry in spec.clusters[]. ClusterName is required and immutable
-// when len(spec.clusters) > 1; optional in the single-cluster case.
+// ClusterSpec is one entry in spec.clusters[]. The cluster name (spec.clusters[].name)
+// is required and immutable when len(spec.clusters) > 1; optional in the single-cluster case.
 // Each field, when set, applies to this cluster; when unset, the operator's
 // per-field default applies.
 type ClusterSpec struct {
-	// ClusterName is the Kubernetes cluster name. Required and immutable
+	// Name is the Kubernetes cluster name. Required and immutable
```
viveksinghggits, line 134: Typo in current form, it should be `ClusterName`.

viveksinghggits (approved):

api/mongodb/v1/search/mongodbsearch_types.go
anandsyncs: Done in 6f3238cb2 — renamed the Go field `ClusterSpec.ClusterName` → `Name` (kept the `json:"name,omitempty"` tag) and updated all references across the search package + tests. Generated CRD is unchanged since the JSON tag was already `name`.

api/mongodb/v1/search/mongodbsearch_types.go
anandsyncs: Fixed in 6f3238cb2 — with the field renamed to `Name`, the doc comment now correctly reads `// Name is the Kubernetes cluster name…` (matches the identifier).

anandsyncs (approved):

PR #1235, opened 2026-06-11
CLOUDP-411215: Replace MDB_DEFAULT_ARCHITECTURE with OperatorConfig.defaultArchitecture

m1kola:
# Based on PR #1011

* PR #1011:
  `master` ← `feature/mc-installation-ux`

  * **PR #1235 (THIS ONE)**:
    `feature/mc-installation-ux` ← `iux-operatorconfig-default-architecture`

# Summary

Part of the Installation UX epic (CLOUDP-260547). Replaces the `MDB_DEFAULT_ARCHITECTURE` environment variable and `operator.mdbDefaultArchitecture` Helm value with `.spec.defaultArchitecture` in the `OperatorConfig` CR, consistent with how `maxConcurrentReconciles` was migrated previously.

- `main.go` reads `defaultArchitecture` from `OperatorConfig` instead of the env var
- Helm chart, public manifests, and `config/manager/manager.yaml` no longer set `MDB_DEFAULT_ARCHITECTURE`
- E2E tests create an `OperatorConfig` CR (via a new `create_operator_config` helper) after operator install when running in static architecture mode, then wait for the operator to restart
- OLM tests remove `MDB_DEFAULT_ARCHITECTURE` from MCK subscription env vars; MEKO subscription retains it since MEKO 1.x still reads that env var

## Proof of Work

E2E tests passing on Evergreen.

## Checklist


m1kola:
We have `e2e_sharded_cluster_scram_sha_1_switch_project` and `e2e_sharded_cluster_scram_sha_256_switch_project` failing, but this is is unrelated to the change.

Everything else seems green. Marking as ready for review.

changelog/20260611_breaking_defaultarchitecture_moved_to_operatorconfig.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: breaking
+date: 2026-06-11
+---
+
+* **Operator**: The `operator.mdbDefaultArchitecture` Helm value and the `MDB_DEFAULT_ARCHITECTURE` environment variable have been removed. Configure the default container architecture using `.spec.defaultArchitecture` in the `OperatorConfig` CR instead. The default value remains `NonStatic`. Users who previously set this value must create or update their `OperatorConfig` resource before upgrading to MCK 2.0.
```
josvazg, line 6: I wonder if we should log a warning or even fail the operator to restart if the user is still setting the env var, and maybe expecting it to work.

josvazg:
**q** : For the Watcher in `pkg/operatorconfig/watcher.go` should we add a comment explaining why restart is preferred over hot-reload?

Not a concern with this change.

docker/mongodb-kubernetes-tests/kubetester/helm.py
```diff
@@ -256,6 +259,41 @@ def apply_crds_from_chart(crds_dir: str):
...
+
+
+def apply_operator_config_crd(api_client=None):
+    """Applies the OperatorConfig CRD and waits for it to be established.
+
+    Used by upgrade tests to ensure the CRD exists before the OperatorConfig CR is created and the
+    operator is upgraded to a version that reads it (legacy operators do not ship this CRD).
+    """
+    with open(OPERATOR_CONFIG_CRD_FILE) as f:
+        crd_body = yaml.safe_load(f)
```
josvazg, line 275: **q**: Should we add 409 support here?

This calls api.create_custom_resource_definition(body=crd_body) without catching a 409. If a test re-runs against an existing cluster where the CRD already exists, it will throw. Compare with create_operator_config() in kubetester.py which correctly catches 409.

docker/mongodb-kubernetes-tests/kubetester/kubetester.py
```diff
@@ -92,6 +92,138 @@ def get_default_architecture() -> str:
...
+    if len(pods) != 1 or not pods[0].status.container_statuses:
+        return None
+    # Select the operator container by name rather than assuming it is the first one. In
+    # multi-cluster deployments the pod also runs an injected istio-proxy sidecar, so the
+    # operator container is not necessarily at index 0 of container_statuses. The operator
+    # container shares its name with the operator deployment (Helm sets both from operator.name).
+    container_statuses = {cs.name: cs for cs in pods[0].status.container_statuses}
+    operator_status = container_statuses.get(name)
+    if operator_status is None:
+        return None
```
josvazg: Returns None if len(pods) != 1, which causes `wait_for_operator_pod_restart` to silently skip waiting.

During a rolling update there could briefly be 2 pods, causing the wait to no-op and the test to proceed before the operator has actually restarted. Right?

docker/mongodb-kubernetes-tests/tests/conftest.py
```diff
@@ -995,7 +1019,6 @@ def install_official_operator(
     helm_args = {
         "registry.imagePullSecrets": operator_installation_config["registry.imagePullSecrets"],
         "managedSecurityContext": managed_security_context,
-        "operator.mdbDefaultArchitecture": operator_installation_config["operator.mdbDefaultArchitecture"],
     }
```
josvazg, line 1024: If any caller still passes this key in operator_installation_config, it will be silently ignored since it's no longer picked up here. Worth checking that all callers/test configs have been cleaned up too.

josvazg (approved):
LGTM, with a few questions and minor checks

changelog/20260611_breaking_defaultarchitecture_moved_to_operatorconfig.md
m1kola: @josvazg we deliberately do not fail the reconciler/operator if old env vars are in use because cleanup is optional (to enable easy rollback to a previous version). But I think increasing visibility in the form of warning makes sense.

I created [CLOUDP-417579](https://jira.mongodb.org/browse/CLOUDP-417579) for that.

josvazg: Makes sense

docker/mongodb-kubernetes-tests/kubetester/kubetester.py
```diff
@@ -92,6 +92,138 @@ def get_default_architecture() -> str:
...
+        "OperatorConfig",
+        "operatorconfigs",
+        "operator.mongodb.com",
+        "v1",
+        api_client=api_client,
+    )
+    resource["spec"] = spec
+    # CustomObject always initialises with status:{}, but OperatorConfig has no status
+    # subresource and create() uses field_validation="Strict", so remove it first.
+    resource.backing_obj.pop("status", None)
```
m1kola: Oh, this is not longer needed. Fixed in master in https://github.com/mongodb/mongodb-kubernetes/pull/1259 and our branch already contains the fix.

docker/mongodb-kubernetes-tests/tests/conftest.py
m1kola: I believe I checked all the callers and we should be good. Also we will add a warning as discussed here https://github.com/mongodb/mongodb-kubernetes/pull/1235#discussion_r3474166680

docker/mongodb-kubernetes-tests/kubetester/helm.py
m1kola: I initially deliberately didn't want to handle 409 when applying CRDs because in the E2E scenarios we know when to apply it (when they don't exist). But after your comment I now reconsidering this approach. I realised that when running/debugging locally it would be great to handle 409 and apply the update.

I'll update here.

Also `create_operator_config` currently just ignores 409. I will look into updating it so that we apply changes to the config.

docker/mongodb-kubernetes-tests/kubetester/kubetester.py
m1kola: Ok, I think the issue here was in the first call to `get_operator_pod_restart_count` where we take a snapshot before waiting for a restart. If it is `None` - we skip in `wait_for_operator_pod_restart` and end up with a potential flake (unlikely buy still).

I tried waiting in `get_operator_pod_restart_count` to get exactly one pod, but I think it complicates things too much without a lot of benefits.

Instead I decided to go with the following: `get_operator_pod_restart_count` can still return `None` when there are no pods or when number of pods is not equal 1. But we fail the whole thing if the initial snapshot is `None`.

The likelihood of this should be very low due to how we run our tests. So I think this simple solution should be enough.

Note that the fact that `get_operator_pod_restart_count` could return `None` should not be an issue in `wait_for_operator_pod_restart` since we wait for it to be more than `previous_restart_count` which is guaranteed to not be `None`.

mircea-cosbuc (approved):

PR #1239, opened 2026-06-11
KUBE-134: clusters[].index is a pure-CRD pin (remove the persisted index mapping)

anandsyncs:
# Summary

`MongoDBSearch.spec.clusters[i].clusterIndex` is now the **single source of truth** for a cluster's resource index. The pin is **required and unique** for multi-cluster (simulated and real), and **optional → 0** for single-cluster. The previously-persisted `ClusterMapping` — and everything that maintained or read it — is removed.

This supersedes the earlier "harden the persisted mapping" approach: the index (`spec.clusters[].index`) is a pure function of the CR, so there is nothing to persist, reconcile, or evict.

## What changed
- **Removed** `AssignClusterIndices` / `cluster_index.go`, the state-ConfigMap `ClusterMapping` field + its writer, and every mapping read.
- **Readers/writers read the pin** directly: `ResolveSingleClusterIndex(search)` = pinned index else 0; `buildRSWorkList` / `buildClusterWorkList` / `CheckSecretsPresence` source `spec.clusters[i].index` via a single `ClusterSpec.ResolveIndex()` accessor (pin else 0).
- **Eliminated the first-reconcile `-1` sentinel** (Envoy reconcile Pending, cleanup skip, per-shard TLS gate) — the pin is present on the CR from creation, so the "wait for the mapping to be written" race cannot occur.
- **Uniqueness** is guaranteed by `validateClustersClusterIndexRequired` (Go) + the CRD CEL rules.

## What is preserved
- The search **state ConfigMap** stays — it still carries KUBE-124's `routingReadyMongotGroups` (read by the Envoy reconciler). Only the `ClusterMapping` field is dropped.

## Size / validation
- Net **-298 LOC** (582 insertions / 880 deletions — mostly deleted mapping machinery + its tests).
- Local gates green: `go build ./...`, `go vet`, `gofmt`, and unit tests for `api/mongodb/v1/search`, `controllers/searchcontroller`, `controllers/operator`, `mongodb-community-operator/controllers`.
- e2e validated (rebased onto ga-base `6a8285bf1`): the gating `e2e_search_simulated_mc_rs`, `e2e_search_simulated_mc_sharded`, and `e2e_search_q2_mc_rs_steady` tasks pass on both `e2e_multi_cluster_kind`/`e2e_multi_cluster_2_clusters` and their static variants.

## Tickets
KUBE-134


anandsyncs:
Validation complete: Evergreen patch [`6a2b1f2b`](https://evergreen.mongodb.com/version/6a2b1f2b3e46790007992c60) — **version success, all tasks green on first execution**: `e2e_search_simulated_mc_rs` + `e2e_search_simulated_mc_sharded` on both `e2e_multi_cluster_kind` and `e2e_static_multi_cluster_kind`, `unit_tests_golang`, `lint_repo` (CRD doc-line regen drift-free).

PR #1240, opened 2026-06-11
[KUBE-127] Add support for mTLS with enc keyfile for scram auth

viveksinghggits:
# Summary

When mongot connects to its sync source over TLS while authenticating with SCRAM (username/password), it used to connect as a plain TLS client which means it didn't present a client certificate during the handshake. This is fine when the source mongod accepts any TLS client, but it breaks when the source is configured to require mutual TLS (i.e. the mongod demands that every connecting client present a certificate signed by a trusted CA). In that setup mongot's SCRAM connection is rejected at the TLS layer, before SCRAM authentication even gets a chance to run.

It is important to understand that authentication and transport are two separate concerns here. SCRAM (username+password) is still the authentication mechanism, but mTLS is a transport-level requirement (mongot may  present a client cert just to establish the connection). So we need a way for a SCRAM-authenticated mongot to also present a client certificate during the TLS handshake.

This PR adds that. Users can now point mongot at a client certificate for the SCRAM connection via a new field on the source spec:

```
  spec:
    source:
      username: <scram-user>
      passwordSecretRef:
        name: <password-secret>
      tls:
        clientCertificateSecretRef:
          name: <client-cert-secret>   # tls.crt, tls.key (+ optional tls.keyFilePassword)
```

When `spec.source.tls.clientCertificateSecretRef` is set, the operator wires the certificate into mongot's `scramAuth.tls` block so mongot presents it during the handshake while still authenticating via SCRAM:

```
  syncSource:
    replicaSet:
      scramAuth:
        username: <scram-user>
        passwordFile: <path>
        tls:
          enabled: true
          caFile: /var/lib/tls/ca/ca-pem
          tlsCertificateKeyFile: /var/lib/tls/scram-client/<hash>.pem
          tlsCertificateKeyFilePasswordFile: /tmp/scram-key-password   # only if encrypted
```

Two more things to note about the behavior:

1. Just like the gRPC keyfile change in #1237, the client certificate's private key can be password-encrypted. If the user adds a `tls.keyFilePassword` key to the client cert secret, the operator mounts it and sets `tlsCertificateKeyFilePasswordFile` so mongot can decrypt the key.
2. For sharded clusters the same client cert and CA are also applied to the router (mongos) connection, so mongot presents the certificate consistently whether it is talking to a shard's replica set or to mongos.

## Proof of Work

E2E is part of the [PR](https://github.com/mongodb/mongodb-kubernetes/pull/1248). 

## Checklist


api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -293,6 +293,11 @@ type MongoDBSource struct {
 	// Mutually exclusive with PasswordSecretRef and Username.
 	// +optional
 	X509 *X509Auth `json:"x509,omitempty"`
+	// TLS configures TLS transport settings for the sync source SCRAM connection.
```
lsierant, line 341: doesn't using mTLS mean x509 and if not then it's scram? I thought when we use client certs it's also x509 auth method and we don't stack both auth mechanisms. Pls check.

viveksinghggits: [discussed](https://mongodb.slack.com/archives/C07F4RWNKNH/p1781647339025629) with the search team and we are going to follow the mongot format and have tls related settings in scram as well.

lsierant (approved):

viveksinghggits:
Last to last commit has almost successful CI run. Last commit just fixes the lint failure.

PR #1232, opened 2026-06-11
search: lifecycle teardown for shard/cluster removal (Phase 6)

anandsyncs:
# Based on PR #1048

## Chain of upstream PRs as of 2026-06-28

* PR #1166:
  `master` ← `lsierant/devcontainer`

  * PR #1048:
    `lsierant/devcontainer` ← `search/ga-base`

    * **PR #1232 (THIS ONE)**:
      `search/ga-base` ← `search/lifecycle-teardown`

## Summary

Closes the Phase 6 lifecycle gaps for MongoDB Search GA: removing a shard or a member cluster now tears down everything search-owned (StatefulSets, Services, ConfigMaps, PVCs, Envoy resources), shard removal waits for the source's two-phase drain, and a pre-existing Envoy routing bug triggered by cluster removal is fixed.

**Tickets:** KUBE-112 (shard-remove teardown), KUBE-113 (drain-aware serving), KUBE-114 (centralized cluster-remove teardown).
**Design:** B12 (teardown order), §16.6 (cluster-removal cleanup), OQ9 (always fresh provision).

## Commits → gaps

| Commit | What it does | Gap |
|---|---|---|
| `6bac923d1` | `component` label stamped on mongot STS/headless Svc/ConfigMap (metadata only; selectors untouched) | enables label-driven sweep |
| `7410c0f29` | Generalized stale sweep: deletes owned-but-unexpected mongot STS/Svc/CM + proxy Services per member cluster, RS + sharded plans, best-effort across clusters | G1 (shard remove leaked everything but the proxy Svc), G2 mongot half |
| `f4bbee5d2` | `persistentVolumeClaimRetentionPolicy: Delete/Delete` on mongot STS | PVC reaping (k8s floor 1.33.7 ≥ 1.27 — feature GA everywhere supported) |
| `2e08ae89a` | Full-reconcile proofs: cluster removal sweeps the removed member, mapping retains its index (never reused), re-add reuses the original index, unreachable member doesn't wedge the CR. **Zero production fixes needed** — the sweep already handled removal | G2 proof |
| `96e3c005c` | Envoy controller: same label+ownership sweep for Envoy Deployment/ConfigMap across member clusters | G2 Envoy half |
| `57ced33e2` | Drain-aware shard set: `GetShardCount = max(spec, status)`; search keeps serving draining shards until the source finishes phase-2 drain; deferral logged | G5 |
[...]

PR #1166, opened 2026-05-30
Devcontainers in isolated worktrees with evg hosts

lsierant:
# Isolated local e2e environments: worktrees + devcontainers (`wt-ctl`)

This introduces devcontainer tooling for running fully isolated local e2e tests in git worktrees. It builds on top of `scripts/dev/create_worktree.sh` (#1037) and adds `scripts/dev/wt-ctl` as a single entrypoint for managing everything worktree/local-e2e related. One command creates a worktree with its own devcontainer and a short-lived Evergreen host, recreates kind clusters, prepares the environment, and starts the operator - several such environments can run in parallel on one machine without stepping on each other.

```bash
scripts/dev/wt-ctl create <branch> --context <variant>     # bring everything up
scripts/dev/wt-ctl status                                  # state of this worktree's stack
scripts/dev/wt-ctl attach                                  # tmux inside the devc: k9s, operator log, e2e log, zsh
scripts/dev/wt-ctl attach -- bash scripts/dev/e2e_run.sh <marker>   # run a test
scripts/dev/wt-ctl delete --all <branch>                   # tear everything down (EVG host included)
```

## What `wt-ctl create` does

It creates the devcontainer and the EVG host in parallel, then executes all the steps a developer would otherwise run manually to get to a working local-operator e2e setup:

- rsyncs the local branch sources to the EVG host
- recreates kind clusters - the multi setup by default (`e2e-operator`, `e2e-cluster-{1,2,3}` + `kind`), or one cluster with `--single-cluster`; `--local-kind` skips the EVG host and runs kind on the laptop instead
- switches the context (`--context`)
- runs `prepare-local-e2e`
- downloads kubeconfigs locally, adjusting ports/proxy-url for the topology (remote EVG host, local kind, or directly on a Linux EVG host in CI)
- starts the operator inside the devcontainer as a local process (new `scripts/dev/op_run.sh`); tests run via the new `scripts/dev/e2e_run.sh`; both capture output under `logs/`

Every phase is resumable: `--resume` continues a failed run, `--restart-from <phase>` forces re-running from a given point. Setup logs land in `logs/setup_worktree/` (`wt-ctl logs <phase> -f`).

## Isolation

- Every worktree gets its own devcontainer stack and a fresh EVG host.
[...]

PR #1241, opened 2026-06-11
KUBE-133: delete member-cluster resources when the MongoDBSearch CR is deleted

anandsyncs:
# Based on PR #1232

* PR #1166:

  * PR #1048:

    * PR #1232:
      `search/ga-base` ← `search/lifecycle-teardown`

      * **PR #1241 (THIS ONE)**:
        `search/lifecycle-teardown` ← `search/lifecycle-ondelete`

## Summary

Stacked on #1232. Deleting a MongoDBSearch CR in **centralized multi-cluster** mode currently leaks every search-owned resource in member clusters — owner references don't cross cluster boundaries, so Kubernetes GC never reaches them. This PR adds the event-driven `OnDelete` deletion path at **exact sharded-MC parity** (template: `ShardedClusterReconcileHelper.OnDelete`). Per explicit decision: **no finalizer** — same accepted semantics as sharded MC (cleanup is best-effort; operator-down or member-unreachable at delete time are accepted windows).

## What it does

- Both the MongoDBSearch and MongoDBSearchEnvoy controllers now watch the CR with `ResourceEventHandler{deleter}` (the same construction as sharded/RS/standalone/multi controllers; verified behavior-identical for Create/Update events — the embedded enqueue handler is an empty struct).
- **Search controller OnDelete**: in MC mode, runs the #1232 stale sweep with empty expectations across central + all member clusters — deletes every owned mongot StatefulSet/headless Service/ConfigMap/proxy Service; PVCs ride along via the `whenDeleted: Delete` retention policy. Per-cluster failures accumulate (`errors.Join`) and never abort remaining clusters. Watched-resource registrations dropped afterward.
- **Envoy controller OnDelete**: reuses the existing teardown (`deleteAllEnvoyResources`) for Envoy Deployments/ConfigMaps + drops its own watch registrations.
- Single-cluster mode is untouched: owner-ref GC handles everything (the MC predicate also keeps legacy stored CRs with empty `clusters[]` on the GC path).
- New shared `searchcontroller.IsMultiClusterMode` predicate replaces three drift-prone inline copies.

## Tests

- Unit (mutation-verified — stubbing the sweep fails all three): full MC delete removes everything on central + both members while foreign-owned look-alikes survive; one erroring member doesn't block the other's cleanup; state-CM-already-GC'd fallback.
- Unit, second pass closing audit gaps (also mutation-verified): single-cluster OnDelete deletes nothing (owner-ref GC parity) and only drops watch registrations; legacy CR with empty `clusters[]` under an MC operator stays on the GC path (pins the `IsMultiClusterMode` conjunct); both controllers drop only the deleted search's watcher entries; both sweeps clean a cluster already removed from `spec.clusters`; `IsMultiClusterMode` truth table.
- The `ResourceEventHandler` registration itself (the delete event actually reaching `OnDelete` through controller-runtime) has no unit-level pin — the repo has no envtest harness — so `e2e_mc_search_cr_deletion` is the load-bearing test for that wiring.
- NEW e2e `e2e_mc_search_cr_deletion` (2-cluster variant, 19 tests): deploy → presence-guard all 7 resource kinds per member (anti-vacuity anchor) → live `$search` → **delete CR → poll all kinds to zero in both members** (incl. PVCs + Envoy), source MongoDBMulti untouched → re-create the CR → Running again with the original cluster indices re-assigned (explicitly asserted — the AC mongotHost target stays valid).

## Notes

- Found during this work, NOT fixed here (follow-up): the **sharded internal source's** search setParameters (`mongotHost`, `searchIndexManagementHostAndPort`) are sticky in the automation config after search CR deletion — they're injected into `desiredShardsConfiguration` from a spec deep-copy and never reach `LastAchievedSpec`, so the removal diff can't see them. The RS source self-cleans. GA MC path (external sources) is unaffected.
- No RBAC changes, no finalizer, no EventRecorder.

## Validation

Evergreen all green on exec0 — [6a2b2c58](https://evergreen.mongodb.com/patch/6a2b2c5854fbb200078d8677): unit go+python, `e2e_mc_search_cr_deletion`, `e2e_search_q2_mc_rs_steady` (regression for the watch-handler swap), `e2e_mc_search_cluster_add_remove` (regression for the shared-helper refactor).

## Ticket

KUBE-133 — MongoDBSearch CR-delete cleanup across all clusters (centralized MC)

### DIFF
--- a/.evergreen-tasks.yml
+++ b/.evergreen-tasks.yml
@@ -1430,6 +1430,11 @@ tasks:
     commands:
       - func: "e2e_test"
 
+  - name: e2e_mc_search_cr_deletion
+    tags: [ "patch-run" ]
+    commands:
+      - func: "e2e_test"
+
   - name: e2e_search_q3_mc_sharded_external_mtls
     tags: [ "patch-run" ]
     commands:
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -1129,6 +1129,7 @@ task_groups:
       - e2e_multi_cluster_2_clusters_replica_set
       - e2e_multi_cluster_2_clusters_clusterwide
       - e2e_search_q2_mc_rs_steady
+      - e2e_mc_search_cr_deletion
       - e2e_search_q3_mc_sharded_external_mtls
       - e2e_search_connectivity_tool_mc_rs
       - e2e_search_connectivity_tool_mc_sharded
--- a/controllers/operator/mongodbsearch_controller.go
+++ b/controllers/operator/mongodbsearch_controller.go
@@ -7,6 +7,7 @@ import (
 
 	"go.uber.org/zap"
 	"golang.org/x/xerrors"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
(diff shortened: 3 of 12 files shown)


anandsyncs:
Superseded by the KUBE-154 lifecycle stack (#1403 → #1362 → #1382 → #1383 → #1393): this PR's CR-delete cleanup (OnDelete routing + member-cluster sweep) is carried by #1382 in a more evolved form. Closing.

PR #1246, opened 2026-06-12
KUBE-123: Fix IsRunningEnterpriseImage for Community CR deployment telemetry

josvazg:
# Summary

All MongoDBCommunity deployment telemetry rows were reporting `IsRunningEnterpriseImage` = true, producing a misleading ~98% enterprise rate for the Community type.

The root cause: addCommunityEvents evaluated Images.IsEnterpriseImage (mongodbImage) where mongodbImage is the env var used only by MEKO Controllers (e.g. non MongoDBCommunity CRD resources), which - by default - always contains the enterprise string — regardless of what MongoDB Community server image the CR itself is actually running.

Fix

pkg/telemetry/collector.go

- Removed **mongodbImage** from addCommunityEvents (it was the wrong signal entirely).
- Added `isCommunityRunningEnterpriseImage(mdb)`, which mirrors the logic already used by guessEnterprise in the community controller: inspect the mongod container image explicitly set in:

`mdb.Spec.StatefulSetConfiguration.SpecWrapper.Spec.Template.Spec.Containers`

If the user has overridden it with an enterprise image → true; otherwise → false (Community CRs use the community server image by default).

The MongoDB (enterprise) CR path in `getMdbEvents` already evaluates enterprise status at the CR level via per-resource annotations. The `addCommunityEvents` now follows the same principle — the check is grounded in what the CR is actually configured to run, not what the operator binary itself is built from.

## Proof of Work

Updated TestAddCommunityEvents in collector_test.go:
- Old tests incorrectly validated behavior driven by the (now-removed) `mongodbImage` parameter.
- New test cases cover: 
  - no image override (→ false)
  - explicit enterprise override (→ true)
  - explicit community override (→ false)
  - and mixed CRs in the same list.

## Checklist

### DIFF
--- a/changelog/20260612_fix_community_deployment_enterprise_image_telemetry.md
+++ b/changelog/20260612_fix_community_deployment_enterprise_image_telemetry.md
@@ -1,0 +1,8 @@
+---
+title: Fix Community deployment enterprise image telemetry
+kind: fix
+date: 2026-06-12
+---
+
+* Fixed a bug where all `MongoDBCommunity` deployment telemetry rows incorrectly reported `IsRunningEnterpriseImage = true`. The field was being evaluated against the operator-level enterprise image rather than the image configured in each CR's spec, causing a misleading ~100% enterprise rate for the Community deployment type.
+* `IsRunningEnterpriseImage` for Community deployments is now derived from the `mongod` container image override in `spec.statefulSet.spec.template.spec.containers`, if present. When no override is set, the field correctly defaults to `false`, reflecting that Community CRs use the community MongoDB server image by default.
--- a/pkg/telemetry/collector.go
+++ b/pkg/telemetry/collector.go
@@ -220,7 +220,7 @@ func collectDeploymentsSnapshot(ctx context.Context, operatorClusterMgr manager.
 	events = append(events, addMultiEvents(ctx, operatorClusterClient, operatorUUID, mongodbImage, databaseNonStaticImage, defaultArchitecture, now)...)
 	// No need to pass databaseNonStaticImage because it is for sure not enterprise image
 	events = append(events, addOmEvents(ctx, operatorClusterClient, operatorUUID, mongodbImage, defaultArchitecture, now)...)
-	events = append(events, addCommunityEvents(ctx, operatorClusterClient, operatorUUID, mongodbImage, now)...)
+	events = append(events, addCommunityEvents(ctx, operatorClusterClient, operatorUUID, now)...)
 	events = append(events, addSearchEvents(ctx, operatorClusterClient, operatorUUID, defaultArchitecture, now)...)
 	return events
 }
@@ -352,7 +352,7 @@ func createEvent(properties FlatProperties, now time.Time, eventType EventType) 
 	}
 }
 
-func addCommunityEvents(ctx context.Context, operatorClusterClient kubeclient.Client, operatorUUID, mongodbImage string, now time.Time) []Event {
+func addCommunityEvents(ctx context.Context, operatorClusterClient kubeclient.Client, operatorUUID string, now time.Time) []Event {
 	var events []Event
 	communityList := &mcov1.MongoDBCommunityList{}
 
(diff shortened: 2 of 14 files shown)


changelog/20260612_fix_community_deployment_enterprise_image_telemetry.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Fix Community deployment enterprise image telemetry
+kind: fix
+date: 2026-06-12
+---
+
+* Fixed a bug where all `MongoDBCommunity` deployment telemetry rows incorrectly reported `IsRunningEnterpriseImage = true`. The field was being evaluated against the operator-level enterprise image rather than the image configured in each CR's spec, causing a misleading ~100% enterprise rate for the Community deployment type.
```
MaciejKaras, line 7: nit: I don't think this is that much customer related fix - it's a telemetry fix that we, MongoDB org, are the sole consumers. If some deployment was improperly tagged with Enterprise image than the fix will only benefit us. I would rather write down a 1-2 sentence entry about a fixes to telemetry reporting for MongoDBCommunity resource and that's it.

pkg/telemetry/collector.go
```diff
@@ -376,6 +376,18 @@ func addCommunityEvents(ctx context.Context, operatorClusterClient kubeclient.Cl
 	return events
 }
 
+// isCommunityRunningEnterpriseImage checks if a MongoDBCommunity CR is configured to run an
```
MaciejKaras, line 379: nit: please remove comments

MaciejKaras (approved):

changelog/20260612_fix_community_deployment_enterprise_image_telemetry.md
nammn: I wouldn't even add a changelog, no customer has something from this

nammn (commented):
Fix is correct, but pr description is not entirely correct; 
"The root cause: addCommunityEvents evaluated Images.IsEnterpriseImage (mongodbImage) where mongodbImage is the operator-level enterprise image (e.g. quay.io/mongodb/mongodb-enterprise-server), which always contains the enterprise string — regardless of what MongoDB server image the CR itself is actually running."

should be
"
The root cause: addCommunityEvents evaluated `Images.IsEnterpriseImage` (mongodbImage) where mongodbImage is the env var used only by MEKO Controllers (e.g. non MongoDBCommunity CRD resources), which - by default - always contains the enterprise string — regardless of what MongoDB Community server image the CR itself is actually running.
"

Having said that - I think using the same code but replacing the env var check with `MDB_COMMUNITY_IMAGE` **might** work as well, since thats what MCO Controllers are using: https://github.com/mongodb/mongodb-kubernetes/blob/5477d57a0149d1c6467eab32e390a67f92e3c9d2/mongodb-community-operator/controllers/construct/mongodbstatefulset.go#L31

**might** - because APPDB runs MCO but with the enterprise-image, so an edge-case to look out for

josvazg:
> Fix is correct, but pr description is not entirely correct; "The root cause: addCommunityEvents evaluated Images.IsEnterpriseImage (mongodbImage) where mongodbImage is the operator-level enterprise image (e.g. quay.io/mongodb/mongodb-enterprise-server), which always contains the enterprise string — regardless of what MongoDB server image the CR itself is actually running."
> 
> should be " The root cause: addCommunityEvents evaluated `Images.IsEnterpriseImage` (mongodbImage) where mongodbImage is the env var used only by MEKO Controllers (e.g. non MongoDBCommunity CRD resources), which - by default - always contains the enterprise string — regardless of what MongoDB Community server image the CR itself is actually running. "
> 
> Having said that - I think using the same code but replacing the env var check with `MDB_COMMUNITY_IMAGE` **might** work as well, since thats what MCO Controllers are using:
> 
> https://github.com/mongodb/mongodb-kubernetes/blob/5477d57a0149d1c6467eab32e390a67f92e3c9d2/mongodb-community-operator/controllers/construct/mongodbstatefulset.go#L31
> 
> **might** - because APPDB runs MCO but with the enterprise-image, so an edge-case to look out for

Updated the description for now.

Do you mean there is a simple fix for this? can you elaborate?

nammn (approved):

PR #1263, opened 2026-06-23
Forward Search metrics to Ops Manager

fealebenpae:
# Based on PR #1048

* PR #1166:

  * PR #1048:

    * **PR #1263 (THIS ONE)**:
      `search/ga-base` ← `fealebenpae/om-search-awareness`

## Summary

MongoDBSearch exposes mongot Prometheus metrics but had no path to surface them in Ops Manager for unified monitoring. This PR adds two related changes.

**`spec.prometheus` → `spec.observability.prometheus`** (breaking API change). The field moves under a new `ObservabilityConfig` struct. A `mode` field (`enabled`/`disabled`, default `enabled`) replaces the previous opt-in pattern — previously an absent `spec.prometheus` meant the metrics endpoint was off; now it is on by default on port 9946. Opt out by setting `spec.observability.prometheus.mode: disabled`.

Since Prometheus is a requirement for the collector to forward Search metrics to Ops Manager, and the metrics forwarder should be enabled as often as possible, this PR flips the Prometheus behavior to be enabled by default, so that goggling the metrics forwarder doesn't cause mongot rollouts.

**New `spec.observability.metricsForwarder`** — a dedicated reconciler (`MongoDBSearchMetricsForwarderReconciler`) that runs alongside the existing search controllers. Per MongoDBSearch CR it deploys one OpenTelemetry Collector Deployment per cluster. The collector is configured via a rendered YAML template (Go embed) and stored in a ConfigMap the reconciler owns; a hash annotation on the Deployment triggers rolling updates on config changes.

The collector scrapes mongot's Prometheus endpoint via DNS SRV service discovery against the headless mongot service (`_prometheus._tcp.<name>-search-<clusterIdx>-svc`), applies a pipeline of processors (filter, rename, metrics generation, cumulative-to-delta), and exports to Ops Manager's OTLP ingest endpoint using agent-API-key basic auth (`groupID:apiKey`). Each mongot pod maps to a distinct host in Ops Manager identified by `sha1(groupID + "-" + namespace + "-" + podName)` — this matches the `mongodb.opsmanager.host.id` resource attribute the collector sets during OTLP ingestion, allowing OM to associate scraped metrics with the right monitored host.

**Topology state machine for host deregistration.** Removing a mongot pod (scale-down, shard removal, CR deletion) requires calling the Ops Manager `agents/api/hosts` delete endpoint. But the OTel collector's Prometheus SD cache can hold a stale reference for up to `SDRefreshInterval + 2×ScrapeInterval` after the pod leaves DNS — deregistering earlier causes ingestion errors for a host OM no longer knows. The reconciler tracks each removed pod through a state machine (`PendingHostDeletions → HostDeletionReadyAfter → Cleaned`) persisted in a dedicated ConfigMap. Deregistration is deferred until the pod is fully gone from Kubernetes *and* the deferral window has elapsed. While deletions are in flight the reconciler requeues every 15 seconds.

**OM project resolution.** For internal `resourceRef` sources the reconciler infers the OM project ConfigMap and agent-key Secret from the source `MongoDB` resource. For external sources (or any source with an explicit `.opsManager` block) the user provides references explicitly. Group ID is pre-resolved from `MongoDB.Status.ProjectId` for internal sources; for explicit configs it is resolved at runtime via the OM agents group API. Cloud Manager and self-hosted OM < 8.0.24 are rejected (forwarder silently disabled in `auto` mode; `Failed` in `enabled` mode).

**Multi-cluster.** The reconciler fans out per entry in the cluster work list (derived from the search state ConfigMap). In a hub-and-spoke install, the central cluster receives the agent-key Secret and OM CA ConfigMap replicated under forwarder-owned labels; each member cluster gets its own Deployment and ConfigMap.

**New CRD fields:**
| Field | Default | Description |
|---|---|---|
| `spec.observability.prometheus.mode` | `enabled` | Prometheus endpoint in mongot |
| `spec.observability.metricsForwarder.mode` | `auto` | `auto`/`enabled`/`disabled` |
| `spec.observability.metricsForwarder.resourceRequirements` | 100m/128Mi req, 250m/256Mi lim | Forwarder container resources |
| `spec.observability.metricsForwarder.deployment` | — | Optional Deployment overrides |
| `spec.observability.metricsForwarder.opsManager` | — | Explicit project ConfigMap + agent credentials Secret |

New status column: `MetricsForwarder` (phase + message). New finalizer (`search-metrics-forwarder`) added while the forwarder is enabled so deletion always triggers host cleanup even if the mode is changed to `disabled` before the CR is deleted.

## Proof of Work

- Unit: mongodbsearch_metrics_controller_test.go, metrics_forwarder_otel_config.go tests
- e2e (pending): `search_enterprise_metrics_forwarder_replicaset.py`, `search_enterprise_metrics_forwarder_sharded.py`, `search_external_metrics_forwarder_replicaset.py`, `search_external_metrics_forwarder_sharded.py`

## Checklist


controllers/searchcontroller/metrics_forwarder_otel_config.yaml.tmpl
```diff
@@ -0,0 +1,342 @@
...
+          - names:
+            {{ range .ShardNames }}
+            - _prometheus._tcp.{{ $.MongotName }}-search-{{ $.ClusterIndex }}-{{ . }}-svc
+            {{ else }}
+            - _prometheus._tcp.{{ $.MongotName }}-search-{{ $.ClusterIndex }}-svc
+            {{ end }}
+            type: SRV
+            refresh_interval: 30s
+          metrics_path: /metrics
+processors:
```
fealebenpae, line 24: Everything between the `memory_limiter` and `resource/ops_manager_discovery_attributes` processors comes from the Ops Manager collaboration and implements the transformation of the mongot Prometheus metrics in the OTLP datapoints that Ops Manager expects. It's safe to skip reviewing.

controllers/operator/mongodbsearch_metrics_controller.go
```diff
@@ -0,0 +1,1353 @@
...
+	// and a second interval provides a safety margin.
+	hostDeletionDeferralWindow = prometheusSDRefreshInterval + 2*prometheusDefaultScrapeInterval
+)
+
+// MongoDBSearchMetricsForwarderReconciler reconciles the metrics forwarder Deployment
+// that forwards mongot Prometheus metrics to Ops Manager.
+type MongoDBSearchMetricsForwarderReconciler struct {
+	kubeClient         kubernetesClient.Client
+	secretClient       secrets.SecretClient
+	watch              *watch.ResourceWatcher
```
lsierant, line 97: how many reconcilers do you expect to be deployed and reconciled potentially in parallel? 
Asking because the Reconciler instance is a footgun as it's effectively a singleton called concurrently by controller sdk. That's why we've introduced a pattern of ReconcilerHelpers, which are instantiated anew with each Reconcile call. 

In this case resource watcher might be a contention point (we've hardened it with mutexes before, but still it might catch you off guard)

fealebenpae: Just one, I think. Do we instantiate multiples of other reconcilers somewhere else?

lsierant: not directly - as said the Reconciler is a singleton. But if you allow parallel reconciles and have more than one CR it will execute Reconcile method of said singleton in parallel in goroutines.

fealebenpae: I understand. This looks like a candidate for field indexers in any case, but I didn’t have the time to sort out the transitive MongoDBSearch -> MongoDB -> ConfigMap and Secret dependency so it’ll be a second pass that hardens this.
Thank you for pointing this out.

lsierant: just make sure we're including reconciliation of this CR in the race unit tests. If we have no races it's fine. 

Also I've just realised that we're also doing too much in the singleton in the MongoDBSearch CR itself, so it's also a candidate for hardening later.

lsierant (approved):

PR #1270, opened 2026-06-24
[KUBE-140] Introduce a CR field `routerHostname` to configure the cluster level hostname in mongos

viveksinghggits:
# Summary

In the earlier implementation to figure out the cluster level hostname that should be configured in the mongos we just removed the `{shardname}` **prefix** from the `externalHostname` field. This was not a good idea and the main reason was the `{shardname}` placeholder can be anywhere in the `externalHostname` string. And even if we improved the function and instead of removing the prefix we just removed `{shardname}` placeholder form entire string, we might get into some format problems. For example `abc.{shardname}-us.example.com` or `abc.{shardname}.example.com`, how would be deterministically remove `{shardname}` from above strings.
To fix this problem, this PR introduces a new CR field `routerHostname` under `clusters.loadBalancer.managed`. Which can be used to configure the the hostname in the externally installed mongos processes.

This `routerHostname` would eventually end up in the envoy config in the filter chain that would select all the clusters (mongot groups) in that K8s cluster.

These  are the validations that are applied to the `routerHostname`, validations happen only in case of external sharded mongodb sources with managed load balancers.

- `spec.clusters[i].loadBalancer.managed.routerHostname` must be specified when using managed load balancer with an external sharded MongoDB source
- No `{shardName}` placeholder: it's the shard-agnostic mongos endpoint, so it must not contain `{shardName}`
- `routerHostname` must be distinct across `spec.clusters[i].loadBalancer.managed`

## Proof of Work

Successful CI.

## Checklist

### DIFF
--- a/api/mongodb/v1/search/mongodbsearch_types.go
+++ b/api/mongodb/v1/search/mongodbsearch_types.go
@@ -265,6 +265,15 @@ type ManagedLBConfig struct {
 	// Required when MongoDB is externally managed. Ignored for operator-managed MongoDB.
 	// +optional
 	ExternalHostname string `json:"externalHostname,omitempty"`
+	// RouterHostname is the host:port a sharded cluster's mongos (router) uses to reach this cluster's
+	// mongot via the managed Envoy LB, and the SNI hostname Envoy matches for the cluster-level filter
+	// chain. Unlike ExternalHostname it is shard-agnostic and must NOT contain a {shardName}
+	// placeholder. This value is per-cluster: in multi-cluster deployments each cluster has its own
+	// Envoy LB and its mongos routes to it, so every cluster's RouterHostname must be distinct.
+	// Required for an external sharded MongoDB source with managed LB; ignored for ReplicaSet sources
+	// and operator-managed MongoDB. Must be covered by the Envoy LB server certificate SANs.
+	// +optional
+	RouterHostname string `json:"routerHostname,omitempty"`
 	// Replicas is the number of Envoy proxy pods to deploy.
 	// Defaults to 1 if not specified.
 	// +optional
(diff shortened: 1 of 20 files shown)


api/mongodb/v1/search/mongodbsearch_validation.go
```diff
@@ -617,6 +641,35 @@ func validateMCExternalHostnames(s *MongoDBSearch) v1.ValidationResult {
 	return v1.ValidationSuccess()
 }
 
+// validateMCRouterHostnames enforces, for an external sharded source with a managed LB, that every
```
lsierant, line 644: curious if we couldn't apply a CEL expression here like for other unique fields we have

viveksinghggits: We can apply CEL for this, it's possible but a little complex, less readable because this validation depends on the source. We are validating this only if the source is external sharded deployment and load balancer is managed because of that the CEL would be moved to the spec level and not the cluster level.

api/mongodb/v1/search/mongodbsearch_types.go
```diff
@@ -265,6 +265,15 @@ type ManagedLBConfig struct {
 	// Required when MongoDB is externally managed. Ignored for operator-managed MongoDB.
 	// +optional
 	ExternalHostname string `json:"externalHostname,omitempty"`
+	// RouterHostname is the host:port a sharded cluster's mongos (router) uses to reach this cluster's
```
anandsyncs, line 268: I think the MCK code skill will make this comment shorter.

viveksinghggits: https://mongodb.slack.com/archives/C0ABQ8BDF0S/p1782383016136349?thread_ts=1782382763.085769&cid=C0ABQ8BDF0S

viveksinghggits: I can do it, but it really feels valuable here.

PR #1271, opened 2026-06-24
CLOUDP-407997: Wire mckci promotion into Evergreen CI

josvazg:
# Summary

Promote the successfully tested images in branch `master` as **releaseable** as is, without further testing or processing.

Initial promotion only marked the operator image, but we need to mark all other images rebuilt and released together with the operator's.

## Proof of Work

CI must pass

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,4 @@
+# MCK Project Conventions
+
+## Go style
+- Struct field docs go on the line above the field, not as trailing inline comments.
--- a/ci/internal/cli/release_promote.go
+++ b/ci/internal/cli/release_promote.go
@@ -9,19 +9,19 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/ci/internal/release"
 )
 
-func newReleasePromoteCmd() *cobra.Command {
+func newPromoteReleaseCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "promote",
 		Short: "Promote candidate image(s) by applying promoted-latest and promoted-{commit}-{version} tags",
 	}
 
-	cmd.AddCommand(newReleasePromoteImageCmd())
-	cmd.AddCommand(newReleasePromoteGroupCmd())
+	cmd.AddCommand(newPromoteImageCmd())
+	cmd.AddCommand(newPromoteImagesCmd())
 
 	return cmd
 }
 
-func newReleasePromoteImageCmd() *cobra.Command {
+func newPromoteImageCmd() *cobra.Command {
 	var (
 		image       string
 		commit      string
@@ -35,7 +35,7 @@ func newReleasePromoteImageCmd() *cobra.Command {
 		Use:   "image",
 		Short: "Promote a single candidate image",
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			return runSinglePromote(cmd, image, commit, version, registryURL, repo, dryRun)
+			return promoteImage(cmd, image, commit, version, registryURL, repo, dryRun)
 		},
 	}
 
@@ -53,7 +53,7 @@ func newReleasePromoteImageCmd() *cobra.Command {
 	return cmd
 }
 
-func newReleasePromoteGroupCmd() *cobra.Command {
+func newPromoteImagesCmd() *cobra.Command {
 	var (
 		buildInfo   string
 		releaseJSON string
@@ -62,10 +62,10 @@ func newReleasePromoteGroupCmd() *cobra.Command {
 	)
 
 	cmd := &cobra.Command{
-		Use:   "group",
+		Use:   "images",
 		Short: "Promote every release image defined in build_info.json at the given commit",
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			return runGroupPromote(cmd, buildInfo, releaseJSON, commit, dryRun)
+			return promoteAllImages(cmd, buildInfo, releaseJSON, commit, dryRun)
 		},
 	}
 
@@ -79,14 +79,14 @@ func newReleasePromoteGroupCmd() *cobra.Command {
 	return cmd
 }
 
-// runGroupPromote promotes every release image in build_info.json at the given
+// promoteAllImages promotes every release image in build_info.json at the given
 // commit, writing promoted tags to each image's primary staging repository.
-func runGroupPromote(cmd *cobra.Command, buildInfo, releaseJSON, commit string, dryRun bool) error {
+func promoteAllImages(cmd *cobra.Command, buildInfo, releaseJSON, commit string, dryRun bool) error {
 	images, err := release.LoadReleaseImages(buildInfo, releaseJSON)
 	if err != nil {
 		return err
 	}
-	results, err := release.PromoteGroup(images, commit, dryRun, release.DefaultRegistryConnector)
+	results, err := release.PromoteImages(images, commit, dryRun, release.DefaultRegistryConnector)
 	if err != nil {
 		return err
 	}
@@ -104,7 +104,7 @@ func runGroupPromote(cmd *cobra.Command, buildInfo, releaseJSON, commit string, 
 	return nil
 }
 
-func runSinglePromote(cmd *cobra.Command, image, commit, version, registryURL, repo string, dryRun bool) error {
+func promoteImage(cmd *cobra.Command, image, commit, version, registryURL, repo string, dryRun bool) error {
 	tags, err := release.Promote(release.PromoteInputs{
 		Image:   image,
 		Commit:  commit,
--- a/ci/internal/cli/release_promote.go
+++ b/ci/internal/cli/release_promote.go
@@ -10,6 +10,18 @@ import (
 )
 
 func newReleasePromoteCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "promote",
+		Short: "Promote candidate image(s) by applying promoted-latest and promoted-{commit}-{version} tags",
+	}
+
+	cmd.AddCommand(newReleasePromoteImageCmd())
+	cmd.AddCommand(newReleasePromoteGroupCmd())
+
+	return cmd
+}
+
+func newReleasePromoteImageCmd() *cobra.Command {
 	var (
 		image       string
 		commit      string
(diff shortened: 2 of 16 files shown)


m1kola (approved):

josvazg:
Turns out I need to add a group promotion, as we ship more images than just. the operator one

ci/internal/release/registry.go
```diff
@@ -9,58 +9,34 @@ import (
...
-// gcrTransport implements Transport via google/go-containerregistry.
-type gcrTransport struct {
+// gcrRegistry implements Registry via google/go-containerregistry.
+type gcrRegistry struct {
 	host     string
 	insecure bool
 }
 
-func (t *gcrTransport) CopyWithTags(srcRef string, dstRepo string, tags []string) error {
+func (t *gcrRegistry) CopyWithTags(srcRef string, dstRepo string, tags []string) error {
```
m1kola: nit: gcr is probably not the best name because it associates with Google Container Registry (GCR).

ci/internal/release/images_promote.go
```diff
@@ -0,0 +1,65 @@
...
+		src := fmt.Sprintf("%s:%s", img.StagingRepo, img.Version)
+		tags, err := Promote(PromoteInputs{
+			Image:   src,
+			Commit:  commit,
+			Version: img.Version,
+			Repo:    path,
+			DryRun:  dryRun,
+		}, connect(host))
+		if err != nil {
+			return nil, fmt.Errorf("promote %s (%s): %w", img.Name, src, err)
```
m1kola, line 47: question: what happens if promotion fails on, let's say, second image? Promotion is not atomic. After fixing the issue, can we re-run the command? Or do we need to bump a version? E.g. if 2.0.1 promotion failed, do we need to do 2.0.2?

m1kola (approved):
Looks good. Just a nit (can be left as is or addressed in a follow up PR) and a question.

ci/internal/release/registry.go
josvazg: Will fix, Will use cRegistry (container Registry)
gcr comes from go container registry (not google), but yes it is unfortunate

ci/internal/release/images_promote.go
josvazg: I seek idempotency here, by default promotions do not error out if the existing image is the same SHA. So the ida is if promotion fails half way, you can retry until it eventually get all images promoted.
Still the stomp detection is not implemented in this PR yet.

m1kola (approved):

ci/internal/cli/release_promote.go
```diff
@@ -10,6 +10,18 @@ import (
...
+		Short: "Promote candidate image(s) by applying promoted-latest and promoted-{commit}-{version} tags",
+	}
+
+	cmd.AddCommand(newReleasePromoteImageCmd())
+	cmd.AddCommand(newReleasePromoteGroupCmd())
+
+	return cmd
+}
+
+func newReleasePromoteImageCmd() *cobra.Command {
```
viveksinghggits: nit: 
```suggestion
func newPromoteImageCmd() *cobra.Command {
```

ci/internal/cli/release_promote.go
```diff
@@ -10,6 +10,18 @@ import (
 )
 
 func newReleasePromoteCmd() *cobra.Command {
```
viveksinghggits:
```suggestion
func newPromoteReleaseCmd() *cobra.Command {
```

I know this change is not part of this PR but just wanted to highlight that `promoteRelease` seems to be conveying the meaning better that `releasePromote`.

ci/internal/cli/release_promote.go
```diff
@@ -20,38 +32,16 @@ func newReleasePromoteCmd() *cobra.Command {
...
+			return runSinglePromote(cmd, image, commit, version, registryURL, repo, dryRun)
 		},
 	}
 
-	cmd.Flags().StringVar(&image, "image", "", "source image reference to promote (required)")
-	cmd.Flags().StringVar(&commit, "commit", "", "commit SHA to encode in the promoted tag (required)")
-	cmd.Flags().StringVar(&version, "version", "", "version to encode in the promoted tag (required)")
+	cmd.Flags().StringVar(&image, "image", "", "source image reference to promote")
+	cmd.Flags().StringVar(&commit, "commit", "", "commit SHA to encode in the promoted tag")
+	cmd.Flags().StringVar(&version, "version", "", "version to encode in the promoted tag")
```
viveksinghggits, line 44: q: what would the version of the promoted image? I am assuming it would be the MCK version that we are releasing right? do we not accept that here? Also what is the significance of `version` flag here?

ci/internal/cli/release_promote.go
```diff
@@ -63,6 +53,82 @@ func newReleasePromoteCmd() *cobra.Command {
 	return cmd
 }
 
+func newReleasePromoteGroupCmd() *cobra.Command {
```
viveksinghggits: nit:
```suggestion
func newReleaseAllImagesCmd() *cobra.Command {
```
what do you think about this?

viveksinghggits: or `promoteAllImagesCmd`.

ci/internal/cli/release_promote.go
```diff
@@ -63,6 +53,82 @@ func newReleasePromoteCmd() *cobra.Command {
...
+	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "print what would happen without copying any images")
+
+	MustNotErr(cmd.MarkFlagRequired("commit"))
+
+	return cmd
+}
+
+// runGroupPromote promotes every release image in build_info.json at the given
+// commit, writing promoted tags to each image's primary staging repository.
+func runGroupPromote(cmd *cobra.Command, buildInfo, releaseJSON, commit string, dryRun bool) error {
```
viveksinghggits:
```suggestion
func promoteAllImages(cmd *cobra.Command, buildInfo, releaseJSON, commit string, dryRun bool) error {
```

viveksinghggits: Basically, I am inclined towards calling all images to AllImages instead of calling that Group. if you agree with this, we should make that change everywhere.

ci/internal/cli/release_promote.go
```diff
@@ -63,6 +53,82 @@ func newReleasePromoteCmd() *cobra.Command {
...
+			}
+			if _, err := fmt.Fprintf(cmd.OutOrStdout(), "%s: %s:%s\n", verb, r.Repo, tag); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func runSinglePromote(cmd *cobra.Command, image, commit, version, registryURL, repo string, dryRun bool) error {
```
viveksinghggits:
```suggestion
func promoteImage(cmd *cobra.Command, image, commit, version, registryURL, repo string, dryRun bool) error {
```

ci/internal/release/group.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"os"
+	"sort"
+)
+
+// AnchorImageName is the image that anchors the release group. The operator is
+// the one image built at merge time; the rest travel with its commit. Its
+// presence in the group is required. (Publish-time "latest" resolution keys off
+// this same anchor, but that lives in a separate change.)
+const AnchorImageName = "operator"
```
viveksinghggits: > AnchorImageName is the image that anchors the release group.

q: What exactly does this mean? Can we simplify this please?

ci/internal/release/group.go
```diff
@@ -0,0 +1,126 @@
...
+			Secondary  []string `json:"secondary-repositories"`
+		} `json:"staging"`
+		Release struct {
+			Repository string `json:"repository"`
+		} `json:"release"`
+	} `json:"images"`
+}
+
+// LoadReleaseImages reads build_info.json and release.json and returns the
+// release group: exactly the images carrying a version-ref. Membership is
```
viveksinghggits:
```suggestion
// LoadReleaseImages reads build_info.json and release.json and returns the
// release group (`[]ReleaseImage`): exactly the images carrying a version-ref. Membership is
```

ci/internal/release/group_promote.go
```diff
@@ -0,0 +1,65 @@
+package release
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// GroupPromoteResult records what was (or would be) promoted for one image.
```
viveksinghggits: q: this comment is a little confusing. how can a type carry what was promoted at one time and at other times what would be promoted?

ci/internal/release/group_promote.go
```diff
@@ -0,0 +1,65 @@
...
+type GroupPromoteResult struct {
+	Name    string   // build_info.json image key
+	Repo    string   // primary staging repository (host/path) the tags were applied to
+	Version string   // release.json version the source image was tagged with
+	Tags    []string // tags applied: promoted-{commit}-{version} and promoted-latest
+}
```
viveksinghggits: nit: I thought why do these comments look different, have I not seen the field level comments in Go types and then it clicked that because they are inline on the fields, usually we add them on top of each field.

ci/internal/release/group_promote.go
```diff
@@ -0,0 +1,65 @@
...
+// promoted-{commit}-{version} and promoted-latest to each image's PRIMARY
+// staging repository only (secondary repositories are intentionally left
+// untouched for now). The source image is the version-tagged image the staging
+// build already pushed, e.g. <staging-repo>:<version>.
+//
+// It hard-fails on the first image that cannot be promoted (e.g. its
+// version-tagged source image is missing), so a broken merge build does not
+// silently promote a partial group. connect resolves a Registry for an image's
+// host; the CLI passes DefaultRegistryConnector and tests inject a fake.
+func PromoteGroup(images []ReleaseImage, commit string, dryRun bool, connect RegistryConnector) ([]GroupPromoteResult, error) {
```
viveksinghggits: Would `PromoteImages` be better here? or even `PromoteImagesGroup`.

ci/internal/cli/release_promote.go
josvazg: Yes, this is the MCK version, but this tool is a bit agnostic and flexible, so it makes no assumptions. Here we are just saying, "take this image and tag it as belonging to this version and this commit sha".

Our usage logic will make sure the version we pass matches the MCK version that was already computed in such commit and that the commit hash is for the code we run and passed all testing.

ci/internal/release/group_promote.go
josvazg: This is used twice: at promotion time and at release time.

At promotion time all related tested images are tagged as promoted. At release time, the same group is picked up from the promoted registry by commit or latest tag, so that 'pinned' group can be published to the final registries with final release tags.

**The release-time half lands in a follow-up PR** (group_publish.go) — this PR only covers the promote side, hence '(or would be)' referring to the dry-run case here.

ci/internal/release/group_promote.go
josvazg: Good catch, agreed — fixed here, and I've added a CLAUDE.md convention so this stays consistent going forward.

ci/internal/release/group_promote.go
josvazg: I'd rather keep PromoteGroup here — it matches GroupPromoteResult, the file name (group_promote.go), and the upcoming PublishGroup/GroupPublishResult pair on the release-publish side. promoteAllImages in the CLI is just the unexported wrapper; it doesn't need to mirror this name. Renaming this to PromoteAllImages would make the func, its own return type, and the sibling publish function all disagree on "Group" vs "AllImages".

PromoteImagesGroup keeps "Group" so it'd stay consistent with GroupPromoteResult/PublishGroup, but it's a bit clunky — Go convention favors verb+noun (PromoteGroup), and "Images" doesn't add disambiguation since the func only ever takes []ReleaseImage. I'd rather keep PromoteGroup.

viveksinghggits: > I'd rather keep PromoteGroup here — it matches GroupPromoteResult, the file name (group_promote.go), and the upcoming PublishGroup/GroupPublishResult pair on the release-publish side.

Like I said [above](https://github.com/mongodb/mongodb-kubernetes/pull/1271#discussion_r3579088947), if someone reads `group` they would not intuitively think that container images are being talked about here. That's why I was wondering if we are talking about container images why do we not just say Images or if we really want mention group we can say `ImageGroup`, it's more readable in my opinion.
And if it's about the file name etc, if you agree that not having group is better we can change file names, that should not be the deciding factor according to me here. 
But I will leave this up to you, I can get in a call as well to quickly settle this 😄 .

josvazg: To be honest, I really need to get this merged ASAP and the name does not seem to me a blocker either way.

I will rename shortly.

josvazg: It is just unfortunate it will kick off the full CI again, for something unrelated to prod.

josvazg: done

CLAUDE.md
viveksinghggits, line 1: do we want this file to be part of current PR? I remember having some discussions about whether there should be a claude.md at nested dirs as well. 
Anyway, even if we want to add this file, it should be part of another PR.

viveksinghggits (approved):

CLAUDE.md
josvazg: will remove

PR #1273, opened 2026-06-25
KUBE-143: delete mongot index PVC on scale-down and CR delete

anandsyncs:
# Based on PR #1048

* PR #1166:

  * PR #1048:

    * **PR #1273 (THIS ONE)**:
      `search/ga-base` ← `search/pvc-retention-policy`

**Jira:** [KUBE-143](https://jira.mongodb.org/browse/KUBE-143)

## What

Sets `persistentVolumeClaimRetentionPolicy: {whenDeleted: Delete, whenScaled: Delete}` on the mongot StatefulSet built by `CreateSearchStatefulSetFunc`.

This is the **first, self-contained slice** of Search GA lifecycle teardown. It is intentionally minimal and has **zero dependency on the cluster-index mapping machinery** removed by #1239, so it lands cleanly on current ga-base.

## Why these values

The mongot search index is a **rebuildable derived** index, not primary data — so freeing its storage promptly is the desired behaviour, and a rebuild from mongod is an acceptable cost.

- **`whenDeleted: Delete`** — when the `MongoDBSearch` CR is deleted, owner-ref GC removes the StatefulSet and the retention policy cascades the PVC deletion, so we don't leak index storage.
- **`whenScaled: Delete`** — on scale-down, the StatefulSet controller immediately removes the orphaned ordinal's PVC. A later scale-up reindexes from mongod.

> **Product decision:** immediate PVC reclaim on scale-down. This deviates from the Kubernetes default (`Retain`) and from data-bearing workloads (mongod/AppDB retain their PVCs) precisely because the search index is rebuildable. **Draining a replica before teardown is a planned follow-up** — this PR does not gate teardown on drain.

## Scope / coverage

- ✅ **Single-cluster teardown**: CR delete → owner-ref GC deletes the StatefulSet → `whenDeleted: Delete` cascades the PVC.
- ✅ **Scale-down (per cluster)**: replica reduction → `whenScaled: Delete` removes the orphaned ordinal's PVC.
- ➡️ **Multi-cluster CR-delete reclaim** activates via the member-cluster `OnDelete` sweep in #1274 (stacked on this PR): member StatefulSets aren't owned by the central CR, so the sweep deletes them and `whenDeleted: Delete` then reclaims their PVCs.
- ⏳ **Shard / cluster scale-down** (per-shard mongot teardown, dropping a cluster from `spec.clusters`) reclaim their PVCs the same way once the operator deletes those StatefulSets — that teardown sweep is tracked separately.

## Convention alignment

- No controller in MCK sets `persistentVolumeClaimRetentionPolicy` today; mongod and AppDB rely on the Kubernetes default `Retain`/`Retain` for primary data. The helper is **search-local** (one consumer) rather than promoted to `pkg/statefulset` — generalizing would imply a delete policy the data-bearing controllers explicitly don't want.
- Overridable per cluster via the existing `clusters[].statefulSet` path if a customer ever needs to retain the index — no new CRD surface added.

## K8s support

Requires the `StatefulSetAutoDeletePVC` feature gate (beta/on-by-default since k8s 1.27, GA 1.32). The operator's stated minimum is 1.30, so the field is honored on every supported cluster.

## Testing

- Unit: retention-policy assertions folded into `TestCreateSearchStatefulSetFunc_DefaultAntiAffinity`.
- **e2e (`search_community_basic`)**: `test_zz_delete_search_cr_reclaims_pvc` deletes the `MongoDBSearch` CR and asserts owner-ref GC removes the mongot StatefulSet and `whenDeleted: Delete` reclaims its data PVC. A presence guard checks the StatefulSet + data PVC exist before delete so the absence assertion can't pass vacuously.
- The `whenScaled`/`whenDeleted` STS-teardown→PVC-reclaim path is additionally exercised end-to-end by the shard scale-down teardown e2e in #1277 (stacked above), and the multi-cluster CR-delete PVC reclaim by `e2e_mc_search_cr_deletion` in #1274.

### DIFF
--- a/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
+++ b/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
@@ -811,10 +811,12 @@ def wait_for_mongot_statefulset_drained(
 ) -> None:
     """Wait until the mongot StatefulSet has 0 ready replicas or is deleted.
 
-    The reconciler deletes the STS entirely when ``spec.clusters[].replicas`` drops to 0,
-    so a 404 from the API is the terminal state. ``api_client`` must target the
-    cluster that hosts the STS — for multi-cluster the mongot StatefulSets live
-    on the member clusters, not the operator/default cluster.
+    On ``spec.clusters[].replicas`` -> 0 the operator keeps the StatefulSet but
+    scales it to 0, so ``ready == desired == 0`` is the normal terminal state; a
+    404 is also accepted for callers that remove the STS outright (e.g. CR
+    delete). ``api_client`` must target the cluster that hosts the STS — for
+    multi-cluster the mongot StatefulSets live on the member clusters, not the
+    operator/default cluster.
     """
     apps_v1 = client.AppsV1Api(api_client=api_client)
 
--- a/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
+++ b/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
@@ -837,6 +837,43 @@ def wait_for_mongot_statefulset_drained(
     )
 
 
+def wait_for_mongot_pvcs_deleted(
+    namespace: str,
+    sts_name: str,
+    *,
+    api_client: Optional[client.ApiClient] = None,
+    timeout: int = 300,
+    sleep_time: int = 5,
+) -> None:
+    """Wait until the PVCs backing a mongot StatefulSet are fully deleted.
+
+    Under ``WhenScaled: Delete`` the STS controller GCs the per-ordinal claim on
+    scale-to-0, but the delete is asynchronous. Scaling back up while the old
+    claim is still ``Terminating`` trips a StatefulSet-controller race: the
+    controller skips pod creation for the deleting claim and is not re-enqueued
+    once it is gone, leaving the pod stuck at ``current: 0``. Block on the claim
+    being gone before restoring replicas. PVCs from a volumeClaimTemplate are
+    named ``<template>-<sts>-<ordinal>``, so we match on the ``-<sts>-`` infix.
+    ``api_client`` must target the cluster hosting the STS.
+    """
+    core = client.CoreV1Api(api_client=api_client)
+
+    def deleted() -> tuple[bool, str]:
+        remaining = [
+            pvc.metadata.name
+            for pvc in core.list_namespaced_persistent_volume_claim(namespace).items
+            if f"-{sts_name}-" in f"-{pvc.metadata.name}"
+        ]
+        return not remaining, ("all deleted" if not remaining else f"remaining={remaining}")
+
+    run_periodically(
+        deleted,
+        timeout=timeout,
+        sleep_time=sleep_time,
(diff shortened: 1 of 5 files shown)


docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -837,6 +839,43 @@ def drained() -> tuple[bool, str]:
...
+    api_client: Optional[client.ApiClient] = None,
+    timeout: int = 300,
+    sleep_time: int = 5,
+) -> None:
+    """Wait until the PVCs backing a mongot StatefulSet are fully deleted.
+
+    Under ``WhenScaled: Delete`` the STS controller GCs the per-ordinal claim on
+    scale-to-0, but the delete is asynchronous. Scaling back up while the old
+    claim is still ``Terminating`` trips a StatefulSet-controller race: the
+    controller skips pod creation for the deleting claim and is not re-enqueued
```
lsierant, line 855: can we handle this on the operator side? Or is the quick-timing issue in e2e test between scaledown and scaleup?

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -116,3 +119,67 @@ def test_search_create_search_index(sample_movies_helper: SampleMoviesSearchHelp
...
+    prefix = f"data-{sts_name}-"
+    return [
+        pvc.metadata.name
+        for pvc in core_v1.list_namespaced_persistent_volume_claim(namespace).items
+        if pvc.metadata.name.startswith(prefix)
+    ]
+
+
+@mark.e2e_search_community_basic
+def test_zy_scale_mongot_offline_reclaims_pvc(namespace: str, mdbs: MongoDBSearch):
```
lsierant, line 137: what's zy?

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -116,3 +119,67 @@ def test_search_create_search_index(sample_movies_helper: SampleMoviesSearchHelp
...
+    wait_for_mongot_pvcs_deleted(namespace, sts_name, timeout=300)
+
+    mdbs.load()
+    mdbs["spec"]["clusters"][0]["replicas"] = 1
+    mdbs.update()
+    mdbs.assert_reaches_phase(Phase.Running, timeout=600)
+
+
+@mark.e2e_search_community_basic
+def test_zz_delete_search_cr_reclaims_pvc(namespace: str, mdbs: MongoDBSearch):
```
lsierant, line 164: is zz a `zx + 1`?  :D

lsierant (approved):

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
anandsyncs: It's just a race issue in test.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: yep :D

PR #1274, opened 2026-06-25
search: sweep member-cluster resources on MongoDBSearch delete

anandsyncs:
[skip-ci]

# Based on PR #1273

* PR #1166:

  * PR #1048:

    * PR #1273:
      `search/ga-base` ← `search/pvc-retention-policy`

      * **PR #1274 (THIS ONE)**:
        `search/pvc-retention-policy` ← `search/mc-cleanup-ondelete`

## What

Adds multi-cluster CR-delete cleanup for `MongoDBSearch`, at **sharded-cluster parity**. Until now, deleting an MC search CR leaked everything in member clusters — owner references don't cross cluster boundaries, and there was no `OnDelete` handler.

A single `OnDelete` on the search controller, wired via the canonical `ResourceEventHandler` delete watch, sweeps the resources this search owns in every member cluster by the search-owner label selector (`mongodb.com/search-name` + `-namespace`), best-effort with error aggregation:

- mongot `StatefulSet`, its `Service`s, and `ConfigMap`
- the Envoy `Deployment` and `ConfigMap`

> Stacked on #1273 (PVC retention). Base will retarget to `search/ga-base` once #1273 merges.

## Design

- **One sweep, not two.** The Envoy resources carry the same search-owner labels, so the single sweep on the parent reconciler reclaims them too. No separate `Deleter` on the Envoy controller — that would be a second controller registering a delete handler on the same parent CR (mck-code rule 5/6). Verified the Envoy controller has no competing CR-delete handler, so there's no double-cleanup race.
- **Gated on `IsMultiClusterMode`.** Single-cluster installs rely on owner-reference GC and need nothing here — matching sharded/mdbmulti, where the explicit sweep is `IsMultiCluster()`-gated.
- **No finalizer (fire-and-forget).** Matches sharded/mdbmulti: if the operator is offline at delete time, member resources leak — the same accepted gap. Deliberate MVP choice.
- **Canonical mechanism.** `ResourceEventHandler{deleter: r}` is the exact delete-watch path used by sharded/replica-set/standalone controllers; the sweep mirrors `deleteClusterResources`. Search needs a local helper (not the shared `deleteClusterResources`) because of its distinct label scheme, the added `Deployment` kind, and to avoid an `operator → searchcontroller` import cycle.

## Not swept (by design)

- **Secrets** — the operator never writes member-cluster Secrets (customers replicate those).
- **PVCs** — reclaimed by the mongot StatefulSet's `whenDeleted: Delete` retention policy (#1273) as the swept StatefulSet goes. **This PR is what activates that latent MC PVC reclaim** I flagged on #1273: the sweep deletes the member StatefulSet, which cascades its PVC.

## Testing

- `searchcontroller`: `TestIsMultiClusterMode` (truth table), `TestCleanupMemberClusterResources` (multi-cluster sweep; asserts owned mongot + Envoy resources reclaimed across two clusters, while Secrets, a different search's resources, and unlabeled resources survive).
- `operator`: `TestMongoDBSearchOnDelete_MultiCluster_SweepsMembersAndDropsWatches`, `TestMongoDBSearchOnDelete_NoSpecClusters_SkipsSweep` (GC parity), `TestMongoDBSearchOnDelete_WrongType_ReturnsError`. Mutation-verified: dropping `Deployment` from the swept kinds fails the sweep test; removing the `IsMultiClusterMode` gate fails the skip test.
- **e2e (`e2e_mc_search_cr_deletion`, 2-cluster variant)**: deploys a centralized MC `MongoDBSearch`, asserts the full per-cluster resource set is present (owner-labelled mongot STS, headless + proxy Services, mongot ConfigMap, Envoy Deployment + ConfigMap, and the backing PVCs), deletes the CR, then polls until every one of those is gone in each member cluster (PVCs reclaimed via the `whenDeleted: Delete` cascade) and the central state ConfigMap is GC'd. The CR is re-created afterwards to prove the deletion left no wreckage. Presence is asserted before delete so the absence polling can't pass vacuously.
- `gofmt`/`go vet`/`golangci-lint-full` clean.


anandsyncs:
Superseded by the KUBE-154 lifecycle stack (#1403 → #1362 → #1382 → #1383 → #1393): the CR-delete member-cluster sweep now lives in #1382 as a label-scoped sweep with UID+ResourceVersion delete preconditions. Closing.

PR #1276, opened 2026-06-25
fix(ci): stabilize flaky e2e tests and speed up member watcher unit tests

nammn:
# Summary

Recent master CI runs showed several unrelated but repeatable e2e and unit-test flakes.

**Ops Manager AppDB validation (`e2e_om_appdb_validation`)**
The test was waiting on the generic MongoDB webhook endpoint, not the OpsManager one, and could assert rejection before the validating webhook was actually enforcing AppDB version rules. We now wait on `validate-mongodb-com-v1-mongodbopsmanager` and poll with a dry-run create until the webhook rejects invalid AppDB versions.

**TLS sharded-cluster cert rotation (`e2e_tls_sharded_cluster_certs_prefix`)**
`assert_abandons_phase(Running)` was racy: cert rotation does not always produce a visible phase departure from `Running`. We capture `last_transition` before the change and assert a state transition happened, then wait for `Running`. The `assert_state_transition_happens` calls in `test_rotate_tls_certificate` and `test_disable_tls` were missing an explicit timeout and fell back to the 600s default -- shorter than the 800s/1200s budgets on the subsequent `assert_reaches_phase` calls. Both are now aligned (800s and 1200s respectively).

**Standalone and sharded-cluster recovery (`e2e_standalone_recovery`, `e2e_sharded_cluster_scram_sha_and_x509`)**
`standalone_recovery.py` and `sharded_cluster_recovery.py` asserted `Phase.Failed` within 20s. The 20s value dated to 2019 when the ConfigMap/credentials were patched before resource creation. The current test structure patches after `mdb.update()`, so the operator reconciles the bad config only after the first good reconcile completes -- which takes longer than 20s on loaded CI hosts. Bumped to 100s, matching the convention already used by `replica_set_config_map.py` and `sharded_cluster_secret.py`.

**Kind cluster RBAC bootstrap race (`e2e_sharded_cluster_scram_sha_and_x509`)**
`kind create cluster --wait 700s` waits for node readiness but not for the apiserver's `rbac/bootstrap-roles` post-start hook. Under host load, `cluster-admin` ClusterRole may not exist yet when the operator's RBAC resources are applied, leaving pods unable to become ready. Added `kind_wait_for_rbac_bootstrap` to `setup_kind_cluster.sh` (called after `kind_wait_for_nodes_are_ready`), which polls until `cluster-admin` is available. Because `recreate_kind_clusters.sh` spawns `setup_kind_cluster.sh` as parallel subprocesses, this automatically covers multi-cluster setups.

**SCRAM SHA upgrade cleanup (`replica_set_scram_sha_upgrade`)**
SCRAM cleanup can take longer than 120s on loaded CI hosts. The delete step also no longer waits on Ops Manager deletion (`mongo_resource_deleted_no_om`) since this test does not use OM.

**IBM smoke startup (s390x / ppc64le)**
IBM runners can need longer image pulls before the test pod reaches `Running`. Startup wait is now configurable via `TEST_APP_STARTUP_TIMEOUT` (8m for IBM smoke), with richer pod/event logging on failure. IBM pipelines no longer download the multi-cluster binary, and the test image only installs it when the artifact is present in the build context.

**Member cluster health watcher unit tests**
Watcher tests spent several seconds in real `time.Sleep` / `assert.Never` windows even though logic completes on the first loop iteration. Tests now run inside `synctest.Test`, and `WatchMemberClusterHealth` sleeps with a cancellable `select` so the watcher goroutine exits cleanly when the test context is cancelled.

## Proof of Work
- ibm patch [passing](https://spruce.corp.mongodb.com/version/6a3d3d749e7df4000759e27d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
- normal ci run part of this patch + a bunch of manual patches all were green

## Checklist

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -440,9 +440,6 @@ tasks:
       - func: setup_building_host_minikube
         vars:
           skip_minikube_setup: true
-      - func: download_multi_cluster_binary
-        vars:
-          platform: linux/s390x
       - func: quay_staging_login
       - func: pipeline
         vars:
@@ -454,9 +451,6 @@ tasks:
       - func: setup_building_host_minikube
         vars:
           skip_minikube_setup: true
-      - func: download_multi_cluster_binary
-        vars:
-          platform: linux/ppc64le
       - func: quay_staging_login
       - func: pipeline
         vars:
--- a/docker/mongodb-kubernetes-tests/Dockerfile
+++ b/docker/mongodb-kubernetes-tests/Dockerfile
@@ -88,4 +88,6 @@ COPY release.json /tests/release.json
 # we use the public directory to automatically test resources samples
 COPY public /mongodb-kubernetes/public
 
-ADD "docker/mongodb-kubernetes-tests/multi-cluster-kube-config-creator_${TARGETARCH}" /usr/local/bin/multi-cluster-kube-config-creator
+RUN if [ -f "/tests/multi-cluster-kube-config-creator_${TARGETARCH}" ]; then \
+    install -m 755 "/tests/multi-cluster-kube-config-creator_${TARGETARCH}" /usr/local/bin/multi-cluster-kube-config-creator; \
+    fi
--- a/docker/mongodb-kubernetes-tests/kubeobject/customobject.py
+++ b/docker/mongodb-kubernetes-tests/kubeobject/customobject.py
@@ -96,16 +96,19 @@ class CustomObject:
         self._register_updated()
         return self
 
-    def create(self) -> CustomObject:
+    def create(self, dry_run: str = None) -> CustomObject:
         """Creates this object in Kubernetes."""
+        kwargs = {"field_validation": "Strict"}
+        if dry_run:
+            kwargs["dry_run"] = dry_run
         obj = self.api.create_namespaced_custom_object(
-            self.group, self.version, self.namespace, self.plural, self.backing_obj, field_validation="Strict"
+            self.group, self.version, self.namespace, self.plural, self.backing_obj, **kwargs
         )
 
-        self.backing_obj = obj
-        self.bound = True
-
-        self._register_updated()
+        if not dry_run:
+            self.backing_obj = obj
+            self.bound = True
+            self._register_updated()
         return self
 
     def update(self) -> CustomObject:
--- a/docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_upgrade.py
+++ b/docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_upgrade.py
@@ -38,7 +38,7 @@ class TestReplicaSetDeleted(KubernetesTester):
       Deletes the Replica Set.
     delete:
       file: replica-set-scram.yaml
-      wait_until: mongo_resource_deleted
+      wait_until: mongo_resource_deleted_no_om
       timeout: 120
     """
 
--- a/docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_upgrade.py
+++ b/docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_upgrade.py
@@ -51,4 +51,4 @@ class TestReplicaSetDeleted(KubernetesTester):
             except AssertionError:
                 return False
 
-        KubernetesTester.wait_until(authentication_was_disabled, timeout=120, sleep_time=5)
+        KubernetesTester.wait_until(authentication_was_disabled, timeout=300, sleep_time=5)
--- a/docker/mongodb-kubernetes-tests/tests/opsmanager/kubernetes_tester/om_appdb_validation.py
+++ b/docker/mongodb-kubernetes-tests/tests/opsmanager/kubernetes_tester/om_appdb_validation.py
@@ -2,7 +2,7 @@ from typing import Optional
 
 import pytest
 from kubernetes.client import ApiException
-from kubetester import wait_for_webhook
+from kubetester import wait_for_webhook, wait_until
 from kubetester.kubetester import KubernetesTester
 from kubetester.kubetester import fixture as yaml_fixture
 from kubetester.opsmanager import MongoDBOpsManager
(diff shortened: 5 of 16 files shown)


docker/mongodb-kubernetes-tests/tests/opsmanager/kubernetes_tester/om_appdb_validation.py
```diff
@@ -17,7 +17,7 @@ def om_resource(namespace: str) -> MongoDBOpsManager:
 
 @pytest.mark.e2e_om_appdb_validation
 def test_wait_for_webhook(namespace: str):
-    wait_for_webhook(namespace)
+    wait_for_webhook(namespace, validation_endpoint="validate-mongodb-com-v1-mongodbopsmanager")
```
nammn, line 20: we are testing the om webhook!

docker/mongodb-kubernetes-tests/tests/tls/tls_sharded_cluster_certs_prefix.py
```diff
@@ -101,12 +101,14 @@ def test_sharded_cluster_has_no_connectivity_without_tls(sc: MongoDB):
...
 @mark.e2e_tls_sharded_cluster_certs_prefix
 def test_rotate_tls_certificate(sc: MongoDB, namespace: str):
+    last_transition = sc.get_status_last_transition_time()
+
     # update the shard cert
     cert = Certificate(name=f"prefix-{MDB_RESOURCE}-0-cert", namespace=namespace).load()
     cert["spec"]["dnsNames"].append("foo")
     cert.update()
 
-    sc.assert_abandons_phase(Phase.Running)
```
nammn, line 109: this is racy!

scripts/dev/contexts/variables/e2e_ibm_smoke_base
```diff
@@ -8,3 +8,4 @@ export CUSTOM_MDB_VERSION=7.0.22-ent # we only have ibm static images starting w
 export KUBE_ENVIRONMENT_NAME=minikube
 export CLUSTER_TYPE=minikube
 export CONTAINER_RUNTIME=podman
+export TEST_APP_STARTUP_TIMEOUT=8m
```
nammn, line 11: ibm needs more time than 2m a lot of times!

docker/mongodb-kubernetes-tests/tests/authentication/replica_set_scram_sha_upgrade.py
```diff
@@ -38,7 +38,7 @@ class TestReplicaSetDeleted(KubernetesTester):
       Deletes the Replica Set.
     delete:
       file: replica-set-scram.yaml
-      wait_until: mongo_resource_deleted
+      wait_until: mongo_resource_deleted_no_om
```
nammn, line 41: its flaky as well, let's not rely on om here

docker/mongodb-kubernetes-tests/Dockerfile
```diff
@@ -88,4 +88,6 @@ COPY release.json /tests/release.json
 # we use the public directory to automatically test resources samples
 COPY public /mongodb-kubernetes/public
 
-ADD "docker/mongodb-kubernetes-tests/multi-cluster-kube-config-creator_${TARGETARCH}" /usr/local/bin/multi-cluster-kube-config-creator
+RUN if [ -f "/tests/multi-cluster-kube-config-creator_${TARGETARCH}" ]; then \
```
nammn, line 91: we don't need this file in multi arch tests, so lets only run this if it exists. So we can remove the related ibm downloads

pkg/multicluster/memberwatch/clusterhealth_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"testing"
+	"testing/synctest"
```
nammn, line 9: memberwatch does a 10s sleep, this can slow down unit tests. I've wrapped them

pkg/multicluster/memberwatch/memberwatch.go
```diff
@@ -167,7 +167,11 @@ func (m *MemberClusterHealthChecker) WatchMemberClusterHealth(ctx context.Contex
 				}
 			}
 		}
-		time.Sleep(10 * time.Second)
+		select {
+		case <-ctx.Done():
+			return
```
nammn, line 172: for faster stop/cleanup

docker/mongodb-kubernetes-tests/tests/opsmanager/kubernetes_tester/om_appdb_validation.py
```diff
@@ -17,14 +17,26 @@ def om_resource(namespace: str) -> MongoDBOpsManager:
...
 
 
 @pytest.mark.e2e_om_appdb_validation
 def test_wrong_appdb_version(namespace: str, custom_version: Optional[str]):
     om = om_resource(namespace)
     om["spec"]["applicationDatabase"]["version"] = "3.6.12"
     om.set_version(custom_version)
+
+    def webhook_enforcing() -> bool:
```
nammn, line 29: https://knative.dev/development/serving/dryrun/

.evergreen.yml
```diff
@@ -440,9 +440,6 @@ tasks:
       - func: setup_building_host_minikube
         vars:
           skip_minikube_setup: true
-      - func: download_multi_cluster_binary
```
nammn, line 443: we don't need that binary

docker/mongodb-kubernetes-tests/kubeobject/customobject.py
```diff
@@ -96,16 +96,19 @@ def load(self) -> Self:
         self._register_updated()
         return self
 
-    def create(self) -> CustomObject:
+    def create(self, dry_run: str = None) -> CustomObject:
```
nammn, line 99: dry run for webhooks

docker/mongodb-kubernetes-tests/tests/opsmanager/kubernetes_tester/om_appdb_validation.py
nammn: we probably should have a common function for this and do it the same for every webhook test 😓

MaciejKaras (approved):
Amazing!

I see only couple of tests being flaky in this whole PR 🚀

josvazg (approved):

PR #1281, opened 2026-06-26
search: block single-operator reconcile of multi-cluster MongoDBSearch (feature-flagged)

lsierant:
[skip-ci]

# Summary

A single operator cannot manage a multi-cluster MongoDBSearch deployment, so it must not act on such a spec. This adds a fast pre-narrow validation gate that rejects it with an informative `Invalid` status before any per-cluster narrowing or source handling.

- Blocks when the operator has **no** `operatorClusterName` **and** `len(spec.clusters) > 1`.
- Per-cluster operators (`operatorClusterName` set) narrow to their own entry via `LocalizeToCluster` and are unaffected; a single-cluster spec is always allowed.
- Gated by the **undocumented** operator env `MDB_SEARCH_ENABLE_MULTI_CLUSTER`; default `false` = block on, so an accidentally-set or zero-value bool fails safe to blocking. The status message never references the flag.
- Helm emits the env only when defined and not `"false"`; existing search suites set it `true` via the dev context, so they are unaffected.

## Proof of Work

Rebased onto `search/ga-base` @ `532b1efa9`. Local `make precommit` + `make all-tests` green (only the unrelated host `wt_ctl` network-registry tests fail, sensitive to the active worktree's registry slot).

**New block e2e — `e2e_search_block_multi_cluster`** (in the `e2e_mdb_kind_ubi_cloudqa_search_block` variant, block on): applies a two-cluster `spec.clusters` to a single-operator MongoDBSearch (RS + sharded sources) and asserts it reaches `Phase.Failed` with the "not supported yet" status — i.e. the block fires.

- The block variant now also runs **on PR patches** (`pr_patch` tag added), and the block-on regression check is trimmed to a **3-test single-cluster smoke subset** (community + replicaset + sharded sources) instead of re-running the full suite, to cut cost.
- PoW-Patch (merged block task group): https://spruce.mongodb.com/version/6a42924bd5bb4100078a9c4b — **16/17**. The merged `e2e_mdb_kind_search_block_task_group` ran cleanly: the block assertion `e2e_search_block_multi_cluster` + the replicaset and sharded smoke tasks all passed. The lone failure is again `e2e_search_community_basic :: test_zz_delete_search_cr_reclaims_pvc` (300s PVC-reclaim timeout) — a base-branch test added by KUBE-143 (#1273), unrelated to the block (gate never fires for single-cluster).
- Prior runs (pre-rename): PoW green https://spruce.mongodb.com/version/6a419232ca864f00073e9bb7 (14/14); full search sweep 159/162 https://spruce.mongodb.com/version/6a41924fa369580007bc4ed8 (3 persistent failures all pre-existing/environmental — x509 `SearchNotEnabled` quiesce race + the known KUBE-124/PR #1215 onboarding-no-outage PodSecurity failure; block off and gate never fired in all of them).
- Unit tests cover: single-operator + >1 clusters blocked (flag unset/false), single cluster allowed, per-cluster operator allowed, flag-true allowed.

## Checklist

---

### DIFF
--- a/controllers/operator/mongodbsearch_controller.go
+++ b/controllers/operator/mongodbsearch_controller.go
@@ -54,6 +54,12 @@ type prepareSearchFunc func(search *searchv1.MongoDBSearch, log *zap.SugaredLogg
 // misconfigured MC specs.
 func newPrepareSearch(operatorClusterName string) prepareSearchFunc {
 	validateSpec := func(search *searchv1.MongoDBSearch, log *zap.SugaredLogger, writeStatus func(workflow.Status) (reconcile.Result, error)) (bool, reconcile.Result, error) {
+		// A single operator (no operatorClusterName) cannot manage a multi-cluster (>1)
+		// search deployment; per-cluster operators narrow to their own entry below.
+		if operatorClusterName == "" && len(search.Spec.Clusters) > 1 && !env.ReadBoolOrDefault(util.SearchEnableMultiClusterEnv, false) { // nolint:forbidigo
+			r, e := writeStatus(workflow.Invalid("multi-cluster MongoDBSearch is not supported yet: spec.clusters must contain a single entry"))
+			return true, r, e
+		}
 		if vErr := search.ValidateSpec(); vErr != nil {
 			r, e := writeStatus(workflow.Invalid("%s", vErr.Error()))
 			return true, r, e
--- a/.evergreen-tasks.yml
+++ b/.evergreen-tasks.yml
@@ -1435,6 +1435,11 @@ tasks:
     commands:
       - func: "e2e_test"
 
+  - name: e2e_search_block_multi_cluster
+    tags: [ "patch-run" ]
+    commands:
+      - func: "e2e_test"
+
   - name: e2e_search_sharded_routed_from_another_shard
     tags: [ "patch-run" ]
     commands:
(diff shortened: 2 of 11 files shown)


controllers/operator/mongodbsearch_controller.go
```diff
@@ -54,6 +54,12 @@ type prepareSearchFunc func(search *searchv1.MongoDBSearch, log *zap.SugaredLogg
 // misconfigured MC specs.
 func newPrepareSearch(operatorClusterName string) prepareSearchFunc {
 	validateSpec := func(search *searchv1.MongoDBSearch, log *zap.SugaredLogger, writeStatus func(workflow.Status) (reconcile.Result, error)) (bool, reconcile.Result, error) {
+		// A single operator (no operatorClusterName) cannot manage a multi-cluster (>1)
+		// search deployment; per-cluster operators narrow to their own entry below.
+		if operatorClusterName == "" && len(search.Spec.Clusters) > 1 && env.ReadBoolOrDefault(util.SearchDisableMultiClusterEnv, true) { // nolint:forbidigo
```
anandsyncs: nit: This flag can be `SearchEnableMultiClusterEnv, false`.
Also because default boolean value is false if the var itself is configured somewhere accidently.

anandsyncs (approved):

Julien-Ben (approved):
Both the logic and the testing look good to me !
Looking at the evergreen config, I just don't fully understand how much duplication we are adding to tests. From what I understand we run the full single cluster search tasks twice ?
But only when merging to master ?

PR #1286, opened 2026-06-29
fix(e2e): bump multi-cluster deletion timeouts and fix broken import/lint

nammn:
### Multi-cluster deletion tests timed out because the test polled faster than the operator could delete

The operator's `OnDelete` handler calls `WaitForReadyState` (90s+ polling) before reaching `deleteClusterResources` to remove StatefulSets. In multi-cluster, cross-cluster agent propagation adds further latency. Tests with 60s polling windows consistently expired before the operator reached the deletion step. Single-cluster counterparts did not fail because there is no cross-cluster propagation.

Bumped poll timeouts from 60s→300s in all multi-cluster deletion/cleanup tests (StatefulSets, Services, ConfigMaps, Secrets, CR resource deletion).

### Infinite poll loops in test infrastructure now have bounded timeouts

`get_pod_when_running` (in `__init__.py`) and `wait_for_condition_stateful_set` (in `kubetester.py`) both had infinite `while True` loops with no timeout. A pod that never became Ready would hang until Evergreen's 2h task executor killed it, producing a less useful failure than an explicit timeout error. Added 600s timeouts with descriptive error messages that include the last-known status.

### CRD replace now retries on HTTP 409 conflicts

`create_or_replace_from_yaml.py` read the CRD's `resourceVersion`, then called replace — but the version could change between read and write, producing a 409 conflict. The patch path already had retry logic for this; the replace path did not. Added a 300s deadline retry loop for the replace path matching the existing patch pattern.

### Community operator e2e: removed `-failfast`

`-failfast` combined with `retry_on_failure` masked which subtests actually failed. Removing it lets all subtests report results independently, so the test pod's exit code and logs reflect every failure, not just the first one.

### Flaky test reporting improvements

- Renamed `get_honeycomb_headers`→`_honeycomb_headers` (internal convention). The test file kept the old import name, causing `ImportError` in CI (`unit_tests_python` task). Fixed.
- Added `total_runs` count to the `FlakyTask` dataclass so the output shows run count alongside flakiness percentage.
- Master failure notifications now annotate each failed task with its known flakiness percentage from the past 7 days (via `get_flaky_tasks`).

### Evergreen API URL routing

Local development uses `evergreen.corp.mongodb.com` (OIDC auth works there), CI uses `evergreen.mongodb.com`. Detection via the `EVR_TASK_ID` env var, which is injected by Evergreen at task runtime and absent locally.

## Proof of Work

- `multi_cluster_replica_set_deletion.py`: StatefulSets still existed 88s after CR deletion, 60s polling timed out
- `sharded_cluster.py` / `sharded_cluster_pv.py`: operator logs confirm `WaitForReadyState` blocks deletion for 90s+

new merged master slack output:
```
BUILD FAILED: 6a429f059bf8b00007a37ee6
Commit:  4d65e5e2 by nam.nguyen
Link:    https://spruce.mongodb.com/version/6a429f059bf8b00007a37ee6

Failed Tasks (29):
Variant                                       Task
---------------------------------------------------------------------------------------------------------
e2e_mdb_kind_ubi_cloudqa                      e2e_disable_tls_and_scale (flaky: 6% over 7d)
e2e_mdb_kind_ubi_cloudqa                      e2e_replica_set_scram_sha_1_upgrade (flaky: 21% over 7d)
```
## Checklist

    - `skip-changelog` label applied (infrastructure/test-only changes, no user-facing behavior)

### DIFF
--- a/docker/mongodb-kubernetes-tests/kubetester/__init__.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/__init__.py
@@ -425,12 +425,14 @@ def get_pod_when_running(
     namespace: str,
     label_selector: str,
     api_client: Optional[kubernetes.client.ApiClient] = None,
+    timeout: int = 600,
 ) -> client.V1Pod:
     """
     Returns a Pod that matches label_selector. It will block until the Pod is in
-    Running state.
+    Running state or timeout is reached.
     """
-    while True:
+    deadline = time.time() + timeout
+    while time.time() < deadline:
         time.sleep(3)
 
         try:
@@ -447,6 +449,10 @@ def get_pod_when_running(
             # The Pod might not exist in Kubernetes yet so skip any 404
             if e.status != 404:
                 raise
+
+    raise Exception(
+        f"Timeout ({timeout}s) waiting for pod with label_selector '{label_selector}' in namespace '{namespace}' to be Running"
+    )
 
 
 def get_pod_when_ready(
--- a/docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
@@ -78,7 +78,7 @@ def patch_from_yaml_single_item(k8s_client, yml_object, namespace="default", **k
                     time.sleep(1)
                     continue
                 raise
-        raise Exception(f"Failed to replace CRD {name} after 30s: {last_error}")
+        raise Exception(f"Failed to replace CRD {name} after 300s: {last_error}")
 
     method = "patch"
     namespaced = hasattr(k8s_api, "{}_namespaced_{}".format("create", kind))
--- a/docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
@@ -54,7 +54,7 @@ def create_or_replace_from_yaml_single_item(k8s_client, yml_object, namespace="d
             pass
 
 
-def patch_from_yaml_single_item(k8s_client, yml_object, namespace="default", **kwargs):
+def patch_from_yaml_single_item(k8s_client, yml_object, namespace="default", timeout=300, **kwargs):
     k8s_api = get_k8s_api(k8s_client, yml_object)
     kind = get_kind(yml_object)
     # Decide which namespace we are going to put the object in,
@@ -64,7 +64,7 @@ def patch_from_yaml_single_item(k8s_client, yml_object, namespace="default", **k
     name = yml_object["metadata"]["name"]
 
     if kind == "custom_resource_definition":
-        deadline = time.time() + 300
+        deadline = time.time() + timeout
         last_error = None
         while time.time() < deadline:
             resource = client.ApiextensionsV1Api().read_custom_resource_definition(name)
--- a/docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import re
+import time
 from os import path
 
 import yaml
@@ -10,7 +11,7 @@ from kubernetes.utils.create_from_yaml import create_from_yaml_single_item
 
 """
 This is a modification of 'create_from_yaml.py' from python kubernetes client library
-It allows to mimic the 'kubectl apply' operation on yaml file. It performs either create or 
+It allows to mimic the 'kubectl apply' operation on yaml file. It performs either create or
 patch on each individual object.
 """
 
@@ -62,16 +63,24 @@ def patch_from_yaml_single_item(k8s_client, yml_object, namespace="default", **k
         namespace = yml_object["metadata"]["namespace"]
     name = yml_object["metadata"]["name"]
 
-    method = "patch"
     if kind == "custom_resource_definition":
-        # fetching the old CRD to make the replace working (has conflict resolution based on 'resourceVersion')
-        # TODO this is prone to race conditions - we need to either loop or use patch with json merge
-        # see https://github.com/helm/helm/pull/6092/files#diff-a483d6c0863082c3df21f4aad513afe2R663
-        resource = client.ApiextensionsV1Api().read_custom_resource_definition(name)
+        deadline = time.time() + 300
+        last_error = None
+        while time.time() < deadline:
+            resource = client.ApiextensionsV1Api().read_custom_resource_definition(name)
+            yml_object["metadata"]["resourceVersion"] = resource.metadata.resource_version
+            try:
+                k8s_api.replace_custom_resource_definition(body=yml_object, name=name, **kwargs)
+                return
+            except client.rest.ApiException as e:
+                if e.status == 409:
+                    last_error = e
+                    time.sleep(1)
+                    continue
+                raise
+        raise Exception(f"Failed to replace CRD {name} after 30s: {last_error}")
 
-        yml_object["metadata"]["resourceVersion"] = resource.metadata.resource_version
-        method = "replace"
-
+    method = "patch"
     namespaced = hasattr(k8s_api, "{}_namespaced_{}".format("create", kind))
     url_path = get_url_path(namespaced, method)
 
--- a/docker/mongodb-kubernetes-tests/kubetester/kubetester.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/kubetester.py
@@ -807,20 +807,26 @@ class KubernetesTester(object):
         return cls.wait_for_condition_stateful_set(cls.get_namespace(), name, attribute, expected_value)
 
     @classmethod
-    def wait_for_condition_stateful_set(cls, namespace, name, attribute, expected_value):
+    def wait_for_condition_stateful_set(cls, namespace, name, attribute, expected_value, timeout=600):
         appsv1 = KubernetesTester.clients("appsv1")
-        ready_to_go = False
-        while not ready_to_go:
+        deadline = time.time() + timeout
+        last_status = None
+        while time.time() < deadline:
             try:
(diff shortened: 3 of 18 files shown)


scripts/evergreen/e2e/single_e2e.sh
```diff
@@ -245,7 +245,7 @@ run_tests() {
     echo
 
     # We need to make sure to access this file after the test has finished
-    kubectl --context "${test_pod_context}" -n "${NAMESPACE}" -c keepalive cp "${TEST_APP_PODNAME}":/tmp/results/myreport.xml logs/myreport.xml
+    kubectl --context "${test_pod_context}" -n "${NAMESPACE}" -c keepalive cp "${TEST_APP_PODNAME}":/tmp/results/myreport.xml logs/myreport.xml || true
```
nammn, line 248: without || true, the script fails an doesn't even upload logs

scripts/funcs/install
```diff
@@ -37,6 +37,24 @@ detect_architecture() {
       esac
 }
 
+kubectl_apply_retry() {
```
nammn, line 40: we apply remote files, we need retries as we can get rate limited/bad network etc.

scripts/dev/install_csi_driver.sh
```diff
@@ -17,7 +18,7 @@ csi_driver_download() {
 
     # Download the tar.gz file
     echo "Downloading ${REPO_URL}..."
-    curl -L -o "${TAR_FILE}" "${REPO_URL}"
+    curl_with_retry -L -o "${TAR_FILE}" "${REPO_URL}"
```
nammn, line 21: all curls occasionally fail

nammn: this one failed in a random patch

nammn:

scripts/evergreen/e2e/setup_cloud_qa.py
```diff
@@ -333,7 +333,7 @@ def clean_unused_keys(org_id: str):
     - currently used by the script (USER_OWNER env variable)
     - containing "EVG" or "NOT_DELETE" in their description
     """
-    keys, newer_keys = get_keys_older_than(org_id, minutes_interval=70)
+    keys, newer_keys = get_keys_older_than(org_id, minutes_interval=140)
```
nammn, line 336: some tests need more than 70 minutes, which causes sometimes teardown to remove the api key of a running test...

docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
```diff
@@ -62,16 +63,24 @@ def patch_from_yaml_single_item(k8s_client, yml_object, namespace="default", **k
...
+            try:
+                k8s_api.replace_custom_resource_definition(body=yml_object, name=name, **kwargs)
+                return
+            except client.rest.ApiException as e:
+                if e.status == 409:
+                    last_error = e
+                    time.sleep(1)
+                    continue
+                raise
+        raise Exception(f"Failed to replace CRD {name} after 30s: {last_error}")
```
josvazg: > docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py:66-83
fix/correct: The deadline is time.time() + 300 (5 min) but the error message says "after 30s" — and the PR description also claims "a 30s deadline retry loop."

scripts/evergreen/notify_master_failures.py
```diff
@@ -346,16 +378,28 @@ def main() -> None:
         print(f"Error fetching tasks: {e}", file=sys.stderr)
         sys.exit(1)
 
+    # Look up flakiness when there are failures (best-effort, non-fatal).
+    flaky_map: dict[tuple[str, str], float] = {}
+    if failed_tasks:
+        print("Looking up flakiness for failed tasks...", file=sys.stderr)
+        try:
+            flaky_tasks_list = get_flaky_tasks()
```
josvazg, line 386: > q/fix: notify_master_failures is careful to send all diagnostics to stderr so that dry-run's json.dumps(message) is the only thing on stdout (line 398). But the newly-imported get_flaky_tasks() does plain print("Querying flaky tasks...") / print("  Found N ...") to stdout. In --dry-run this corrupts the JSON payload for anything parsing stdout. Either route those prints to stderr or suppress them in the master-failures path.

scripts/python/evergreen_api.py
```diff
@@ -4,6 +4,8 @@
 from evergreen.api import EvergreenApi, EvgAuth
 
 EVERGREEN_API = "https://evergreen.mongodb.com/api"
+if not os.environ.get("EVR_TASK_ID"):
+    EVERGREEN_API = "https://evergreen.corp.mongodb.com/api"
```
josvazg, line 10: > q: The URL is swapped at import time based on EVR_TASK_ID. Any importer picks up the corp URL if that env var is absent, even non-Evergreen tooling. Works, but a module-level import side-effect keyed on an env var is easy to trip over later — worth a comment right there explaining the local-vs-CI intent (it's in the docstring of get_evergreen_api, but the branch that matters is 20 lines up).

scripts/evergreen/notify_flaky_tests.py
```diff
@@ -22,95 +22,82 @@
...
-        "X-Honeycomb-Team": api_key,
-        "Content-Type": "application/json",
-    }
+    return {"X-Honeycomb-Team": api_key, "Content-Type": "application/json"}
 
 
-def create_and_run_query(headers: dict, query_spec: dict, dataset: str = DATASET) -> dict[str, Any]:
-    """Create a query, run it, and return the results."""
-    # Create query
+def _honeycomb_query(query_spec: dict, dataset: str = DATASET) -> dict:
```
josvazg, line 42: > nitpick: _honeycomb_headers() is now re-invoked on every request (3× per query). Harmless, but caching it once at the top of the function reads cleaner and re-reads the env var only once.

docker/mongodb-kubernetes-tests/kubetester/__init__.py
josvazg, line 506: > side-note: list_matching_pods / pod_is_ready / wait_for_pods_ready / wait_for_no_pods_ready are a sizable well-tested-looking addition but I don't see them referenced by the deletion-timeout story in the PR summary. If they're used by the search e2e work they belong there; if they're unused dead code on this branch, drop them or note where they're consumed.

josvazg (approved):
LGTM, with minor comments

scripts/evergreen/notify_flaky_tests.py
nammn: done

scripts/python/evergreen_api.py
nammn: done

docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
nammn: done

docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
```diff
@@ -62,16 +63,24 @@ def patch_from_yaml_single_item(k8s_client, yml_object, namespace="default", **k
...
-    method = "patch"
     if kind == "custom_resource_definition":
-        # fetching the old CRD to make the replace working (has conflict resolution based on 'resourceVersion')
-        # TODO this is prone to race conditions - we need to either loop or use patch with json merge
-        # see https://github.com/helm/helm/pull/6092/files#diff-a483d6c0863082c3df21f4aad513afe2R663
-        resource = client.ApiextensionsV1Api().read_custom_resource_definition(name)
-
-        yml_object["metadata"]["resourceVersion"] = resource.metadata.resource_version
-        method = "replace"
+        deadline = time.time() + 300
```
MaciejKaras: nit: can we add this `300` timeout as default param in the function itself? This would be aligned with other functions and keep the it configurable

docker/mongodb-kubernetes-tests/kubetester/__init__.py
```diff
@@ -425,12 +425,14 @@ def get_pod_when_running(
     namespace: str,
     label_selector: str,
     api_client: Optional[kubernetes.client.ApiClient] = None,
+    timeout: int = 600,
```
MaciejKaras, line 428: do we need such big timeout here?

docker/mongodb-kubernetes-tests/kubetester/kubetester.py
```diff
@@ -807,21 +807,27 @@ def wait_for_condition_string(cls, condition):
         return cls.wait_for_condition_stateful_set(cls.get_namespace(), name, attribute, expected_value)
 
     @classmethod
-    def wait_for_condition_stateful_set(cls, namespace, name, attribute, expected_value):
+    def wait_for_condition_stateful_set(cls, namespace, name, attribute, expected_value, timeout=600):
```
MaciejKaras, line 810: can we have lower value here i.e. 300 like in other places?

docker/mongodb-kubernetes-tests/kubetester/__init__.py
nammn: hmm fair 300 should suffice

nammn: but then I don't worry much about 5m vs 10m in this place

docker/mongodb-kubernetes-tests/kubetester/kubetester.py
nammn: for now I would like to use a longer one, since we didn't even had one in the past. I want to get some metrics in before deciding for a smaller one

docker/mongodb-kubernetes-tests/kubetester/create_or_replace_from_yaml.py
nammn: ah ahh yeah

nammn: done!

MaciejKaras (approved):
LGTM!

PR #1301, opened 2026-06-30
Only run e2e tests on PRs when production code is touched

josvazg:
# Summary

Split `pr_patch` into **static** and **e2e**. The e2e even is only triggered when the go files un production dirs are touched.

## Proof of Work

CI passes.
✅ [Patch with pr_patch_static skips e2e tests](https://spruce.corp.mongodb.com/version/6a43a003e23a2600071f537f/tasks)
✅ [Patch with both PR patch aliases includes e2e tests](https://spruce.corp.mongodb.com/patch/6a43a2289e8efc0007cec79f)

## Checklist

### DIFF
--- a/ci/internal/evergreen/aliases_test.go
+++ b/ci/internal/evergreen/aliases_test.go
@@ -1,0 +1,164 @@
+package evergreen_test
+
+import (
+	"os"
+	"regexp"
+	"testing"
+
+	"gopkg.in/yaml.v3"
+)
+
+func TestE2EFilePattern(t *testing.T) {
+	re := loadE2EFilePattern(t)
+
+	tests := []struct {
+		desc        string
+		files       []string
+		shouldMatch bool
+	}{
+		// --- should trigger e2e ---
+		{
+			desc:        "root-level Go file",
+			files:       []string{"main.go"},
+			shouldMatch: true,
+		},
+		{
+			desc:        "root-level Python file",
+			files:       []string{"generate_ssdlc_report_test.py"},
+			shouldMatch: true,
+		},
+		{
+			desc:        "api package",
+			files:       []string{"api/v1/types.go"},
+			shouldMatch: true,
+		},
+		{
+			desc:        "controllers package",
+			files:       []string{"controllers/operator/foo.go"},
+			shouldMatch: true,
+		},
+		{
+			desc:        "internal package",
+			files:       []string{"internal/util/foo.go"},
+			shouldMatch: true,
+		},
+		{
+			desc:        "pkg package",
+			files:       []string{"pkg/kube/client.go"},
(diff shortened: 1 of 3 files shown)


josvazg:
@MaciejKaras @nammn Claude tells me I must merge this before we can really test it, as it is still working with the evergreen.yml in `master`. But still I am waiting of the frozen branch to be lifted before proceeding with this. In teh meantime, what do you think of this change?

Please, ignore the **go.mod** and **rbac** related changes, I am not sure why I needed to add those unless `master` is broken right now.

MaciejKaras:
@josvazg I would rather wait for the release to finish. It should happen today, right?

josvazg:
> @josvazg I would rather wait for the release to finish. It should happen today, right?

Yes, no rush, but still I want your opinion.

This will not be merged until after the unfreeze.

ci/internal/evergreen/aliases_test.go
```diff
@@ -0,0 +1,154 @@
...
+				return regexp.MustCompile(entry.FilePattern)
+			}
+		}
+	}
+	t.Fatal("no pr_patch_e2e entry found in github_pr_aliases")
+	return nil
+}
+
+func TestE2EFilePattern(t *testing.T) {
+	re := loadE2EFilePattern(t)
```
nammn, line 12: we should also trigger e2e for any python code (e.g. fixing or adding tests)?

josvazg: you mean for flakiness fighting where the issue is in the test and not in the production code?
yes, but I think that might make the regexp too complex, maybe we should think about allow it to be decided from the PR tags.

Let me check options

josvazg: BTW this change touches no production and is unable to go green, never ever

nammn: not only for flakiness fighting, but also if you add a new test or change an existing test

josvazg: Done

ci/internal/evergreen/aliases_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	tests := []struct {
+		desc        string
+		files       []string
+		shouldMatch bool
+	}{
+		// --- should trigger e2e ---
+		{
```
nammn, line 20: can we add a test for root-level python file?

josvazg: sure!

josvazg: Done

nammn (approved):

roothorp (approved):

PR #1305, opened 2026-06-30
search: allow shared managed-LB hostname across clusters (cross-AZ failover)

lsierant:
search: allow shared managed-LB hostname across clusters (cross-AZ failover)

## Summary
Relaxes the multi-cluster MongoDBSearch managed-LB validation so a given shard can use the **same SNI hostname in every cluster**.

- `validateMCExternalHostnames`: dropped the cross-cluster *distinct-hostname* check; kept the `{shardName}` placeholder requirement and the managed-LB guard.
- `validateMCRouterHostnames`: this function was *only* the cross-cluster distinct check, so it's removed entirely (its call site too). Per-field routerHostname rules remain in `validateRouterHostname`.

## Why
A cross-AZ search **failover** proxy is an L4 SNI-passthrough Envoy that fronts multiple clusters' managed Envoys behind one hostname. Because L4 can't rewrite SNI, the preserved client SNI must match whichever AZ the connection lands on — i.e. the shard's hostname has to be identical across clusters. The old distinct-hostname rule forbade exactly that. Each cluster still runs its own Envoy, so a shared SNI `server_name` does not collide within a cluster.

## Proof of Work
- PoW-Patch (unit + lint): **green** — https://spruce.mongodb.com/version/6a43a79337a14b0007188ead
- Operator image (`build_operator_ubi`): green — https://spruce.mongodb.com/version/6a43a5606b27e8000731add6
- Local: `golangci-lint` (full + fmt) + `govulncheck` clean on the change; `go build ./...` + `go test ./api/mongodb/v1/search/... ./controllers/searchcontroller/...` green.

## Checklist

### DIFF
--- a/api/mongodb/v1/search/mongodbsearch_validation.go
+++ b/api/mongodb/v1/search/mongodbsearch_validation.go
@@ -106,7 +106,6 @@ func multiClusterValidators() []func(*MongoDBSearch) v1.ValidationResult {
 		validateMCRequiresExternalSource,
 		validateMCRequiresManagedLB,
 		validateMCExternalHostnames,
-		validateMCRouterHostnames,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


viveksinghggits (approved):

anandsyncs (approved):

viveksinghggits:
evergreen retry

PR #1306, opened 2026-06-30
CLOUDP-411215: Add voyageais to OperatorConfig watchedResources enum

m1kola:
# Summary

Master catch-up for the OperatorConfig work. Since the technical design was frozen, master introduced a new VoyageAI controller (KUBE-126) with its own watched CRD, and PR #1287 renamed that CRD's plural to `voyageais`. The `OperatorConfig` CRD's `watchedResources` enum did not know about it.

This adds `voyageais` to the `WatchedResource` enum in `api/operator/v1/operatorconfig_types.go` and regenerates the CRD manifests (`config/crd/bases`, `helm_chart/crds`, `public/crds.yaml`) so the enum stays in sync with the operator's watch list (`voyageAICRDPlural = "voyageais"`).

Note: the `watchedResources` runtime wiring is a separate work item; this only authors the enum value, consistent with the rest of the already-authored OperatorConfig spec.

## Proof of Work

- `go build ./...` passes.
- `go test ./pkg/operatorconfig/...` passes.
- Regenerated CRDs verified to contain `voyageais` in the `watchedResources` enum, with no other schema drift.

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

PR #1310, opened 2026-07-01
CLOUDP-411215: Move watchedResources setting to OperatorConfig

m1kola:
# Summary

Continues migrating operator settings into the `OperatorConfig` CR. Moves the set of CRDs the operator reconciles off the `-watch-resource` flag / `operator.watchedResources` Helm value onto `OperatorConfig.spec.watchedResources`, defaulting to all known CRDs. The previously implicit enablement of `mongodbmulticluster` (from `multiCluster.clusters`) and `clustermongodbroles` (from `operator.enableClusterMongoDBRoles`) is now expressed explicitly in that list. Member-cluster discovery now tolerates a missing member-list ConfigMap, so single-cluster operators start even with `mongodbmulticluster` watched.

Note: this branch also carries the commit from #1306 (voyageais catchup) — close #1306 in favour of this PR.

## Proof of Work

- `pkg/operatorconfig/operatorconfig_test.go` — defaulting (absent CR → all CRDs; explicit subset preserved)
- `helm template helm_chart` renders with no `-watch-resource` args across default / `multiCluster.clusters` / `enableClusterMongoDBRoles=false`
- e2e `operator_partial_crd` and `multicluster_search` migrated to `OperatorConfig`

## Checklist

---

### DIFF
--- a/api/operator/v1/operatorconfig_types.go
+++ b/api/operator/v1/operatorconfig_types.go
@@ -28,6 +28,19 @@ const (
 	WatchedResourceVoyageAI            WatchedResource = "voyageais"
 )
 
+// AllWatchedResources lists every CRD the operator can reconcile. It is the default value
+// for OperatorConfig.spec.watchedResources when the field is omitted.
+var AllWatchedResources = []WatchedResource{
+	WatchedResourceMongoDB,
+	WatchedResourceOpsManagers,
+	WatchedResourceMongoDBUsers,
+	WatchedResourceMongoDBCommunity,
+	WatchedResourceMongoDBSearch,
+	WatchedResourceMongoDBMultiCluster,
+	WatchedResourceClusterMongoDBRoles,
+	WatchedResourceVoyageAI,
+}
+
 // Architecture defines the container architecture used by operator-managed workloads.
 // +kubebuilder:validation:Enum=NonStatic;Static
 type Architecture string
--- a/api/operator/v1/operatorconfig_types.go
+++ b/api/operator/v1/operatorconfig_types.go
@@ -14,7 +14,7 @@ const (
 )
 
 // WatchedResource identifies a CRD that the operator will actively reconcile.
-// +kubebuilder:validation:Enum=mongodb;opsmanagers;mongodbusers;mongodbcommunity;mongodbsearch;mongodbmulticluster;clustermongodbroles
+// +kubebuilder:validation:Enum=mongodb;opsmanagers;mongodbusers;mongodbcommunity;mongodbsearch;mongodbmulticluster;clustermongodbroles;voyageais
 type WatchedResource string
 
 const (
@@ -25,6 +25,7 @@ const (
 	WatchedResourceMongoDBSearch       WatchedResource = "mongodbsearch"
 	WatchedResourceMongoDBMultiCluster WatchedResource = "mongodbmulticluster"
 	WatchedResourceClusterMongoDBRoles WatchedResource = "clustermongodbroles"
+	WatchedResourceVoyageAI            WatchedResource = "voyageais"
 )
 
 // Architecture defines the container architecture used by operator-managed workloads.
--- a/docker/mongodb-kubernetes-tests/kubetester/kubetester.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/kubetester.py
@@ -201,6 +201,7 @@ def apply_operator_config_from_test_env(
     api_client=None,
     name: str = "mongodb-kubernetes-operator",
     wait_for_ready: Optional[Callable[[], None]] = None,
+    extra_spec: Optional[dict] = None,
 ) -> bool:
     """Creates the OperatorConfig CR from test env vars (if any non-default settings exist) and waits for
     the operator to restart and reload its configuration.
@@ -222,6 +223,8 @@ def apply_operator_config_from_test_env(
     from tests.conftest import local_operator
 
     spec = build_operator_config_spec_from_test_env()
+    if extra_spec:
+        spec.update(extra_spec)
     if not spec:
         return False
 
--- a/docker/mongodb-kubernetes-tests/tests/operator/operator_partial_crd.py
+++ b/docker/mongodb-kubernetes-tests/tests/operator/operator_partial_crd.py
@@ -4,6 +4,8 @@ from typing import Dict
 import pytest
 from kubernetes import client
 from kubetester.create_or_replace_from_yaml import create_or_replace_from_yaml
+from kubetester.helm import apply_operator_config_crd
+from kubetester.kubetester import create_operator_config
 from kubetester.operator import Operator, delete_operator_crds, list_operator_crds
 from pytest import fixture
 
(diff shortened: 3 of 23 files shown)


main.go
```diff
@@ -521,6 +498,12 @@ func getMemberClusters(ctx context.Context, cfg *rest.Config, currentNamespace s
 
 	m := corev1.ConfigMap{}
 	err = c.Get(ctx, types.NamespacedName{Name: util.MemberListConfigMapName, Namespace: currentNamespace}, &m)
+	if apierrors.IsNotFound(err) {
+		// No multi-cluster configuration present: run as single-cluster with no member clusters.
+		// The member-list ConfigMap is absent on single-cluster installs (e.g. when
+		// mongodbmulticluster is watched by default). Callers handle an empty member list.
+		return nil, nil
+	}
```
josvazg, line 506: > 1. main.go:501-506 — getMemberClusters swallowing NotFound conflates "not configured" with "misconfigured." This function only runs when mongodbmulticluster is watched — which is now the default case for every install, single- or multi-cluster. So a genuinely broken multi-cluster install (ConfigMap deleted, wrong namespace, RBAC returning a 404-shaped error) now silently degrades to "0 member clusters, single-cluster mode" with just a Warnf, instead of failing fast as it did before. Worth a second look: is there a way to distinguish "single-cluster, no multi-cluster CM expected" from "multi-cluster intended but the CM went missing"? Right now there isn't — both look identical.

josvazg:
> 2. Real upgrade gotcha (documented, but sharp edge): existing OperatorConfig CRs with a restricted watchedResources subset don't get migrated. withDefaults only fills in AllWatchedResources when the field is empty, not merged into an existing non-empty list. So anyone who previously restricted resources via Helm (which never included mongodbmulticluster/clustermongodbroles unless separately opted in) and already has an OperatorConfig CR with an explicit subset will silently lose clustermongodbroles/mongodbmulticluster watching post-upgrade unless they manually update the CR first. The changelog does call this out explicitly ("must create or update their OperatorConfig resource before upgrading"), so this is intentional and documented — flagging it so you make sure release notes/upgrade docs downstream actually surface it prominently, since it's easy to miss in a changelog file.

pkg/operatorconfig/operatorconfig.go
```diff
@@ -35,5 +36,8 @@ func withDefaults(cfg operatorv1.OperatorConfig) operatorv1.OperatorConfig {
 	if cfg.Spec.MaxConcurrentReconciles == 0 {
 		cfg.Spec.MaxConcurrentReconciles = 1
 	}
+	if len(cfg.Spec.WatchedResources) == 0 {
```
josvazg, line 39: >3. Minor: empty list vs. omitted field are indistinguishable. if len(cfg.Spec.WatchedResources) == 0 { ... = AllWatchedResources } means a user who explicitly sets watchedResources: [] (intending "watch nothing") silently gets "watch everything" instead. Given the field is x-kubernetes-list-type: set, an empty list is a legal, distinct value from omission. Low-severity edge case, but worth a one-line doc comment or explicit test if "watch nothing" is ever a supported configuration.

api/operator/v1/operatorconfig_types.go
```diff
@@ -25,8 +25,22 @@ const (
...
 	WatchedResourceMongoDBMultiCluster WatchedResource = "mongodbmulticluster"
 	WatchedResourceClusterMongoDBRoles WatchedResource = "clustermongodbroles"
+	WatchedResourceVoyageAI            WatchedResource = "voyageais"
 )
 
+// AllWatchedResources lists every CRD the operator can reconcile. It is the default value
+// for OperatorConfig.spec.watchedResources when the field is omitted.
+var AllWatchedResources = []WatchedResource{
+	WatchedResourceMongoDB,
+	WatchedResourceOpsManagers,
```
josvazg, line 35: non blocking
> this list now has to stay in sync with the enum marker above and the plural constants in main.go by hand — worth a test that iterates the kubebuilder enum values (or a go:generate) so adding a CRD can't silently miss one of the three spots.

josvazg (approved):
LGTM with minor comments

pkg/operatorconfig/operatorconfig.go
m1kola: "watch nothing" makes no sense for our operator. I don't think it is worth adding complexity to distinguish between empty list and omitted value.

main.go
m1kola: This is a transient state to keep our E2Es working on the E2E branch. The endgame is to get rid of the ConfigMap and migrate to MemberCluster CRs.

With MemberCluster CRDs our target logic is exactly that - if there are no MemberCluster CRs - we are running in a single-cluster.

m1kola:
> 2. Real upgrade gotcha (documented, but sharp edge): existing OperatorConfig CRs with a restricted watchedResources subset don't get migrated. withDefaults only fills in AllWatchedResources when the field is empty, not merged into an existing non-empty list. So anyone who previously restricted resources via Helm (which never included mongodbmulticluster/clustermongodbroles unless separately opted in) and already has an OperatorConfig CR with an explicit subset will silently lose clustermongodbroles/mongodbmulticluster watching post-upgrade unless they manually update the CR first. The changelog does call this out explicitly ("must create or update their OperatorConfig resource before upgrading"), so this is intentional and documented — flagging it so you make sure release notes/upgrade docs downstream actually surface it prominently, since it's easy to miss in a changelog file.

@josvazg true - this is slightly annoying. But we will:
1. Implement a migration script and it should handle this.
2. We will document in the migration guide for users who won't be using the migration script.

docker/mongodb-kubernetes-tests/kubetester/kubetester.py
```diff
@@ -201,6 +201,7 @@ def apply_operator_config_from_test_env(
     api_client=None,
     name: str = "mongodb-kubernetes-operator",
     wait_for_ready: Optional[Callable[[], None]] = None,
+    extra_spec: Optional[dict] = None,
```
mircea-cosbuc, line 204: nit: any reason we're not doing a `**kwargs` pattern, so we can use a call similar to something like this:
```
apply_operator_config_from_test_env(
            self.namespace, api_client=self.api_client, name=self.name, wait_for_ready=wait_for_ready
            self.namespace,
            api_client=self.api_client,
            name=self.name,
            wait_for_ready=wait_for_ready,
            watchedResources={},
            anotherSpecValue=false,
        )
```

docker/mongodb-kubernetes-tests/tests/operator/operator_partial_crd.py
```diff
@@ -24,12 +26,14 @@ def operator_only_ops_manager_and_mongodb(
     namespace: str,
     operator_installation_config: Dict[str, str],
 ) -> Operator:
-    helm_args = operator_installation_config.copy()
-    helm_args["operator.watchedResources"] = "{opsmanagers,mongodb}"
+    # Restrict which CRDs the operator reconciles via OperatorConfig. The CR must exist before the
+    # operator starts: with no CR it defaults to watching all CRDs and would crash on the ones that
+    # aren't installed in this test.
+    create_operator_config(namespace, {"watchedResources": ["opsmanagers", "mongodb"]})
```
mircea-cosbuc, line 32: It's great to already see this shaping up in the config being equivalent to the helm args.

mircea-cosbuc (approved):
LGTM

docker/mongodb-kubernetes-tests/kubetester/kubetester.py
m1kola: I went with an explicit `extra_spec` dict rather than `**kwargs` because the keys are `OperatorConfig` spec fields (camelCase like `watchedResources`), so a dict mirrors the CR spec 1:1 and keeps those out of the function's snake_case signature. `**kwargs` would also sit next to real params like `multi_cluster` and `wait_for_ready`.

PR #1314, opened 2026-07-01
CLOUDP-411215: Move memberClusterClientTimeout setting to OperatorConfig

m1kola:
# Summary

Part of the Installation UX epic: operator settings are being moved out of Helm values / environment variables and into the `OperatorConfig` CR so they can be configured post-installation, consistently across Helm, OLM and raw YAML. This PR migrates the member-cluster API client timeout.

Previously the timeout was set via the `multiCluster.clusterClientTimeout` Helm value / `CLUSTER_CLIENT_TIMEOUT` environment variable and read directly from the environment at two sites. Now it is read once at startup from `.spec.multiCluster.memberClusterClientTimeout` and threaded into the member-cluster clients (`CreateMemberClusterClients`) and the member-cluster health checker (`MemberClusterHealthChecker`, via a new `WithTimeout` option). The Helm value, template block and rendered env var in the public manifests have been removed. The default remains `10` seconds.

Because `spec.multiCluster` is an optional pointer block, `operatorconfig.withDefaults` now materialises it and defaults the timeout to `10`, so an omitted block cannot leave clients with a zero (i.e. no) timeout.

This is a breaking change for anyone who set `multiCluster.clusterClientTimeout` / `CLUSTER_CLIENT_TIMEOUT`; a `breaking` changelog entry is included.

## Proof of Work

- `pkg/operatorconfig/operatorconfig_test.go` — added `TestLoad_MemberClusterClientTimeout` (omitted block defaults to `10`; explicit value preserved) and extended `TestLoad_AbsentCR` to assert the defaulted timeout.
- `go build ./...`, `go vet` on the touched packages, and `go test ./pkg/operatorconfig/... ./pkg/multicluster/...` all pass locally.
- `grep` confirms no remaining `CLUSTER_CLIENT_TIMEOUT` / `clusterClientTimeout` references outside the changelog; the `OperatorConfig` CRD YAML is unchanged (the field already existed).

## Checklist

---

### DIFF
--- a/pkg/multicluster/multicluster.go
+++ b/pkg/multicluster/multicluster.go
@@ -25,9 +25,8 @@ import (
 
 const (
 	// kubeconfig path holding the credentials for different member clusters
-	DefaultKubeConfigPath   = "/etc/config/kubeconfig/kubeconfig"
-	KubeConfigPathEnv       = "KUBE_CONFIG_PATH"
-	ClusterClientTimeoutEnv = "CLUSTER_CLIENT_TIMEOUT"
+	DefaultKubeConfigPath = "/etc/config/kubeconfig/kubeconfig"
+	KubeConfigPathEnv     = "KUBE_CONFIG_PATH"
 )
 
 type KubeConfig struct {
@@ -61,7 +60,7 @@ func (k KubeConfig) LoadKubeConfigFile() (KubeConfigFile, error) {
 }
 
 // CreateMemberClusterClients creates a client(map of cluster-name to client) to talk to the API-Server corresponding to each member clusters.
-func CreateMemberClusterClients(clusterNames []string, kubeConfigPath string) (map[string]*restclient.Config, error) {
+func CreateMemberClusterClients(clusterNames []string, kubeConfigPath string, clientTimeoutSeconds int) (map[string]*restclient.Config, error) {
 	clusterClientsMap := map[string]*restclient.Config{}
 
 	for _, c := range clusterNames {
@@ -72,7 +71,7 @@ func CreateMemberClusterClients(clusterNames []string, kubeConfigPath string) (m
 		if clientset == nil {
 			return nil, xerrors.Errorf("failed to get clientset for cluster: %s", c)
 		}
-		clientset.Timeout = time.Duration(env.ReadIntOrDefault(ClusterClientTimeoutEnv, 10)) * time.Second // nolint:forbidigo
+		clientset.Timeout = time.Duration(clientTimeoutSeconds) * time.Second
 		clusterClientsMap[c] = clientset
 	}
 	return clusterClientsMap, nil
--- a/pkg/operatorconfig/operatorconfig.go
+++ b/pkg/operatorconfig/operatorconfig.go
@@ -39,5 +39,15 @@ func withDefaults(cfg operatorv1.OperatorConfig) operatorv1.OperatorConfig {
 	if len(cfg.Spec.WatchedResources) == 0 {
 		cfg.Spec.WatchedResources = slices.Clone(operatorv1.AllWatchedResources)
 	}
+	// MultiCluster is a pointer, so an omitted block leaves the API server's
+	// nested defaults (e.g. memberClusterClientTimeout=10) unapplied. Ensure the
+	// block exists and the timeout is defaulted so member-cluster clients never
+	// end up with a zero (i.e. no) timeout.
+	if cfg.Spec.MultiCluster == nil {
+		cfg.Spec.MultiCluster = &operatorv1.MultiClusterConfig{}
+	}
+	if cfg.Spec.MultiCluster.MemberClusterClientTimeout == 0 {
+		cfg.Spec.MultiCluster.MemberClusterClientTimeout = 10
+	}
 	return cfg
 }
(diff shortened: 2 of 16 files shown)


josvazg:
> Heads-up (non-blocking, just for awareness): This commit removes the multiCluster.clusterClientTimeout Helm value. There's a leftover reference to it in docker/mongodb-kubernetes-tests/tests/multicluster_search/conftest.py:96, where it's listed in a helm_args.pop("multiCluster.clusterClientTimeout", None).

> It's not part of this diff and it's harmless — .pop(..., None) is a no-op now that the key is gone — so nothing breaks. Flagging only in case you want to tidy up the dead reference while this ticket is fresh, since it's the same value being removed here. Fine to leave for a follow-up.

pkg/operatorconfig/operatorconfig.go
```diff
@@ -39,5 +39,15 @@ func withDefaults(cfg operatorv1.OperatorConfig) operatorv1.OperatorConfig {
...
 		cfg.Spec.WatchedResources = slices.Clone(operatorv1.AllWatchedResources)
 	}
+	// MultiCluster is a pointer, so an omitted block leaves the API server's
+	// nested defaults (e.g. memberClusterClientTimeout=10) unapplied. Ensure the
+	// block exists and the timeout is defaulted so member-cluster clients never
+	// end up with a zero (i.e. no) timeout.
+	if cfg.Spec.MultiCluster == nil {
+		cfg.Spec.MultiCluster = &operatorv1.MultiClusterConfig{}
+	}
+	if cfg.Spec.MultiCluster.MemberClusterClientTimeout == 0 {
```
josvazg, line 49: > == 0 defaulting prevents a legitimate "no timeout" value
> **q**: withDefaults treats MemberClusterClientTimeout == 0 as unset and forces it to 10. Combined with the CRD's Minimum=1, this means a user can never express a 0 (= no timeout) value. This was intentional right?

pkg/operatorconfig/operatorconfig.go
```diff
@@ -39,5 +39,15 @@ func withDefaults(cfg operatorv1.OperatorConfig) operatorv1.OperatorConfig {
 	if len(cfg.Spec.WatchedResources) == 0 {
 		cfg.Spec.WatchedResources = slices.Clone(operatorv1.AllWatchedResources)
 	}
+	// MultiCluster is a pointer, so an omitted block leaves the API server's
+	// nested defaults (e.g. memberClusterClientTimeout=10) unapplied. Ensure the
+	// block exists and the timeout is defaulted so member-cluster clients never
+	// end up with a zero (i.e. no) timeout.
+	if cfg.Spec.MultiCluster == nil {
+		cfg.Spec.MultiCluster = &operatorv1.MultiClusterConfig{}
```
josvazg, line 47: >  Asymmetric defaulting within MultiClusterConfig

> followup: **withDefaults** materializes the nil MultiCluster pointer and defaults MemberClusterClientTimeout, but not MemberClusterRequiredHealthyStreak, even though both fields have kubebuilder defaults. No impact today because RequiredHealthyStreak is still read from the RequiredHealthyStreakEnv env var (main.go:431, mongodbsearch_controller.go:352) rather than from OperatorConfig — but that itself is a latent inconsistency: the CRD field exists but isn't consumed. Worth a follow-up or a clarification of the intent going forward.

pkg/multicluster/multicluster.go
```diff
@@ -72,7 +71,7 @@ func CreateMemberClusterClients(clusterNames []string, kubeConfigPath string) (m
 		if clientset == nil {
 			return nil, xerrors.Errorf("failed to get clientset for cluster: %s", c)
 		}
-		clientset.Timeout = time.Duration(env.ReadIntOrDefault(ClusterClientTimeoutEnv, 10)) * time.Second // nolint:forbidigo
+		clientset.Timeout = time.Duration(clientTimeoutSeconds) * time.Second
```
josvazg, line 74: >  No guard against negative timeout on the absent-CR path

> nit-pick: The seconds value is converted to a Duration without a lower-bound check. The CRD's Minimum=1 covers the normal path, but the absent-CR / fake-client path relies solely on the == 0 → 10 default, so a negative value slipping past validation would yield a negative client timeout. Very low risk; noting in case defense-in-depth is wanted here (and at the parallel ClientTimeout conversions in the two controllers).

josvazg (approved):
LGTM, minor comments

pkg/operatorconfig/operatorconfig.go
m1kola: @josvazg  `MemberClusterRequiredHealthyStreak` is handled separately in #1316. We set the default there.

pkg/operatorconfig/operatorconfig.go
m1kola: Yes, intentional. Our minimum timeout is 1 and it can never be "no timeout".

pkg/multicluster/multicluster.go
m1kola: The value either comes from a CR which is validated on admission by API server with min value 1 or comes from a default defined in code which is positive value. I don't think we need to do anything more sophisticated here.

m1kola:
> It's not part of this diff and it's harmless — .pop(..., None) is a no-op now that the key is gone — so nothing breaks. Flagging only in case you want to tidy up the dead reference while this ticket is fresh, since it's the same value being removed here. Fine to leave for a follow-up.

I'll clean up in a follow up PR.

Edit: https://github.com/mongodb/mongodb-kubernetes/pull/1328

mircea-cosbuc (approved):

PR #1315, opened 2026-07-01
Replace pre-commit with prek

nammn:
## Summary
- Swap the Python `pre-commit` framework for `prek` (https://github.com/j178/prek) across local dev setup, git hooks, Makefile, and Evergreen CI.
- Add `scripts/evergreen/setup_prek.sh` to download and checksum-verify the `prek` binary into `${PROJECT_DIR}/bin`, and wire it into `install.sh` and the `setup_prek` Evergreen function/task.
- `check_precommit.sh` now fails fast with instructions if `prek` isn't installed, instead of `pip install`-ing on the fly.

## Why
`prek` is a Rust reimplementation of `pre-commit` that's a drop-in replacement for our existing `.pre-commit-config.yaml` setup, with a few concrete wins:
- **Faster hook runs**: no Python interpreter startup or per-hook virtualenv bootstrapping, which matters both for the local git hook (runs on every commit) and for the `lint_repo` Evergreen task.
- **No Python/pip dependency for the tool itself**: it's a single static binary fetched from GitHub releases and checksum-verified, so it's no longer coupled to `requirements.txt`/venv state. This removes a class of "my venv is stale so pre-commit is broken" local dev issues. (that annoyed me locally the most 😓 )
- **Same config, no migration cost**: it reads our existing `.pre-commit-config.yaml` as-is, so this is purely a runtime swap, not a rewrite of our lint/format hook definitions.
- **Fails loud instead of silently reinstalling**: CI now errors clearly if `setup_prek` didn't run, rather than masking setup issues with an implicit `pip install` on every run.

## Benchmarks
**Prek**
```
❯ time prek -a 
update-mongodb-operator-version..........................................Passed
generate-manifests.......................................................Passed
update-mco-tests.........................................................Passed
regenerate-public-rbac-multi-cluster.....................................Passed
update-licenses..........................................................Passed
check-kubebuilder-annotations............................................Passed
update-release-json......................................................Passed
update-values-yaml.......................................................Passed
generate-standalone-yaml.................................................Passed
validate-snippets........................................................Passed
validate-helm-charts.....................................................Passed
validate-go-version......................................................Passed
check for added large files..............................................Passed
check for merge conflicts................................................Passed
check json...............................................................Passed
black....................................................................Passed
isort....................................................................Passed
ShellCheck v0.10.0.......................................................Passed
ty.......................................................................Passed
govulncheck..............................................................Passed
golangci-lint-full.......................................................Passed
golangci-lint-fmt........................................................Passed

```
```
Executed in    7.28 secs    fish           external
```

**Pre-Commit**
```
Executed in   18.80 secs    fish           external
```

## Test plan
- [x] `make precommit` / `make precommit-full` run via `prek` locally
- [x] `.githooks/pre-commit` runs `prek run --all-files` successfully
- [x] Evergreen `lint_repo`/precommit task installs `prek` via `setup_prek` and passes


.githooks/pre-commit
```diff
@@ -14,5 +14,5 @@ if [ -f "${PROJECT_DIR}/venv/bin/activate" ]; then
   source "${PROJECT_DIR}/venv/bin/activate"
```
m1kola, line 15: Do we still need to activate venv with prek?

nammn: no! We can remove!

.pre-commit-config.yaml
```diff
@@ -20,35 +25,71 @@ repos:
...
         language: system
         pass_filenames: false
         stages: [pre-commit]
+        priority: 4
 
-      # Remaining generation jobs run in parallel
-      - id: generate-and-update
-        name: generate-and-update (parallel)
-        entry: scripts/dev/generate_files.sh generate_all
+      - id: generate-manifests
```
josvazg, line 35: > q: — .pre-commit-config.yaml:54 (generate-manifests now priority: 1)
In the old config, the parallel batch (manifests, mco, rbac, licenses, kubebuilder) ran last (step 5), i.e. after generate-standalone-yaml (step 4). But generate-standalone-yaml reads helm_chart/crds/ which generate-manifests writes — so the old ordering looks like it fed stale CRDs into standalone. The new P1-before-P4 ordering fixes that. Worth confirming with the author this reordering was intentional (and ideally calling it out in the PR description as a latent-bug fix, not just a mechanical swap), since it's a behavioral change hiding inside a "pure runtime swap."

.pre-commit-config.yaml
```diff
@@ -20,35 +25,71 @@ repos:
...
+        stages: [pre-commit]
+        priority: 1
+
+      - id: check-kubebuilder-annotations
+        name: check-kubebuilder-annotations
+        entry: scripts/dev/generate_files.sh check_kubebuilder_annotations
+        language: system
+        pass_filenames: false
+        stages: [pre-commit]
+        priority: 1
```
josvazg, line 73: >q: — .pre-commit-config.yaml:22 & :54-92 (six hooks all at priority: 1, run concurrently)
update-mongodb-operator-version writes release.json and now runs concurrently with the other P1 hooks (previously it was strictly step 1, before the batch). I checked: make manifests, create_mco_tests.py, licenses, rbac and the kubebuilder grep don't read release.json, and update-release-json is correctly P2 (serial, after) — so no race today. But that safety is implicit. A one-line note in the config comment ("P1 hooks are mutually file-independent; only update-release-json at P2 consumes release.json") would protect this against someone later adding a P1 hook that reads release.json.

scripts/evergreen/setup_prek.sh
```diff
@@ -0,0 +1,52 @@
...
+
+echo "Verifying checksum..."
+# ponytail: sha256sum on Linux, shasum -a 256 on macOS
+if command -v sha256sum >/dev/null 2>&1; then
+  sha256sum -c "${tarball}.sha256" --status
+else
+  shasum -a 256 -c "${tarball}.sha256" --status
+fi
+
+tar --strip-components=1 -xf "${tarball}"
```
josvazg, line 46: > nitpick: — scripts/evergreen/setup_prek.sh:30-46
The downloaded ${tarball} and .sha256 are left behind in ${PROJECT_DIR}/tmp — no cleanup trap. Minor, but every install.sh run leaves artifacts.

scripts/evergreen/setup_prek.sh
```diff
@@ -0,0 +1,52 @@
...
+# Downloads and installs prek (Rust reimplementation of pre-commit)
+# from GitHub releases into ${PROJECT_DIR}/bin.
+#
+
+set -Eeou pipefail
+
+source scripts/dev/set_env_context.sh
+source scripts/funcs/install
+
+PREK_VERSION="v0.4.6"
```
josvazg, line 12: > nitpick: — scripts/evergreen/setup_prek.sh:30-46
The downloaded ${tarball} and .sha256 are left behind in ${PROJECT_DIR}/tmp — no cleanup trap. Minor, but every install.sh run leaves artifacts.

nammn: after a run usually PROJECT_DIR either gets cleaned up or the whole vm

scripts/evergreen/check_precommit.sh
```diff
@@ -28,9 +28,9 @@ export EVERGREEN_MODE=true
 # Store the current state of the index and working directory
 initial_index_state=$(git diff --name-only --cached --diff-filter=AM)
 
-# Run pre-commit on all files
+# Run prek on all files
```
josvazg, line 31: > nitpick: — scripts/evergreen/check_precommit.sh:31 & :37
Leftover naming: the title still says "Running pre-commit checks" and it echoes "Pre-commit hook has completed.". Cosmetic, but since the whole point is the prek migration, worth updating for grep-ability.

Should we say precommit or prek now? or both?

scripts/evergreen/setup_prek.sh
nammn: same as below

scripts/evergreen/check_precommit.sh
```diff
@@ -14,10 +14,10 @@ if [ -f "${PROJECT_DIR}/venv/bin/activate" ]; then
   source "${PROJECT_DIR}/venv/bin/activate"
```
josvazg, line 14: > q: — scripts/evergreen/check_precommit.sh:14
It still sources venv/bin/activate. Since one of prek's selling points (per the PR description) is decoupling from venv state, is this still needed here? (I think yes — the language: system local hooks shell out to run_python.sh — but if so it's worth a comment, otherwise it looks like a leftover contradicting the PR's stated motivation.)

scripts/evergreen/check_precommit.sh
nammn: hmmm hmmmm i think pre-commit makes more sense naming-wise.
Prek is just a utility

josvazg (approved):
LGTM, minor comments and question, but feel free to merge

.githooks/pre-commit
nammn: Confirmed — venv activation was already removed from .githooks/pre-commit in this PR (line 11 now goes straight to `exec prek -a`). prek is a Rust binary on PATH via set_env_context.sh, and the hooks it spawns either self-activate (run_python.sh, generate_files.sh) or are stdlib-only (validate_snippets.py). Resolving.

.pre-commit-config.yaml
nammn: Confirmed intentional and a latent-bug fix. Verified via `prek -a --dry-run` locally: generate-manifests (P1) runs before generate-standalone-yaml (P4), so helm_chart/crds/ is fresh when standalone reads it. The old serial config ran the parallel batch (step 5) *after* generate-standalone-yaml (step 4), so standalone consumed stale CRDs. Documented this in the config comment header along with the P1/P4 ordering note. Resolving.

.pre-commit-config.yaml
nammn: Good catch — added the guard note to the config comment header: "P1 hooks are mutually file-independent; only update-release-json at P2 consumes release.json. Do not add a P1 hook that reads release.json without moving it to P2+." Verified the current safety: make manifests (no release.json read), create_mco_tests.py (no release.json read), licenses/rbac/kubebuilder-grep don't touch it, and update-release-json is P2 (serial, after the P1 barrier). Resolving.

scripts/evergreen/check_precommit.sh
nammn: Not needed — removed in the latest push.

MaciejKaras (approved):

PR #1317, opened 2026-07-02
CLOUDP-400897: Remove operator env from helm charts

roothorp:
# Summary

Remove the operator env variable (dev/prod) from helm charts. Keep the functionality in the Operator, so it can be used by manually setting the env var.

## Checklist

---

### DIFF
--- a/docker/mongodb-kubernetes-tests/kubetester/operator.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/operator.py
@@ -277,6 +277,7 @@ def list_operator_crds() -> List[V1CustomResourceDefinition]:
         key=lambda crd: crd.metadata.name,
     )
 
+
 def add_to_custom_env_vars_value(helm_args: dict, key: str, value: str) -> None:
     existing = helm_args.get("customEnvVars")
     env_vars = existing.split("&") if existing else []
--- a/docker/mongodb-kubernetes-tests/kubetester/operator.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/operator.py
@@ -60,7 +60,10 @@ class Operator(object):
             helm_args = {}
 
         helm_args["namespace"] = namespace
-        helm_args["operator.env"] = "dev"
+        if helm_args.get("customEnvVars"):
+            helm_args["customEnvVars"] += "\&OPERATOR_ENV=dev"
+        else:
+            helm_args["customEnvVars"] = "OPERATOR_ENV=dev"
 
         # the import is done here to prevent circular dependency
         from tests.conftest import local_operator
--- a/docker/mongodb-kubernetes-tests/kubetester/operator.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/operator.py
@@ -60,10 +60,8 @@ class Operator(object):
             helm_args = {}
 
         helm_args["namespace"] = namespace
-        if helm_args.get("customEnvVars"):
-            helm_args["customEnvVars"] += "\&OPERATOR_ENV=dev"
-        else:
-            helm_args["customEnvVars"] = "OPERATOR_ENV=dev"
+
+        add_to_custom_env_vars_value(helm_args, "OPERATOR_ENV", "dev")
 
         # the import is done here to prevent circular dependency
         from tests.conftest import local_operator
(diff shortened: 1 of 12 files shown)


m1kola (changes requested):
There are some dependencies on this env var in the operator code (in telemetry, pprof. Probably somewhere else).

Our plan is to remove it as an interface from Helm chart. But we want to keep it for internal use (e..g pprof). We will likely need to update scripts alongside removal from helm.

I also see that some E2E tests (e.g. `e2e_olm_meko_operator_upgrade_with_resources`, `e2e_olm_operator_upgrade_with_resources`, `e2e_olm_operator_upgrade` - maybe others too) set it. Need to check if it is necessary to set it in these tests or not.

And we also need to regenerate examples.

helm_chart/templates/operator.yaml
```diff
@@ -72,8 +72,6 @@ spec:
             {{- toYaml .Values.operator.securityContext | nindent 12 }}
           {{- end }}
           env:
-            - name: OPERATOR_ENV
-              value: {{ .Values.operator.env }}
```
m1kola, line 76: Following up on this comment: https://github.com/mongodb/mongodb-kubernetes/pull/1317#pullrequestreview-4615735081

So if we remove this, do we still propagate `OPERATOR_ENV` when set to `dev` to the operator? I see that we set it in root context here:

https://github.com/mongodb/mongodb-kubernetes/blob/1f284f3ec58d96b07b58df2b8c983e13da227de9/scripts/dev/contexts/root-context#L38

Would the operator still pick it up when running locally? Or in tests?

m1kola: Also I think the operator will panic here if we do not pass it at all:

https://github.com/mongodb/mongodb-kubernetes/blob/1f284f3ec58d96b07b58df2b8c983e13da227de9/main.go#L550

I think we should default to prod in the operator, but make it possible to override it using the env var for local dev via contexts.

roothorp: I believe I have now defaulted us to `prod`, and set to `dev` (using `customEnvVars`) where appropriate.

I will see how the tests look after this change.

docker/mongodb-kubernetes-tests/kubetester/operator.py
```diff
@@ -59,7 +59,10 @@ def __init__(
             helm_args = {}
 
         helm_args["namespace"] = namespace
-        helm_args["operator.env"] = "dev"
+        if helm_args.get("customEnvVars"):
+            helm_args["customEnvVars"] += "\&OPERATOR_ENV=dev"
+        else:
+            helm_args["customEnvVars"] = "OPERATOR_ENV=dev"
```
m1kola: nit: would be great not to duplicate this on line 63 if `customEnvVars` already non-empty.

m1kola (approved):
Small nit. Other than that - looks good to me.

josvazg (commented):
q: missing changelog entry?

docker/mongodb-kubernetes-tests/kubetester/operator.py
roothorp: I'm not sure I follow - we need to do 2 different things here because we want to append (with an `&`) if there are existing env vars (which we do on ln 63), but if there are no existing vars, we cannot append to an empty string so we need to initialise it (ln 65).

Am I misunderstanding what you want cleaned up here?

docker/mongodb-kubernetes-tests/kubetester/operator.py
```diff
@@ -59,7 +59,10 @@ def __init__(
             helm_args = {}
 
         helm_args["namespace"] = namespace
-        helm_args["operator.env"] = "dev"
+        if helm_args.get("customEnvVars"):
+            helm_args["customEnvVars"] += "\&OPERATOR_ENV=dev"
```
m1kola: `\&` - is it correct? Do we need to escape `&` here?

docker/mongodb-kubernetes-tests/kubetester/operator.py
m1kola: I was wondering if we could avoid duplicating `OPERATOR_ENV=dev` somehow. E.g. maybe use a `join`?

Or even define a separate function to work with `customEnvVars` (we _might_ need to add more things to `customEnvVars` later, but not sure about it):

```python
def add_to_custom_env_vars_value(helm_args: dict, key: str, value: str) -> None:
    existing = helm_args.get("customEnvVars")
    env_vars = existing.split("&") if existing else []
    env_vars.append(f"{key}={value}")
    helm_args["customEnvVars"] = "&".join(env_vars)
```

docker/mongodb-kubernetes-tests/kubetester/operator.py
roothorp: Yes, I think we pass these straight into shell wholesale so we need to escape it - and the `&` is what we're using to split the env vars up.

m1kola (approved):

PR #1319, opened 2026-07-02
Fix flake in `TestStreakResetsOnUnhealthyWhenAlmostRecovered` unit test

m1kola:
# Summary

This test appears to be slow and flaky (example [here](https://spruce.corp.mongodb.com/task/__feature_mc_installation_ux_unit_tests_unit_tests_golang_patch_c257dc894cb71bc6c033e16d09ae5529c3106487_6a461391df3f460007946b47_26_07_02_07_30_27/logs?execution=1)) due to waits. This is an attempt to solve this with `synctest.Wait()`.

## Proof of Work

N/A

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

---

PR #1320, opened 2026-07-02
Harden remaining memberwatch clusterhealth tests against synctest scheduling races

m1kola:
# Summary

Stacked on #1319, which fixed a flake in `TestStreakResetsOnUnhealthyWhenAlmostRecovered` caused by `assert.Eventually`/`assert.Never` racing the watcher goroutine inside a `synctest` bubble: testify spawns a fresh goroutine per poll, so two non-atomic effects of the same watcher iteration (streak update, then annotation write) can be observed out of order.

This PR applies the same `synctest.Wait()` pattern to the other tests in `clusterhealth_test.go` that used `assert.Never`/`assert.Eventually` pairs, for consistency and to remove the arbitrary 500ms/5s polling timeouts. Tests that already synchronize correctly via `<-watchChannel` (a genuine happens-before point, since the channel send happens after the relevant writes in the watcher's program order) are left unchanged.

## Proof of Work

- `pkg/multicluster/memberwatch/clusterhealth_test.go` — ran the full package with `go test ./pkg/multicluster/memberwatch/... -race -count=20` and `go test ./pkg/multicluster/... -race -count=5`; all pass with no flakes.

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title? — N/A, test-only deflake follow-up
- [x] Have you checked whether your jira ticket required DOCSP changes? — N/A, no jira ticket
    - `skip-changelog` label applied (test-only change, no user-facing behaviour)

### DIFF
--- a/pkg/multicluster/memberwatch/clusterhealth_test.go
+++ b/pkg/multicluster/memberwatch/clusterhealth_test.go
@@ -112,13 +112,13 @@ func TestAnnotationIsAdded(t *testing.T) {
 
 		go checker.WatchMemberClusterHealth(ctx, zap.S(), watchChannel, central, nil)
 
-		require.Eventually(t, func() bool {
-			got := &mdbmulti.MongoDBMultiCluster{}
-			if err := fakeClient.Get(ctx, types.NamespacedName{Name: "mdbmc", Namespace: "ns"}, got); err != nil {
-				return false
-			}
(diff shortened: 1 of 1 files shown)


lucian-tosa (approved):

Julien-Ben (approved):

m1kola:
There are some red E2Es, but these are flakes. I'm going to merge because this PR only changes one unit test.

PR #1334, opened 2026-07-03
Fix only run e2e tests on PRs when production code is touched

josvazg:
# Summary

Should fix #1301 which was ineffective.

The unit test introduced in #1301 is no longer needed as actual file patterns now are way simpler, harder to get wrong.

## Proof of Work

CI passes. Next PRs not touching production code skip e2e tests.

## Checklist


.evergreen.yml
```diff
@@ -257,7 +260,7 @@ github_pr_aliases:
     task: ".*"
   - variant_tags: [ "pr_patch_e2e" ]
     task: ".*"
-    file_pattern: '^(api|cmd|controllers|docker|internal|mongodb-community-operator|pkg|scripts)/.*\.(go|py)$|^[^/]+\.(go|py)$'
+    task: ".*"
```
MaciejKaras: nit: redundant

ci/internal/evergreen/aliases_test.go
```diff
@@ -2,14 +2,14 @@ package evergreen_test
...
-	"regexp"
+	"strings"
 	"testing"
 
 	"gopkg.in/yaml.v3"
 )
 
-func TestE2EFilePattern(t *testing.T) {
-	re := loadE2EFilePattern(t)
+func TestE2EVariantPaths(t *testing.T) {
```
MaciejKaras: It feels a little bit odd to test the evergreen logic here. I would either drop this completely or use `evergreen evaluate` command to get the full evergreen config and verify variants are configured properly.

```
evergreen evaluate - prints the given project configuration with tags and included files expanded (excluding files included from a separate module)
```

Maybe you can present a use case where this test provides value?

.evergreen.yml
```diff
@@ -1474,6 +1477,10 @@ buildvariants:
   - name: e2e_mdb_community
     display_name: e2e_mdb_community
     tags: [ "pr_patch_e2e", "staging", "e2e_test_suite" ]
+    paths:
```
MaciejKaras: Instead of copying the same pattern in every variant you could use yaml anchors we have in `.evergreen-functions.yml`:

```yaml
variables:
  - &production_code_paths
    paths:
      - "**/*.go"
      - "**/*.py"
      - "!ci/**"

buildVariants:
  - name: build_om80_images
    display_name: build_om80_images
    tags: [ "pr_patch_static", "staging" ]
    <<: *production_code_paths
    run_on:
      - ubuntu2404-small
    tasks:
      - name: build_om_images
```

ci/internal/evergreen/aliases_test.go
josvazg: makes sense the regex now are very simple

.evergreen.yml
```diff
@@ -1474,6 +1477,10 @@ buildvariants:
...
+      - "**/*.go"
+      - "**/*.py"
+      - "!ci/**"
```
MaciejKaras: We should also run e2e tests if someone modifies:

- evergreen.yml (s)
- build_info.json
- build_info_agent.json
- release.json
- go.mod
- Dockerfiles

... and probably many more. 

Before we merge this we need a documented way of triggering missing e2e tests so that engineers and agents can add them. It would be great to have label based triggers or some other way, but I know that is a long shot. For now let's just document this ignore behaviour and explain how to trigger missing tests in EVERGREEN.md

.evergreen.yml
josvazg: yep, much batter like that

.evergreen.yml
josvazg: Good point. The initial pattern is narrow (*.go / *.py) and make adding new triggers cheap, so I'd rather add more triggers on demand than up front.

For now I've documented this in EVERGREEN.md, including how to trigger e2e manually and where to add new file patterns when the need arises.

MaciejKaras (approved):
LGTM!

.evergreen.yml
```diff
@@ -11,6 +14,14 @@ include:
   - filename: .evergreen-release.yml
 
 variables:
+  - &production_code_paths
+    paths:
+      - "**/*.go"
```
lsierant: IMO this should define it otherwise: just don't trigger evg in case some non-production paths were touched. 

Specifying like this leaky. From the top of my head: you might change a core CI script in scripts/dev/ (those are sometimes without extensions) and you won't be checked for that.

josvazg: ok, let me pivot to a denylist approach instead. I will the one adding exceptions as needed instead

josvazg: done, it is a denylist now

MaciejKaras: If we are using denylist now, we should be better off using global settings rather per taskvariant.

https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#project-level-file-ignoring

.evergreen.yml
```diff
@@ -1,5 +1,16 @@
...
+# Ensure path filtering only applies to PR patches, not merge queue or mainline.
+disable_merge_queue_path_filtering: true
+
+# Never schedule any build for commits/PRs that only touch these files.
+ignore:
+  - "**/*.md"
+  - ".gitignore"
+  - "**/OWNERS"
+  - ".github/**"
+  - "changelog/**"
```
MaciejKaras: for changelogs I have mixed opinion here. If someone adds/changes a changelog file and it has impact on the operator semver, we should still trigger the precommit (part of lint-repo) to validate committed operator version in `release.json`, right? Maybe it's safer to move it to `production_code_paths`?

MaciejKaras: Other option is I thought is to move `lint-repo` task to GitHub Actions and run it always, no matter what is the change.

.evergreen.yml
josvazg: added ignored for changes that do not need any CI, like docs

.evergreen.yml
josvazg: That seems way more complex, with no upside to pay for it

.evergreen.yml
```diff
@@ -11,6 +22,15 @@ include:
   - filename: .evergreen-release.yml
 
 variables:
+  # NOTE: This is a denylist to prevent "vertigo" about missing untracked scripts.
```
lsierant: >This is a denylist to prevent "vertigo" about missing untracked scripts.

I don't understand this sentence.

josvazg: will reword

josvazg: reworded for clarity

lsierant (approved):

.evergreen.yml
MaciejKaras: Ok, but leaving `changelog/**` (and `**/*.md` which also captures the changelog.md files) will prevent the CI to run. I would keep the `**/*.md`, but revert the changelog path to also include them `"!changelog/**"`

josvazg: Done

MaciejKaras (approved):
LGTM

PR #1337, opened 2026-07-06
CLOUDP-411215: Move proxy env propagation setting to OperatorConfig

m1kola:
# Summary

As part of the Installation UX epic, operator settings are moving out of
installation-time environment variables and into the `OperatorConfig` CR so they
can be configured on a running operator. This PR migrates the proxy environment
variable propagation toggle.

`MDB_PROPAGATE_PROXY_ENV` is removed and replaced by
`.spec.proxy.envPropagationPolicy` (`Propagate`/`NoPropagation`, default
`NoPropagation`). When set to `Propagate`, the operator's `HTTP_PROXY`,
`HTTPS_PROXY` and `NO_PROXY` values are propagated onto managed database
workloads exactly as before; the proxy values themselves remain environment
variables on the operator container. Behaviour and defaults are unchanged.

## Proof of Work

- Unit tests pass locally: `pkg/operatorconfig` (proxy defaulting in
  `withDefaults`), `controllers/operator/construct` (proxy env construction),
  and standalone/replicaset reconciler tests.
- `make manifests generate` produces no CRD/deepcopy diff — the
  `proxy.envPropagationPolicy` schema already existed in the CRD.
- e2e: pending — `e2e_operator_proxy` (updated to enable propagation via the
  OperatorConfig CR) to be run via a targeted Evergreen patch.

## Checklist

### DIFF
--- a/changelog/20260706_breaking_proxy_envpropagation_moved_to_operatorconfig.md
+++ b/changelog/20260706_breaking_proxy_envpropagation_moved_to_operatorconfig.md
@@ -1,0 +1,6 @@
+---
+kind: breaking
+date: 2026-07-06
+---
+
+* **Operator**: The `MDB_PROPAGATE_PROXY_ENV` environment variable has been removed. Configure propagation of the operator's `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` environment variables onto managed database workloads using `.spec.proxy.envPropagationPolicy` (`Propagate`/`NoPropagation`) in the `OperatorConfig` CR instead. The default behaviour remains unchanged (`NoPropagation`).
--- a/controllers/operator/authentication_test.go
+++ b/controllers/operator/authentication_test.go
@@ -48,7 +48,7 @@ func TestX509CanBeEnabled_WhenThereAreOnlyTlsDeployments_ReplicaSet(t *testing.T
 
 	addKubernetesTlsResources(ctx, kubeClient, rs)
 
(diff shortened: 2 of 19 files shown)


josvazg:
Follow up: We might consider for refactor controller constructors to take an options/config struct instead of positional args — The new*Reconciler / Add*Controller / *ReconcilerHelper signatures have accumulated many adjacent positional bools (forceEnterprise, enableClusterMongoDBRoles, agentDebug, …, propagateProxyEnv, automaticRecoveryEnabled), which are easy to transpose.

As we keep moving settings into OperatorConfig, thread a struct (or pass operatorCfg.Spec / a derived subset) so new fields don't require editing every signature + test call site.

Not an issue for this PR, but maybe we could tackle this next, before we add more changes like this.

josvazg (approved):
LGTM

m1kola:
> Follow up: We might consider for refactor controller constructors to take an options/config struct instead of positional args

@josvazg  Yes, this is getting out of hand.

> As we keep moving settings into OperatorConfig, thread a struct (or pass operatorCfg.Spec / a derived subset)

I would prefer not to pass a struct. Maybe using `opts` pattern where we pass options to apply. But not sure yet.



> Not an issue for this PR, but maybe we could tackle this next, before we add more changes like this.

For now I'm trying to focus on implementing changes. I'll see how things go and might refactor this before we merge into master, if we have time for this.

roothorp (approved):

PR #1338, opened 2026-07-06
CLOUDP-411215: Move telemetry settings to OperatorConfig

m1kola:
# Summary

Part of the Installation UX epic: operator settings are being moved off Helm
values and environment variables onto the `OperatorConfig` CR so a single base
deployment can be reconfigured post-installation across Helm, OLM and raw YAML.
This PR migrates the telemetry configuration.

Telemetry is now configured via `OperatorConfig.spec.telemetry`: master `mode`,
`collection.frequency`, `collection.kubeTimeout`, per-type
`collection.{clusters,deployments,operators}.mode`, `send.mode` and
`send.frequency`. The `operator.telemetry.*` Helm values and the
`MDB_OPERATOR_TELEMETRY_*` environment variables are removed; the operator reads
the resolved config once at startup and the existing OperatorConfig watcher
restarts it on change. Telemetry remains opt-out (absence of config keeps it
enabled) and all default values are unchanged. `MDB_OPERATOR_TELEMETRY_KUBE_TIMEOUT`
is promoted to `spec.telemetry.collection.kubeTimeout`;
`MDB_OPERATOR_TELEMETRY_SEND_BASEURL` is retained as an internal, test-only knob
(settable via `operator.telemetry.send.baseUrl`).

## Proof of Work

- Unit tests `go test ./pkg/telemetry/... ./pkg/operatorconfig/...` pass, covering
  telemetry defaulting in `withDefaults`, the resolved `Config` threading, and the
  send-frequency / kube-timeout behaviour.
- `helm template` renders with no telemetry env vars by default, emits
  `MDB_OPERATOR_TELEMETRY_SEND_BASEURL` only when `operator.telemetry.send.baseUrl`
  is set, and still deploys the telemetry ClusterRole.
- e2e: pending — to be exercised via a targeted Evergreen patch on this draft.

## Checklist

---

### DIFF
--- a/api/operator/v1/operatorconfig_types.go
+++ b/api/operator/v1/operatorconfig_types.go
@@ -136,6 +136,13 @@ type TelemetryCollectionConfig struct {
 	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('1m')",message="frequency must be at least 1m"
 	Frequency *metav1.Duration `json:"frequency,omitempty"`
 
+	// KubeTimeout is the timeout for Kubernetes API calls made while collecting telemetry.
+	// Duration using h (hours), m (minutes) and s (seconds) units, e.g. "30s", "5m". Must be at least 1s; defaults to 5m when omitted.
+	// +optional
+	// +kubebuilder:default="5m"
+	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('1s')",message="kubeTimeout must be at least 1s"
+	KubeTimeout *metav1.Duration `json:"kubeTimeout,omitempty"`
+
 	// Clusters controls collection of cluster-level telemetry.
 	// +optional
 	Clusters *TelemetryCollectionClustersConfig `json:"clusters,omitempty"`
--- a/api/operator/v1/zz_generated.deepcopy.go
+++ b/api/operator/v1/zz_generated.deepcopy.go
@@ -348,6 +348,11 @@ func (in *TelemetryCollectionConfig) DeepCopyInto(out *TelemetryCollectionConfig
 		*out = new(metav1.Duration)
 		**out = **in
 	}
+	if in.KubeTimeout != nil {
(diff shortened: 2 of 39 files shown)


josvazg:
Should we double check for left over usages of `MDB_OPERATOR_TELEMETRY_SEND_ENABLED` ?
maybe also
`MDB_OPERATOR_TELEMETRY_SEND_ENABLED`
`MDB_OPERATOR_TELEMETRY_COLLECTION_FREQUENCY`

josvazg:
> The changelog claims operator.telemetry.* Helm values "have been removed," and values.yaml dropped the block. But helm_chart/templates/operator-roles-telemetry.yaml:2-6 still gates the telemetry ClusterRole/binding on:
> {{- $telemetry := default dict .Values.operator.telemetry }}
> {{- if ne $telemetry.enabled false }}
>   {{- if ne $telemetry.installClusterRole false}}
> Default behavior is preserved (empty dict → both ne … false are true → ClusterRole still created), so it doesn't break rendering. But it's now inconsistent: telemetry enable/disable moved to the CR, yet the RBAC is still tied to a removed Helm value. Either this template should be reconciled with the new model, or the changelog should note enabled/installClusterRole still exist for RBAC gating.

josvazg:
> scripts/dev/print_operator_env.sh still emits removed vars (minor, dev-only)

> Lines 28-29 still set MDB_OPERATOR_TELEMETRY_COLLECTION_FREQUENCY and MDB_OPERATOR_TELEMETRY_SEND_ENABLED, which are now ignored — so local dev runs silently change telemetry behavior (send defaults back to enabled). Worth cleaning up so dev env matches the new config path.

m1kola:
I'm moving back to draft. `e2e_multi_cluster_om_appdb_no_mesh` is failing and I suspect it might be a legitimate failure.

+ need to address feedback above.

josvazg (approved):

roothorp (approved):

PR #1285, opened 2026-06-28
[KUBE-156] Add per cluster component status in Search resource

viveksinghggits:
# Summary

This PR adds a new field `status.clusters` in the MongoDBSearch resource to make sure that we can inform users about the per cluster status of search sts, load balancer and metrics forwarder components.

Here are some of the examples of how this status would look like for some scenarios
```
status:
  clusters:
  - index: 0
    loadBalancer: Running
    name: kind-e2e-cluster-1
    search: Pending
    searchMessage: StatefulSet not ready
  - index: 1
    loadBalancer: Running
    name: kind-e2e-cluster-2
    search: Pending
    searchMessage: StatefulSet not ready

status:
  clusters:
  - index: 0
    loadBalancer: Pending
    loadBalancerMessage: 'Load balancer deployment mdb-mc-rs-ext-lb-search-search-lb-0
      not ready: 0/1 replicas ready'
    name: kind-e2e-cluster-1
    search: Pending
    searchMessage: StatefulSet not ready
  - index: 1
    loadBalancer: Pending
    loadBalancerMessage: 'Load balancer deployment mdb-mc-rs-ext-lb-search-search-lb-1
      not ready: 0/1 replicas ready'
    name: kind-e2e-cluster-2
    search: Pending
    searchMessage: StatefulSet not ready

status:
  clusters:
  - index: 0
    loadBalancer: Running
    metricsForwarder: Running
    search: Pending
    searchMessage: StatefulSet not ready

status:
  clusters:
  - index: 0
    loadBalancer: Running
    metricsForwarder: Running
    search: Running

status:
  clusters:
  - index: 0
    loadBalancer: Running
    metricsForwarder: Pending
    metricsForwarderMessage: 'Metrics forwarder deployment mdb-rs-search-metrics-forwarder-0
      not ready: 0/1 replicas ready'
    search: Pending
    searchMessage: StatefulSet not ready
```

To make sure that `status.clusters` is not updated by different controller search controller reads the status from the clusters and then accumulates them in proper format. This would makes sure that different controllers are not overriding each other's changes.

## Proof of Work

E2E is part of [this PR](https://github.com/mongodb/mongodb-kubernetes/pull/1340). 

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you added changelog file?
    - use `skip-changelog` label if not needed
[...]

PR #1340, opened 2026-07-06
[KUBE-162] Add E2E test for multi cluster status in search resource

viveksinghggits:
# Summary

This PR adds [E2E test](https://github.com/mongodb/mongodb-kubernetes/pull/1285) for the change that add a field `status.clusters` in `MongoDBSearch` CR to have cluster level status of search components.

## Proof of Work

Successful CI.

## Checklist

---

### DIFF
--- a/docker/mongodb-kubernetes-tests/kubetester/mongodb_search.py
+++ b/docker/mongodb-kubernetes-tests/kubetester/mongodb_search.py
@@ -101,20 +101,20 @@ class MongoDBSearch(MongoDB, CustomObject):
             logger.info(f"MongoDBSearch {self.name}: loadBalancer status correctly absent")
 
     def get_cluster_statuses(self) -> list[dict]:
-        """Returns the status.clusterStatuses list, or [] if absent.
+        """Returns the status.clusters list, or [] if absent.
 
         The search controller is the sole writer and recomputes the whole list
-        every reconcile (one entry per spec.clusters[], keyed by clusterIndex).
+        every reconcile (one entry per spec.clusters[], keyed by index).
         """
         try:
-            return list(self["status"]["clusterStatuses"] or [])
+            return list(self["status"]["clusters"] or [])
         except KeyError:
             return []
 
     def get_cluster_status(self, cluster_index: int) -> Optional[dict]:
-        """Returns the per-cluster status entry with the given clusterIndex, or None."""
+        """Returns the per-cluster status entry with the given index, or None."""
         for cs in self.get_cluster_statuses():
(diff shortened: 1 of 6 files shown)


Julien-Ben (approved):

docker/mongodb-kubernetes-tests/kubetester/mongodb_search.py
```diff
@@ -100,6 +100,173 @@ def assert_lb_status(self):
...
+        self.load()
+        statuses = self.get_cluster_statuses()
+        spec_indexes = self._spec_cluster_indexes()
+
+        want_count = expected_count if expected_count is not None else len(spec_indexes)
+        assert len(statuses) == want_count, f"expected {want_count} clusters entries, got {len(statuses)}: {statuses}"
+
+        got_indexes = [cs.get("index") for cs in statuses]
+        assert len(set(got_indexes)) == len(got_indexes), f"duplicate index in status.clusters: {got_indexes}"
+        if expected_count is None:
```
anandsyncs: Nit: Every new caller passes `expected_count`, so this skips the check that status indexes match `spec.clusters`. Cluster names are not checked either, which means a swapped `(name, index)` mapping can still pass. Could we always compare the exact `(name, index)` pairs from spec and status, and use `expected_count` only as an additional count check?

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -1116,3 +1124,145 @@ def execute_vector_search(_idx: int = cluster_index) -> tuple:
...
+    We need this to set unreasonable memeory request to simulate the faulty envoy deployment to test the search status.
+    """
+    mdbs.load()
+    managed = mdbs["spec"]["clusters"][list_pos]["loadBalancer"]["managed"]
+    if memory is None:
+        managed.pop("resourceRequirements", None)
+        managed.pop("deployment", None)
+    else:
+        managed["resourceRequirements"] = {"requests": {"memory": memory}}
+        managed["deployment"] = {"spec": {"strategy": {"type": "Recreate"}}}
```
anandsyncs: Forcing `Recreate` here hides the Deployment rollout case discussed in #1285. Production uses the default `RollingUpdate`, which can keep the old Envoy pod ready while the new pod is unschedulable; the current `ReadyReplicas`-only check then continues reporting `Running`. Once #1285 checks `UpdatedReplicas` and `ObservedGeneration`, could we remove this override so the E2E exercises the production rollout behavior?

anandsyncs (approved):

PR #1344, opened 2026-07-07
KUBE-154: don't set cross-cluster owner references on member-cluster resources

anandsyncs:
## Summary

Kubernetes garbage collection does not span clusters. In multi-cluster MongoDBSearch, several per-cluster write paths were unconditionally setting an `OwnerReference` back to the central `MongoDBSearch` CR on resources written to **member** clusters (mongot StatefulSets, Services, ConfigMaps, and operator-managed TLS secrets). Since the CR's UID only exists in the central cluster's etcd, the member cluster's own GC controller resolves the owner reference, finds no matching object, and deletes the child resource within ~1 second — an infinite create/delete crash loop. Verified on real (multi-apiserver) EKS clusters; this does not reproduce in the kind e2e harness, which shares a single apiserver across all MC test contexts.

Fix mirrors the guard already shipped in `MongoDBSearchEnvoyReconciler`: only call `SetOwnerReference`/`SetOwnerReferences` when writing to the central cluster (`clusterName == ""`). Member-cluster resource lifecycle continues to rely on the existing component-label sweep, unchanged.

- `createOrUpdateStatefulSet`, `ensureSearchService`, `ensureMongotConfig` — guarded.
- `buildHeadlessService`, `buildProxyService`, `buildClusterLevelProxyService` — these Service builders had their own separate unconditional `SetOwnerReferences` call (the actual site where the Service owner ref was set); now guarded.
- `ensureIngressTlsConfig` / `ensureX509ClientCertConfig` / `ensureEgressTlsConfig` → `tls.EnsureTLSSecret` — threaded `clusterName` down and guarded.
- `search_construction.go`'s `CreateSearchStatefulSetFunc` had a redundant unconditional owner-ref call alongside `createOrUpdateStatefulSet`'s own (correct) one — removed rather than guarded, since the latter is the sole choke point that persists the StatefulSet.
- `search_state.go`'s state ConfigMap owner-ref call is untouched: that ConfigMap is only ever written via the central client, never fanned out per-cluster.

Fixes KUBE-154.

## Topology behavior

- **Single-operator (vanilla) installs**: `spec.clusters` is mandatory (minItems=1) and the single entry may be **named**. An operator with no `OPERATOR_CLUSTER_NAME` and no member-cluster clients can only write locally (`clientForCluster` falls back to the central client for every name when the member map is empty), so `prepareSearch` now blanks the in-memory cluster names in that mode: a named single entry behaves exactly like `clusters: [{}]` — same index-based resource names, owner references present, CR-delete GC (including PVC reclaim) intact. Without this normalization the name-based guard would silently drop owner refs for named single-cluster installs.
- **Hub-and-spoke multi-cluster**: member-cluster resources (`clusterName != ""`) no longer carry the owner reference — the CR's UID doesn't exist in the member apiserver, so GC was deleting them (the bug above).
- **Simulated MC (one operator per cluster)**: also drops owner references. Here the CR is local to the cluster, so GC would actually have worked — dropping them anyway is intentional, for consistency with the Envoy controller and the label-sweep-based lifecycle (CR-delete cleanup arrives separately).
- **Upgrades / pre-existing resources**: existing resources keep whatever owner references they already carry — the StatefulSet/ConfigMap mutate is in-place and does not touch `ownerReferences`. That is moot for hub-and-spoke: multi-cluster reconcile is feature-flag-gated (`MDB_SEARCH_ENABLE_MULTI_CLUSTER`, default off), so no GA install has member-cluster resources with stale cross-cluster refs. And it is benign-or-useful for simulated MC: pre-existing refs there are local and valid, so CR-delete GC keeps working for existing objects while newly created objects converge to the label-managed lifecycle.

## Test plan

- [x] `go build ./...`
- [x] `go test ./controllers/searchcontroller/... ./controllers/operator/... ./pkg/tls/... ./api/mongodb/v1/search/...` — all green
- [x] `gofmt -l` on changed files — clean
- [x] Extended `TestReconcilePlan_UsesPerClusterClient` to assert member-cluster StatefulSet/ConfigMap/Service objects carry no owner reference
- [x] Extended `TestEnsureX509ClientCertConfig_MongotAndStsModification` to assert central-cluster secret keeps its owner ref while a member-cluster secret does not
- [x] `assertSearchOwnerLabels` (controller tests) now also asserts zero `ownerReferences` on every non-legacy object it checks — covers the hub-and-spoke controller path AND the operator-per-cluster (simulated-MC/`LocalizeToCluster`) path for mongot STS/ConfigMap/proxy/headless/cluster-level proxy Service plus the Envoy Deployment/ConfigMap
- [x] Member-cluster cluster-level proxy Service asserted ref-free in `TestReconcileShardedMC_FanOutUsesPerClusterClient`
- [x] Ingress TLS operator secret asserted through the full reconcile threading: ref-free on the member cluster (`TestReconcileShardedMC_AllUnitsAppliedBeforeReadinessCheck`) and ref-PRESENT on the legacy central cluster (`TestReconcileSharded_RoutingSwitchOneWay`)
- [x] Legacy ref-PRESENT side completed in `TestReconcileReplicaSet_CreatesResources`: headless Service and mongot ConfigMap keep their `MongoDBSearch` owner ref (STS was already asserted)
- [x] Negative sanity: temporarily inverting the STS and `EnsureTLSSecret` guards makes the new assertions fail as expected
- [x] `TestMongoDBSearchReconcile_Success` gained a `named single cluster entry behaves as unnamed` row; every row now runs `assertSearchOwnerLabels` on STS + headless/proxy Services + mongot ConfigMap, and the helper now asserts an owner ref is PRESENT when `clusterName == ""` (hardens all legacy call sites)
- [x] Envoy: owner-ref-present assertions folded into `TestReconcile_RoutingReadyFromState_DrivesFallbackRoutes` (vanilla operator + named single entry → Envoy ConfigMap/Deployment carry the ref)
- [x] Metrics: `TestReconcile_EnterpriseSource_CreatesDeploymentAndConfigMap` switched to a named single entry; forwarder Deployment/ConfigMap assert the ref
- [x] Negative sanity: disabling the `prepareSearch` normalization block fails all three extended tests
- [x] e2e: `q2_mc_rs_steady` (hub-and-spoke) and `simulated_mc_rs` (operator-per-cluster) now assert `metadata.ownerReferences` is empty on each member/local cluster's mongot StatefulSet, headless + proxy Services, mongot ConfigMap, and the operator-managed TLS secret. kind's shared apiserver can't reproduce the GC deletion itself, but these pin the contract (no refs written) on a real cluster.

skip-ci


changelog/20260707_fix_mongodbsearch_member_cluster_owner_references.md
```diff
@@ -0,0 +1,6 @@
+---
```
anandsyncs: remove this file

anandsyncs: Removed in e0a5466c2.

anandsyncs:
Superseded by the KUBE-154 lifecycle stack: #1382 implements the owner-reference locality policy (local cluster keeps canonical refs as GC backstop, member clusters get none) at every search write site. Closing.

PR #1345, opened 2026-07-07
CLOUDP-411215: Set maxConcurrentReconciles via OperatorConfig in perf tests

m1kola:
# Summary

`maxConcurrentReconciles` was moved to the `OperatorConfig` CR, so the operator no longer reads the `MDB_MAX_CONCURRENT_RECONCILES` env var or the `operator.maxConcurrentReconciles` Helm value. The e2e operator perf variants still set `MDB_MAX_CONCURRENT_RECONCILES` (10 / 30), but nothing wired it through anymore — so they were silently running at the default concurrency of `1`, no longer measuring what they were meant to.

This PR restores that wiring through the established test pattern (the one `defaultArchitecture` uses):

- `single_e2e.sh` passes `MDB_MAX_CONCURRENT_RECONCILES` to the test pod (only when set), and the test-app template renders it conditionally.
- `build_operator_config_spec_from_test_env` reads it and puts it into the `OperatorConfig` CR spec, which the operator install applies.
- The value is set only where needed: the global default in `root-context` and the redundant one-thread overrides are removed, so non-perf tests run at the operator default with **no** CR and **no** operator restart.

It also removes the dead `operator.maxConcurrentReconciles` Helm value and the leftover script/env plumbing from when the setting lived on the operator `Deployment`.

## Proof of Work

- Evergreen perf run: https://spruce.corp.mongodb.com/version/6a4ccaf7da6f370007c4be70

## Checklist

- [x] Have you checked whether your jira ticket required DOCSP changes? (not required)
- [x] Have you added changelog file? (`skip-changelog` — test/CI-only change)

---


docker/mongodb-kubernetes-tests/kubetester/kubetester.py
```diff
@@ -135,6 +135,9 @@ def build_operator_config_spec_from_test_env() -> dict:
     spec: dict = {}
     if is_default_architecture_static():
         spec["defaultArchitecture"] = "Static"
+    max_concurrent_reconciles = os.getenv("MDB_MAX_CONCURRENT_RECONCILES")
+    if max_concurrent_reconciles:
+        spec["maxConcurrentReconciles"] = int(max_concurrent_reconciles)
```
josvazg, line 140: **nitpick**: No int() validation. int(max_concurrent_reconciles) in kubetester.py:139 will raise ValueError if the env var is ever a non-numeric string. Given it only flows from controlled context files this is acceptable, but a malformed value would fail with a raw traceback rather than a clear message.

josvazg (approved):

roothorp (approved):

PR #1346, opened 2026-07-07
KUBE-157: sweep stale per-shard TLS secrets on member clusters

anandsyncs:
## Summary

Stacked on top of #1344 (KUBE-154).

Found during code review of KUBE-154: the stale-shard sweep (`cleanupStaleShardResources`) reaps per-shard StatefulSet/Service/ConfigMap left behind by shard scale-down, but not `Secret`. Now that KUBE-154 correctly stops setting a cross-cluster owner reference on member-cluster resources, the per-shard ingress TLS secret (built via `pkg/tls/tls_secret.go`'s `EnsureTLSSecret`, `perShardTLSResource` adapter) has no cleanup path left on member clusters when its shard is removed:

- Before KUBE-154: a dangling owner reference meant the member cluster's own GC controller eventually deleted the secret as a side effect of the crash-loop bug.
- After KUBE-154: that's correctly gone, but nothing replaced it — the secret leaks permanently on shard removal.

This PR closes that gap the same way ConfigMap already is: stamp the secret with the owner + component labels (`searchOwnerLabels` + `componentLabelKey: mongotComponent`), and add it as a 5th entry to the `sweeps` slice.

Scoped to the per-shard **ingress** TLS secret only — the x509 client-cert and SCRAM client-cert secrets are single, CR-scoped (not per-shard-named), so shard-removal sweeping doesn't apply to them; both call sites now pass `nil` labels and are otherwise unaffected.

Fixes KUBE-157.

## Test plan

- [x] `go build ./...`
- [x] `go test ./controllers/searchcontroller/... ./pkg/tls/... ./controllers/operator/...` — all green
- [x] `gofmt -l` / `go vet` — clean
- [x] Extended `TestCleanupStaleShardResources` (single-cluster) and `TestCleanupStaleShardResources_MCFanOut` (2-cluster MC) with a stale/live/different-owner TLS secret fixture, mirroring the existing ConfigMap assertions
- [x] e2e coverage: `e2e_search_sharded_scale_shards_managed_lb`'s `verify_shard_resources_deleted` now also polls the removed shard's operator TLS secret (`<name>-search-0-<shard>-certificate-key`, per `TLSOperatorSecretForClusterShard`) to 404, with a matching existence assertion after scale-up so the absence check can't pass vacuously

## CI

Rebased onto the new #1344 head (changelog entry + comment trim) and force-pushed, which retriggers the Evergreen run — the previous run was aborted.

Restacked again onto the latest #1344 head (`9e7ed3dec`, vanilla cluster-name normalization); no changes to this PR's own three commits.

skip-ci

### DIFF
--- a/controllers/searchcontroller/mongodbsearch_reconcile_helper.go
+++ b/controllers/searchcontroller/mongodbsearch_reconcile_helper.go
@@ -802,10 +802,11 @@ func (r *MongoDBSearchReconcileHelper) pruneRoutingReady(ctx context.Context, li
 }
 
 // cleanupStaleShardResources reaps the per-shard mongot resources of shards that no
-// longer exist (shard scale-down): proxy Service, headless Service, ConfigMap, and the
-// mongot StatefulSet — whose whenDeleted=Delete policy then cascades its PVC. For every
-// live shard it rebuilds the expected resource names; anything we own carrying the
-// matching scope label but not in those sets belongs to a removed shard and is deleted.
+// longer exist (shard scale-down): proxy Service, headless Service, ConfigMap, ingress
+// TLS Secret, and the mongot StatefulSet — whose whenDeleted=Delete policy then cascades
+// its PVC. For every live shard it rebuilds the expected resource names; anything we own
+// carrying the matching scope label but not in those sets belongs to a removed shard and
+// is deleted.
 //
 // It fans out over the central client and every member client; per-kind/per-cluster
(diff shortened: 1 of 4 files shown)


anandsyncs:
Superseded by the KUBE-154 lifecycle stack: #1382 stamps owner labels on the per-shard operator TLS secrets and reaps stale ones via the stale-shard sweep — same approach this PR introduced. Closing.

PR #1348, opened 2026-07-07
CLOUDP-420747: Track backport branches

josvazg:
# Summary

Add initial backporting track info and supporting command.

The YAML allows us to track our backporting branches and the tool queries it.

**Note**: old-master is totally fictional at this point. Still it might be used later on for testing before we actually have a real public backport branch to go to.

## Proof of Work

CI unit tests included and:

```shell
$ ./scripts/mckci backport from master
old-master

$ ./scripts/mckci backport from old-master
$

$ ./scripts/mckci backport from main
Error: "main": branch is not a tracked backporting branch

$ ./scripts/mckci backport from master --json |jq .
{
  "name": "old-master",
  "description": "Test backporting branch",
  "version": "0.x",
  "internal-only": true
}

```

## Checklist

---

### DIFF
--- a/ci/backporting.yaml
+++ b/ci/backporting.yaml
@@ -1,0 +1,10 @@
+# Backporting branches information
+branches:
+  - name: master
+    description: Master branch for latest version
+    version: 1.x
+  - name: old-master
+    description: Test backporting branch
+    version: 0.x
+    internal-only: true
+
--- a/ci/config.go
+++ b/ci/config.go
@@ -1,0 +1,11 @@
+// Package ci embeds repo-level CI configuration so mckci can read it without
+// resolving filesystem paths at runtime. The embedded data reflects the commit
+// the binary was built from.
+package ci
+
+import _ "embed"
+
+// BackportingYAML is the raw contents of ci/backporting.yaml.
+//
+//go:embed backporting.yaml
+var BackportingYAML []byte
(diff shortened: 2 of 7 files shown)


ci/backporting.yaml
```diff
@@ -0,0 +1,10 @@
+# Backporting branches information
```
m1kola, line 1: Do we need to keep this file updated across all backporting branches?

MaciejKaras:
@josvazg I have general question: why do we need to track backported branches? Is there a TD with the design and requirements where it's explained more generally?

ci/backporting.yaml
josvazg: yes, it is only updated after very major release, should not be too often.

josvazg:
> @josvazg I have general question: why do we need to track backported branches? Is there a TD with the design and requirements where it's explained more generally?

not at the moment, but I can add a section in the TD. It is quite simple though

josvazg:
Making it a draft, seems more controversial than expected. Will clarify before proceeding.

josvazg:
We finally decided to drop this for now and move it as a nice to have or YAGNI. We will manually cherry pick bug fixes and therefore automation does not need to track the backporting branches.

Alternative, we can just alter on bugfixes merging to avoid missing them, but not automate the PR creation.

PR #1349, opened 2026-07-07
KUBE-122: Cap default mongot JVM heap at 30GB

Julien-Ben:
## Summary

The default JVM heap for mongot (half of the memory request) is now capped at 30GB, per the [mongot sizing guidance](https://www.mongodb.com/docs/manual/tutorial/mongot-sizing/advanced-guidance/hardware/#jvm-heap-sizing). Above ~30GB the JVM loses compressed object pointers and mongot takes a performance hit.

User-provided heap flags are not affected.

## Proof of work

Unit tests pass, including new cases at and above the cap.

### DIFF
--- a/changelog/20260707_fix_mongodbsearch_default_jvm_heap_capped_at_30gb.md
+++ b/changelog/20260707_fix_mongodbsearch_default_jvm_heap_capped_at_30gb.md
@@ -3,4 +3,4 @@ kind: fix
 date: 2026-07-07
 ---
 
-* **MongoDBSearch**: The default JVM heap size (half of the memory request) is now capped at 30GB, following the [mongot sizing guidance](https://www.mongodb.com/docs/manual/tutorial/mongot-sizing/advanced-guidance/hardware/#jvm-heap-sizing). Heap sizes above ~30GB prevent the JVM from using compressed object pointers and degrade performance. User-provided heap flags are not affected.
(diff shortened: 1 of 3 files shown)


changelog/20260707_fix_mongodbsearch_default_jvm_heap_capped_at_30gb.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2026-07-07
+---
+
+* **MongoDBSearch**: The default JVM heap size (half of the memory request) is now capped at 30GB, following the [mongot sizing guidance](https://www.mongodb.com/docs/manual/tutorial/mongot-sizing/advanced-guidance/hardware/#jvm-heap-sizing). Heap sizes above ~30GB prevent the JVM from using compressed object pointers and degrade performance. User-provided heap flags are not affected.
```
lucian-tosa: The same guidance mentions that 48gb+ is also a good value for this, but we are capping it at 30. Does this need additional consideration, or is the jvm flags workaround acceptable?

lsierant: @lucian-tosa great catch! 
@Julien-Ben let's add after "User-provided heap flags are not affected, so if more than 30GB heap is required, use jvmFlags" or similar

lsierant (approved):

changelog/20260707_fix_mongodbsearch_default_jvm_heap_capped_at_30gb.md
Julien-Ben: True, it's a bit confusing these recommendations. I also followed what we do in Atlas, but I've put more details only in Jira: https://jira.mongodb.org/browse/KUBE-122

Julien-Ben: I added the recommendation to use jvmflags if more is needed

lucian-tosa (approved):

PR #1358, opened 2026-07-08
CLOUDP-407997: Publish MCK promoted MCK images

josvazg:
# Summary

Prepare CI tool to publish previously promoted images and unify naming.

New `mckci release publish` and `mckci release publish images` commands for publishing single and all release images at a specific commit (or promoted-latest). Stomp protection added to `mckci release promote images`.

## Stomp protection (both promote images + publish images)

Every image's immutable tag is checked for conflicts BEFORE any writes. If any image conflicts and `--force` is not set, the entire operation is refused — no partial writes.

## Latest Marker

To support backport branches you might want to use different "latest" markers with `--latestMarker` such as `latest`, `latest-v1`, etc.

## Examples

```bash
# Publish all release images from latest promoted image/commit
mckci release publish images 

# Publish all release images at a specific commit
mckci release publish images \
  --commit a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2

# Publish single image (omit --commit for promoted-latest)
mckci release publish \
  --staging-image quay.io/mongodb/staging/mongodb-kubernetes \
  --commit a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2

# Promote all release images with force override
mckci release promote images --commit a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2 --force

# Dry-run all images
mckci release promote images --commit a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2 --dry-run
```

## Proof of Work

CI must pass, changes fully self tested.

## Checklist

### DIFF
--- a/ci/internal/cli/release_promote.go
+++ b/ci/internal/cli/release_promote.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -23,19 +24,29 @@ func newPromoteReleaseCmd() *cobra.Command {
 
 func newPromoteImageCmd() *cobra.Command {
 	var (
-		image       string
-		commit      string
-		version     string
-		registryURL string
-		repo        string
-		dryRun      bool
+		image        string
+		commit       string
+		version      string
+		registryURL  string
+		repo         string
+		latestMarker string
+		force        bool
+		dryRun       bool
 	)
 
 	cmd := &cobra.Command{
 		Use:   "image",
 		Short: "Promote a single candidate image",
+		Long: `Promotes a single candidate image, tagging it promoted-{commit}-{version}
+and promoted-{latest-marker} (--latest-marker is required).
+
+The promoted-{latest-marker} tag is a mutable pointer that always moves.
+When promoting a backport (e.g. patching an older release branch after a
+newer version has already been promoted), pass a distinct --latest-marker
+(e.g. "latestv1") so the backport doesn't steal the "latest" pointer away
+from the newest promoted version.`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			return promoteImage(cmd, image, commit, version, registryURL, repo, dryRun)
+			return promoteImage(cmd, image, commit, version, registryURL, repo, latestMarker, force, dryRun)
 		},
 	}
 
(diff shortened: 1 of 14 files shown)


josvazg:
Need to fix promotion first upstream, we need to promote and release all associated images not just the operator one.

ci/internal/cli/release_publish.go
```diff
@@ -0,0 +1,153 @@
...
+		dryRun       bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "image",
+		Short: "Publish a single promoted candidate image",
+		Long: `Resolves the promoted candidate for the given commit (or the latest promoted
+commit if --commit is omitted), then retags it in the production registry as
+:{version} and :latest. The version is derived from the promoted-{commit}-{version}
+tag already present in the staging registry — no --version flag is needed.
```
m1kola: Is it a behaviour we want with backporting?

So if we have 2.0.1 right now as `latest` and then publish 2.0.2 it becomes `latest`. So far so good. But I I understand the description correctly, backporting something to 1.x and say publishing 1.10.1 - will move `latest` to that version.

Is my understanding correct? If yes - then I think it is not what should be happening and `latest` should always post to the highest version possible within the current branch. In the example above it will be 2.0.2 because it is the latest version despite 1.10.1 being the latest in the timeline of releases.

ci/internal/release/images_promote.go
```diff
@@ -6,57 +6,79 @@ import (
...
+		if err != nil {
+			return nil, fmt.Errorf("check %s (%s): %w", img.Name, srcRef, err)
+		}
+		if conflict != "" {
+			conflicts = append(conflicts, fmt.Sprintf("%s: %s", img.Name, conflict))
+		}
+		prep = append(prep, prepared{img: img, reg: reg, host: host, repoPath: path, srcRef: srcRef, versionTag: versionTag})
+	}
+	if len(conflicts) > 0 && !force {
+		return nil, fmt.Errorf("refusing to promote images; tag conflicts found (use --force to override):\n%s", strings.Join(conflicts, "\n"))
```
m1kola: nit: I would not encourage use of `--force`

ci/internal/release/publish.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// Immutable tag first: only once that succeeds do we move :latest, so a
+	// failure never leaves latest ahead of the version it's supposed to point at.
+	if err := reg.CopyWithTags(srcRef, inputs.ProdRepo, []string{version}); err != nil {
+		return PublishResult{}, fmt.Errorf("publish %s: %w", srcRef, err)
+	}
+	if err := reg.CopyWithTags(srcRef, inputs.ProdRepo, []string{"latest"}); err != nil {
+		return PublishResult{}, fmt.Errorf("publish %s (latest): %w", srcRef, err)
+	}
```
m1kola: Related to https://github.com/mongodb/mongodb-kubernetes/pull/1358/changes#r3614108429

Doesn't seem right in the context of backporting: If we have 2.x as a current version, publishing in 1.x must not move `latest` to that image.

m1kola (changes requested):
I think something is wrong around logic with `latest`. See questions below.

ci/internal/cli/release_publish.go
josvazg: Yes I think we are going to need more than one "latest" I can add a flag to pass a custom "latest" marker such as `latestv1`.

The latest marker is useful as it allows to pin the most up to date promoted image without knowing the commit in advance. But you are right for backports we need to extend this a bit.

My proposal is a new optional flag names `--latest-marker` (or similarly named) that defaults to latest but we can set to "latestv1", "latestv2" etc the "latest" remains the marker for the current release.

ci/internal/release/images_promote.go
josvazg: I can soften the wording, but still, the flag exists and is a separate thing to ensure safe usage. The assumption is if you are using it, is because you must.

`double check and use --force to override if required` would be test here?

ci/internal/release/publish.go
josvazg: agreed, we will be able to have several "latest" pointers.

ci/internal/release/images_promote.go
m1kola: My logic is that if you must use it - you are aware of it or you at least looked at `--help`. So it is more of a deliberate decision because it requires some research before using it.

But don't worry about this comment too much. it is ok as is. Not a big deal.

ci/internal/cli/release_publish.go
m1kola: I see. If we need to have some sort of latest marker, I think we can use major version for that. E.g. if 2.0.2 is the latest, it will be tagged with `2.0.2`, `2.x` (or just `2`), `latest` and anything in 1.x tagged with `1.<minor>.<patch>` (say `1.10.1` from example above), `1.x` (or just `1`) and that's it.

josvazg: At the tool level we can accept anything. This PR is at the tool level, the wiring is a separate PR. We can discuss conventions there.

ci/internal/cli/release_promote.go
```diff
@@ -44,6 +55,8 @@ func newPromoteImageCmd() *cobra.Command {
 	cmd.Flags().StringVar(&version, "version", "", "version to encode in the promoted tag")
 	cmd.Flags().StringVar(&registryURL, "registry", "https://quay.io", "target OCI registry base URL")
 	cmd.Flags().StringVar(&repo, "repo", "mongodb/mongodb-kubernetes-operator", "target image repository")
+	cmd.Flags().StringVar(&latestMarker, "latest-marker", "latest", "marker used for the mutable promoted-{marker} pointer tag (use a distinct value, e.g. \"latestv1\", for backports)")
```
m1kola: Should we make it harder to shoot ourselves in the foot? I feel like having this as a required flag and not setting a default value will help to prevent accidental override of `latest` to an undesired version (e.g. when backporting to 1.x while the latest is 2.x).

I realise it is more likely to happen if we run it manually (which we won't be doing under normal circumstances), but still might be better to be explicit about the intent.

Same applies to all other commands where we have `latest-marker`.

ci/internal/release/publish.go
```diff
@@ -0,0 +1,155 @@
...
+// production).
+func Publish(inputs PublishInputs, host string, reg Registry) (PublishResult, error) {
+	if inputs.StagingImage == "" {
+		return PublishResult{}, errors.New("staging-image is required")
+	}
+	if inputs.ProdRepo == "" {
+		return PublishResult{}, errors.New("prod-repo is required")
+	}
+
+	marker := normalizeLatestMarker(inputs.LatestMarker)
```
m1kola: Stylistic nit: here and in other places, it feels like it belongs to CLI validation. I think cobra even has `Args` callback for that on `cobra.Command`.

m1kola (approved):
A nit and a suggestion, but nothing blocking.

ci/internal/cli/release_promote.go
josvazg: Not sure, from the point of view of the tool itself, it does not make too much sense to enforce you to always set this flag. The tool is relatively agnostic which allows us some wiggle room to change the wiring without having to change the tool for every decision.

If you are going to shoot yourself in the foot by using the default, you might as well do it by using the flag wrong.

In other words, I do not see a big win for enforcing this required always.

ci/internal/release/publish.go
josvazg: Simplified, thanks for the heads up

ci/internal/cli/release_promote.go
m1kola: > If you are going to shoot yourself in the foot by using the default, you might as well do it by using the flag wrong.

Absolutely, but this field being required makes the choice explicit and deliberate. Whereas the current default is a hidden gotcha.

> from the point of view of the tool itself

The tool does not exist in isolation. It is a custom-made tool for a specific purpose/workflow. We clearly know that we will be using it in the context of backporting and we've decided that `latest` must always be set to the latest/highest version (= must not be set to a version from the previous major version).

So if we can slightly reduce the chance of a human error by explicitly requiring a value for this flag it might be worth doing it.

I won't block the PR on this, but I'm in favour of this being explicit.

josvazg: ok, will make it required.

My argument is, if this were a more generic tool, or later gets promoted to a generic tool, `latest` is a perfectly reasonable default for most case, which not always include backports. And when you do backports, everybody knows that the marker needs to be per version, when the share registry path, which might not always be the case.

mircea-cosbuc (approved):

mircea-cosbuc (approved):

mircea-cosbuc (approved):

PR #1362, opened 2026-07-09
KUBE-154: establish MongoDBSearch lifecycle identity

anandsyncs:
# Summary

MongoDBSearch resources need durable identity so reconciliation, watch routing, state adoption, and cleanup remain safe across the resource lifecycle. This PR stamps owner name, namespace, cluster, and component labels on operator-managed workloads and copies. Writers merge user/override labels first and stamp the managed labels last — the managed set always wins, with no separate restore/reapply pass.

## Proof of Work

- `go build ./...` clean; unit suites for `controllers/operator`, `controllers/searchcontroller`, `pkg/handler`, `pkg/tls`, `api/mongodb/v1/search` pass; each of the 2 commits builds and passes its package tests in isolation.
- **Round-10 gates** at head `c59f3392a`: `go build ./...` + unit suites for the operator, searchcontroller, and handler trees green (17 packages ok across the stack tip); golangci-lint unchanged (one pre-existing finding in a file untouched by the stack).
- Race suite (`gotestsum -- -race` over controllers/api/pkg): 2,537 tests, 0 failures, 0 data races.
- `golangci-lint` clean on the diff (2 pre-existing findings in untouched files); black-box verified: the nil-map fix is load-bearing (test panics without it).
- Label precedence pinned by the compact table test (member + single-cluster rows); adoption semantics pinned at the `MutateSearchState` boundary (legacy label repair; no-op mutate does not create state). Test delta vs master: +164 lines.
- Evergreen: the canonical 68-task matrix at this head `486dea11c` concluded **green** — [`6a6752e80286710007f64f21`](https://spruce.corp.mongodb.com/version/6a6752e80286710007f64f21): 68/68, all at execution 0, zero restarts. The stack's current topmost matrix (round 7, at #1393's `4352719aa`, built on this head) also concluded **green** — [`6a67d2fb23f20c000767e00e`](https://spruce.corp.mongodb.com/version/6a67d2fb23f20c000767e00e): 69/69 at execution 0, zero restarts. Prior revision `bee40edd2` also concluded green — [`6a6364f9e0037d000798157f`](https://spruce.corp.mongodb.com/version/6a6364f9e0037d000798157f): 68/68, 67 at execution 0, the one auto-restart environmental (OpenShift autoscaler double scale-up, triaged in the review thread).

## Checklist

---

### DIFF
--- a/controllers/operator/mongodbsearch_controller.go
+++ b/controllers/operator/mongodbsearch_controller.go
@@ -189,7 +189,7 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
 		// A concurrent writer bumped the ConfigMap between read and update; retry
 		// instead of marking the CR Failed over a transient race.
 		if apierrors.IsConflict(err) {
-			return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Pending("Search state was modified concurrently, retrying").Requeue(), log)
+			return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Pending("Search state was modified concurrently, re-queuing ").Requeue(), log)
 		}
 		return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Failed(xerrors.Errorf("failed to read or repair search state: %w", err)), log)
 	}
--- a/controllers/operator/mongodbsearch_controller.go
+++ b/controllers/operator/mongodbsearch_controller.go
@@ -180,9 +180,18 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
 		}
 	}
 
-	state, err := searchcontroller.ReadSearchState(ctx, r.kubeClient, mdbSearch)
+	// The no-op mutation reads the state and, as a side effect, repairs legacy
+	// owner labels on the state ConfigMap.
+	state, err := searchcontroller.MutateSearchState(ctx, r.kubeClient, mdbSearch, func(*searchcontroller.SearchDeploymentState) bool {
+		return false
+	})
 	if err != nil {
-		return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Failed(xerrors.Errorf("failed to read search state: %w", err)), log)
+		// A concurrent writer bumped the ConfigMap between read and update; retry
+		// instead of marking the CR Failed over a transient race.
+		if apierrors.IsConflict(err) {
+			return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Pending("Search state was modified concurrently, retrying").Requeue(), log)
+		}
+		return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Failed(xerrors.Errorf("failed to read or repair search state: %w", err)), log)
 	}
 
 	reconcileHelper := searchcontroller.NewMongoDBSearchReconcileHelper(r.kubeClient, mdbSearch, searchSource, r.operatorSearchConfig, r.memberClusterClientsMap, state)
--- a/controllers/operator/mongodbsearch_metrics_controller.go
+++ b/controllers/operator/mongodbsearch_metrics_controller.go
@@ -1008,6 +1008,9 @@ func (r *MongoDBSearchMetricsForwarderReconciler) ensureMetricsForwarderDeployme
 			dep.Labels = merge.StringToStringMap(dep.Labels, deploymentOverride.MetadataWrapper.Labels)
 			dep.Annotations = merge.StringToStringMap(dep.Annotations, deploymentOverride.MetadataWrapper.Annotations)
 		}
+		// Identity labels are merged after user label overrides so users cannot
+		// detach the Deployment from its owning MongoDBSearch.
+		dep.Labels = merge.StringToStringMap(dep.Labels, labels)
 
 		if clusterName == "" {
 			return controllerutil.SetOwnerReference(search, dep, c.Scheme())
@@ -1123,16 +1126,7 @@ func metricsForwarderResourceRequirements(search *searchv1.MongoDBSearch) corev1
 // metricsForwarderLabelsForCluster returns resource labels including cross-cluster enqueue labels.
 // clusterName=="" (single-cluster/central): cluster-name label is omitted.
 func metricsForwarderLabelsForCluster(search *searchv1.MongoDBSearch, clusterName string, clusterIndex int) map[string]string {
-	labels := map[string]string{
-		"app":                                search.MetricsForwarderDeploymentNameForCluster(clusterIndex),
-		"component":                          metricsForwarderLabelName,
-		khandler.MongoDBSearchOwnerNameLabel: search.Name,
-		khandler.MongoDBSearchOwnerNamespaceLabel: search.Namespace,
-	}
-	if clusterName != "" {
-		labels[khandler.MongoDBSearchClusterNameLabel] = clusterName
-	}
-	return labels
+	return khandler.SearchManagedLabels(search, search.MetricsForwarderDeploymentNameForCluster(clusterIndex), metricsForwarderLabelName, clusterName)
 }
 
 func metricsForwarderPodLabelsForCluster(search *searchv1.MongoDBSearch, clusterIndex int) map[string]string {
--- a/controllers/operator/mongodbsearch_metrics_controller.go
+++ b/controllers/operator/mongodbsearch_metrics_controller.go
@@ -1126,7 +1126,7 @@ func metricsForwarderResourceRequirements(search *searchv1.MongoDBSearch) corev1
 // metricsForwarderLabelsForCluster returns resource labels including cross-cluster enqueue labels.
 // clusterName=="" (single-cluster/central): cluster-name label is omitted.
 func metricsForwarderLabelsForCluster(search *searchv1.MongoDBSearch, clusterName string, clusterIndex int) map[string]string {
-	return khandler.SearchManagedLabels(search, search.MetricsForwarderDeploymentNameForCluster(clusterIndex), metricsForwarderLabelName, clusterName)
+	return khandler.SearchOwnershipLabels(search, search.MetricsForwarderDeploymentNameForCluster(clusterIndex), metricsForwarderLabelName, clusterName)
 }
 
 func metricsForwarderPodLabelsForCluster(search *searchv1.MongoDBSearch, clusterIndex int) map[string]string {
--- a/controllers/operator/mongodbsearch_metrics_controller_test.go
+++ b/controllers/operator/mongodbsearch_metrics_controller_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/controllers/operator/watch"
 	"github.com/mongodb/mongodb-kubernetes/controllers/searchcontroller"
 	mdbcv1 "github.com/mongodb/mongodb-kubernetes/mongodb-community-operator/api/v1" //nolint:depguard
+	khandler "github.com/mongodb/mongodb-kubernetes/pkg/handler"
 	kubernetesClient "github.com/mongodb/mongodb-kubernetes/pkg/kube/client"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/merge"
@@ -367,11 +368,16 @@ func TestMetricsForwarderLabels(t *testing.T) {
 
 	labels := metricsForwarderLabels(search)
 	assert.Equal(t, "my-search-search-metrics-forwarder-0", labels["app"])
-	assert.Equal(t, metricsForwarderLabelName, labels["component"])
+	assert.Equal(t, metricsForwarderLabelName, labels[khandler.MongoDBSearchComponentLabel])
+	assert.NotContains(t, labels, khandler.MongoDBSearchClusterNameLabel)
+
+	memberLabels := metricsForwarderLabelsForCluster(search, "us-east", 3)
+	assert.Equal(t, search.MetricsForwarderDeploymentNameForCluster(3), memberLabels["app"])
+	assert.Equal(t, "us-east", memberLabels[khandler.MongoDBSearchClusterNameLabel])
 
 	podLabels := metricsForwarderPodLabels(search)
 	assert.Equal(t, "my-search-search-metrics-forwarder-0", podLabels["app"])
-	assert.NotContains(t, podLabels, "component")
+	assert.NotContains(t, podLabels, khandler.MongoDBSearchComponentLabel)
 }
 
 func TestMetricsForwarderResourceRequirements_Defaults(t *testing.T) {
--- a/controllers/operator/mongodbsearch_metrics_controller_test.go
+++ b/controllers/operator/mongodbsearch_metrics_controller_test.go
@@ -1677,11 +1677,17 @@ func newTestMongoDBSearchWithReplicas(name, namespace, mdbName string, replicas 
 }
 
 // TestReconcileTopologyState_FirstReconcile_RecordsCurrentReplicas verifies that the first call
-// creates a state ConfigMap recording the current replica count and returns pending=false.
+// records the current replica count in the state ConfigMap and returns pending=false. The
+// pre-existing ConfigMap carries nil Data (as after a hand-edit): it must read as fresh state
+// and must not panic the state write.
 func TestReconcileTopologyState_FirstReconcile_RecordsCurrentReplicas(t *testing.T) {
 	search := newTestMongoDBSearchWithReplicas(testSearchName, testNamespace, testMDBName, 2)
 	agentSecret := newTestAgentKeySecret("agent-key-secret", testNamespace)
-	r, fakeClient := newMetricsForwarderReconciler(testDefaultImage, search, agentSecret)
+	emptyStateCM := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{
+		Name:      fmt.Sprintf("%s-metrics-forwarder-state", search.Name),
+		Namespace: testNamespace,
+	}}
+	r, fakeClient := newMetricsForwarderReconciler(testDefaultImage, search, agentSecret, emptyStateCM)
 
 	var deletedHostIDs []string
 	r.omRequester = recordingDeleteHostsRequester(&deletedHostIDs)
--- a/controllers/operator/mongodbsearchenvoy_controller.go
+++ b/controllers/operator/mongodbsearchenvoy_controller.go
@@ -555,6 +555,9 @@ func (r *MongoDBSearchEnvoyReconciler) ensureDeployment(ctx context.Context, sea
 			dep.Labels = merge.StringToStringMap(dep.Labels, managedLB.Deployment.MetadataWrapper.Labels)
 			dep.Annotations = merge.StringToStringMap(dep.Annotations, managedLB.Deployment.MetadataWrapper.Annotations)
 		}
+		// Identity labels are merged after user label overrides so users cannot
+		// detach the Deployment from its owning MongoDBSearch.
+		dep.Labels = merge.StringToStringMap(dep.Labels, labels)
 
 		if clusterName == "" {
 			return controllerutil.SetOwnerReference(search, dep, c.Scheme())
@@ -780,17 +783,7 @@ func defaultEnvoyResourceRequirements() corev1.ResourceRequirements {
 // clusterIndex is used for the "app" label (resource name); clusterName is used
 // for the cluster-name label so label selectors remain name-keyed.
 func envoyLabelsForCluster(search *searchv1.MongoDBSearch, clusterName string, clusterIndex int) map[string]string {
-	labels := map[string]string{
-		"app":                                search.LoadBalancerDeploymentNameForCluster(clusterIndex),
-		"component":                          labelName,
-		khandler.MongoDBSearchOwnerNameLabel: search.Name,
-		khandler.MongoDBSearchOwnerNamespaceLabel: search.Namespace,
-	}
-	// In single-cluster legacy mode (clusterName==""), omit the per-cluster label so existing watchers continue to match.
-	if clusterName != "" {
-		labels[khandler.MongoDBSearchClusterNameLabel] = clusterName
-	}
-	return labels
+	return khandler.SearchManagedLabels(search, search.LoadBalancerDeploymentNameForCluster(clusterIndex), labelName, clusterName)
 }
 
 // envoyReplicas returns the cluster's desired Envoy replica count.
--- a/controllers/operator/mongodbsearchenvoy_controller.go
+++ b/controllers/operator/mongodbsearchenvoy_controller.go
@@ -783,7 +783,7 @@ func defaultEnvoyResourceRequirements() corev1.ResourceRequirements {
 // clusterIndex is used for the "app" label (resource name); clusterName is used
 // for the cluster-name label so label selectors remain name-keyed.
 func envoyLabelsForCluster(search *searchv1.MongoDBSearch, clusterName string, clusterIndex int) map[string]string {
-	return khandler.SearchManagedLabels(search, search.LoadBalancerDeploymentNameForCluster(clusterIndex), labelName, clusterName)
+	return khandler.SearchOwnershipLabels(search, search.LoadBalancerDeploymentNameForCluster(clusterIndex), labelName, clusterName)
 }
 
 // envoyReplicas returns the cluster's desired Envoy replica count.
--- a/controllers/operator/state_store.go
+++ b/controllers/operator/state_store.go
@@ -57,6 +57,11 @@ func (s *StateStore[S]) read(ctx context.Context) error {
 	}
 
 	s.data = cm.Data
+	// An empty or hand-edited ConfigMap can carry nil Data; setDataValue would
+	// panic writing into a nil map.
+	if s.data == nil {
+		s.data = map[string]string{}
+	}
 	return nil
 }
 
--- a/controllers/searchcontroller/mongodbsearch_reconcile_helper.go
+++ b/controllers/searchcontroller/mongodbsearch_reconcile_helper.go
@@ -46,6 +46,7 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/pkg/statefulset"
 	"github.com/mongodb/mongodb-kubernetes/pkg/tls"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/maputil"
+	"github.com/mongodb/mongodb-kubernetes/pkg/util/merge"
 )
 
 const (
@@ -126,27 +127,16 @@ func NewMongoDBSearchReconcileHelper(
 // don't cross cluster boundaries; labels are the only path back from a member
 // resource to its central CR (for watch routing and label-based GC).
 func searchOwnerLabels(search *searchv1.MongoDBSearch, clusterName string) map[string]string {
-	labels := map[string]string{
-		khandler.MongoDBSearchOwnerNameLabel:      search.Name,
-		khandler.MongoDBSearchOwnerNamespaceLabel: search.Namespace,
-	}
-	if clusterName != "" {
-		labels[khandler.MongoDBSearchClusterNameLabel] = clusterName
-	}
-	return labels
+	return khandler.SearchManagedLabels(search, "", "", clusterName)
 }
 
-// withSearchOwnerLabels merges the search-owner labels into the STS ObjectMeta
-// labels. It does NOT touch the selector/pod-template (which is immutable on
-// existing STS); the pod-routing labels stay in the unit's podLabels.
+// withSearchOwnerLabels adds the managed identity labels to the StatefulSet
+// labels. It runs last in the modification chain so user labels merged earlier
+// cannot win over them. It does not touch the selector/pod-template (immutable
+// on an existing STS); the pod-routing labels stay in the unit's podLabels.
 func withSearchOwnerLabels(search *searchv1.MongoDBSearch, clusterName string) statefulset.Modification {
 	return func(set *appsv1.StatefulSet) {
-		if set.Labels == nil {
-			set.Labels = map[string]string{}
-		}
-		for k, v := range searchOwnerLabels(search, clusterName) {
-			set.Labels[k] = v
-		}
+		set.Labels = merge.StringToStringMap(set.Labels, khandler.SearchManagedLabels(search, "", mongotComponent, clusterName))
 	}
 }
 
@@ -668,22 +658,25 @@ func (r *MongoDBSearchReconcileHelper) applyReconcileUnit(
 		}
 	}
 
+	tlsSecretLabels := searchOwnerLabels(r.mdbSearch, unit.clusterName)
+	ingressTLSSecretLabels := khandler.SearchManagedLabels(r.mdbSearch, "", mongotComponent, unit.clusterName)
+
 	// Per-unit ingress TLS: each shard may have its own secret, so this cannot
 	// be hoisted out of the loop.
-	ingressTlsMongotModification, ingressTlsStsModification, err := r.ensureIngressTlsConfig(ctx, unitClient, unit.tlsResource)
+	ingressTlsMongotModification, ingressTlsStsModification, err := r.ensureIngressTlsConfig(ctx, unitClient, unit.tlsResource, ingressTLSSecretLabels)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	// Per-unit egress TLS (SCRAM CA + optional client cert): uses unitClient so the
 	// operator-managed secret is created on the cluster where mongot pods run.
-	egressTlsMongotModification, egressTlsStsModification, err := r.ensureEgressTlsConfig(ctx, unitClient)
+	egressTlsMongotModification, egressTlsStsModification, err := r.ensureEgressTlsConfig(ctx, unitClient, tlsSecretLabels)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	// Per-unit x509 client cert: uses unitClient for the same reason as egress TLS.
-	x509MongotModification, x509StsModification, err := r.ensureX509ClientCertConfig(ctx, unitClient)
+	x509MongotModification, x509StsModification, err := r.ensureX509ClientCertConfig(ctx, unitClient, tlsSecretLabels)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -730,7 +723,6 @@ func (r *MongoDBSearchReconcileHelper) applyReconcileUnit(
 		unitClient,
 		unit.stsName,
 		stsFunc,
-		withSearchOwnerLabels(r.mdbSearch, unit.clusterName),
 		mods.passwordAuthSts,
 		configHashModification,
 		mods.keyfileSts,
@@ -738,7 +730,8 @@ func (r *MongoDBSearchReconcileHelper) applyReconcileUnit(
 		egressTlsStsModification,
 		x509StsModification,
 		mods.embeddingConfigSts,
-		stsOverride, // must be last: see StatefulSetOverrideModification
+		stsOverride,
+		withSearchOwnerLabels(r.mdbSearch, unit.clusterName),
 	)
 	if err != nil {
 		return nil, nil, err
@@ -1029,7 +1022,10 @@ func (r *MongoDBSearchReconcileHelper) cleanupStaleShardResources(ctx context.Co
 		expected map[string]bool
 		kind     string
 	}{
-		{func() client.ObjectList { return &appsv1.StatefulSetList{} }, client.MatchingLabels(searchOwnerLabels(r.mdbSearch, "")), expectedSTS, "StatefulSet"},
+		{func() client.ObjectList { return &appsv1.StatefulSetList{} }, client.MatchingLabels{
+			khandler.MongoDBSearchOwnerNameLabel:      r.mdbSearch.Name,
+			khandler.MongoDBSearchOwnerNamespaceLabel: r.mdbSearch.Namespace,
+		}, expectedSTS, "StatefulSet"},
 		{func() client.ObjectList { return &corev1.ServiceList{} }, client.MatchingLabels{componentLabelKey: proxyServiceComponent}, expectedProxy, "proxy Service"},
 		{func() client.ObjectList { return &corev1.ServiceList{} }, client.MatchingLabels{componentLabelKey: mongotComponent}, expectedHeadless, "headless Service"},
 		{func() client.ObjectList { return &corev1.ConfigMapList{} }, client.MatchingLabels{componentLabelKey: mongotComponent}, expectedConfig, "ConfigMap"},
@@ -1383,7 +1379,7 @@ func mongotServicePorts(search *searchv1.MongoDBSearch) []corev1.ServicePort {
 const (
 	appLabelKey           = "app"
 	shardLabelKey         = "shard"
-	componentLabelKey     = "component"
+	componentLabelKey     = khandler.MongoDBSearchComponentLabel
 	proxyServiceComponent = "search-proxy"
 	// mongotComponent is the component-label value on the per-shard mongot headless
 	// Service and ConfigMap (the StatefulSet is swept by owner label instead).
@@ -1696,12 +1692,12 @@ func setupMongotContainerArgsForAPIKeys() container.Modification {
 // ensureIngressTlsConfig processes TLS configuration for any mongot deployment.
 // For non-sharded deployments, pass r.mdbSearch as the tlsResource.
 // For sharded deployments, pass a perShardTLSResource adapter.
-func (r *MongoDBSearchReconcileHelper) ensureIngressTlsConfig(ctx context.Context, kubeClient kubernetesClient.Client, tlsResource tls.TLSConfigurableResource) (mongot.Modification, statefulset.Modification, error) {
+func (r *MongoDBSearchReconcileHelper) ensureIngressTlsConfig(ctx context.Context, kubeClient kubernetesClient.Client, tlsResource tls.TLSConfigurableResource, labels map[string]string) (mongot.Modification, statefulset.Modification, error) {
 	if r.mdbSearch.Spec.Security.TLS == nil {
 		return mongot.NOOP(), statefulset.NOOP(), nil
 	}
 
-	certFileName, err := tls.EnsureTLSSecret(ctx, kubeClient, tlsResource)
+	certFileName, err := tls.EnsureTLSSecret(ctx, kubeClient, tlsResource, labels, tlsResource.GetOwnerReferences())
 	if err != nil {
 		return nil, nil, err
 	}
@@ -1775,7 +1771,7 @@ func (x *x509AuthResource) TLSOperatorSecretNamespacedName() types.NamespacedNam
 
 // ensureX509ClientCertConfig processes x509 client certificate configuration for the sync source in case of mongot to mongod communication.
 // When x509 is configured, it replaces username/password auth with x509 certificate auth.
-func (r *MongoDBSearchReconcileHelper) ensureX509ClientCertConfig(ctx context.Context, kubeClient kubernetesClient.Client) (mongot.Modification, statefulset.Modification, error) {
+func (r *MongoDBSearchReconcileHelper) ensureX509ClientCertConfig(ctx context.Context, kubeClient kubernetesClient.Client, labels map[string]string) (mongot.Modification, statefulset.Modification, error) {
 	if !r.mdbSearch.IsX509Auth() {
 		return mongot.NOOP(), statefulset.NOOP(), nil
 	}
@@ -1788,7 +1784,7 @@ func (r *MongoDBSearchReconcileHelper) ensureX509ClientCertConfig(ctx context.Co
 	}
 
 	x509Resource := &x509AuthResource{MongoDBSearch: r.mdbSearch}
-	certFileName, err := tls.EnsureTLSSecret(ctx, kubeClient, x509Resource)
+	certFileName, err := tls.EnsureTLSSecret(ctx, kubeClient, x509Resource, labels, x509Resource.GetOwnerReferences())
 	if err != nil {
 		return nil, nil, err
 	}
@@ -1884,7 +1880,7 @@ func (p *perShardTLSResource) TLSOperatorSecretNamespacedName() types.Namespaced
 	return p.TLSOperatorSecretForClusterShard(p.clusterIndex, p.shardName)
 }
 
-func (r *MongoDBSearchReconcileHelper) ensureEgressTlsConfig(ctx context.Context, kubeClient kubernetesClient.Client) (mongot.Modification, statefulset.Modification, error) {
+func (r *MongoDBSearchReconcileHelper) ensureEgressTlsConfig(ctx context.Context, kubeClient kubernetesClient.Client, labels map[string]string) (mongot.Modification, statefulset.Modification, error) {
 	tlsSourceConfig := r.db.TLSConfig()
 	if tlsSourceConfig == nil {
 		return mongot.NOOP(), statefulset.NOOP(), nil
@@ -1896,7 +1892,7 @@ func (r *MongoDBSearchReconcileHelper) ensureEgressTlsConfig(ctx context.Context
 	scramKeyPasswordSecret := r.mdbSearch.ScramKeyFilePasswordSecret()
 	if r.mdbSearch.HasScramClientCert() {
 		scramCertResource := &scramClientCertResource{MongoDBSearch: r.mdbSearch}
-		certFileName, err := tls.EnsureTLSSecret(ctx, kubeClient, scramCertResource)
+		certFileName, err := tls.EnsureTLSSecret(ctx, kubeClient, scramCertResource, labels, scramCertResource.GetOwnerReferences())
 		if err != nil {
 			return nil, nil, err
 		}
--- a/controllers/searchcontroller/mongodbsearch_reconcile_helper.go
+++ b/controllers/searchcontroller/mongodbsearch_reconcile_helper.go
@@ -127,7 +127,7 @@ func NewMongoDBSearchReconcileHelper(
 // don't cross cluster boundaries; labels are the only path back from a member
 // resource to its central CR (for watch routing and label-based GC).
 func searchOwnerLabels(search *searchv1.MongoDBSearch, clusterName string) map[string]string {
-	return khandler.SearchManagedLabels(search, "", "", clusterName)
+	return khandler.SearchOwnershipLabels(search, "", "", clusterName)
 }
 
 // withSearchOwnerLabels adds the managed identity labels to the StatefulSet
@@ -136,7 +136,7 @@ func searchOwnerLabels(search *searchv1.MongoDBSearch, clusterName string) map[s
 // on an existing STS); the pod-routing labels stay in the unit's podLabels.
 func withSearchOwnerLabels(search *searchv1.MongoDBSearch, clusterName string) statefulset.Modification {
 	return func(set *appsv1.StatefulSet) {
-		set.Labels = merge.StringToStringMap(set.Labels, khandler.SearchManagedLabels(search, "", mongotComponent, clusterName))
+		set.Labels = merge.StringToStringMap(set.Labels, khandler.SearchOwnershipLabels(search, "", mongotComponent, clusterName))
 	}
 }
 
@@ -659,7 +659,7 @@ func (r *MongoDBSearchReconcileHelper) applyReconcileUnit(
 	}
 
 	tlsSecretLabels := searchOwnerLabels(r.mdbSearch, unit.clusterName)
-	ingressTLSSecretLabels := khandler.SearchManagedLabels(r.mdbSearch, "", mongotComponent, unit.clusterName)
+	ingressTLSSecretLabels := khandler.SearchOwnershipLabels(r.mdbSearch, "", mongotComponent, unit.clusterName)
 
 	// Per-unit ingress TLS: each shard may have its own secret, so this cannot
 	// be hoisted out of the loop.
--- a/controllers/searchcontroller/mongodbsearch_reconcile_helper_test.go
+++ b/controllers/searchcontroller/mongodbsearch_reconcile_helper_test.go
@@ -33,6 +33,7 @@ import (
 	khandler "github.com/mongodb/mongodb-kubernetes/pkg/handler"
 	kubernetesClient "github.com/mongodb/mongodb-kubernetes/pkg/kube/client"
 	"github.com/mongodb/mongodb-kubernetes/pkg/mongot"
+	"github.com/mongodb/mongodb-kubernetes/pkg/statefulset"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/maputil"
 )
 
@@ -104,6 +105,25 @@ func newTestFakeClient(objects ...client.Object) kubernetesClient.Client {
 
 	clientBuilder.WithObjects(objects...)
 	return kubernetesClient.NewClient(clientBuilder.Build())
+}
+
+// assertSearchManagedLabels asserts the managed identity on every local
+// mongot resource: owner labels, component, cluster-name label presence,
+// and the owner reference back to the MongoDBSearch CR.
+func assertSearchManagedLabels(t *testing.T, obj client.Object, search *searchv1.MongoDBSearch, clusterName string) {
+	t.Helper()
+	labels := obj.GetLabels()
+	assert.Equal(t, search.Name, labels[khandler.MongoDBSearchOwnerNameLabel])
+	assert.Equal(t, search.Namespace, labels[khandler.MongoDBSearchOwnerNamespaceLabel])
+	assert.Equal(t, mongotComponent, labels[khandler.MongoDBSearchComponentLabel])
+	if clusterName == "" {
+		assert.NotContains(t, labels, khandler.MongoDBSearchClusterNameLabel)
+	} else {
+		assert.Equal(t, clusterName, labels[khandler.MongoDBSearchClusterNameLabel])
+	}
+	assert.True(t, slices.ContainsFunc(obj.GetOwnerReferences(), func(ref metav1.OwnerReference) bool {
+		return ref.Kind == "MongoDBSearch" && ref.Name == search.Name && ref.UID == search.UID
+	}), "expected owner reference back to the MongoDBSearch CR")
 }
 
 func reconcileMongoDBSearch(ctx context.Context, fakeClient kubernetesClient.Client, mdbSearch *searchv1.MongoDBSearch, mdbc *mdbcv1.MongoDBCommunity, operatorConfig OperatorSearchConfig) workflow.Status {
@@ -613,6 +633,49 @@ func TestMongoDBSearchReconcileHelper_ServiceCreation(t *testing.T) {
 
 			assertServiceBasicProperties(t, svc, mdbSearch)
 			assertServicePorts(t, svc, tc.expectedPorts)
+		})
+	}
+}
+
+func TestSearchManagedLabelsWinOverOverrides(t *testing.T) {
+	tests := []struct {
+		name        string
+		clusterName string
+	}{
+		{name: "member cluster", clusterName: "member-a"},
+		{name: "legacy single cluster"},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			overrideLabels := map[string]string{
+				"custom-label":                            "custom-value",
+				khandler.MongoDBSearchOwnerNameLabel:      "wrong-name",
+				khandler.MongoDBSearchOwnerNamespaceLabel: "wrong-namespace",
+				khandler.MongoDBSearchComponentLabel:      "wrong-component",
+			}
+			if tc.clusterName != "" {
+				overrideLabels[khandler.MongoDBSearchClusterNameLabel] = "wrong-cluster"
+			}
+			search := newTestMongoDBSearch("test-search", "test-ns", func(search *searchv1.MongoDBSearch) {
+				search.UID = "search-uid"
+				search.Spec.Clusters = []searchv1.ClusterSpec{{
+					Name: tc.clusterName,
+					StatefulSetConfiguration: &v1.StatefulSetConfiguration{
+						MetadataWrapper: v1.StatefulSetMetadataWrapper{
+							Labels: overrideLabels,
+						},
+					},
+				}}
+			})
+			sizing := search.EffectiveClusters()[0]
+			sts := statefulset.New(
+				CreateSearchStatefulSetFunc(search, sizing, "test-search-search-0", search.Namespace, "test-search-search-0-svc", "test-search-search-0-config", map[string]string{"app": "test-search-search-0"}, "mongot:latest", false),
+				StatefulSetOverrideModification(sizing.StatefulSetConfiguration),
+				withSearchOwnerLabels(search, tc.clusterName),
+			)
+
+			assert.Equal(t, "custom-value", sts.Labels["custom-label"])
+			assertSearchManagedLabels(t, &sts, search, tc.clusterName)
 		})
 	}
 }
@@ -2600,7 +2663,7 @@ func TestEnsureX509ClientCertConfig_NoopWhenNotConfigured(t *testing.T) {
 	fakeClient := newTestFakeClient(search)
 	helper := NewMongoDBSearchReconcileHelper(fakeClient, search, nil, newTestOperatorSearchConfig(), nil, nil)
 
-	mongotMod, stsMod, err := helper.ensureX509ClientCertConfig(t.Context(), fakeClient)
+	mongotMod, stsMod, err := helper.ensureX509ClientCertConfig(t.Context(), fakeClient, searchOwnerLabels(search, ""))
 	require.NoError(t, err)
 
 	// Apply modifications and verify no changes
@@ -2641,13 +2704,14 @@ func TestEnsureX509ClientCertConfig_ErrorWhenTLSNotConfigured(t *testing.T) {
 	fakeClient := newTestFakeClient(search)
 	helper := NewMongoDBSearchReconcileHelper(fakeClient, search, dbSource, newTestOperatorSearchConfig(), nil, nil)
 
-	_, _, err := helper.ensureX509ClientCertConfig(t.Context(), fakeClient)
+	_, _, err := helper.ensureX509ClientCertConfig(t.Context(), fakeClient, searchOwnerLabels(search, ""))
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "tls must be enabled")
 }
 
 func TestEnsureX509ClientCertConfig_MongotAndStsModification(t *testing.T) {
 	search := newTestMongoDBSearch("test-search", "test-ns", func(s *searchv1.MongoDBSearch) {
+		s.UID = "search-uid"
 		s.Spec.Source.X509 = &searchv1.X509Auth{
 			ClientCertificateSecret: corev1.LocalObjectReference{Name: "x509-cert"},
 		}
@@ -2666,8 +2730,11 @@ func TestEnsureX509ClientCertConfig_MongotAndStsModification(t *testing.T) {
 	fakeClient := newTestFakeClient(search, x509Secret)
 	helper := NewMongoDBSearchReconcileHelper(fakeClient, search, dbSource, newTestOperatorSearchConfig(), nil, nil)
 
-	mongotMod, stsMod, err := helper.ensureX509ClientCertConfig(t.Context(), fakeClient)
+	mongotMod, stsMod, err := helper.ensureX509ClientCertConfig(t.Context(), fakeClient, searchOwnerLabels(search, ""))
 	require.NoError(t, err)
+	operatorSecret, err := fakeClient.GetSecret(t.Context(), search.X509OperatorManagedSecret())
+	require.NoError(t, err)
+	assert.Subset(t, operatorSecret.Labels, searchOwnerLabels(search, ""))
 
 	// Apply mongot modification to a config with both ReplicaSet and Router (sharded scenario)
 	config := &mongot.Config{
@@ -2765,7 +2832,7 @@ func TestEnsureX509ClientCertConfig_KeyPassword(t *testing.T) {
 	fakeClient := newTestFakeClient(search, x509Secret, keyPasswordSecret)
 	helper := NewMongoDBSearchReconcileHelper(fakeClient, search, dbSource, newTestOperatorSearchConfig(), nil, nil)
 
-	mongotMod, stsMod, err := helper.ensureX509ClientCertConfig(t.Context(), fakeClient)
+	mongotMod, stsMod, err := helper.ensureX509ClientCertConfig(t.Context(), fakeClient, searchOwnerLabels(search, ""))
 	require.NoError(t, err)
 
 	// Verify mongot config has key password path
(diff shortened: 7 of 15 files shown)


controllers/operator/mongodbsearch_controller.go
```diff
@@ -362,7 +495,6 @@ func AddMongoDBSearchController(
 		watchedTypes := []client.Object{
```
anandsyncs, line 355: **major (correctness):** `appsv1.Deployment` was removed from the member `watchedTypes` (and the central watch list never had it), but this controller is the only thing that sweeps Deployments — the Envoy/metrics controllers no-op on NotFound. If a Deployment write races the CR delete, or a Deployment delete fails and the operator restarts before the retry, nothing ever re-enqueues the sweeper and the Deployment leaks permanently. Every other swept kind self-heals via its watch. Suggest adding `&appsv1.Deployment{}` to both the central and member watch lists with the same label mapper + predicate.

anandsyncs (commented):
Automated 3-track review (operator/watch Go, searchcontroller sweep/TLS Go, Python e2e) against head `6a8e018bb`, focused on correctness, simplicity, and reusability — inline comments follow.

**Owner-ref removal status:** complete at every search write site — mongot STS/CM and state CM strip legacy refs explicitly, Services/Secrets via full-object replace; no `EnqueueRequestForOwner` remains and tests assert refs empty. The one gap is the Envoy/metrics mutate funcs not stripping pre-PR legacy refs (inline).

**Must-fix (3):** Deployment swept-but-not-watched leak; `ResourcesHandler.Delete` now enqueues CM/Secret deletes for every controller in the operator (not just search); Envoy/metrics legacy owner refs never stripped on upgrade.

**PR description corrections:** the UID label is not consulted by either sweep — recreated-CR protection comes from the uncached confirmation (UID's only functional consumer is the state-CM incarnation guard); and the cross-controller delete-routing behavior change should be called out (or gated away).

**Solid:** fail-closed uncached confirm with a test pinning the uncached reader; label stamping verified at every write site; customer/source secrets un-sweepable; multicluster-first sweep with error aggregation and Pod/PVC negative assertions; e2es assert recreation by uid-change with presence guards and correct per-cluster clients.

controllers/operator/watch/config_change_handler.go
```diff
@@ -83,16 +86,18 @@ func (c *ResourcesHandler) doHandle(namespace, name string, q workqueue.TypedRat
...
+func (c *ResourcesHandler) doMap(ctx context.Context, obj client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	if c.MapFunc == nil {
 		return
 	}
+	for _, request := range c.MapFunc(ctx, obj) {
+		q.Add(request)
+	}
+}
 
+func (c *ResourcesHandler) Delete(ctx context.Context, e event.TypedDeleteEvent[client.Object], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
```
anandsyncs, line 87: **major (correctness, operator-wide):** base deliberately ignored ConfigMap/Secret **delete** events here; routing them through `doHandle` unconditionally changes behavior for every controller that registers `ResourcesHandler` (OM, ReplicaSet, ShardedCluster, Standalone, MongoDBUser, mdbmulti). Deleting a user-managed credentials/TLS secret now drives all dependents into Failed + endless rate-limited retries, and namespace teardown floods the workqueues. Only the search handlers need delete routing — suggest gating on `c.MapFunc != nil` (set by all three search handlers, no legacy handler) to keep base semantics elsewhere. Worth a mention in the PR description either way.

controllers/operator/mongodbsearchenvoy_controller.go
```diff
@@ -456,11 +465,6 @@ func buildReplicaSetRouteForCluster(search *searchv1.MongoDBSearch, clusterIndex
 // ensureConfigMap creates or updates the Envoy ConfigMap with three files:
 // bootstrap.json (static), cds.json (dynamic clusters), and lds.json (dynamic listener).
 // Kubernetes ConfigMap updates are atomic (symlink swap), so all files update together.
-//
-// Cross-cluster ownership note: Kubernetes garbage collection does not span
-// clusters, so we only set an OwnerReference when writing into the central
-// cluster (clusterName == ""). Cleanup of member-cluster objects is handled
-// explicitly in deleteEnvoyResources.
 func (r *MongoDBSearchEnvoyReconciler) ensureConfigMap(ctx context.Context, search *searchv1.MongoDBSearch, bootstrapJSON, cdsJSON, ldsJSON, clusterName string, clusterIndex int, c kubernetesClient.Client, log *zap.SugaredLogger) error {
```
anandsyncs, line 464: **major (upgrade path):** this mutate stopped *adding* owner refs but doesn't strip pre-existing ones, and `CreateOrUpdate` preserves untouched metadata. A cluster upgraded from a pre-PR operator keeps `ownerRef → old CR UID` on this ConfigMap/Deployment; on delete + quick recreate the sweep is skipped (uncached confirm sees the new CR), reconcile adopts the object in place, and native GC then deletes it out from under the new CR — the same bug class this PR fixes for mongot. Suggest `OwnerReferences = nil` here and in `ensureDeployment` (L507), matching the mongot/state-CM sites.

controllers/operator/mongodbsearch_metrics_controller.go
```diff
@@ -942,8 +946,6 @@ func mongotHostID(groupID, namespace, podName string) string {
 }
 
 // ensureMetricsForwarderConfigMap creates or updates the metrics forwarder ConfigMap for one cluster.
-// OwnerReference is set only for the central cluster (clusterName == ""); member-cluster objects
-// use labels for cross-cluster GC.
 func (r *MongoDBSearchMetricsForwarderReconciler) ensureMetricsForwarderConfigMap(ctx context.Context, search *searchv1.MongoDBSearch, configYAML []byte, clusterName string, clusterIndex int, c kubernetesClient.Client, log *zap.SugaredLogger) error {
```
anandsyncs, line 947: **major (upgrade path):** same as the Envoy controller — legacy owner refs from pre-PR operators are never stripped here or in `ensureMetricsForwarderDeployment` (L970); `OwnerReferences = nil` in both mutate funcs closes the delete+fast-recreate GC race on upgraded clusters.

controllers/operator/mongodbsearch_controller.go
```diff
@@ -238,6 +274,99 @@ func (r *MongoDBSearchReconciler) getSourceMongoDBForSearch(ctx context.Context,
...
+	}
+
+	var errs error
+	for _, item := range items {
+		obj, ok := item.(client.Object)
+		if !ok || !ownsSearchForSweep(search, obj) {
+			continue
+		}
+		log.Infof("Deleting %s %s for deleted MongoDBSearch %s on cluster %q", kind, obj.GetName(), search, clusterName)
+		if err := deleteClient.Delete(ctx, obj); err != nil && !apierrors.IsNotFound(err) {
```
anandsyncs: **minor:** CR absence is confirmed once before the whole multi-kind/multi-cluster sweep, so a CR recreated mid-sweep can have its fresh resources deleted by the sweep tail (watches recreate them, but it's a one-cycle outage). `deleteClient.Delete(ctx, obj, client.Preconditions{UID: ptr.To(obj.GetUID())})` makes the sweep exact for free — and would give the UID label an actual consumer in the sweep path (today the description's recreated-CR-protection claim rests on the uncached confirm, not the UID label).

controllers/operator/mongodbsearch_controller.go
```diff
@@ -238,6 +274,99 @@ func (r *MongoDBSearchReconciler) getSourceMongoDBForSearch(ctx context.Context,
...
+			errs = errors.Join(errs, xerrors.Errorf("failed deleting %s %s for deleted MongoDBSearch %s on cluster %q: %w", kind, obj.GetName(), search, clusterName, err))
+		}
+	}
+	return errs
+}
+
+func ownsSearchForSweep(search types.NamespacedName, obj client.Object) bool {
```
anandsyncs: **minor (simplicity/reuse):** `ownsSearchForSweep` re-checks exactly the two labels the `List` already filtered on (server-side and in the fake client), so it can never be false here. It's also a copy of `ownsForSweep` in the reconcile helper — this sweep and the shard scale-down sweep (`sweepStaleShardResources`) share the List→owner-check→NotFound-tolerant-Delete shape but diverge on `errors.Join` vs `multierror.Append` and duplicate the owner check. One shared sweep helper in `searchcontroller` would serve both and drop ~40 lines.

controllers/operator/mongodbsearch_controller.go
```diff
@@ -319,22 +449,25 @@ func AddMongoDBSearchController(
...
-		{&corev1.ConfigMap{}, &watch.ResourcesHandler{ResourceType: watch.ConfigMap, ResourceWatcher: r.watch}},
-		{&appsv1.StatefulSet{}, ownerHandler},
-		{&corev1.Service{}, ownerHandler},
-		{&corev1.Secret{}, ownerHandler},
+		{&appsv1.StatefulSet{}, handler.EnqueueRequestsFromMapFunc(khandler.EnqueueMemberClusterObjectToSearch)},
+		{&corev1.Service{}, handler.EnqueueRequestsFromMapFunc(khandler.EnqueueMemberClusterObjectToSearch)},
+		{&corev1.Secret{}, &watch.ResourcesHandler{ResourceType: watch.Secret, ResourceWatcher: r.watch, MapFunc: khandler.EnqueueMemberClusterObjectToSearch}},
+		{&corev1.ConfigMap{}, &watch.ResourcesHandler{ResourceType: watch.ConfigMap, ResourceWatcher: r.watch, MapFunc: khandler.EnqueueMemberClusterObjectToSearch}},
 	}
 	for _, w := range centralWatches {
```
anandsyncs, line 342: **nit:** the `switch w.obj.(type)` here smuggles per-entry configuration out of the `centralWatches` table — a `predicates []predicate.Predicate` field on the entries keeps the table self-describing.

controllers/operator/state_store.go
```diff
@@ -19,7 +19,22 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/pkg/kube/configmap"
 )
 
-const stateKey = "state"
+const (
+	stateKey         = "state"
+	stateOwnerUIDKey = "ownerUID"
+)
+
+type stateStoreOptions struct {
```
anandsyncs: **minor (proportionality):** functional-options machinery (`stateStoreOptions`/`StateStoreOption`/`WithStateStoreOwnerUID` + the `if opt != nil` guard no caller exercises) for one string option with one consumer. A plain `ownerUID string` parameter does the job — and the caller's `search.UID != ""` special-case in the metrics controller can go too, since an empty UID is already a no-op in `write`/`isStaleOwnerUID`.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -144,7 +145,9 @@ func withSearchOwnerLabels(search *searchv1.MongoDBSearch, clusterName string) s
 		if set.Labels == nil {
```
anandsyncs: **minor (reuse):** this hand-rolls what `khandler.ReapplyProtectedSearchLabels` already does, with weaker semantics — it overwrites protected keys but never removes a user-injected one (e.g. a single-cluster `statefulSet` override setting `mongodb.com/cluster-name: bogus` sticks forever). Building the desired map and calling the shared helper keeps the invariant identical across all three controllers.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -724,7 +733,8 @@ func (r *MongoDBSearchReconcileHelper) applyReconcileUnit(
 		egressTlsStsModification,
 		x509StsModification,
 		mods.embeddingConfigSts,
-		stsOverride, // must be last: see StatefulSetOverrideModification
+		stsOverride,
+		withSearchOwnerLabels(r.mdbSearch, unit.clusterName),
```
anandsyncs, line 734: **nit:** `withSearchOwnerLabels` is applied both before the override (L728) and here after it; nothing in between reads the labels and this application sets the same keys, so the earlier one is dead — drop it and keep this restore-after-override one.

controllers/searchcontroller/search_state.go
```diff
@@ -50,6 +51,17 @@ func searchStateFromCM(cm *corev1.ConfigMap) (*SearchDeploymentState, error) {
 	return state, nil
 }
 
+func searchStateLabels(search *searchv1.MongoDBSearch) map[string]string {
+	labels := searchOwnerLabels(search, "")
```
anandsyncs: **minor (simplicity):** with the UID now included in `searchOwnerLabels`, `searchStateLabels` is byte-identical to `searchOwnerLabels(search, "")` — the function can be deleted and both call sites switched.

pkg/handler/labels_search.go
```diff
@@ -15,11 +15,36 @@ import (
 const (
 	MongoDBSearchOwnerNameLabel      = "mongodb.com/search-name"
 	MongoDBSearchOwnerNamespaceLabel = "mongodb.com/search-namespace"
+	MongoDBSearchOwnerUIDLabel       = "mongodb.com/search-uid"
+	MongoDBSearchComponentLabel      = "component"
```
anandsyncs, line 17: **nit:** the searchcontroller package has its own `componentLabelKey` for the same `"component"` key used here — alias one to the other so a future rename can't silently split the sweep selectors from the stamped labels.

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
```diff
@@ -517,6 +584,79 @@ def test_per_cluster_search_query(
...
+
+    assert_workload_ready_in_cluster(
+        mcc,
+        namespace,
+        {sts_name: desired_mongot_replicas},
+        envoy_name,
+        timeout=300,
+    )
+
+    Operator(namespace=namespace, name=SIMULATED_OPERATOR_NAME, api_client=mcc.api_client).restart_operator_deployment()
```
anandsyncs: **major (test discriminance):** every post-restart assertion here can pass without the restarted operator doing any work — `assert_reaches_phase(Running)` reads the stale status already in etcd, the workloads were ready before the restart, and the `$search` query was already passing. An operator that comes up pod-ready but never registers watches or reconciles would pass. Suggest creating drift the restarted process must heal, e.g. delete the mongot ConfigMap *after* `restart_operator_deployment()` returns and wait for a new UID.

docker/mongodb-kubernetes-tests/tests/search/search_sharded_scale_shards_managed_lb.py
```diff
@@ -553,6 +558,29 @@ def test_scale_down_verify_stale_resources_cleaned(namespace: str):
...
+                return True, f"operator-managed TLS Secret {operator_tls_secret_name} deleted"
+            raise
+
+    run_periodically(
+        operator_tls_secret_gone,
+        timeout=300,
+        sleep_time=10,
+        msg=f"operator-managed TLS Secret {operator_tls_secret_name} deletion",
+    )
+    core_v1.read_namespaced_secret(source_tls_secret_name, namespace)
```
anandsyncs: **major (test discriminance):** this source secret is cert-manager-backed — if the sweep wrongly deleted it, cert-manager re-issues it within seconds, and this name-only read (after up to 300s of polling above) would still pass. Capture the source secret's `metadata.uid` at scale-up (`verify_shard_resources_exist` already reads it) and assert the same uid here.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -121,6 +147,68 @@ def test_search_assert_search_query(sample_movies_helper: SampleMoviesSearchHelp
...
+        except ApiException as exc:
+            if exc.status == 404:
+                return False, f"{description} is absent"
+            raise
+
+    run_periodically(check, timeout=timeout, sleep_time=5, msg=f"{description} recreation")
+
+
+@mark.e2e_search_community_basic
+def test_recreate_mongot_top_level_resources(
```
anandsyncs: **flake risk:** deleting the mongot STS triggers `whenDeleted: Delete` PVC reaping (blocked by pvc-protection until pod-0 exits) while the operator's watch recreates the same-named STS within seconds — the exact Terminating-PVC race documented in `connectivity.py::wait_for_mongot_pvcs_deleted` and dodged by `test_scale_mongot_offline_reclaims_pvc` below. This test can't apply the dodge (recreation is the feature), so watch CI for a stuck `current: 0` timeout here and in `simulated_mc_rs.py`'s STS-delete test. If it flakes, the mitigation probably belongs in the operator — a customer `kubectl delete sts` self-heal hits the same state.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -515,6 +516,55 @@ def _per_cluster_envoy_configmap_name(mdbs_name: str, cluster_index: int) -> str
     return f"{mdbs_name}-search-lb-{cluster_index}-config"
 
 
+def _operator_managed_tls_secret_name(mdbs_name: str) -> str:
```
anandsyncs: **minor (reuse):** `_wait_for_not_found`/`_wait_for_search_deleted`/`_mongot_data_pvc_names`/`_operator_managed_tls_secret_name` are copy-pasted verbatim into `simulated_mc_rs.py` (and `search_community_basic.py` has the same waiters under different names) — 2–3 consumers each, past the lift threshold. Waiters belong beside `wait_for_mongot_pvcs_deleted` in `tests/common/search/connectivity.py`; the TLS-copy name helper in `search_resource_names.py`.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -1116,3 +1276,79 @@ def execute_vector_search(_idx: int = cluster_index) -> tuple:
             sleep_time=5,
             msg=f"cluster {cluster_index}: $vectorSearch query to succeed",
         )
+
+
+@mark.e2e_search_q2_mc_rs_steady
+def test_delete_search_resource_cleans_all_member_cluster_artifacts(
```
anandsyncs: **minor (coverage):** this delete test asserts only deletion, never *non-deletion* of user-owned inputs — an over-broad sweep selector that deleted the source cert secret, sync-user password secret, or CA ConfigMap on member clusters would pass. One read per kind after the last `_wait_for_not_found` closes it. Also: this ~70-line names-dict + seven-kind sweep block is near-identical to `simulated_mc_rs.py`'s — one shared `assert_search_artifacts_swept(...)` helper would serve both.

docker/mongodb-kubernetes-tests/kubetester/operator.py
```diff
@@ -206,12 +206,13 @@ def restart_operator_deployment(self):
         )
 
         # wait till there are 0 operator pods
-        count = 0
-        while count < 6:
+        for _ in range(20):
             pods = self.list_operator_pods()
```
anandsyncs, line 211: **nit:** good catch on the old loop never incrementing `count` (latent infinite hang). The new bound is 20×3s=60s shared by all callers — operator pod termination is normally <35s but consider ~120s of slack for loaded CI nodes.

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
```diff
@@ -69,6 +73,12 @@
 
 MEMBERS_PER_CLUSTER: List[int | None] = [3, 3]
 MONGOT_REPLICAS_PER_CLUSTER = 3
+RecreatedResource = (
```
anandsyncs: **nit:** `RecreatedResource` is used once and `_wait_for_recreated` only touches `.metadata.uid` — the alias can go.

controllers/operator/mongodbsearch_controller.go
anandsyncs: Added `Deployment` to both central and member label-routed Search watches. The routing tests now cover central/member create, update, protected-label removal, and delete events for Deployments.

controllers/operator/watch/config_change_handler.go
anandsyncs: Gated delete handling on `MapFunc != nil`, which cleanly distinguishes the Search handlers from the legacy handlers in this repository. Tests verify Search deletes enqueue reconciliation while legacy ConfigMap/Secret deletes remain ignored.

controllers/operator/mongodbsearchenvoy_controller.go
anandsyncs: Both Envoy ConfigMap and Deployment mutate functions now clear pre-PR owner references before applying desired state. Upgrade-adoption tests cover each resource with a stale owner reference.

controllers/operator/mongodbsearch_metrics_controller.go
anandsyncs: Both metrics-forwarder ConfigMap and Deployment mutate functions now clear pre-PR owner references, with upgrade-adoption coverage for each. The existing host-deregistration sequencing is unchanged.

controllers/operator/mongodbsearch_controller.go
anandsyncs: Sweep deletes now carry the listed object's UID as a Kubernetes delete precondition. Deterministic tests verify the precondition is sent and that a same-name replacement survives the resulting conflict.

controllers/operator/mongodbsearch_controller.go
anandsyncs: Removed the redundant post-list owner-label check. The exact owner name/namespace selector remains on the list, while UID delete preconditions independently protect against list/delete replacement races.

controllers/operator/mongodbsearch_controller.go
anandsyncs: Made each central watch entry self-describing by carrying its own predicates alongside the object and handler. The registration loop no longer switches on object type.

controllers/operator/state_store.go
anandsyncs: Replaced the single-option StateStore functional-options API with a direct `ownerUID string` parameter. Search passes its UID and existing non-Search callers pass an empty string.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: StatefulSet override restoration now reuses `khandler.ReapplyProtectedSearchLabels`. The focused legacy single-cluster test verifies an injected protected cluster label is removed while unrelated user labels remain.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: Removed the dead pre-override owner-label application. Protected labels are applied once after the user override, where they cannot be replaced.

controllers/searchcontroller/search_state.go
anandsyncs: Deleted `searchStateLabels` and switched both call sites to `searchOwnerLabels(search, "")`.

pkg/handler/labels_search.go
anandsyncs: Aliased the searchcontroller component key to `khandler.MongoDBSearchComponentLabel` so label stamping and sweep selection share one constant.

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
anandsyncs: The simulated multi-cluster test now deletes the managed mongot ConfigMap only after `restart_operator_deployment()` returns and requires recreation with a different UID. This makes successful post-restart watch registration and reconciliation observable before readiness/query checks.

docker/mongodb-kubernetes-tests/tests/search/search_sharded_scale_shards_managed_lb.py
anandsyncs: Scale-up now records the source certificate Secret UID per shard. After the stale managed copy is removed, scale-down asserts the source Secret still has that exact UID, detecting delete-and-reissue.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: I kept the recreation scenario unchanged rather than adding speculative operator behavior: the test requires a new StatefulSet UID, readiness, and restored queryability, and the prior priority-30 community patch completed without reproducing a terminating-PVC collision. The separate scale-to-zero test continues to verify the intended PVC reclaim behavior explicitly.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
anandsyncs: Consolidated the repeated lifecycle waiters, PVC lookup, and Search artifact assertions into the existing `tests/common/search/connectivity.py`. Managed TLS naming is now shared through the existing `search_resource_names.py`; no helper module was added.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
anandsyncs: The Q2 delete test now captures and later compares UIDs for the source TLS Secret, sync-user password Secret, and CA ConfigMap on each member client. That detects both deletion and transparent recreation while still verifying all operator-managed artifacts are removed.

docker/mongodb-kubernetes-tests/kubetester/operator.py
anandsyncs: Increased the shared operator termination bound to 40 attempts at three seconds, providing approximately 120 seconds for loaded CI nodes.

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
anandsyncs: Removed the one-use `RecreatedResource` alias; the waiter accepts the resource directly and reads its UID.

controllers/operator/watch/config_change_handler.go
anandsyncs: Follow-up in `e8f83a44e`: the gate is now type-specific—only ConfigMap/Secret deletes with a nil `MapFunc` retain the legacy ignore behavior. Tracked MongoDB and other non-ConfigMap/Secret deletes still call `doHandle` with a nil map function, with a regression test covering that path.

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
```diff
@@ -571,3 +717,32 @@ def test_cross_cluster_isolation_absence(
...
+            "svc": search_resource_names.mongot_service_name_for_cluster(MDBS_RESOURCE_NAME, idx),
+            "proxy": search_resource_names.mc_proxy_svc_name(MDBS_RESOURCE_NAME, idx),
+            "mongot_cm": search_resource_names.mongot_configmap_name_for_cluster(MDBS_RESOURCE_NAME, idx),
+            "envoy_deployment": search_resource_names.lb_deployment_name(MDBS_RESOURCE_NAME, idx),
+            "envoy_cm": search_resource_names.lb_configmap_name(MDBS_RESOURCE_NAME, idx),
+            "operator_tls_secret": search_resource_names.operator_managed_tls_secret_name(MDBS_RESOURCE_NAME),
+        }
+
+        assert_search_artifacts_present(mcc, namespace, names)
+
```
anandsyncs: **minor (coverage):** the q2 delete test got the protected-input survival asserts (source cert secret / sync-user password / CA ConfigMap uid-compare after the sweep), but this delete test didn't — an over-broad sweep selector on member clusters would still pass here. Same pattern as q2: capture the uids before `mdbs.delete()` and re-read after `wait_for_search_artifacts_deleted`.

anandsyncs (commented):
Re-reviewed at `133f2c37c` — 18 of the 19 previous findings are properly fixed, each of the three majors with a load-bearing test:

- Deployment watches added centrally + per-member with routing tests; CM/Secret delete routing gated on `MapFunc != nil` with the legacy-ignores-delete semantics pinned by test (verified every `ResourcesHandler` registration in the repo keeps baseline behavior); all four Envoy/metrics mutate funcs strip legacy owner refs, with tests pre-seeding stale refs so the assertion is load-bearing.
- Sweep deletes now carry UID preconditions (interceptor test proves a recreated object survives); dead helpers/options machinery removed with no dangling references; `ReapplyProtectedSearchLabels` reused.
- The restart e2e now proves the **new** operator process reconciled (pod-proxy `/metrics`: `reconcile_total > 0` + idle workqueue after a hard zero-pods barrier) before creating drift; the TLS source-secret uid flows from scale-up to the post-sweep assert, so cert-manager re-issuance can no longer mask a wrong deletion.

Remaining: one coverage gap + three nits inline. One validation note: the pod-proxy metrics barrier in `simulated_mc_rs.py` is a first use of that path in this suite — statically verified (port/labels/controller name vs controller-runtime v0.20.4) and it fails loud rather than vacuously, but the green Evergreen patches in the description predate these commits, so please run `e2e_search_simulated_mc_rs` (ideally also the community + sharded-scale tasks) once at this head before merging.

controllers/operator/watch/config_change_handler.go
```diff
@@ -83,16 +86,24 @@ func (c *ResourcesHandler) doHandle(namespace, name string, q workqueue.TypedRat
...
+	switch e.Object.(type) {
+	case *corev1.ConfigMap, *corev1.Secret:
```
anandsyncs: **nit:** "has `MapFunc`" as the proxy for "gets CM/Secret delete routing" is correct for every registration today, but a future legacy handler that adds a MapFunc for create/update mapping would silently opt into delete routing. A one-line comment stating the invariant here would keep that from happening by accident.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -515,6 +524,11 @@ def _per_cluster_envoy_configmap_name(mdbs_name: str, cluster_index: int) -> str
     return f"{mdbs_name}-search-lb-{cluster_index}-config"
```
anandsyncs, line 515: **nit:** `_per_cluster_mongot_config_name` / `_per_cluster_envoy_deployment_name` / `_per_cluster_envoy_configmap_name` duplicate `search_resource_names.mongot_configmap_name_for_cluster` / `lb_deployment_name` / `lb_configmap_name` — `simulated_mc_rs.py` already uses the shared ones; this file is the last holdout after the helper lift.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -912,6 +911,118 @@ def deleted() -> tuple[bool, str]:
...
+            return resource.metadata.uid != old_uid, f"{what} UID is {resource.metadata.uid}"
+        except client.exceptions.ApiException as exc:
+            if exc.status == 404:
+                return False, f"{what} is absent"
+            raise
+
+    run_periodically(recreated, timeout=timeout, sleep_time=5, msg=f"{what} recreation")
+
+
+def wait_for_search_deleted(mdbs: "MongoDBSearch", timeout: int = 300) -> None:
```
anandsyncs: **nit:** `wait_for_search_deleted` re-implements the loop above — it can delegate to `wait_for_resource_deleted(mdbs.load, ...)`.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
anandsyncs: Removed all three local naming helpers and replaced every call site with `mongot_configmap_name_for_cluster`, `lb_deployment_name`, and `lb_configmap_name` from `search_resource_names`.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
anandsyncs: `wait_for_search_deleted` now delegates directly to `wait_for_resource_deleted(mdbs.load, ...)`, preserving the timeout and API error behavior.

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
anandsyncs: Added shared `protected_search_input_uids` coverage: simulated MC now captures the source TLS Secret, sync-user password Secret, and CA ConfigMap UIDs through each member client before deletion and requires the same UIDs after managed-artifact cleanup. Q2 now uses the same helper.

controllers/operator/watch/config_change_handler.go
anandsyncs: Added the invariant comment: `MapFunc` is intentionally the Search-only opt-in for ConfigMap/Secret delete routing, and legacy handlers must not set it merely for create/update mapping.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -760,6 +760,116 @@ def test_verify_per_cluster_envoy_deployment(
         logger.info(f"Envoy Deployment {envoy_deployment_name} ready in cluster {mcc.cluster_name} (idx={cluster_idx})")
 
 
+@mark.e2e_search_q2_mc_rs_steady
+def test_named_single_cluster_shape_no_owner_refs_and_cleanup(
```
anandsyncs: **coverage gap 1 (highest risk):** self-healing recreation is never exercised in true central+member topology — the remote-member-cache → central-operator label-routed event path at the heart of this PR is only recreation-verified in simulated MC, where central ≡ member. Cheapest fold, right here: before `named_single.delete()`, delete the mongot ConfigMap + Envoy Deployment (and the operator TLS secret copy, which has a new watch but no recreation test in any topology) on the target member and `wait_for_resource_recreated` on each.

controllers/operator/mongodbsearch_metrics_controller.go
```diff
@@ -984,6 +979,7 @@ func (r *MongoDBSearchMetricsForwarderReconciler) ensureMetricsForwarderDeployme
 	}
 
 	_, err := controllerutil.CreateOrUpdate(ctx, c, dep, func() error {
+		dep.OwnerReferences = nil
```
anandsyncs: **coverage gap 2:** with the owner refs gone, the label sweep is now the *only* thing that deletes the metrics-forwarder Deployment/ConfigMap on CR delete — and no e2e in any topology asserts they're removed (`search_enterprise_metrics_forwarder_replicaset.py::test_deleteing_search_resource_deletes_hosts` only checks OM host deregistration). Two `wait_for_resource_deleted` calls folded into that test close it.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -85,6 +105,17 @@ def test_create_search_resource(mdbs: MongoDBSearch):
     mdbs.assert_reaches_phase(Phase.Running, timeout=600)
 
 
+@mark.e2e_search_community_basic
+def test_search_statefulset_preserves_custom_and_protected_labels(namespace: str, mdbs: MongoDBSearch):
```
anandsyncs: **coverage gap 3:** the new label-removal routing (stripping owner labels triggers reconcile + restoration) has unit coverage only. Fold here: patch the STS to drop `mongodb.com/search-name`, then poll until the operator restores it — that exercises the predicate's label-removal path against a real cluster.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -52,6 +68,10 @@ def mdbs(namespace: str) -> MongoDBSearch:
     )
 
     try_load(resource)
+    resource["spec"]["clusters"][0]["statefulSet"] = {
+        "metadata": {"labels": {CUSTOM_STATEFULSET_LABEL: CUSTOM_STATEFULSET_LABEL_VALUE}},
```
anandsyncs: **coverage nit:** the override merge is additive, so this test still passes if `ReapplyProtectedSearchLabels` were dropped entirely. Adding `"component": "bogus"` to this override map makes the existing `component == "mongot"` assertion discriminating — one line.

anandsyncs (commented):
Behavior→e2e coverage matrix for the PR (every new operator behavior mapped to the e2e whose assertions fail if it breaks). Overall: coverage is strong in single-cluster and simulated-MC — the delete sweep, recreation with uid-change asserts, restart reconvergence with the metrics barrier, and TLS-copy cleanup are all genuinely discriminating. Unit-only is justified for the stale-cache/UID-precondition races, fail-closed confirmation, state-CM UID guard, and the negative side of delete-routing gating (not observable as a cluster outcome).

Four gaps worth closing (inline), ranked; 1–2 I'd treat as pre-merge, 3–4 as cheap folds:

1. no recreation test in true central+member MC (the PR's core routing path)
2. metrics-forwarder Deployment/CM sweep-on-delete unasserted in any topology (owner refs no longer delete them)
3. label-strip → restore untested e2e
4. one-line fixture change to make the protected-label test discriminating

Follow-ups that live outside this diff (no inline anchor): `tests/upgrades/operator_upgrade_search.py` proves post-upgrade convergence but never asserts owner refs were stripped, nor does a post-upgrade CR delete — folding owner-refs-empty asserts plus a final `mdbs.delete()` + absence waits there covers the adoption path end-to-end. Minor: state-ConfigMap deletion and proxy-Service recreation are never asserted.

controllers/operator/mongodbsearch_metrics_controller.go
anandsyncs: The deletion case now re-enables the metrics forwarder after the disable scenario, waits for Running, presence-guards its Deployment and ConfigMap, preserves host-deregistration sequencing, and then requires both resources to be deleted by the Search sweep.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: Added a live label-removal test that merge-patches `mongodb.com/search-name` off the mongot StatefulSet, confirms the patch removed it, and polls until the operator restores the correct Search owner label.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
anandsyncs: Added true central/member recreation coverage in Q2: the target member now deletes the main Search mongot ConfigMap, Envoy Deployment, and operator-managed TLS copy, requires changed UIDs for all three, waits for Envoy readiness, and confirms Search returns to Running before cleanup continues.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: Added `component: bogus` to the user StatefulSet metadata override. The existing `component == mongot` assertion is now load-bearing for protected-label reapplication while the unrelated custom label remains preserved.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -668,22 +669,28 @@ func (r *MongoDBSearchReconcileHelper) applyReconcileUnit(
 		}
 	}
 
+	tlsSecretLabels := searchResourceIdentityLabels(r.mdbSearch, unit.clusterName)
+	ingressTLSSecretLabels := maps.Clone(tlsSecretLabels)
+	if unit.shardName != "" {
+		ingressTLSSecretLabels[componentLabelKey] = mongotComponent
+	}
```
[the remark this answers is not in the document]
anandsyncs: Fixed in `e0d3deb2a`: every operator-managed ingress TLS Secret now receives `component=mongot` independent of shard topology. Added table-driven replica-set and sharded coverage for owner name/namespace/UID/component identity and legacy single-cluster cluster-label omission.

controllers/operator/state_store.go
```diff
@@ -57,14 +64,39 @@ func (s *StateStore[S]) read(ctx context.Context) error {
...
 func (s *StateStore[S]) write(ctx context.Context, log *zap.SugaredLogger) error {
+	if s.data == nil {
+		s.data = map[string]string{}
+	}
+	if s.ownerUID != "" {
+		s.data[stateOwnerUIDKey] = s.ownerUID
+	}
+
+	ownerReferences := s.ownerReferences
+	if s.ownerUID != "" && len(s.readOwnerRefs) > 0 {
```
anandsyncs: **Stack-boundary churn (applies to the whole owner-ref layer in this PR):** this `readOwnerRefs` merge machinery — plus the dual-owner-ref adoption in `search_state.go MutateSearchState`, `searchResourceIdentityLabels` in `mongodbsearch_reconcile_helper.go`, the owner-ref plumbing in `pkg/tls/tls_secret.go`, and their dedicated tests — is deleted wholesale by #1382 in the next layer of this stack. Since the layers land together, consider trimming this PR to the durable primitives only (the state UID data key, the `mongodb.com/search-uid` label, the `labels` param on `EnsureTLSSecret`, and the UID-mismatch reset) and leaving owner-ref handling untouched so #1382 removes it in one motion. Zero behavior delta at the tip of the stack, and it removes a few hundred lines of throwaway review surface from both PRs.

controllers/searchcontroller/search_state_test.go
```diff
@@ -80,6 +85,119 @@ func TestRoutingSwitch_StateCMWrites(t *testing.T) {
...
+		}))
+
+		helper := newHelper(c)
+		require.NoError(t, helper.markRoutingReady(ctx, "sh-0"))
+		assert.Equal(t, []string{"sh-0"}, readState(t, c).RoutingReadyMongotGroups)
+		assert.False(t, switchedOn(helper, "stale-shard"))
+
+		cm := &corev1.ConfigMap{}
+		require.NoError(t, c.Get(ctx, stateCMName, cm))
+		require.Len(t, cm.OwnerReferences, 2)
```
anandsyncs: **Test-side of the stack-boundary churn (companion to the state_store.go comment):** the owner-reference assertions added here — plus the `Len(2)`/controller-ref asserts in the metrics `TestOpenTopologyStateStore_UIDSemantics` and the owner-ref asserts added to `TestReconcileReplicaSet_CreatesResources`/`TestReconcile_IngressTLSSecretIdentity`/`TestEnsureX509ClientCertConfig_*` — are all rewritten to `assert.Empty(OwnerReferences)` by #1382. That's ~40-50 test lines pinning a design that never reaches the tip of the stack. If this PR is trimmed to the durable primitives, these tests should assert the label-based end state once and never mention owner refs.

controllers/operator/mongodbsearch_metrics_controller.go
```diff
@@ -1124,10 +1125,11 @@ func metricsForwarderResourceRequirements(search *searchv1.MongoDBSearch) corev1
 // clusterName=="" (single-cluster/central): cluster-name label is omitted.
 func metricsForwarderLabelsForCluster(search *searchv1.MongoDBSearch, clusterName string, clusterIndex int) map[string]string {
 	labels := map[string]string{
-		"app":                                search.MetricsForwarderDeploymentNameForCluster(clusterIndex),
-		"component":                          metricsForwarderLabelName,
-		khandler.MongoDBSearchOwnerNameLabel: search.Name,
+		"app":                                     search.MetricsForwarderDeploymentNameForCluster(clusterIndex),
+		khandler.MongoDBSearchComponentLabel:      metricsForwarderLabelName,
```
viveksinghggits: naming of this const `metricsForwarderLabelName` is not part of this PR but it should be `metricsForwarderLabelValue` right? Let's change that maybe?

controllers/operator/mongodbsearchenvoy_controller.go
```diff
@@ -555,6 +555,7 @@ func (r *MongoDBSearchEnvoyReconciler) ensureDeployment(ctx context.Context, sea
 			dep.Labels = merge.StringToStringMap(dep.Labels, managedLB.Deployment.MetadataWrapper.Labels)
 			dep.Annotations = merge.StringToStringMap(dep.Annotations, managedLB.Deployment.MetadataWrapper.Annotations)
 		}
+		dep.Labels = khandler.ReapplyProtectedSearchLabels(dep.Labels, labels)
```
viveksinghggits: Should `ReapplyProtectedSearchLabels` not be called `appendProtectedSearchlabels` or something like that? Or `reapply` is significant?

controllers/operator/state_store.go
```diff
@@ -29,6 +33,8 @@ type StateStore[S any] struct {
 	resourceName    string
 	ownerLabels     map[string]string
 	ownerReferences []metav1.OwnerReference
+	readOwnerRefs   []metav1.OwnerReference
```
viveksinghggits: q: what does `readOwnerRefs` have?

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -136,17 +134,20 @@ func searchOwnerLabels(search *searchv1.MongoDBSearch, clusterName string) map[s
 	return labels
 }
 
+func searchResourceIdentityLabels(search *searchv1.MongoDBSearch, clusterName string) map[string]string {
```
viveksinghggits: I like the idea of UID but was just wondering if we get into any issues if we just relied on name/namespace combination to figure out if a particular search resource created the children resource.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -136,17 +134,20 @@ func searchOwnerLabels(search *searchv1.MongoDBSearch, clusterName string) map[s
...
+	labels := searchOwnerLabels(search, clusterName)
+	labels[khandler.MongoDBSearchOwnerUIDLabel] = string(search.UID)
```
viveksinghggits: should this also add the `component: mongot` label?

controllers/searchcontroller/search_state.go
```diff
@@ -50,6 +54,11 @@ func searchStateFromCM(cm *corev1.ConfigMap) (*SearchDeploymentState, error) {
 	return state, nil
 }
 
+func searchStateHasCurrentUID(cm *corev1.ConfigMap, search *searchv1.MongoDBSearch) bool {
+	recordedUID, ok := cm.Labels[khandler.MongoDBSearchOwnerUIDLabel]
+	return !ok || recordedUID == string(search.UID)
+}
```
viveksinghggits: nit:
```suggestion
func searchStateHasUID(cm *corev1.ConfigMap, uid *type) bool {
	recordedUID, ok := cm.Labels[khandler.MongoDBSearchOwnerUIDLabel]
	return !ok || recordedUID == string(search.UID)
}
```
I don't think having current is important here without defining what current is.

controllers/searchcontroller/search_state.go
```diff
@@ -99,18 +113,46 @@ func MutateSearchState(ctx context.Context, c kubernetesClient.Client, search *s
...
+	uidMatches := searchStateHasCurrentUID(cm, search)
+	state := NewSearchDeploymentState()
+	if uidMatches {
+		state, err = searchStateFromCM(cm)
+		if err != nil {
+			return nil, err
+		}
 	}
-	if !mutate(state) {
+
```
viveksinghggits, line 107: I think some small comment here would be helpful for later, to understand what's happening.

viveksinghggits (commented):
looks pretty good.

controllers/operator/state_store.go
anandsyncs: Keeping this stack boundary is intentional. Layer 1 must be independently safe while it introduces UID identity and adoption, so it preserves the existing central-cluster GC backstop. Layer 2 removes owner references only after the uncached absence-confirmed, UID/RV-preconditioned sweep exists. Dropping them here would create a cleanup gap between layers. Final Layer-1 head `fc9a4bb6` passed the exact 69/69 matrix.

controllers/searchcontroller/search_state_test.go
anandsyncs: Same disposition on the tests: these assertions pin Layer 1's transitional safety contract, including preserving the existing controller reference while adding the current UID adoption reference. Layer 2 deliberately replaces that contract with label-only cleanup and its own tests. Keeping both makes each stacked layer valid independently; Layer 1 passed 69/69 at `fc9a4bb6`.

controllers/operator/mongodbsearch_metrics_controller.go
anandsyncs: Agreed that `Value` would be the more literal suffix. The constant predates this change and the name has no behavioral ambiguity at its use sites, so I left the unrelated rename out rather than invalidate the reviewed/evidenced stack for cosmetic churn.

controllers/operator/mongodbsearchenvoy_controller.go
anandsyncs: `Reapply` is significant here: the helper runs after user metadata is merged and overwrites the protected identity keys while preserving unrelated labels. `append` would imply that a clobbered or removed protected value is not repaired. The lifecycle E2E removes `search-name` and verifies the complete protected set plus the custom user label is restored.

controllers/operator/state_store.go
anandsyncs: At the final Layer-1 head, `StateStore.read` clones the ConfigMap's existing owner references when UID tracking is enabled. `write` preserves that existing controller reference and appends the current UID as a non-controller adoption reference when missing. That avoids replacing the old controller during same-name adoption until Layer 2's sweep-based cleanup takes over.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: Name/namespace is insufficient because a MongoDBSearch can be deleted and recreated with the same key. The UID is the load-bearing generation fence: stale resources and stale reconciles must never act on the replacement CR. The final stack stamps `mongodb.com/search-uid`, rejects mismatches, and exercises same-name recreation in both focused tests and the successful 69-task matrices.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: Addressed. `withSearchOwnerLabels` now adds `MongoDBSearchComponentLabel=mongot` before reapplying the protected labels, and the matching resource/TLS identity coverage verifies the component label. Final Layer-1 evidence is 69/69.

controllers/searchcontroller/search_state.go
anandsyncs: Addressed in the final head: the helper is now `searchStateUIDMatchesOrMissing`, which states both the UID comparison and the accepted legacy missing-label case without relying on an ambiguous notion of current.

controllers/searchcontroller/search_state.go
anandsyncs: Addressed. `MutateSearchState` now documents the resourceVersion-checked read/modify/write, UID-mismatch reset, and legacy missing-UID adoption behavior; the metadata block also explains why an existing controller reference is preserved during adoption.

viveksinghggits (approved):

pkg/handler/labels_search.go
```diff
@@ -3,23 +3,65 @@ package handler
...
+	}
+	if clusterName != "" {
+		labels[MongoDBSearchClusterNameLabel] = clusterName
+	}
+	return labels
+}
+
+// ReapplyProtectedSearchLabels restores labels used for Search event routing
+// and cleanup after user metadata overrides while preserving unrelated labels.
+func ReapplyProtectedSearchLabels(labels, desired map[string]string) map[string]string {
```
anandsyncs: Updated target: please remove this protected-label abstraction and its controller-specific tests. Each writer can merge user labels first and then apply the required owner/component/cluster labels last; that ordering protects identity without a second reapply pass or a maintained protected-key list.

controllers/operator/mongodbsearch_metrics_controller_test.go
```diff
@@ -223,6 +229,186 @@ func getTopologyState(t *testing.T, c client.Client, search *searchv1.MongoDBSea
 	return state.Clusters[""]
 }
 
+func TestOpenTopologyStateStore_UIDSemantics(t *testing.T) {
```
anandsyncs: Please test the generic StateStore UID contract once at the StateStore boundary with a compact table (matching marker, mismatched marker, and markerless state with a matching controller owner) instead of four metrics-specific blocks spanning roughly 180 lines. Controller tests should cover controller behavior, not re-test the shared store implementation.

pkg/handler/labels_search.go
anandsyncs: Done - `ReapplyProtectedSearchLabels` is removed. Writers now merge user/override labels first and stamp the managed owner/component/cluster labels last, all through the single `MongoDBSearchManagedLabels` helper; the label-drift e2e is replaced by one compact table test (`TestSearchManagedLabelsWinOverOverrides`) covering override precedence for member and single-cluster shapes.

controllers/operator/mongodbsearch_metrics_controller_test.go
anandsyncs: Done - the four metrics-specific blocks (~186 lines) are replaced by one table-driven `TestStateStoreOwnerUIDContract` at the store boundary (marker match reads; marker mismatch rejects despite a matching owner ref; markerless adoption/rejection rows; empty store-owner legacy semantics; each row also proves the next write stamps the marker).

controllers/searchcontroller/search_state_test.go
```diff
@@ -142,3 +260,71 @@ func TestRoutingSwitch_StateCMWrites(t *testing.T) {
...
+
+func TestReadSearchState_UIDGuardBehavior(t *testing.T) {
+	ctx := context.Background()
+	search := newTestMongoDBSearch("mysearch", mock.TestNamespace)
+	search.UID = "new-search-uid"
+
+	storedStateRaw, err := json.Marshal(SearchDeploymentState{RoutingReadyMongotGroups: []string{"stored-shard"}})
+	require.NoError(t, err)
+
+	tests := []struct {
```
anandsyncs: `TestStateStoreOwnerUIDContract` already covers the explicit-marker matrix. Keep this Search-specific test only for semantics that differ from generic StateStore: markerless/no-owner is adopted, markerless/foreign-controller resets, explicit mismatch resets. Move the read-only metadata assertion to one representative case instead of re-reading and checking owner refs in every row. This preserves the distinct Search contract in a much smaller table.

controllers/searchcontroller/search_state.go
```diff
@@ -50,6 +54,15 @@ func searchStateFromCM(cm *corev1.ConfigMap) (*SearchDeploymentState, error) {
 	return state, nil
 }
 
+func searchStateUIDMatchesOrMissing(cm *corev1.ConfigMap, search *searchv1.MongoDBSearch) bool {
```
anandsyncs: Please remove UID-mismatch and markerless stale-state discrimination together with their tests. Keep identity stamping and ordinary in-place upgrade adoption, but deserialize existing state without same-name replacement fencing; that replacement race is explicitly unsupported.

controllers/operator/state_store.go
```diff
@@ -103,6 +120,32 @@ func (s *StateStore[S]) ReadState(ctx context.Context) (*S, error) {
 	}
 }
 
+func (s *StateStore[S]) isStaleOwnerUID() bool {
```
anandsyncs: Same-name replacement/UID replay recovery is out of scope. Please retain StateStore’s ordinary read/write semantics and remove the ownerUID marker, mismatch rejection, constructor parameter, readOwnerReferences snapshot, and dedicated contract test.

controllers/operator/mongodbsearchenvoy_controller.go
```diff
@@ -555,6 +555,9 @@ func (r *MongoDBSearchEnvoyReconciler) ensureDeployment(ctx context.Context, sea
 			dep.Labels = merge.StringToStringMap(dep.Labels, managedLB.Deployment.MetadataWrapper.Labels)
 			dep.Annotations = merge.StringToStringMap(dep.Annotations, managedLB.Deployment.MetadataWrapper.Annotations)
 		}
+		// The managed identity labels are stamped last so user label overrides
```
viveksinghggits: Let's simplify this comment to make it human readable.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
```diff
@@ -126,27 +127,17 @@ func NewMongoDBSearchReconcileHelper(
...
 }
 
-// withSearchOwnerLabels merges the search-owner labels into the STS ObjectMeta
-// labels. It does NOT touch the selector/pod-template (which is immutable on
-// existing STS); the pod-routing labels stay in the unit's podLabels.
+// withSearchOwnerLabels stamps the managed identity labels onto the STS
+// ObjectMeta labels. It runs last in the modification chain so user override
+// labels merged earlier can never win over them. It does NOT touch the
+// selector/pod-template (which is immutable on existing STS); the pod-routing
+// labels stay in the unit's podLabels.
```
viveksinghggits: Let's make it human understable, i.e. use simple words not the words like stamps etc.

pkg/handler/labels_search.go
```diff
@@ -3,23 +3,44 @@ package handler
...
+	MongoDBSearchComponentLabel      = "component"
+	// MongoDBSearchClusterNameLabel records the target member cluster.
 	MongoDBSearchClusterNameLabel = "mongodb.com/cluster-name"
 )
 
+// MongoDBSearchManagedLabels returns the managed identity labels shared by
+// Search resource writers, cleanup selectors, and event routing. Writers merge
+// user labels first and stamp these labels last so user metadata overrides can
+// never detach a resource from its owning MongoDBSearch.
+func MongoDBSearchManagedLabels(search metav1.Object, app, component, clusterName string) map[string]string {
```
viveksinghggits: ownerLabelsForSearch(search,) ownerLabels

controllers/searchcontroller/search_state.go
anandsyncs: Done — the UID fence and marker discrimination are removed entirely; state reads are master's plain read-by-name. Per the follow-up decision this went further: the search-uid label and all uid-based delete selectivity are removed too (the UID survives only inside ownerReferences, which garbage-collect the state ConfigMaps on CR delete in every operator mode). Same-name-successor overlap within the deletion window is documented as an accepted race at the sweep sites. Validated by the canonical matrix at head.

controllers/operator/state_store.go
anandsyncs: Done — state_store.go is back to master's shape (plus a nil-map normalize on read from the bot review); the UID contract table test is deleted with the mechanism, and NewStateStore is back to its 3-argument signature.

controllers/operator/state_store.go
```diff
@@ -57,10 +63,18 @@ func (s *StateStore[S]) read(ctx context.Context) error {
 	}
 
 	s.data = cm.Data
+	s.readOwnerReferences = append([]metav1.OwnerReference(nil), cm.OwnerReferences...)
 	return nil
```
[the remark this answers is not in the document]
anandsyncs: Fixed — read() now normalizes nil ConfigMap data to an empty map; the assertion is folded into the existing state-store test (which panics without the fix).

controllers/operator/mongodbsearch_controller.go
```diff
@@ -182,7 +182,9 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
 		}
 	}
 
-	state, err := searchcontroller.ReadSearchState(ctx, r.kubeClient, mdbSearch)
+	state, err := searchcontroller.MutateSearchState(ctx, r.kubeClient, mdbSearch, func(*searchcontroller.SearchDeploymentState) bool {
+		return false
+	})
 	if err != nil {
 		return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Failed(xerrors.Errorf("failed to read search state: %w", err)), log)
 	}
```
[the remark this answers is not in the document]
anandsyncs: Fixed — the caller now requeues (Pending) on apierrors.IsConflict instead of marking the CR Failed, matching the operator's conflict conventions; covered in the existing reconcile test.

controllers/searchcontroller/search_state_test.go
anandsyncs: Done — with the fencing removed, TestStateStoreOwnerUIDContract and the read-path guard table are deleted outright. The residual Search-specific adoption semantics are three plain subtests: legacy label repair on write; foreign-owned CM adoption preserving the existing controller reference (the one representative case carrying the full owner-ref/label assertions); and no-op mutate creating no state.

controllers/operator/mongodbsearchenvoy_controller.go
anandsyncs: Reworded — the comment now states plainly that the operator's identity labels are applied after the user's label overrides so users cannot remove or change them.

controllers/searchcontroller/mongodbsearch_reconcile_helper.go
anandsyncs: Reworded in plain language (no "stamps"); kept the two constraints it documents: it runs last so user labels can't win, and it deliberately doesn't touch the selector/pod template (immutable on an existing StatefulSet).

pkg/handler/labels_search.go
anandsyncs: Renamed to `SearchManagedLabels`, keeping it exported — it has cross-package callers in all three controllers and the API package, so a lowercase `ownerLabelsForSearch` wouldn't compile. Kept "managed" rather than "owner" because it returns more than the owner pair: it optionally adds the app, component, and cluster-name routing/selection labels. (The uid entry is gone entirely per the successor-race decision.)

controllers/searchcontroller/search_state.go
```diff
@@ -103,9 +108,35 @@ func MutateSearchState(ctx context.Context, c kubernetesClient.Client, search *s
 	if err != nil {
 		return nil, err
 	}
-	if !mutate(state) {
+
+	stateChanged := mutate(state)
+	metadataChanged := false
+	// Adopt legacy metadata without replacing an existing controller owner reference.
```
anandsyncs: This still adopts a state ConfigMap whose controller owner is a previous MongoDBSearch UID by appending the current CR as a second, non-controller owner. Same-name replacement recovery is explicitly out of scope, and ordinary in-place 1.9.x upgrade state already has the current controller owner. Please remove this foreign-owner adoption block and its previous-CR test; keep only owner-label repair for the normal upgrade path.

anandsyncs:
Go-test proportionality is improved but still high: the isolated delta is +227/-17, with a realistic target of roughly +92–116. The largest remaining reductions are in search_state_test.go (remove previous-CR adoption and unrelated no-op rows; keep creation, one metadata repair, append/prune/conflict) and mongodbsearch_reconcile_helper_test.go (one representative identity writer instead of repeated Service/StatefulSet/ConfigMap checks). This preserves every supported invariant while removing about 110–135 test lines.

controllers/operator/mongodbsearch_controller.go
```diff
@@ -180,9 +182,18 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
 		}
 	}
 
-	state, err := searchcontroller.ReadSearchState(ctx, r.kubeClient, mdbSearch)
+	// The no-op mutation reads the state and, as a side effect, repairs legacy
```
anandsyncs, line 183: [Explanation] This applies new labels one time because GA deployments won't have them.

pkg/tls/tls_secret.go
```diff
@@ -31,7 +31,7 @@ type TLSConfigurableResource interface {
 // ensureTLSSecret will create or update the operator-managed Secret containing
```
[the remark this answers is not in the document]
anandsyncs: Fixed — the doc comment now starts with the exported name EnsureTLSSecret.

controllers/operator/mongodbsearch_controller.go
```diff
@@ -182,8 +182,17 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
...
+	state, err := searchcontroller.MutateSearchState(ctx, r.kubeClient, mdbSearch, func(*searchcontroller.SearchDeploymentState) bool {
+		return false
+	})
 	if err != nil {
+		// A concurrent writer bumped the ConfigMap between read and update; retry
+		// instead of marking the CR Failed over a transient race.
+		if apierrors.IsConflict(err) {
+			return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Pending("Search state was modified concurrently, retrying").Requeue(), log)
+		}
 		return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Failed(xerrors.Errorf("failed to read search state: %w", err)), log)
```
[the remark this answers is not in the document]
anandsyncs: Fixed — the message is now "failed to read or repair search state", matching what MutateSearchState can actually fail doing (it also creates the ConfigMap and repairs legacy owner labels).

controllers/searchcontroller/search_state.go
```diff
@@ -103,9 +108,35 @@ func MutateSearchState(ctx context.Context, c kubernetesClient.Client, search *s
...
+	if !slices.ContainsFunc(cm.OwnerReferences, func(ref metav1.OwnerReference) bool {
+		return ref.UID == search.UID
+	}) {
+		cm.OwnerReferences = append(cm.OwnerReferences, metav1.OwnerReference{
+			APIVersion: v1.SchemeGroupVersion.String(),
+			Kind:       "MongoDBSearch",
```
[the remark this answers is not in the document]
anandsyncs: Moot in the final shape — the owner-reference append this literal lived in is removed entirely (verified at tags 1.9.0/1.9.1: released operators already stamp a controller owner reference on every state ConfigMap write, so the adoption append was dead on all supported paths). The only remaining owner-ref construction goes through kube.BaseOwnerReference, which already uses GetKind().

controllers/searchcontroller/search_state.go
anandsyncs: Done — the foreign-owner adoption block and its previous-CR test are removed; MutateSearchState now only repairs legacy owner labels on existing ConfigMaps. Verified at tags 1.9.0/1.9.1: both state ConfigMaps get a controller owner reference on every write, so ordinary upgrade state never needed the append — and in the same-name-successor window the append actually rescued the predecessor's stale state from garbage collection, contradicting the accepted-race stance. The label repair stays because it is load-bearing: the removed-operator state-CM sweep selects by the owner labels, which GA-written ConfigMaps lack. PR body corrected accordingly (GA CMs lack owner labels, not references).

anandsyncs:
Round-5 addressed: test delta is now +164 vs master (from +227). The adoption-append block and its
test rows are removed outright — verified at tags 1.9.0/1.9.1 that both state ConfigMaps carry
controller owner references on every write, so adoption reduces to owner-label repair (which stays:
the cleanup sweeps' label-based List cannot see GA-written ConfigMaps without it). The remaining
delta pins label precedence and the repair contract; the residual gap to ~+100 is sole coverage.
All three Copilot items are addressed (the GetKind suggestion went moot with the block's removal).

controllers/operator/mongodbsearch_controller.go
```diff
@@ -180,9 +182,18 @@ func (r *MongoDBSearchReconciler) Reconcile(ctx context.Context, request reconci
...
+	// owner labels on the state ConfigMap.
+	state, err := searchcontroller.MutateSearchState(ctx, r.kubeClient, mdbSearch, func(*searchcontroller.SearchDeploymentState) bool {
+		return false
+	})
 	if err != nil {
-		return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Failed(xerrors.Errorf("failed to read search state: %w", err)), log)
+		// A concurrent writer bumped the ConfigMap between read and update; retry
+		// instead of marking the CR Failed over a transient race.
+		if apierrors.IsConflict(err) {
+			return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Pending("Search state was modified concurrently, retrying").Requeue(), log)
```
viveksinghggits:
```suggestion
			return commoncontroller.UpdateStatus(ctx, r.kubeClient, mdbSearch, workflow.Pending("Search state was modified concurrently, re-queuing ").Requeue(), log)
```

viveksinghggits: retry gives the feeling that it's being retried immediately but we that's not the case. re-queuing actually depicts what's happening.

pkg/handler/labels_search.go
```diff
@@ -6,20 +6,40 @@ import (
...
-	// per-cluster member resources (Envoy Deployment + ConfigMap).
+	MongoDBSearchComponentLabel      = "component"
+	// MongoDBSearchClusterNameLabel records the target member cluster.
 	MongoDBSearchClusterNameLabel = "mongodb.com/cluster-name"
 )
 
+// SearchManagedLabels returns the managed identity labels shared by Search
+// resource writers, cleanup selectors, and event routing. Writers merge user
+// labels first and apply these labels last so user metadata overrides can
+// never detach a resource from its owning MongoDBSearch.
```
viveksinghggits: nit:
This comment focuses on where these labels are used. We should instead focus on what this function does.
I would recommend to just mention that this returns labels that identifies the passed search resource and owner.

pkg/handler/labels_search.go
```diff
@@ -6,20 +6,40 @@ import (
...
+func SearchManagedLabels(search metav1.Object, app, component, clusterName string) map[string]string {
```
viveksinghggits:
```suggestion
func ManagedBySearchLabels(search metav1.Object, app, component, clusterName string) map[string]string {
```

what do you think about this?

viveksinghggits (commented):
Looks good to me, left some small comments.

controllers/searchcontroller/mongodbsearch_reconcile_helper_test.go
```diff
@@ -617,6 +637,49 @@ func TestMongoDBSearchReconcileHelper_ServiceCreation(t *testing.T) {
 	}
 }
 
+func TestSearchManagedLabelsWinOverOverrides(t *testing.T) {
```
anandsyncs: this test can be removed

anandsyncs: verify that it is being tested indirectly in full reconciler tests

pkg/handler/labels_search.go
anandsyncs: user `SearchOwnershipLabels`

anandsyncs:
```suggestion
func SearchOwnershipLabels(search metav1.Object, app, component, clusterName string) map[string]string {
```

ashortt-mdb (approved):

pkg/handler/labels_search.go
anandsyncs: Done in c59f3392a — renamed across all call sites (call sites added by the later stack layers are renamed in their own round-10 commits).

controllers/searchcontroller/mongodbsearch_reconcile_helper_test.go
anandsyncs: Checked: it was not covered indirectly — no full-reconcile test set user label overrides. Folded the assertion into `TestReconcileReplicaSet_CreatesResources` (a user StatefulSet label override carrying a custom label plus wrong-valued ownership keys: the custom label survives, the ownership labels win — through the real reconcile path) and removed the standalone test. Done in c59f3392a.

PR #1363, opened 2026-07-09
chore(e2e): Add transport-level retry for Kubernetes API calls on 5xx

nammn:
## Summary

E2E tests run ~40 minutes. Transient Kubernetes API failures (connection refused, timeout, 5xx) kill tests mid-run, forcing expensive re-runs. This PR adds transport-level retry to the Kubernetes client so ALL API reads get automatic retry on transient failures with zero per-function overhead.

### Why transport-level

- **Catches the actual failure at the actual boundary.** The thing that fails is the network call, not the Python function.
- **Safe by default.** `allowed_methods` stays at urllib3 default (GET/HEAD/OPTIONS only). Writes are never retried at transport — no double-create risk.

## Proof of Work

- `make precommit`: black, isort, ty all pass
- ci

## Checklist

- [x] Have you added changelog file? (skip-changelog label applied — E2E test infra only)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -228,8 +228,8 @@ repos:
       # staticcheck) see the full codebase and pre-existing violations are not silently skipped.
       - id: golangci-lint-full
         args: [--timeout=10m]
-        language_version: 1.26.4
+        language_version: 1.26.5
         priority: 6
       - id: golangci-lint-fmt
-        language_version: 1.26.4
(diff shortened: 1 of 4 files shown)


Julien-Ben (approved):

.pre-commit-config.yaml
nammn, line 231: master has that already changed

MaciejKaras (approved):

PR #1365, opened 2026-07-09
CLOUDP-420746: automate RedHat operator bundle publishing

josvazg:
# Summary

Add scripts/release/publish-openshift-bundles-to-rh.sh:
- Downloads the certified OLM bundle from S3
- Prepares/pushes a branch on the mongodb-forks certified-operators and community-operators repos
- Prints the upstream PR links (PR creation stays manual).

Wire it into Evergreen via a new publish_openshift_bundles_to_rh func, task, and variant using github.generate_token.

## Proof of Work

CI must pass.

✅ Local run passes

```shell
% KEEP_WORKDIR=true VERSION=1.8.1 RH_DRYRUN=true scripts/release/publish-openshift-bundles-to-rh.sh
...
Click the links below to open the upstream PR (mark it as a DRAFT), then ask the team
for a quick look; certified merges on green CI, community may need a close/re-open:
  https://github.com/redhat-openshift-ecosystem/certified-operators/compare/main...mongodb-forks:certified-operators:mongodb-kubernetes-1.8.1?expand=1
  https://github.com/k8s-operatorhub/community-operators/compare/main...mongodb-forks:community-operators:mongodb-kubernetes-1.8.1?expand=1
Workdir kept for inspection: /var/folders/8g/h28ztw4j64x4msqcmg4p9t0h0000gp/T/tmp.GbC0n9rGUb
```

## Checklist

### DIFF
--- a/.evergreen-release.yml
+++ b/.evergreen-release.yml
@@ -213,7 +213,7 @@ buildvariants:
     tags: [ "release" ]
     run_on:
       - ubuntu2404-small
-    allowed_requesters: [ "patch", "github_tag" ]
+    allowed_requesters: [ "github_tag" ]
     depends_on:
       - name: run_conditionally_prepare_and_upload_openshift_bundles
         variant: prepare_openshift_bundles
--- a/.evergreen-release.yml
+++ b/.evergreen-release.yml
@@ -94,6 +94,14 @@ tasks:
           variant: prepare_openshift_bundles
           task: prepare_and_upload_openshift_bundles
 
+  - name: publish_openshift_bundles_to_rh
+    tags: [ "openshift_bundles" ]
+    allowed_requesters: [ "patch", "github_tag" ]
+    commands:
+      - func: clone
+      - func: update_evergreen_expansions
+      - func: publish_openshift_bundles_to_rh
+
   - name: release_kubectl_mongodb_plugin
     tags: [ "binary_release" ]
     allowed_requesters: [ "patch", "github_tag" ]
@@ -196,6 +204,21 @@ buildvariants:
         variant: preflight_release_images
     tasks:
       - name: run_conditionally_prepare_and_upload_openshift_bundles
+
+  # Publishing to the RedHat forks is a separate, manually-triggered variant: the bundle is
+  # pulled from S3, so it does not need to share a host with the prepare task, and the PRs
+  # are draft + human-reviewed so we do not auto-chain it after every release build.
+  - name: publish_openshift_bundles_to_rh
+    display_name: publish_openshift_bundles_to_rh
+    tags: [ "release" ]
+    run_on:
+      - ubuntu2404-small
+    allowed_requesters: [ "patch", "github_tag" ]
+    depends_on:
+      - name: run_conditionally_prepare_and_upload_openshift_bundles
+        variant: prepare_openshift_bundles
+    tasks:
+      - name: publish_openshift_bundles_to_rh
 
   - name: prerelease_kind_code_snippets
     display_name: prerelease_kind_code_snippets
--- a/ci/internal/cli/github_apptoken.go
+++ b/ci/internal/cli/github_apptoken.go
@@ -1,0 +1,68 @@
+package cli
+
+import (
+	"encoding/base64"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/mongodb/mongodb-kubernetes/ci/internal/github"
+)
+
+func newGHAppTokenCmd() *cobra.Command {
+	var (
+		appID          string
+		installationID string
+		pemBase64      string
+		pemFile        string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "app-token",
+		Short: "Mint a short-lived GitHub App installation access token and print it to stdout",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if appID == "" || installationID == "" || (pemBase64 == "" && pemFile == "") {
+				return fmt.Errorf("--app-id, --installation-id, and --pem-base64 (or --pem-file) are required")
+			}
+
+			pem, err := resolvePEM(pemFile, pemBase64)
+			if err != nil {
+				return err
+			}
+
+			token, err := github.MintInstallationToken(cmd.Context(), github.AppTokenInputs{
+				AppID:          appID,
+				InstallationID: installationID,
+				PEM:            pem,
+			})
+			if err != nil {
+				return err
+			}
+			_, err = fmt.Fprintln(cmd.OutOrStdout(), token)
+			return err
+		},
+	}
+
+	cmd.Flags().StringVar(&appID, "app-id", "", "GitHub App ID")
+	cmd.Flags().StringVar(&installationID, "installation-id", "", "installation ID the token is scoped to")
+	cmd.Flags().StringVar(&pemBase64, "pem-base64", "", "base64-encoded PEM private key")
+	cmd.Flags().StringVar(&pemFile, "pem-file", "", "path to a PEM private key file")
+
+	return cmd
+}
+
+func resolvePEM(pemFile, pemBase64 string) ([]byte, error) {
+	if pemFile != "" {
+		return os.ReadFile(pemFile)
+	}
+	if pemBase64 == "" {
+		return nil, fmt.Errorf("no private key provided: pass --pem-file or --pem-base64")
+	}
+	pem, err := base64.StdEncoding.DecodeString(strings.TrimSpace(pemBase64))
+	if err != nil {
+		return nil, fmt.Errorf("decode --pem-base64: %w", err)
+	}
+	return pem, nil
+}
--- a/scripts/release/publish-openshift-bundles-to-rh.sh
+++ b/scripts/release/publish-openshift-bundles-to-rh.sh
@@ -1,14 +1,16 @@
 #!/usr/bin/env bash
 # Publishes the OpenShift OLM bundle to the RedHat operator repositories.
 #
-# It downloads the unified certified bundle tarball (produced and uploaded by the
+# It downloads the unified bundle tarball (produced and uploaded by the
 # prepare_and_upload_openshift_bundles Evergreen task) and, for each RedHat fork,
 # pushes a branch containing the bundle laid out under operators/<package>/<version>/.
+#
 # It intentionally does NOT open the pull requests: it prints the "create PR" URLs and
 # the manual review checklist so a human can raise the DRAFT PRs and drive the review.
 #
-# The same certified bundle is used for both the certified-operators and the
+# The same bundle is used for both the certified-operators and the
 # community-operators repositories (the file contents are identical between them).
+# The "certified" term is used in the tarball file for historical reasons.
 #
 # Usage (runnable standalone from a CLI or from Evergreen):
 #   GH_TOKEN=<token-with-write-to-forks> VERSION=1.3.0 RH_DRYRUN=true \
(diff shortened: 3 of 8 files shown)


scripts/release/publish-openshift-bundles-to-rh.sh
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
+# Publishes the OpenShift OLM bundle to the RedHat operator repositories.
+#
+# It downloads the unified certified bundle tarball (produced and uploaded by the
+# prepare_and_upload_openshift_bundles Evergreen task) and, for each RedHat fork,
```
viveksinghggits: the bundle is not already certified right? or is it?

scripts/release/publish-openshift-bundles-to-rh.sh
```diff
@@ -0,0 +1,152 @@
...
+# pushes a branch containing the bundle laid out under operators/<package>/<version>/.
+# It intentionally does NOT open the pull requests: it prints the "create PR" URLs and
+# the manual review checklist so a human can raise the DRAFT PRs and drive the review.
```
viveksinghggits, line 9: Why do we not open the PR? Are we not able to using the app account that we have or there are any other reasons?

viveksinghggits (commented):
Looks pretty good.

ci/internal/cli/github_apptoken.go
viveksinghggits, line 1: are we using these Go files by the script publish-openshift-bundles-to-rh.sh?

viveksinghggits: we don't seem to be using this, so where are they used exactly?

scripts/release/publish-openshift-bundles-to-rh.sh
josvazg: It was a guardrail to avoid unwanted PRs when testing or by mistake. I can reconsider though, as we could draft the PR instead, without actually putting it for review. That would solve the the issue that the evergreen job does not have a good way to give the PR opening links, other that checking the build logs.

scripts/release/publish-openshift-bundles-to-rh.sh
josvazg: What do you lean exactly?
Just to double check, we create both PRs to redHat from a single tarball, the certified one. That is what I did for 1.8.1 AFAICR and it worked.

ci/internal/cli/github_apptoken.go
josvazg: Good catch! The publishing code requires a GitHub App token, evergreen only allows to wire one app by project, and MCK already had one set up. This new app was created adhoc for releases. The go code here takes a github app credential and mints a session token to use a GH_TOKEN before the release process can proceed.

This was not surfaced on local tests as I was minting the token manually and I had not managed to test this in CI as there the tests were blocked by the preflight failing, as I was testing an old image that broke the preflight test.

All this to say you catched a bug here. I need to make sure CI wires the GH_TOKEN minting before calling `scripts/release/publish-openshift-bundles-to-rh.sh`

will fix

josvazg: This is now fixed

scripts/release/publish-openshift-bundles-to-rh.sh
josvazg: Actually I do not see a proper way to drop the releaser into the "open PR page", because I do not really want to create PRs in RH without human users being aware of them. That has to be a conscious decision and tests or mistakes could pollute RedHat without us noticing until it is too late.

scripts/release/publish-openshift-bundles-to-rh.sh
viveksinghggits: What I wanted to highlight is, is the bundle that is uploaded to AWS bucket by our evergreen pipeline already certified? If not then the comment is wrong and we should change that.

scripts/release/publish-openshift-bundles-to-rh.sh
viveksinghggits: > because I do not really want to create PRs in RH without human users being aware of them.

If someone is starting the release, they know that two PRs are going to be raised for OLM bundles as part of the release process.

> That has to be a conscious decision and tests or mistakes could pollute RedHat without us noticing until it is too late

Is this really possible? This workflow would ONLY be executed once the release is triggered, is there any other way to execute this workflow.

I think we should just create the PR in draft because that's what we are recommended to do in our release guide. The PRs would not even be in ready for review, once the PRs are raised in draft status we can ask for reviews and then we can make them ready for review manually.

> that the evergreen job does not have a good way to give the PR opening links, other that checking the build logs

I am not sure what exactly you meant here but if you mean you are not easily able to create the draft PR from evergreen, we do that already for the helm release step. To release helm [we create](https://github.com/mongodb/mongodb-kubernetes/blob/master/scripts/release/create_chart_release_pr.py#L31) a PR to the helm-chart repo automatically.

scripts/release/publish-openshift-bundles-to-rh.sh
m1kola: I believe Red Hat's certification is happening when we open a PR against their certified operator repo. I suspect "certified bundle tarball" here means something else.

I would change wording to avoid confusion like this ^

scripts/release/publish-openshift-bundles-to-rh.sh
m1kola: +1 for a draft PR

.evergreen-release.yml
```diff
@@ -197,6 +205,21 @@ buildvariants:
...
 
+  # Publishing to the RedHat forks is a separate, manually-triggered variant: the bundle is
+  # pulled from S3, so it does not need to share a host with the prepare task, and the PRs
+  # are draft + human-reviewed so we do not auto-chain it after every release build.
+  - name: publish_openshift_bundles_to_rh
+    display_name: publish_openshift_bundles_to_rh
+    tags: [ "release" ]
+    run_on:
+      - ubuntu2404-small
+    allowed_requesters: [ "patch", "github_tag" ]
```
m1kola: Do we need it on `patch`?

scripts/release/publish-openshift-bundles-to-rh.sh
josvazg: This all comes from the wording on the current release document:
> OLM bundles and RedHat Pull Requests
> Download created the **unified** OLM bundle from S3 and open a PR:
> Download the bundle (it’s named mck-operator-**certified**-x.y.z.tgz, untar it) from S3 link that is logged in the prepare_and_upload_openshift_bundles evg task from release.
> We have to create PRs for our OLM bundle in the repositories https://github.com/k8s-operatorhub/community-operators and https://github.com/redhat-openshift-ecosystem/certified-operators both of these repositories are forked already at the https://github.com/mongodb-forks/community-operators and https://github.com/mongodb-forks/certified-operators respectively.

AFAIU we create a single **unified** bundle tarball, but we call it **certified**. I have no problem changing the wording here, but the terms were confusing beforehand. I was also surprised the same tarball works in both repos as is, but yeah, that is the process we seem to follow and I completed for 1.8.1.

As for whether it was certified or not, I think the same as @m1kola here, the repo in RedHat's side is called certified-operators, I assume that they are considered certified once they are merged there, never before.

josvazg: To be clear, I will be changing wording here, but will not rename the tarball which predates this change.

scripts/release/publish-openshift-bundles-to-rh.sh
josvazg: > If someone is starting the release, they know that two PRs are going to be raised for OLM bundles as part of the release process.

The release process is done by evg in the background, and we will also be testing it. It is easy to trigger by mistake, by a test or forget about cleaning up.

> I think we should just create the PR in draft because that's what we are recommended to do in our release guide. The PRs would not even be in ready for review, once the PRs are raised in draft status we can ask for reviews and then we can make them ready for review manually.

I was almost convinced to go the PR draft way, until I checked with gemini and confirmed that even a PR draft kicks things on the RedHat side. We should not be doing that:

> 1. Tekton/Prow CI Pipelines Will Still Trigger
> Red Hat’s automated certification and community pipelines (operator-pipelines) listen to GitHub webhook events. When a PR is created on their repos—even as a draft—it triggers webhook events that kick off their Tekton/Prow validation pipelines, run bundle checks, and reserve test cluster resources.
> 
> 2. It Hits Red Hat’s "One Open PR" Limit
> Red Hat enforces a strict rule across certified-operators, community-operators-prod, and related repos: you may only have one open pull request at a time per package/project. A draft PR counts as an open PR. If your automation opens a draft PR and it sits waiting for human approval, your release pipeline is locked. If another automated build runs in the meantime, it will fail outright because an open PR already exists.
> 
> 3. Upstream Noise and Artifact Pollution
> Creating a draft PR publishes a public PR under the redhat-openshift-ecosystem organization. If your internal human review decides not to release or needs to regenerate the bundle from scratch, you have to close or abandon that draft PR, leaving unnecessary clutter and failed pipeline records on Red Hat’s upstream repositories.

I do not think we need to care for point 2, that is on us. But points 1 and 3 makes us a noisy contributor and we do not want that.

> I am not sure what exactly you meant here but if you mean you are not easily able to create the draft PR from evergreen, we do that already for the helm release step. To release helm [we create](https://github.com/mongodb/mongodb-kubernetes/blob/master/scripts/release/create_chart_release_pr.py#L31) a PR to the helm-chart repo automatically.

I mean evergreen is a job runner, it will create the PR, but it has no interactive way (AFAIK) to tell you "here are the links"... which is precisely the same limitation to the DX for just the links to create the PRs instead.

I think we should leverage the already interactive release workflow in GitHub to "show the create PR links" once evergreen is done creating the branches required. That makes that part of the process totally synchronous and manual for the releaser, which usually only means approve in a couple of clicks per PR.

.evergreen-release.yml
josvazg: maybe not, will double check

thanks for the heads up

josvazg: You were right, seems better to remove. We should not kick this off manually

scripts/release/publish-openshift-bundles-to-rh.sh
viveksinghggits: Even if we are worried about CI getting called for testing, I think we won't need more than 1 PR to test whether we are able to open the PR in that repo or not and I am pretty sure that 1 PR is not going to be a problem to anyone and wouldn't make un noisy contributor 😄. And after that one PR we know that the PR will only be raised for releases.


> it will create the PR, but it has no interactive way (AFAIK) to tell you "here are the links"... which is precisely the same limitation to the DX for just the links to create the PRs instead.

We can log the link in the evergreen patch and that link can be clicked on. An example from the PR that we create to release the chart
image
So I am not really sure what you meant by interactive way.

But it's ok, I think you are inclined towards just showing the link that can be used to open the PR, which is fine as well, it's just that if we are automating the things I didn't see a downside completely automating this one as well.

josvazg: Yes, I prefer to show a link to create the PR than create a draft PR on a background process. Because evergreen is a background process by design. You need to dig up the logs to find links, that is an awful DX.
Our release process is kicked by a GitHub workflow, which is a much nicer interactive process. You go to a web page that summarizes the state and you do not need to dig up logs, at least not when all is going well.

So my proposal for this PR is let evergreen do the automation part, I will fix the DX separately, as part of the new release process workflow. Such new release process will have an approval pause, just like the old, and before that pause the links to create the PRs are shown, if everything worked out.

viveksinghggits (approved):

m1kola (approved):

PR #1220, opened 2026-06-10
Add multi-cluster Search code snippets (docs/search scenarios 12 and 13, kind)

ashortt-mdb:
# Summary

Multi-cluster is the one Search GA topology with no CI-validated code snippets: the existing `docs/search` scenarios (01–11) are all single-Kubernetes-cluster, and the multi-cluster install content cannot be sourced from Self-Managed docs (it is MCK-specific). This PR adds the two missing scenarios, following the `docs/search` family conventions, as snippet input for the GA multi-cluster Search install guide (DOCSP-60800, MCK 1.9.0):

- **`docs/search/12-search-rs-multi-cluster`** — `MongoDBSearch` over a `MongoDBMultiCluster` replica set (external source, per-cluster managed Envoy LB, one mongot per member cluster)
- **`docs/search/13-search-sharded-multi-cluster`** — `MongoDBSearch` over a `topology: MultiCluster` sharded cluster (per-(cluster,shard) mongot, per-shard + mongos routing)

They run on the same 2-cluster kind e2e infrastructure as the multi-cluster search e2e tests, via a new **`private_kind_multi_cluster_code_snippets`** build variant (`pr_patch` — runs on every PR patch, no manual scheduling needed). The variant context sources `e2e_multi_cluster_2_clusters`, so the existing `setup_kubernetes_environment` provisions the interconnected kind clusters with an Istio mesh; the setup group additionally downloads the kubectl-mongodb plugin for `kubectl mongodb multicluster setup`.

## Design notes

- Multi-cluster does **not** auto-wire mongod→mongot, so the source CRs set the search `setParameter`s declaratively. Scenario 13 (sharded) uses `shardOverrides[].additionalMongodConfig` per shard + `mongos.additionalMongodConfig` for the routers; scenario 12 (RS) uses a single spec-level `mongotHost` routing all members to **cluster 0's** Envoy (`MongoDBMultiCluster` has no per-cluster `additionalMongodConfig`).
- CR shapes mirror the `e2e_search_q2_mc_rs_steady` and `e2e_search_q3_mc_sharded_external_mtls` e2e tests.
- Search Secrets (mongot cert, LB certs, sync-user password) are not auto-replicated by the operator; a dedicated snippet copies them to the member clusters.
[...]

PR #1368, opened 2026-07-13
Add multi-cluster Search code snippets (docs/search scenarios 12 and 13, kind)

anandsyncs:
# Summary

Adds public, CI-validated MongoDB Search snippets for two multi-cluster topologies:

- `docs/search/12-search-rs-multi-cluster`: MongoDB Search for an existing Ops Manager-managed replica set.
- `docs/search/13-search-sharded-multi-cluster`: MongoDB Search for an existing Ops Manager-managed sharded cluster.

The public flows assume the customer already has a running MongoDB deployment. CI-only helpers deploy source MongoDB resources for automated testing and are kept out of customer instructions.

## UX and consistency

- Separates customer-facing external deployment variables from internal CI variables.
- Lets customers configure existing deployment and shard names, with sensible defaults.
- Uses full customer-provided MongoDB connection strings for shared query scenarios.
- Reuses the normal replica-set and sharded query snippets because applications connect through MongoDB normally.
- Keeps public index inspection simple while CI readiness polling remains internal.
- Documents Ops Manager Search parameters for existing `mongod` and `mongos` processes.
- Replicates only the TLS Secrets required by each member cluster.

## Validation

Focused Evergreen patch [`6a54b213bae0e80007c40494`](https://spruce.corp.mongodb.com/version/6a54b213bae0e80007c40494) passed all selected Search snippet tasks, including both multi-cluster scenarios.

Supersedes #1220.

skip-ci


docs/search/13-search-sharded-multi-cluster/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+   `MDB_EXTERNAL_CLUSTER_NAME` defaults to `mdb-mc-sh`. Change it to the
+   existing deployment name. Also set the existing shard names and every
+   `MDB_EXTERNAL_*_HOST_*` endpoint to match the source deployment.
+
+2. In Ops Manager, open **Deployment > Processes** and click **Modify** for
+   each process listed below.
+3. Under **Advanced Configuration Options**, click **Add Option**, select
+   `setParameter`, and add `mongotHost` and
+   `searchIndexManagementHostAndPort` with these values:
```
anandsyncs: **Missing required `setParameter`s.** This section documents only `mongotHost` and `searchIndexManagementHostAndPort`, but the CI path (`13_0310_internal_create_mongodb_mc_sharded.sh`) sets six parameters on every mongos and shard mongod: those two **plus** `searchTLSMode: requireTLS`, `useGrpcForSearch: true`, `skipAuthenticationToMongot: false`, `skipAuthenticationToSearchIndexManagementServer: false`. The operator always writes all six (`buildSearchSetParameters`; gRPC is required for the mongos→mongot path), and scenario 07's README documents the TLS/gRPC ones. A customer following this page verbatim gets a failing TLS/gRPC handshake, and CI can't catch it because CI never exercises the README path. Extend the table to the full parameter set from `13_0310`.

docs/search/12-search-rs-multi-cluster/README.md
```diff
@@ -0,0 +1,91 @@
...
+   ```
+
+   `MDB_EXTERNAL_CLUSTER_NAME` defaults to `mdb-mc-rs`. Change it to the
+   existing replica-set name, and set each `MDB_EXTERNAL_HOST_*` value to an
+   endpoint that the Kubernetes clusters can reach.
+
+2. In Ops Manager, open **Deployment > Processes** and click **Modify** for
+   each `mongod` in the replica set.
+3. Under **Advanced Configuration Options**, click **Add Option**, select
+   `setParameter`, and add both of these parameters:
```
anandsyncs: Same gap as 13's README: the CI CR (`12_0310`) also sets `searchTLSMode: requireTLS`, `useGrpcForSearch: true`, `skipAuthenticationToMongot: false`, and `skipAuthenticationToSearchIndexManagementServer: false` — all required for the mongod→Envoy/mongot path, so the customer instructions need the full set, not just these two.

docs/search/13-search-sharded-multi-cluster/README.md
```diff
@@ -0,0 +1,100 @@
...
+   Enter the expanded values from `env_variables.sh`, not the variable names.
+4. Click **Save**, review the changes, and deploy the updated configuration.
+   Do not add these Search parameters to config server processes.
+
+Ops Manager applies `setParameter` startup options to the processes it manages.
+See [Advanced Options for MongoDB Deployments](https://www.mongodb.com/docs/ops-manager/current/reference/deployment-advanced-options/).
+
+Processes use cluster 0's proxies because the source deployment has one shared
+value per process role. Processes in cluster 1 therefore send Search traffic
+across the service mesh to cluster 0.
```
anandsyncs: **Worth an explicit call: everything is pinned to cluster 0's proxies.** As written, cluster 1's entire search stack (mongot StatefulSets, Envoy, certs, replicated secrets) — which the customer is instructed to build — never receives query traffic, and cluster 0's Envoy becomes a cross-mesh SPOF for `$search` in both clusters. The rationale sentence is also not quite right: "one shared value per process role" is true of the CR's `additionalMongodConfig`, but these instructions use OM's per-process **Modify** flow, which supports per-process values — so per-cluster pinning (`...-search-1-...` hosts for cluster-1 processes) is available. If single-stable-target is the deliberate choice (it does simplify cluster add/remove), keep it but fix this sentence and state that cluster 1's search stack is effectively a warm standby; otherwise add `MDB_PROXY_HOST_1*` vars and instruct local-cluster pinning.

docs/search/12-search-rs-multi-cluster/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+   Enter the expanded value from `env_variables.sh`, not the variable name.
+4. Click **Save**, review the changes, and deploy the updated configuration.
+
+Ops Manager applies `setParameter` startup options to the processes it manages.
+See [Advanced Options for MongoDB Deployments](https://www.mongodb.com/docs/ops-manager/current/reference/deployment-advanced-options/).
+
+All replica-set members use cluster 0's proxy because the source deployment
+has one shared value for each Search parameter. Members in cluster 1 therefore
+send Search traffic across the service mesh to cluster 0.
```
anandsyncs: Same consideration as 13's README: OM's per-process Modify flow would allow pinning cluster-1 members to a cluster-1 proxy, so "has one shared value" isn't the real constraint. If single-stable-target is deliberate, adjust the rationale and note cluster 1's stack is a warm standby.

docs/search/03-search-query-usage/README.md
```diff
@@ -0,0 +1,60 @@
...
+./test.sh
+```
+
+## Step-by-step snippets
+
+1. `03_0410_run_mongodb_tools_pod.sh` — deploy tools pod.
+2. `03_0420_import_movies_mflix_database.sh` — import sample data.
+3. `03_0430_create_search_index.sh` — create text search index.
+4. `03_0435_create_vector_search_index.sh` — create vector index.
+5. `03_0440_wait_for_search_index_ready.sh` — wait for index readiness.
```
anandsyncs: **This step documents the `sleep 60` snippet while CI now uses a real poll.** `03_0440_wait_for_search_index_ready.sh` is literally `sleep 60` under a comment claiming index status can't be checked — which this PR falsifies (`03_0439_internal_wait_for_search_indexes.sh` and the public `08_0440_list_search_indexes.sh` both read `getSearchIndexes()[].status`). `test.sh` now swaps in the internal poll, so this documented step is no longer CI-exercised and will rot. Suggest mirroring the 08 treatment: drop the sleep snippet and point step 5 at a public status check.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0326_internal_verify_envoy_deployment.sh
```diff
@@ -0,0 +1,37 @@
...
+  envoy_deployment="${MDB_SEARCH_RESOURCE_NAME}-search-lb-${ci}"
+  proxy_service="${MDB_SEARCH_RESOURCE_NAME}-search-${ci}-proxy-svc"
+
+  echo "== cluster ${ci} (${ctx}) =="
+
+  echo "Checking mongot StatefulSet..."
+  if kubectl get statefulset "${mongot_sts}" -n "${MDB_NS}" --context "${ctx}" &>/dev/null; then
+    echo "  [ok] StatefulSet '${mongot_sts}' exists"
+  else
+    echo "  [FAIL] StatefulSet '${mongot_sts}' NOT FOUND" >&2
```
anandsyncs, line 16: The `[FAIL] ... NOT FOUND` branches only echo to stderr — the script still ends with `[ok] ... deployed and healthy` and exits 0, so a missing StatefulSet/Deployment/Service passes CI silently (only a `rollout status` timeout on an *existing* Deployment can fail). `12_0040_validate_env.sh` already has the right idiom: accumulate `has_error`, end with `false`. Same in `13_0326`.

docs/search/13-search-sharded-multi-cluster/code_snippets/13_0326_internal_verify_envoy_deployment.sh
```diff
@@ -0,0 +1,51 @@
...
+      echo "    Waiting for Envoy pods..."
+      kubectl rollout status deployment/"${envoy_deployment}" -n "${MDB_NS}" --context "${ctx}" --timeout=120s
+    fi
+  else
+    echo "  [FAIL] Deployment '${envoy_deployment}' NOT FOUND" >&2
+  fi
+
+  echo "Checking per-shard proxy Services..."
+  for shard_name in "${MDB_SHARD_0_NAME}" "${MDB_SHARD_1_NAME}" "${MDB_SHARD_2_NAME}"; do
+    proxy_service="${MDB_SEARCH_RESOURCE_NAME}-search-${ci}-${shard_name}-proxy-svc"
```
anandsyncs, line 45: This loop verifies only the three per-shard proxy Services, never the router proxy Service (`${MDB_SEARCH_RESOURCE_NAME}-search-${ci}-proxy-svc`) that `13_0320`'s `routerHostname` and every mongos's `mongotHost` depend on — scenario 12 checks its equivalent. Separately, the Envoy readiness check here kept the jsonpath-compare + fallback pattern while `12_0326` was simplified to a bare `rollout status --timeout=120s`; the simpler form is sufficient — align the two.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0325_wait_for_search_resource.sh
```diff
@@ -0,0 +1,18 @@
...
+  --timeout=900s
+
+echo "[ok] MongoDBSearch '${MDB_SEARCH_RESOURCE_NAME}' is Running"
+
+echo ""
+echo "MongoDBSearch pods per member cluster:"
+for ctx in "${K8S_CTX_0}" "${K8S_CTX_1}"; do
+  echo "== ${ctx} =="
+  kubectl get pods -n "${MDB_NS}" --context "${ctx}" \
+    -l "app.kubernetes.io/component=mongot"
```
anandsyncs: This selector matches nothing: mongot pods only carry `app: <headless-svc-name>` (plus `shard:` for sharded sources) — nothing in the operator sets `app.kubernetes.io/component` on pods. So this customer-facing "show search pods" section always prints `No resources found`. Inherited from 09/10/11, but this PR copies it onto two new public pages. Use `-l "app=${MDB_SEARCH_RESOURCE_NAME}-search-${ci}-svc"` here (tracking a `ci` counter like `12_0326` does), and e.g. `-l shard` in `13_0325`.

scripts/code_snippets/tests/test_kind_search_rs_multi_cluster_snippets.sh
```diff
@@ -0,0 +1,63 @@
...
+# shellcheck disable=SC1090
+test -f "${test_dir}/env_variables_${CODE_SNIPPETS_FLAVOR}.sh" && source "${test_dir}/env_variables_${CODE_SNIPPETS_FLAVOR}.sh"
+
+# The query module is single-context: point it at the central cluster, where
+# the tools pod and the CA ConfigMap live.
+export K8S_CTX="${K8S_CTX_0}"
+
+${test_dir}/test.sh
+
+# Phase 3: per-cluster query verification — prove each member cluster serves
```
anandsyncs: The comment overstates what Phase 3 proves: (a) every member's `mongotHost` points at cluster 0's proxy, so both iterations exercise cluster-0's Envoy/mongot — cluster 1's search stack never serves query traffic in this topology; what's actually proven is that a mongod on each cluster can reach proxy-0 across the mesh. (b) `03_0450`/`03_0455` don't assert result count > 0, so an empty result set still passes. Suggest rewording, and adding a small `itcount() > 0` assertion instead of reusing the print-only snippets (same gap that #1299 closed for the pytest MC e2es).

scripts/code_snippets/tests/test_kind_search_sharded_multi_cluster_snippets.sh
```diff
@@ -0,0 +1,64 @@
...
+test -f "${test_dir}/env_variables_${CODE_SNIPPETS_FLAVOR}.sh" && source "${test_dir}/env_variables_${CODE_SNIPPETS_FLAVOR}.sh"
+
+# The query module is single-context: point it at the central cluster, where
+# the tools pod and the CA ConfigMap live. MDB_ADMIN_CONNECTION_STRING and
+# MDB_USER_CONNECTION_STRING are exported by scenario 13's env_variables.sh.
+export K8S_CTX="${K8S_CTX_0}"
+
+${test_dir}/test.sh
+
+# Phase 3: per-cluster query verification — prove each member cluster's mongos
```
anandsyncs: Same as the RS runner: all mongos processes target cluster-0 proxies, and `08_0450`/`08_0455` can't fail on empty results — note the known routed_from_another_shard failure mode manifests as *empty* results, not an error, so a count>0 assertion is what would actually catch it.

docs/search/03-search-query-usage/code_snippets/03_0439_internal_wait_for_search_indexes.sh
```diff
@@ -0,0 +1,19 @@
+indexes=("movies default" "embedded_movies vector_index")
+[[ -n "${EMBEDDING_MODEL:-}" ]] && indexes+=("movies vector_auto_embed_index")
+
+for collection_and_index in "${indexes[@]}"; do
+  read -r collection index_name <<< "${collection_and_index}"
+
+  for _ in {1..30}; do
+    status="$(kubectl exec --context "${K8S_CTX}" -n "${MDB_NS}" mongodb-tools-pod \
```
anandsyncs, line 8: With the inherited `set -eou pipefail`, a single transient `kubectl exec` failure aborts the whole wait instead of consuming one of the 30 retries — an `|| true` inside the substitution lets the retry loop actually retry. Same in `08_0439`.

.evergreen-snippets.yml
```diff
@@ -46,6 +76,10 @@ variables:
         variant: init_test_run
       - name: build_database_image_ubi
         variant: init_test_run
+      - name: build_readiness_probe_image
```
anandsyncs, line 81: These two images are only pulled by MongoDBCommunity workloads — scenarios 12/13 (MongoDBMultiCluster + MongoDBSearch) don't need them — and adding them to this *shared* `base_om8_dependency` copy silently re-gates the sibling `private_kind_code_snippets` variant while diverging further from the `.evergreen.yml` original, whose comment says the two copies must stay mirrored. Either sync both anchors or add a comment justifying the divergence (if this fixes a latent image-pull race for the community scenario, say so).

anandsyncs (commented):
Full review pass over the PR (scenarios 12/13, the 03/08 changes, and the Evergreen wiring), focused on correctness, simplicity/efficiency, and customer UX of the non-internal snippets. **No blockers.** Two customer-facing doc gaps worth fixing before this becomes public docs source (missing required `setParameter`s in the 12/13 READMEs; 03's README still documenting the sleep-60 wait), one topology-guidance decision to make explicit (everything pinned to cluster-0 proxies), and a set of minor correctness/polish items — all as inline comments.

**What checks out** (verified against the operator source): every derived name — per-(cluster,shard) mongot cert secrets, LB server/client certs, proxy Services, StatefulSets, Envoy Deployments — matches the API naming functions exactly; LB cert SANs cover the SNI hostnames; `*_0317` replicates exactly the four secrets cluster 1 needs and nothing more; the env three-way split is the cleanest of the search scenarios so far (OM creds/passwords finally out of the customer file); `0040_validate_env` is a genuinely good customer-UX addition; the 08 wait refactor fixes a wait that previously could never fail (printed ERROR, exited 0); all five existing consumers of 03/08 still satisfy the new env contracts; and the Evergreen task-group wiring (inotify rationale, context/variant naming, teardown parity) is correct end-to-end.

One nit with no good inline anchor: the new snippets `03_0439`, `08_0439`, and `08_0440_list_search_indexes.sh` are mode 644 while most sibling snippets are 755 — customers running `./code_snippets/<file>.sh` per the docs style get "permission denied".

.evergreen-snippets.yml
```diff
@@ -208,3 +261,12 @@ buildvariants:
     <<: *base_om8_dependency
     tasks:
       - name: kind_code_snippets_task_group
+
+  - name: private_kind_multi_cluster_code_snippets
```
anandsyncs, line 271: This variant gets the kubectl-mongodb plugin artifact via `download_multi_cluster_binary`, but build ordering is only guaranteed transitively (`base_om8_dependency` → `build_test_image` → nested `build_kubectl_mongodb_plugin` dep). `private_gke_code_snippets` above declares the dep explicitly — worth doing the same so a future change to `build_test_image`'s deps can't reintroduce a 404 race.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0100_install_operator.sh
```diff
@@ -0,0 +1,33 @@
+echo "Configuring multi-cluster service accounts and roles..."
+
+KUBECONFIG="${MDB_PLUGIN_KUBECONFIG:-${KUBECONFIG:-${HOME}/.kube/config}}" \
```
anandsyncs: `MDB_PLUGIN_KUBECONFIG` exists only in `env_variables_e2e_private.sh` (kind node-IP kubeconfig workaround), so this published snippet shows customers an unexplained CI variable. Suggest a bare invocation here (like `ra-02_0200` on the multi-cluster architecture pages) and exporting `KUBECONFIG` from the private env file instead. Same in `13_0100`.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0045_create_namespaces.sh
```diff
@@ -0,0 +1,7 @@
+for ctx in "${K8S_CTX_0}" "${K8S_CTX_1}"; do
+  kubectl create namespace "${MDB_NS}" --context "${ctx}" --dry-run=client -o yaml | \
+    kubectl apply --context "${ctx}" -f -
+  kubectl label namespace "${MDB_NS}" istio-injection=enabled --overwrite --context "${ctx}"
```
anandsyncs, line 4: Public snippet hardcodes Istio while the README prerequisite is a generic "service mesh". Inert on other meshes, but a one-line README note ("snippets assume Istio; adjust the injection label for your mesh") would avoid confusion. Same in `13_0045`.

docs/search/12-search-rs-multi-cluster/env_variables.sh
```diff
@@ -0,0 +1,41 @@
...
+# Existing TLS resources. Keep the issuer and CA ConfigMap aligned with the
+# source deployment's trusted CA.
+export CERT_MANAGER_NAMESPACE="cert-manager"
+export MDB_TLS_CERT_SECRET_PREFIX="certs"
+export MDB_TLS_CA_CONFIGMAP="${MDB_EXTERNAL_CLUSTER_NAME}-ca"
+export MDB_TLS_CA_ISSUER="my-ca-issuer"
+
+# MongoDB Search configuration.
+export MDB_MONGOT_REPLICAS_PER_CLUSTER=1
+export ENVOY_PROXY_PORT="27028"
```
anandsyncs, line 31: `ENVOY_PROXY_PORT` reads as configurable, but `ManagedLBConfig` has no port field — 27028 is the fixed operator default and the MongoDBSearch CR never sets a port, so a customer who edits this silently points `mongotHost` at a dead port. Annotate it "(operator default — do not change)" like scenario 10, or inline it into the host vars. Same in 13's env file.

docs/search/13-search-sharded-multi-cluster/env_variables.sh
```diff
@@ -0,0 +1,56 @@
+# Kubernetes contexts for the two member clusters. Cluster 0 is also the
+# central cluster where the operator and MongoDBSearch resource run.
+export K8S_CTX_0="<central/member-0 cluster context>"
+export K8S_CTX_1="<member-1 cluster context>"
+export MDB_NS="mongodb"
+
+# Existing external sharded cluster and the MongoDBSearch resource to create.
+export MDB_EXTERNAL_CLUSTER_NAME="mdb-mc-sh"
+export MDB_SEARCH_RESOURCE_NAME="${MDB_EXTERNAL_CLUSTER_NAME}-search"
```
anandsyncs: This default yields doubled names in everything the customer sees (`...-search-search-0-<shard>-proxy-svc`). Scenario 07 used a plain name (`ext-sh`); a suffix-free default reads much better.

docs/search/12-search-rs-multi-cluster/test.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
+# Test runner script for MongoDB Search with a multi-cluster (MongoDBMultiCluster)
+# replica set source + per-cluster managed Envoy LB
+#
+# This script executes all code snippets in order to test the full deployment flow.
+# It can be run manually or as part of automated E2E testing.
+#
+# Usage:
+#   ./test.sh                    # Run with env_variables.sh
```
anandsyncs: Running `test.sh` manually also requires `env_variables_internal.sh` (OM plumbing, member counts) — worth mentioning both files in the usage header. Same in 13.

docs/search/12-search-rs-multi-cluster/code_snippets/12_9010_internal_delete_namespace.sh
```diff
@@ -0,0 +1,14 @@
+echo "WARNING: This will delete namespace '${MDB_NS}' and all its resources in every member cluster."
```
anandsyncs, line 1: Unreachable: not run by `test.sh`, not referenced by the README or the runner. Scenario 10's `test.sh` ends with a "To clean up, run: ..." hint — add the same to both new scenarios, or drop the file.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0315a_internal_create_mongodb_users.sh
```diff
@@ -0,0 +1,108 @@
+echo "Creating password secrets for MongoDB users..."
```
anandsyncs, line 1: Second file with the `12_0316` stem (with `0316a`/`0316b` following), and `test.sh` runs this one *before* the lexicographically-first public `12_0316_create_search_sync_secret.sh` — the numbering no longer encodes execution order. No functional impact (run order is explicit), but renumbering this to e.g. `12_0315a_internal_...` keeps the scheme meaningful. Same in 13.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0316b_create_lb_tls_certificates.sh
```diff
@@ -0,0 +1,68 @@
+echo "Creating TLS certificates for managed load balancer (Envoy)..."
+
+for i in 0 1; do
```
anandsyncs: Only divergence from the otherwise byte-identical `13_0316b` is the loop variable (`i` here, `ci` there) — align on `ci` to match `0316a`/`0326`.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0040_validate_env.sh
```diff
@@ -0,0 +1,57 @@
...
+has_error=false
+if (( ${#missing_vars[@]} )); then
+  echo "ERROR: Missing required environment variables:" >&2
+  for m in "${missing_vars[@]}"; do echo "  - ${m}" >&2; done
+  echo "Please edit env_variables.sh and set these values before proceeding." >&2
+  has_error=true
+fi
+
+missing_contexts=()
+for ctx in "${K8S_CTX_0:-}" "${K8S_CTX_1:-}"; do
```
anandsyncs, line 33: When `K8S_CTX_*` still holds a `<placeholder>`, this loop also runs and reports the literal placeholder as a missing context, duplicating the missing-variable error above — skip values already flagged as placeholders. (Also worth considering: validating that the `MDB_TLS_CA_CONFIGMAP` ConfigMap exists on both contexts — it's a hard README prerequisite whose non-placeholder default the placeholder check can't catch.) Same in `13_0040`.

docs/search/03-search-query-usage/README.md
```diff
@@ -0,0 +1,60 @@
+# Replica Set MongoDB Search — Query Usage
+
+Use this module to import sample data, create Search/Vector Search indexes, and run query snippets against a replica-set MongoDB Search deployment.
+
+Run this query module after one of these infrastructure modules:
```
anandsyncs, line 5: This list omits `02-search-enterprise-deploy` and `04-search-external-mongod`, both of which chain into 03 in CI (`test_kind_search_enterprise_snippets.sh`, `test_kind_search_external_mongod_snippets.sh`).

docs/search/13-search-sharded-multi-cluster/README.md
anandsyncs: Added all four missing TLS/gRPC/authentication parameters to the sharded customer instructions alongside the two routing parameters.

docs/search/12-search-rs-multi-cluster/README.md
anandsyncs: Added the complete six-parameter set to the replica-set customer instructions.

docs/search/13-search-sharded-multi-cluster/README.md
anandsyncs: Kept the deliberately simple stable cluster-0 target, but corrected the rationale and documented cluster 1 as warm standby plus the per-process local-proxy alternative.

docs/search/12-search-rs-multi-cluster/README.md
anandsyncs: Updated the replica-set guidance to describe the deliberate stable cluster-0 target accurately, including warm-standby behavior and the local-proxy alternative.

docs/search/03-search-query-usage/README.md
anandsyncs: Replaced the fixed sleep with a public index-status command and added it to the exercised test flow. The established filename remains intact because published docs reference it directly.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0326_internal_verify_envoy_deployment.sh
anandsyncs: Both verification scripts now accumulate missing/unready resources and return failure instead of printing a false success.

docs/search/13-search-sharded-multi-cluster/code_snippets/13_0326_internal_verify_envoy_deployment.sh
anandsyncs: Added router proxy Service verification and aligned Envoy readiness with the rollout-status check.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0325_wait_for_search_resource.sh
anandsyncs: Replaced the ineffective selector with the actual per-cluster app labels, including per-shard labels for scenario 13.

scripts/code_snippets/tests/test_kind_search_rs_multi_cluster_snippets.sh
anandsyncs: Reworded Phase 3 to state that both MongoDB endpoints use cluster 0’s stable proxy, and added internal-only non-empty text/vector result assertions.

scripts/code_snippets/tests/test_kind_search_sharded_multi_cluster_snippets.sh
anandsyncs: Added internal non-empty text and vector result assertions for both mongos endpoints and corrected the Phase 3 claim.

docs/search/03-search-query-usage/code_snippets/03_0439_internal_wait_for_search_indexes.sh
anandsyncs: Both internal readiness loops now tolerate transient kubectl failures and continue through their retry budget.

.evergreen-snippets.yml
anandsyncs: Documented why the shared snippet dependency needs these images: the single-cluster task group includes Community consumers that pull them directly.

.evergreen-snippets.yml
anandsyncs: Added a direct plugin dependency while preserving the complete inherited dependency set; the evaluated Evergreen variant confirms all dependencies remain present.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0100_install_operator.sh
anandsyncs: Removed the CI-only variable from both public install snippets. The private environment now exports the generated kubeconfig through standard KUBECONFIG.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0045_create_namespaces.sh
anandsyncs: Added an explicit Istio-label note to both READMEs with guidance to adjust it for other meshes.

docs/search/12-search-rs-multi-cluster/env_variables.sh
anandsyncs: Marked port 27028 as the fixed operator default that customers should not change in both environment files.

docs/search/13-search-sharded-multi-cluster/env_variables.sh
anandsyncs: Changed the public Search resource defaults to suffix-free ext-sh and ext-rs names, avoiding doubled search segments.

docs/search/12-search-rs-multi-cluster/test.sh
anandsyncs: Updated both test headers to show that public and internal environment files must be sourced for manual CI-style execution.

docs/search/12-search-rs-multi-cluster/code_snippets/12_9010_internal_delete_namespace.sh
anandsyncs: Both test runners now print the corresponding cleanup command, making the internal cleanup snippets reachable.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0315a_internal_create_mongodb_users.sh
anandsyncs: Renamed both internal user-creation snippets to the 0315a step and updated their runners; public filenames were unchanged.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0316b_create_lb_tls_certificates.sh
anandsyncs: Aligned the replica-set certificate loops on the ci variable used by the sharded and neighboring snippets.

docs/search/12-search-rs-multi-cluster/code_snippets/12_0040_validate_env.sh
anandsyncs: Placeholder contexts are now skipped by the context lookup, avoiding duplicate errors. I left CA ConfigMap existence out because CI validates before its internal setup creates that ConfigMap.

docs/search/03-search-query-usage/README.md
anandsyncs: Added both the enterprise and external-mongod infrastructure modules to the scenario 03 consumer list.

anandsyncs:
Superseded by the three-PR stack: #1378 (KUBE-169 shared query snippets), #1379 (KUBE-170 replica-set multi-cluster Search), and #1380 (KUBE-171 sharded multi-cluster Search). Full-stack Evergreen validation: https://spruce.corp.mongodb.com/version/6a579a5d8cfab20007061024

PR #1371, opened 2026-07-13
CLOUDP-420747: Prepare CI to work on backport branches as well as master

josvazg:
# Summary

Ensure the CI keeps working the same on `master` than on future back-porting branches v1, v2, v3, pattern `v*`. For testing `v*-test` branches are also supported, but they publish to test registries.

## Proof of Work

TBD

## Checklist

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -4,10 +4,12 @@ on:
   push:
     branches:
       - master
+      - v*
   pull_request:
     branches:
       - master
       - release-*
+      - v*
     paths:
       - '**/*.go'
       - 'go.mod'
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -4,10 +4,11 @@ on:
   push:
     branches:
       - master
+      - release-v*
   pull_request:
     branches:
       - master
-      - release-*
+      - release-v*
     paths:
       - '**/*.go'
       - 'go.mod'
--- a/.github/workflows/preview_release_notes.yml
+++ b/.github/workflows/preview_release_notes.yml
@@ -4,10 +4,12 @@ on:
     branches:
       - master
       - release-*
+      - v*
   pull_request:
     branches:
       - master
       - release-*
+      - v*
 
 jobs:
   preview_release_notes:
@@ -39,8 +41,10 @@ jobs:
           RELEASE_INITIAL_VERSION: ${{ env.RELEASE_INITIAL_VERSION }}
       - name: Add disclaimer to release notes preview for Pull Requests
         if: github.event_name == 'pull_request'
+        env:
+          TARGET_BASE_REF: ${{ github.base_ref }}
         run: |
-          echo -e "_:warning: (this preview might not be accurate if the PR is not rebased on current master branch)_\n" > release_notes_tmp.md
+          echo -e "_:warning: (this preview might not be accurate if the PR is not rebased on current $TARGET_BASE_REF branch)_\n" > release_notes_tmp.md
           cat release_notes_preview.md >> release_notes_tmp.md
           mv -f release_notes_tmp.md release_notes_preview.md
       - name: Summarize results
--- a/.github/workflows/preview_release_notes.yml
+++ b/.github/workflows/preview_release_notes.yml
@@ -3,11 +3,11 @@ on:
   push:
     branches:
       - master
-      - release-*
+      - release-v*
   pull_request:
     branches:
       - master
-      - release-*
+      - release-v*
 
 jobs:
   preview_release_notes:
@@ -39,8 +39,10 @@ jobs:
           RELEASE_INITIAL_VERSION: ${{ env.RELEASE_INITIAL_VERSION }}
       - name: Add disclaimer to release notes preview for Pull Requests
         if: github.event_name == 'pull_request'
+        env:
+          TARGET_BASE_REF: ${{ github.base_ref }}
         run: |
-          echo -e "_:warning: (this preview might not be accurate if the PR is not rebased on current master branch)_\n" > release_notes_tmp.md
+          echo -e "_:warning: (this preview might not be accurate if the PR is not rebased on current $TARGET_BASE_REF branch)_\n" > release_notes_tmp.md
           cat release_notes_preview.md >> release_notes_tmp.md
           mv -f release_notes_tmp.md release_notes_preview.md
       - name: Summarize results
--- a/scripts/check-ocp-version-drift.sh
+++ b/scripts/check-ocp-version-drift.sh
@@ -17,6 +17,9 @@
 #   REMOTE                 Git remote to push to and create the PR against.
 #                          Defaults to "origin".  Set to e.g. "mongodb" when
 #                          testing against an upstream remote from a fork.
+#   TARGET_BRANCH          Branch to base the update PR on and merge into.
+#                          Defaults to "master". Set to a backporting branch
+#                          (e.g. "v1") when running this check on that branch.
 #
 # Prerequisites (normal mode): oc (already logged in), jq, gh (authenticated via GH_TOKEN)
 # Prerequisites (dry-run + --actual-version): jq
@@ -27,6 +30,7 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROJECT_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
 CONFIG_FILE="${CONFIG_FILE:-${PROJECT_ROOT}/kubernetes-versions.json}"
 REMOTE="${REMOTE:-origin}"
+TARGET_BRANCH="${TARGET_BRANCH:-master}"
 
 # --- argument parsing --------------------------------------------------------
 
@@ -125,8 +129,8 @@ create_update_pr() {
     local repo
     repo=$(remote_repo)
 
-    git fetch "${REMOTE}" master --quiet || die "git fetch ${REMOTE} master failed"
-    git checkout -b "${branch}" "${REMOTE}/master"
+    git fetch "${REMOTE}" "${TARGET_BRANCH}" --quiet || die "git fetch ${REMOTE} ${TARGET_BRANCH} failed"
+    git checkout -b "${branch}" "${REMOTE}/${TARGET_BRANCH}"
 
     local tmp
     tmp=$(mktemp)
@@ -157,7 +161,7 @@ recorded a different minor version.  This PR updates the file to match reality.
 - [ ] Verify the OpenShift CI tasks pass with the updated version.
 EOF
         )" \
-        --base master \
+        --base "${TARGET_BRANCH}" \
         --head "${branch}"
 }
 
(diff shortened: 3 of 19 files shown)


.github/workflows/preview_release_notes.yml
```diff
@@ -4,10 +4,12 @@ on:
     branches:
       - master
       - release-*
+      - v*
```
m1kola: I'm confused - why we need `v*` branches here and in other places? I thought `release-*` was meant for backporting. E.g. `release-1.x`?

scripts/evergreen/should_release_agents_on_ecr.sh
```diff
@@ -8,20 +8,20 @@ no_local_changes() {
...
   echo "skipping since release.json has not changed"
   exit 1
 else
   echo "release.json has changed, triggering ecr agent release"
   echo "git diff"
   git diff -- release.json
-  echo "git diff master"
-  git diff origin/master -- release.json
+  echo "git diff ${branch_name:-master}"
+  git diff "origin/${branch_name:-master}" -- release.json
```
m1kola: Seems like too many places with defaulting. Maybe worth a a variable? Or even perhaps make it always required?

scripts/check-ocp-version-drift.sh
```diff
@@ -27,6 +30,7 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROJECT_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
 CONFIG_FILE="${CONFIG_FILE:-${PROJECT_ROOT}/kubernetes-versions.json}"
 REMOTE="${REMOTE:-origin}"
+TARGET_BRANCH="${TARGET_BRANCH:-master}"
```
m1kola: Is it going to be present in evg? I see we only pass `GH_TOKEN` in `check_ocp_version_drift` function.

Is it supposed to be `branch_name` evg expansion?

scripts/create-go-bump-pr.sh
```diff
@@ -21,6 +21,7 @@ fi
 version="${1#go}"
 branch="auto/bump-go-${version}"
 title="Bump Go version to ${version}"
+base_branch="${branch_name:-$(git rev-parse --abbrev-ref HEAD)}"
```
m1kola: Is it correct? I believe `branch_name` is evg expansion and we run this script in GH actions.

scripts/test/bash/backport_branch/test_backport_branch_ci.bats
```diff
@@ -0,0 +1,91 @@
...
+    git commit -q -m "add kubectl-mongodb path only on v1"
+    git push -q origin v1
+
+    tree_listing=$(env branch_name=v1 git ls-tree -r "origin/v1" --name-only)
+
+    [[ "${tree_listing}" == *"pkg/kubectl-mongodb/only_on_v1.go"* ]]
+
+    master_tree_listing=$(git ls-tree -r "origin/master" --name-only)
+    [[ "${master_tree_listing}" != *"pkg/kubectl-mongodb/only_on_v1.go"* ]]
+}
```
m1kola: Are we supposed to run `RBAC_SCRIPT` here?

scripts/test/bash/backport_branch/test_backport_branch_ci.bats
```diff
@@ -0,0 +1,91 @@
...
+    cd "${CLONE_DIR}"
+    git checkout -q v1
+
+    # release.json on v1 differs from master's, but nothing has changed
+    # *within* v1 itself — comparing against origin/master (the old bug)
+    # would incorrectly report a change here.
+    run env branch_name=v1 "${SCRIPT}"
+
+    [ "$status" -eq 1 ]
+}
```
m1kola: What is the difference between the two?

.github/workflows/preview_release_notes.yml
m1kola: Also - I can think of branches that start with `v`, but not version branches. E.g. `vuln-fix` or something like that.

scripts/dev/contexts/evg-private-context
```diff
@@ -137,6 +146,10 @@ case ${BUILD_SCENARIO} in
 
   "${BUILD_SCENARIO_STAGING}")
     COMMIT_SHA_SHORT=$(git rev-parse --short=8 HEAD)
+    if [ "${IS_TEST_BRANCH}" == "true" ]; then
+      REGISTRY="${REGISTRY:-${TEST_BRANCH_REGISTRY:-268558157000.dkr.ecr.us-east-1.amazonaws.com/test}}"
+      export FLAGS="${FLAGS:-} --registry ${REGISTRY}"
+    fi
```
m1kola, line 162: Q: Do we need to handle `IS_TEST_BRANCH` somehow in release scenario? E.g. exit/error if `IS_TEST_BRANCH` is set to true in these for some reason?

I'm not too worried since it is mostly automated and a chance of manual error is minimal. Just a question.

.github/workflows/preview_release_notes.yml
m1kola: Also - we have `v*`, `^v[0-9]+-test$` and `^v(\d+)(-test)?$` in different places. Each serves their purpose, but I imagine it will be a nightmare if we need to change the pipeline ever again. How do you find all the instances/cross reference them?

josvazg: I see your concern. I will revisit the backporting branches names, but vX was the most obvious branch name to me.

Maybe release-v* is ok, a bit long but descriptive enough.

Will revisit this.

scripts/check-ocp-version-drift.sh
josvazg: The way evergreen works for this is a bit of a PITA, AFAIK we will need a project in evg PER brackporting branch.

I already have one for the test backport branch and will need to rename the branch. I guess this value is wired to the project.
Sample:
https://spruce.corp.mongodb.com/project/mongodb-kubernetes-v1-test/settings/general

scripts/evergreen/should_release_agents_on_ecr.sh
josvazg: Let me see if we can simplify

.github/workflows/preview_release_notes.yml
josvazg: moving to `release-v*-[-test]`

On the expression sprawl: The 3 pattern forms, affect 12 files, but they're all greppable with one string `release-v`. The duplication is due to different runtime contexts (YAML, shell, Python), not sloppiness. And release-v*-test is the right suffix convention. One more reason to avoid adding more porograming languages into a project. But I do not see it like a huge issue.

scripts/check-ocp-version-drift.sh
m1kola: Yes, but still - where does `TARGET_BRANCH` come from? Don't we need to pas it to the task?

josvazg: You selected it by choosing the evergreen project, each project is tied to a GitHub branch.

scripts/evergreen/should_release_agents_on_ecr.sh
josvazg: Tried to make it simpler by defaulting once per script.

scripts/check-ocp-version-drift.sh
josvazg: Sorry I misunderstood your question here. This requires a fix so that it takes the proper value from evg, regardless of how evg project ties itself to branches. I will pack it in the same commit as the prior simplication.

scripts/create-go-bump-pr.sh
josvazg: Good catch. Fixed!

scripts/test/bash/backport_branch/test_backport_branch_ci.bats
josvazg: You are right, that was dead code and the test name is misleading.

scripts/test/bash/backport_branch/test_backport_branch_ci.bats
josvazg: none really, copy paste issue, removed now

scripts/dev/contexts/evg-private-context
josvazg: The only plausible accident is someone has BUILD_SCENARIO=staging on a branch but the branch name accidentally matches the -test suffix. That's a project-level config error (wrong Evergreen project / branch misnamed), not a code bug. Worst case: images land in the test ECR registry, we notice, rename the branch, re-run. Nothing is corrupted, nothing in production is touched.

Note you'd have to name the branch as a test branch, which isn't something that happens by accident.

.github/workflows/govulncheck.yml
```diff
@@ -4,10 +4,12 @@ on:
   push:
     branches:
       - master
+      - release-v*
   pull_request:
     branches:
       - master
       - release-*
+      - release-v*
```
m1kola: Now we have `release-*` and `release-v*`. What is the difference?

josvazg: typo, will fix

josvazg: wait those were preexisting, because release-* was already a pattern before
need to find a clean naming pattern for backporting branches

josvazg: We could use branch-v* instead for branches, the release-* are for tags

josvazg: Looking deeper, seems there are some release-* branches left over, but currently there is no semantic associated with that we need to care about, not even tags. So probably we can just drop the `release-*` globs.

scripts/test/bash/backport_branch/test_backport_branch_ci.bats
```diff
@@ -0,0 +1,77 @@
...
+@test "git ls-tree: lists release-v1's own tree, not master's" {
+    cd "${CLONE_DIR}"
+    git checkout -q release-v1
+    mkdir -p pkg/kubectl-mongodb
+    echo "marker" > pkg/kubectl-mongodb/only_on_v1.go
+    git add pkg/kubectl-mongodb
+    git commit -q -m "add kubectl-mongodb path only on release-v1"
+    git push -q origin release-v1
+
+    tree_listing=$(env branch_name=release-v1 git ls-tree -r "origin/release-v1" --name-only)
```
m1kola: nit: why `env` here?

scripts/create-go-bump-pr.sh
m1kola: @josvazg I see you made a change and it is now just:

```bash
base_branch="$(git rev-parse --abbrev-ref HEAD)"
```

I believe `actions/checkout` checks out as detached head and `git rev-parse --abbrev-ref HEAD` will not return branch name when in detached head. I think this is not what you want here.

Also my understanding is that `base_branch` is supposed to be a our target branch (e.g. `master` or `release-v*` branch. Not the currently checked out commit/branch.

Or am I missing something?

josvazg: fixing

josvazg: changed to use GITHUB_REF_NAME when available

m1kola (approved):

PR #1376, opened 2026-07-15
CLOUDP-361051: Remove imageType logic

mircea-cosbuc:
# Summary

Remove the UBI8/UBI9 branching logic, now exclusively using UBI9. This means that the (EoL) MongoDB 6 is no longer supported (as it does not have images available for UBI9).

This includes cleaning up the `annotations` and `defaultArchitecture` parameters for `GetOfficialImage`, removing `MDB_APPDB_ASSUME_OLD_FORMAT`, and updating tests to accommodate.

## Checklist


pkg/images/Imageurls.go
```diff
@@ -114,42 +114,32 @@ func ContainerImage(imageUrls ImageUrls, imageName string, version string) strin
...
 		r := regexp.MustCompile("-.+$")
 		if !r.MatchString(version) {
 			version = version + "-" + imageType
 		}
+		// 5.0.6-ubi8 -> 5.0.6-ubi9
 		if found, suffix := architectures.HasSupportedImageTypeSuffix(version); found {
 			version = fmt.Sprintf("%s%s", strings.TrimSuffix(version, suffix), imageType)
 		}
-		// if neither, let's not change it: 5.0.6-ubi8 -> 5.0.6-ubi8
+		// any other suffix is left untouched: 5.0.6-ubuntu -> 5.0.6-ubuntu
```
m1kola, line 142: ubuntu? Is that right?

pkg/images/Imageurls.go
```diff
@@ -114,42 +114,32 @@ func ContainerImage(imageUrls ImageUrls, imageName string, version string) strin
...
-
-	if architectures.IsRunningStaticArchitecture(annotations, defaultArchitecture) {
-		imageType = string(architectures.ImageTypeUBI9)
-	} else {
-		// For non-static architecture, we need to default to UBI8 to support customers running MongoDB versions < 6.0.4,
-		// which don't have UBI9 binaries.
-		imageType = string(architectures.ImageTypeUBI8)
-	}
-
+	imageType := string(architectures.ImageTypeUBI9)
```
m1kola, line 121: Reminder to double check with Yavor: is it still compatible with what he was doing (I believe he was looking into supporting UBI micro).

m1kola (commented):
First very quick pass.

docker/mongodb-kubernetes-database/Dockerfile
```diff
@@ -33,8 +39,6 @@ RUN microdnf install -y --disableplugin=subscription-manager \
         xz-libs \
         findutils
 
-RUN ln -s /usr/lib64/libsasl2.so.3 /usr/lib64/libsasl2.so.2
-
 # Set the required perms
 RUN mkdir -p "${MMS_LOG_DIR}" \
         && chmod 0775 "${MMS_LOG_DIR}" \
```
josvazg, line 44: **q**: Is this ln removal not problematic?

On ubi9, only .so.3 is shipped. If any binary in the container (agent, server, vendored tool) was built against ubi8 and explicitly links libsasl2.so.2, it will fail at runtime. 
I am guessing we do not have such dependency from Go, right?

pkg/images/Imageurls_test.go
```diff
@@ -54,18 +53,17 @@ func TestContainerImage(t *testing.T) {
...
 	tests := []struct {
-		name        string
-		input       string
-		annotations map[string]string
-		want        string
-		setupEnvs   func(t *testing.T)
+		name      string
+		input     string
+		want      string
+		setupEnvs func(t *testing.T)
```
josvazg, line 61: **q**:  Should we add a explicit test for -ubi9 input being idempotent?
  
`pkg/images/Imageurls_test.go` tests rewriting -ubi8 → -ubi9, but doesn't explicitly test that a version already with -ubi9 is correctly left unchanged. The function is idempotent in practice, but an explicit test would prevent regressions.

josvazg (approved):
LGTM with minor comments

PR #1382, opened 2026-07-16
KUBE-179: clean up and self-heal MongoDBSearch resources

anandsyncs:
# Summary

MongoDBSearch resources must clean up after themselves across lifecycle events: CR deletion, member-cluster removal, and shard scale-down. This PR builds on #1362 (identity + state adoption), which is now **merged** — the branch is rebased directly onto master and the PR targets master. The design is deliberately simple: cached clients everywhere, owner-reference GC on the local cluster, and one best-effort cleanup pass where GC cannot reach. Rapid replacement/remove-readd races, self-healing after out-of-band tampering, and post-restart recovery are explicitly out of scope; the quick same-name-successor race is accepted by design (see #1362).

```mermaid
flowchart LR
  A[master incl. merged PR 1362] --> C[PR 1382 lifecycle cleanup - this PR]
  C --> D[PR 1383 lifecycle e2e]
  D --> E[PR 1393 GA upgrade e2e]
```

## The model

A cluster entry's identity is its **index**: pod names, resource suffixes, and OM host IDs all derive from CR name + index. The cluster **name** is display metadata plus the member-client routing key and never gates lifecycle decisions. An index change means the old cluster was deleted and a new one added; a rename at the same index is a pure metadata update; re-pointing an index at a different physical cluster is unsupported and rejected at admission (see the CEL rule below).

## What this PR does

**Shared primitives (one mechanism, not four):**
- `searchcontroller.DeleteOwnedResource` (Get → ownership + component-label check → Delete; `found` = existed-before-delete, which the forwarder-before-host Pending gate relies on) and `DeleteAllOwnedResources` (owner-label list/delete), both over one unexported predicate core. The stale-shard keep-set and the customer-Secret name-shape guards are predicate callers of the same core. All three controllers — main, metrics, Envoy — delete through these; the Envoy teardown deliberately **skips with a warning** on a same-name object owned by someone else instead of erroring (an ownership collision it can never resolve must not requeue LB teardown forever).
- One ownership gate: `khandler.HasSearchOwnership` — search-name/namespace label equality over `SearchOwnershipLabels` (the Envoy gate adds the component check). No UID-based selectivity (successor race accepted in #1362); no cluster-name labels (names never gate lifecycle).

[...]

PR #1383, opened 2026-07-16
KUBE-180: validate MongoDBSearch lifecycle recovery

anandsyncs:
# Summary

End-to-end coverage for the MongoDBSearch lifecycle behaviors introduced by the identity layer (#1362) and the cleanup layer (#1382). Test-only (zero Go), targets master.

- **CR-delete cleanup**: full per-member-cluster artifact sweep with customer inputs surviving UID-unchanged (2-cluster RS, `q2_mc_rs_steady.py`); named single-cluster shape (q2); PVC reclaim (community); metrics-forwarder teardown incl. removal of all six controller-injected mongot `setParameter`s and `searchMongotHostsRemovalFinalizer` release, plus the disable-retention contract.
- **Cluster-entry remove/re-add** with survivor-UID pinning: RS (q2), sharded (`q3_mc_sharded_external_mtls.py`), and removed-own-entry sweeps on real clusters (`simulated_mc_rs.py` / `simulated_mc_sharded.py`, incl. central state-CM lifecycle).
- **Shard scale-down** reaping; **managed→none LB** both sides: single-cluster teardown vs the MC rejection contract (`validateMCRequiresManagedLB`); missing-client warn-skip policy.
- Deletion asserted via concrete (kind, name) identity polls first, label-scoped emptiness List as backstop; ownership via `mongodb.com/search-name`/`search-namespace` (+cluster/component) labels.

Notes:

- GC-with-operator-scaled-down coverage lives in #1393's upgrade suite, not this PR.

## Proof of Work

- **Head `eb6da7b25`: full PR patch green (28/28 checks)**, including the three previously red ones (`unit_tests`/lint, `e2e_multi_cluster_2_clusters`, `e2e_om80_kind_ubi_large`).
- Earlier validated lineage: full 69-task matrices green twice ([`6a67d2fb`](https://spruce.corp.mongodb.com/version/6a67d2fb23f20c000767e00e), [`6a678ef1`](https://spruce.corp.mongodb.com/version/6a678ef18e26f100074ce4d8)); targeted runs 16/16 ([`6a69c980`](https://spruce.corp.mongodb.com/version/6a69c980ea4e0e000783fce9)) and 12/12 ([`6a6a1a9b`](https://spruce.corp.mongodb.com/version/6a6a1a9b17eec10007b1083b)), all at execution 0.

## Checklist

---

### DIFF
--- a/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
+++ b/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
@@ -23,6 +23,7 @@ from tests.common.search.search_tester import SearchTester
 
 if TYPE_CHECKING:
     from kubetester.mongodb_search import MongoDBSearch
+    from kubetester.multicluster_client import MultiClusterClient
 
 logger = test_logger.get_test_logger(__name__)
 
@@ -984,6 +985,101 @@ def wait_for_resource_deleted(read_fn: Callable[[], Any], what: str, timeout: in
 
 def wait_for_search_deleted(mdbs: "MongoDBSearch", timeout: int = 300) -> None:
     wait_for_resource_deleted(mdbs.load, f"MongoDBSearch {mdbs.name}", timeout=timeout)
+
+
+def protected_search_input_uids(
+    core_v1: Any,
+    namespace: str,
+    source_tls_secret_name: str,
+    sync_user_secret_name: str,
+    ca_configmap_name: str,
+    *,
+    additional_secret_names: tuple[str, ...] = (),
+) -> dict[str, str]:
+    def uid(resource: Any, what: str) -> str:
+        value = resource.metadata.uid
+        assert value, f"{what} has no UID"
+        return value
+
+    uids = {
+        "source_tls_secret": uid(
+            core_v1.read_namespaced_secret(source_tls_secret_name, namespace),
+            f"Secret {source_tls_secret_name}",
+        ),
+        "sync_user_secret": uid(
+            core_v1.read_namespaced_secret(sync_user_secret_name, namespace),
+            f"Secret {sync_user_secret_name}",
+        ),
+        "ca_configmap": uid(
+            core_v1.read_namespaced_config_map(ca_configmap_name, namespace),
+            f"ConfigMap {ca_configmap_name}",
+        ),
+    }
+    uids.update(
+        {
+            f"secret/{name}": uid(core_v1.read_namespaced_secret(name, namespace), f"Secret {name}")
+            for name in additional_secret_names
+        }
+    )
+    return uids
+
+
+def search_artifact_uids(
+    mcc: "MultiClusterClient",
+    namespace: str,
+    names: dict[str, str],
+) -> dict[str, str]:
+    apps = mcc.apps_v1_api()
+    core = mcc.core_v1_api()
+    resources = {
+        f"StatefulSet/{names['sts']}": mcc.read_namespaced_stateful_set(names["sts"], namespace),
+        f"Service/{names['svc']}": mcc.read_namespaced_service(names["svc"], namespace),
+        f"Service/{names['proxy']}": mcc.read_namespaced_service(names["proxy"], namespace),
+        f"ConfigMap/{names['mongot_cm']}": mcc.read_namespaced_config_map(names["mongot_cm"], namespace),
+        f"Deployment/{names['envoy_deployment']}": apps.read_namespaced_deployment(
+            names["envoy_deployment"], namespace
+        ),
+        f"ConfigMap/{names['envoy_cm']}": mcc.read_namespaced_config_map(names["envoy_cm"], namespace),
+        f"Secret/{names['operator_tls_secret']}": core.read_namespaced_secret(names["operator_tls_secret"], namespace),
+    }
+    pvc_names = mongot_data_pvc_names(namespace, names["sts"], api_client=mcc.api_client)
+    assert pvc_names, f"[{mcc.cluster_name}] expected mongot data PVCs for {names['sts']}"
+    resources.update(
+        {
+            f"PersistentVolumeClaim/{name}": core.read_namespaced_persistent_volume_claim(name, namespace)
+            for name in pvc_names
+        }
+    )
+
+    uids = {what: resource.metadata.uid for what, resource in resources.items()}
+    assert all(uids.values()), f"[{mcc.cluster_name}] Search artifact without UID: {uids}"
+    return uids
+
+
+def wait_for_search_artifacts_deleted(
+    mcc: "MultiClusterClient",
+    namespace: str,
+    sts_name: str,
+    search_name: str,
+    *,
+    timeout: int = 600,
+) -> None:
+    """Label-scoped absence is the authoritative cleanup assertion; the exact-name
+    STS check on top catches accidental label stripping on the primary artifact."""
+    wait_for_resource_deleted(
+        lambda: mcc.read_namespaced_stateful_set(sts_name, namespace),
+        f"STS {sts_name} in {mcc.cluster_name}",
+        timeout=timeout,
+    )
+    wait_for_mongot_pvcs_deleted(namespace, sts_name, api_client=mcc.api_client, timeout=300)
+    wait_for_search_owned_resources_deleted(
+        mcc.apps_v1_api(),
+        mcc.core_v1_api(),
+        namespace,
+        search_name,
+        where=mcc.cluster_name,
+        timeout=timeout,
+    )
 
 
 def hard_kill_pods_by_label(
--- a/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
+++ b/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
@@ -10,7 +10,7 @@ import re
 import time
 import uuid
 from dataclasses import dataclass, field
-from typing import Any, Callable, Optional
+from typing import TYPE_CHECKING, Any, Callable, Optional
 
 import pymongo
 import pymongo.errors
@@ -20,6 +20,9 @@ from kubetester.kubetester import run_periodically
 from pymongo.read_preferences import Nearest
 from tests import test_logger
 from tests.common.search.search_tester import SearchTester
+
+if TYPE_CHECKING:
+    from kubetester.mongodb_search import MongoDBSearch
 
 logger = test_logger.get_test_logger(__name__)
 
@@ -39,6 +42,9 @@ FAILURE_SEARCH_NOT_ENABLED = "search_not_enabled"
 # only an explicit index-state rejection proves the bad Ready-while-syncing mode.
 FAILURE_MONGOT_UNREACHABLE = "mongot_unreachable"
 FAILURE_OTHER = "other"
+
+SEARCH_OWNER_NAME_LABEL = "mongodb.com/search-name"
+SEARCH_OWNER_NAMESPACE_LABEL = "mongodb.com/search-namespace"
 
 _CURSOR_LOST_MESSAGE_RE = re.compile(
     r"cursor id .*?(not found|was killed)|remote error from mongot|rst_stream",
@@ -894,14 +900,9 @@ def wait_for_mongot_pvcs_deleted(
     named ``<template>-<sts>-<ordinal>``, so we match on the ``-<sts>-`` infix.
     ``api_client`` must target the cluster hosting the STS.
     """
-    core = client.CoreV1Api(api_client=api_client)
 
     def deleted() -> tuple[bool, str]:
-        remaining = [
-            pvc.metadata.name
-            for pvc in core.list_namespaced_persistent_volume_claim(namespace).items
-            if f"-{sts_name}-" in f"-{pvc.metadata.name}"
-        ]
+        remaining = mongot_data_pvc_names(namespace, sts_name, api_client=api_client)
         return not remaining, ("all deleted" if not remaining else f"remaining={remaining}")
 
     run_periodically(
@@ -910,6 +911,79 @@ def wait_for_mongot_pvcs_deleted(
         sleep_time=sleep_time,
         msg=f"mongot PVCs for STS {sts_name} to be deleted",
     )
+
+
+def _search_owned_top_level_resources(
+    apps_v1: Any,
+    core_v1: Any,
+    namespace: str,
+    search_name: str,
+) -> dict[str, list[Any]]:
+    selector = f"{SEARCH_OWNER_NAME_LABEL}={search_name},{SEARCH_OWNER_NAMESPACE_LABEL}={namespace}"
+    return {
+        "StatefulSet": apps_v1.list_namespaced_stateful_set(namespace, label_selector=selector).items,
+        "Deployment": apps_v1.list_namespaced_deployment(namespace, label_selector=selector).items,
+        "Service": core_v1.list_namespaced_service(namespace, label_selector=selector).items,
+        "ConfigMap": core_v1.list_namespaced_config_map(namespace, label_selector=selector).items,
+        "Secret": core_v1.list_namespaced_secret(namespace, label_selector=selector).items,
+    }
+
+
+def wait_for_search_owned_resources_deleted(
+    apps_v1: Any,
+    core_v1: Any,
+    namespace: str,
+    search_name: str,
+    *,
+    where: str,
+    timeout: int = 600,
+) -> None:
+    def deleted() -> tuple[bool, str]:
+        resources = _search_owned_top_level_resources(apps_v1, core_v1, namespace, search_name)
+        remaining = [
+            f"{kind}/{resource.metadata.name}(uid={resource.metadata.uid})"
+            for kind, items in resources.items()
+            for resource in items
+        ]
+        return not remaining, f"{where}: remaining={remaining}"
+
+    run_periodically(
+        deleted,
+        timeout=timeout,
+        sleep_time=5,
+        msg=f"{where} Search-owned top-level resource cleanup",
+    )
+
+
+def mongot_data_pvc_names(
+    namespace: str,
+    sts_name: str,
+    *,
+    api_client: Optional[client.ApiClient] = None,
+) -> list[str]:
+    core = client.CoreV1Api(api_client=api_client)
+    return [
+        pvc.metadata.name
+        for pvc in core.list_namespaced_persistent_volume_claim(namespace).items
+        if f"-{sts_name}-" in f"-{pvc.metadata.name}"
+    ]
+
+
+def wait_for_resource_deleted(read_fn: Callable[[], Any], what: str, timeout: int = 300) -> None:
+    def deleted() -> tuple[bool, str]:
+        try:
+            read_fn()
+            return False, f"{what} still present"
+        except client.exceptions.ApiException as exc:
+            if exc.status == 404:
+                return True, f"{what} deleted"
+            raise
+
+    run_periodically(deleted, timeout=timeout, sleep_time=5, msg=f"{what} cleanup")
+
+
+def wait_for_search_deleted(mdbs: "MongoDBSearch", timeout: int = 300) -> None:
+    wait_for_resource_deleted(mdbs.load, f"MongoDBSearch {mdbs.name}", timeout=timeout)
 
 
 def hard_kill_pods_by_label(
--- a/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
+++ b/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
@@ -1029,6 +1029,15 @@ def protected_search_input_uids(
     *,
     additional_secret_names: tuple[str, ...] = (),
 ) -> dict[str, str]:
+    """Snapshot the UIDs of customer-provided ("protected") Search inputs.
+
+    These are resources the customer creates (source TLS Secret, sync-user password
+    Secret, CA ConfigMap, ...) that the operator consumes but must never delete or
+    recreate — cleanup sweeps only remove operator-owned artifacts. Tests capture
+    these UIDs before a lifecycle transition and assert afterwards that they are
+    unchanged, proving the inputs survived untouched.
+    """
+
     def uid(resource: Any, what: str) -> str:
         value = resource.metadata.uid
         assert value, f"{what} has no UID"
--- a/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
+++ b/docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
@@ -24,6 +24,7 @@ from tests.common.search.search_tester import SearchTester
 if TYPE_CHECKING:
     from kubetester.mongodb_search import MongoDBSearch
     from kubetester.multicluster_client import MultiClusterClient
+    from kubetester.phase import Phase
 
 logger = test_logger.get_test_logger(__name__)
 
@@ -987,6 +988,38 @@ def wait_for_search_deleted(mdbs: "MongoDBSearch", timeout: int = 300) -> None:
     wait_for_resource_deleted(mdbs.load, f"MongoDBSearch {mdbs.name}", timeout=timeout)
 
 
+def wait_for_deployment_recreated(apps: Any, namespace: str, name: str, previous_uid: str, timeout: int = 300) -> None:
+    """Poll until Deployment `name` exists with a NEW uid and all desired replicas ready."""
+
+    def recreated_and_ready() -> tuple[bool, str]:
+        try:
+            deployment = apps.read_namespaced_deployment(name, namespace)
+        except client.exceptions.ApiException as exc:
+            if exc.status == 404:
+                return False, f"{name} absent"
+            raise
+        desired = deployment.spec.replicas or 0
+        ready = deployment.status.ready_replicas or 0
+        replaced = deployment.metadata.uid != previous_uid
+        return replaced and desired > 0 and ready == desired, f"replaced={replaced}, ready={ready}/{desired}"
+
+    run_periodically(recreated_and_ready, timeout=timeout, sleep_time=5, msg=f"Deployment {name} recreation")
+
+
+def wait_for_metrics_forwarder_phase(mdbs: "MongoDBSearch", phase: "Phase", timeout: int = 120) -> None:
+    def reached() -> bool:
+        mdbs.reload()
+        status = mdbs.get_metrics_forwarder_status()
+        return status is not None and status.get("phase") == phase.name
+
+    run_periodically(
+        reached,
+        timeout=timeout,
+        sleep_time=10,
+        msg=f"metrics forwarder status to reach {phase.name}",
+    )
+
+
 def protected_search_input_uids(
     core_v1: Any,
     namespace: str,
@@ -1024,24 +1057,37 @@ def protected_search_input_uids(
     return uids
 
 
+def _mc_search_artifact_readers(
+    mcc: "MultiClusterClient",
+    namespace: str,
+    names: dict[str, str],
+) -> dict[str, Callable[[], Any]]:
+    """Direct readers for the concrete (kind, name) identities of one cluster's
+    Search artifact set (keys match `mc_search_artifact_names`)."""
+    apps = mcc.apps_v1_api()
+    core = mcc.core_v1_api()
+    return {
+        f"StatefulSet/{names['sts']}": lambda: mcc.read_namespaced_stateful_set(names["sts"], namespace),
+        f"Service/{names['svc']}": lambda: mcc.read_namespaced_service(names["svc"], namespace),
+        f"Service/{names['proxy']}": lambda: mcc.read_namespaced_service(names["proxy"], namespace),
+        f"ConfigMap/{names['mongot_cm']}": lambda: mcc.read_namespaced_config_map(names["mongot_cm"], namespace),
+        f"Deployment/{names['envoy_deployment']}": lambda: apps.read_namespaced_deployment(
+            names["envoy_deployment"], namespace
+        ),
+        f"ConfigMap/{names['envoy_cm']}": lambda: mcc.read_namespaced_config_map(names["envoy_cm"], namespace),
+        f"Secret/{names['operator_tls_secret']}": lambda: core.read_namespaced_secret(
+            names["operator_tls_secret"], namespace
+        ),
+    }
+
+
 def search_artifact_uids(
     mcc: "MultiClusterClient",
     namespace: str,
     names: dict[str, str],
 ) -> dict[str, str]:
-    apps = mcc.apps_v1_api()
     core = mcc.core_v1_api()
-    resources = {
-        f"StatefulSet/{names['sts']}": mcc.read_namespaced_stateful_set(names["sts"], namespace),
-        f"Service/{names['svc']}": mcc.read_namespaced_service(names["svc"], namespace),
-        f"Service/{names['proxy']}": mcc.read_namespaced_service(names["proxy"], namespace),
-        f"ConfigMap/{names['mongot_cm']}": mcc.read_namespaced_config_map(names["mongot_cm"], namespace),
-        f"Deployment/{names['envoy_deployment']}": apps.read_namespaced_deployment(
-            names["envoy_deployment"], namespace
-        ),
-        f"ConfigMap/{names['envoy_cm']}": mcc.read_namespaced_config_map(names["envoy_cm"], namespace),
-        f"Secret/{names['operator_tls_secret']}": core.read_namespaced_secret(names["operator_tls_secret"], namespace),
-    }
+    resources = {what: read() for what, read in _mc_search_artifact_readers(mcc, namespace, names).items()}
     pvc_names = mongot_data_pvc_names(namespace, names["sts"], api_client=mcc.api_client)
     assert pvc_names, f"[{mcc.cluster_name}] expected mongot data PVCs for {names['sts']}"
     resources.update(
@@ -1059,19 +1105,36 @@ def search_artifact_uids(
 def wait_for_search_artifacts_deleted(
     mcc: "MultiClusterClient",
     namespace: str,
-    sts_name: str,
+    names: dict[str, str],
     search_name: str,
     *,
     timeout: int = 600,
 ) -> None:
-    """Label-scoped absence is the authoritative cleanup assertion; the exact-name
-    STS check on top catches accidental label stripping on the primary artifact."""
-    wait_for_resource_deleted(
-        lambda: mcc.read_namespaced_stateful_set(sts_name, namespace),
-        f"STS {sts_name} in {mcc.cluster_name}",
+    """Identity-based absence is the authoritative cleanup assertion: every concrete
+    (kind, name) captured before deletion must read back 404 — a resource that loses
+    its owner labels drops out of a label selector but not out of this poll. The
+    label-scoped emptiness List stays as a secondary check for labeled orphans
+    outside the captured inventory."""
+    readers = _mc_search_artifact_readers(mcc, namespace, names)
+
+    def identities_deleted() -> tuple[bool, str]:
+        remaining = []
+        for what, read in readers.items():
+            try:
+                read()
+                remaining.append(what)
+            except client.exceptions.ApiException as exc:
+                if exc.status != 404:
+                    raise
+        return not remaining, f"{mcc.cluster_name}: remaining={remaining}"
+
+    run_periodically(
+        identities_deleted,
         timeout=timeout,
+        sleep_time=5,
+        msg=f"{mcc.cluster_name} Search artifact identity cleanup",
     )
-    wait_for_mongot_pvcs_deleted(namespace, sts_name, api_client=mcc.api_client, timeout=300)
+    wait_for_mongot_pvcs_deleted(namespace, names["sts"], api_client=mcc.api_client, timeout=300)
     wait_for_search_owned_resources_deleted(
         mcc.apps_v1_api(),
         mcc.core_v1_api(),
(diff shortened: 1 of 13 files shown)


docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -912,6 +911,132 @@ def deleted() -> tuple[bool, str]:
...
+    mcc.read_namespaced_config_map(names["mongot_cm"], namespace)
+    mcc.apps_v1_api().read_namespaced_deployment(name=names["envoy_deployment"], namespace=namespace)
+    mcc.read_namespaced_config_map(names["envoy_cm"], namespace)
+    mcc.core_v1_api().read_namespaced_secret(names["operator_tls_secret"], namespace)
+    assert mongot_data_pvc_names(
+        namespace, names["sts"], api_client=mcc.api_client
+    ), f"expected at least one mongot data PVC for {names['sts']} before delete"
+
+
+def wait_for_search_artifacts_deleted(
```
anandsyncs, line 1114: **Systemic tightening for all four cleanup tests:** every absence check in this helper (and its callers) is by exact resource NAME. An orphan the operator leaves under an unexpected name — e.g. a stale TLS copy from a previous UID, or a legacy-named leftover — is invisible to every cleanup assertion in the stack. Adding one list-by-label-selector (`mongodb.com/search-name=<name>` [+ uid]) emptiness check per kind at the end of this helper closes that hole globally.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -92,6 +117,41 @@ def test_verify_per_cluster_status(mdbs: MongoDBSearch):
...
+    sts = get_statefulset(namespace, search_resource_names.mongot_statefulset_name(mdbs.name))
+    assert sts.metadata.labels[CUSTOM_STATEFULSET_LABEL] == CUSTOM_STATEFULSET_LABEL_VALUE
+    assert sts.metadata.labels["mongodb.com/search-name"] == mdbs.name
+    assert sts.metadata.labels["mongodb.com/search-namespace"] == namespace
+    assert sts.metadata.labels["mongodb.com/search-uid"] == mdbs["metadata"]["uid"]
+    assert sts.metadata.labels["component"] == "mongot"
+
+
+@mark.e2e_search_community_basic
+def test_search_statefulset_restores_removed_owner_label(namespace: str, mdbs: MongoDBSearch):
```
anandsyncs: This waits only for the stripped `mongodb.com/search-name` label to return. A regression where the reapply restores that one label but drops `search-uid`/`component` or clobbers the custom user label would pass. After restoration, re-read once and assert the full protected set + `CUSTOM_STATEFULSET_LABEL` are all intact (reuse the tuple from the coexistence test above). Also 300s is generous for a next-reconcile reapply; 60-90s would catch latency regressions.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -128,16 +188,40 @@ def test_search_assert_search_query(sample_movies_helper: SampleMoviesSearchHelp
...
-    ``data`` -> ``data-<sts>-<ordinal>``)."""
-    core_v1 = client.CoreV1Api()
-    prefix = f"data-{sts_name}-"
-    return [
-        pvc.metadata.name
-        for pvc in core_v1.list_namespaced_persistent_volume_claim(namespace).items
-        if pvc.metadata.name.startswith(prefix)
-    ]
+@mark.e2e_search_community_basic
+def test_recreate_mongot_top_level_resources(
```
anandsyncs: Recreated objects are validated by UID-change + readiness, but never re-checked for protected labels. A self-heal that recreates the STS without the owner labels (breaking future sweep/routing) passes. One extra assert on the recreated resources' labels closes it — same applies to the q2 and simulated-MC recreation tests.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -760,6 +764,155 @@ def test_verify_per_cluster_envoy_deployment(
...
+        lambda: target_cluster.read_namespaced_config_map(cm_name, namespace),
+        f"ConfigMap {cm_name} in {target_cluster_name}",
+        timeout=600,
+    )
+    wait_for_resource_deleted(
+        lambda: target_cluster.core_v1_api().read_namespaced_secret(operator_tls_secret, namespace),
+        f"operator-managed TLS secret {operator_tls_secret} in {target_cluster_name}",
+        timeout=600,
+    )
+    wait_for_mongot_pvcs_deleted(namespace, sts_name, api_client=target_cluster.api_client, timeout=300)
```
anandsyncs: Cross-talk gap: after deleting `named_single`, the test never re-asserts that `mdbs`'s resources **survived** the sweep — the core 'label-scoped delete of CR A must not touch CR B' claim is only implicit in later tests. Re-read `mdbs`'s STS/Envoy/TLS-secret here and assert present with unchanged UIDs.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_metrics_forwarder_replicaset.py
```diff
@@ -206,7 +207,34 @@ def check_per_cluster_metrics_forwarder_absent():
...
+    mdbs["spec"]["observability"]["metricsForwarder"]["mode"] = "enabled"
+    mdbs.update()
+    mdbs.assert_reaches_phase(Phase.Running, timeout=300)
+
+    def metrics_forwarder_running():
+        mdbs.reload()
+        status = mdbs.get_metrics_forwarder_status()
+        return status is not None and status.get("phase") == Phase.Running.name
+
+    run_periodically(metrics_forwarder_running, timeout=120, interval=10)
```
anandsyncs: `run_periodically` reads `sleep_time=`, not `interval=` — this kwarg is silently ignored and it polls every 2s. Harmless (still bounded) but copied from the pre-existing `interval=` sites in this file; worth fixing to `sleep_time=10` (and adding `msg=`) here and in the siblings.

anandsyncs:
**Coverage-gap audit (from the stack review; ranked by customer risk).** P0 — cheap, kind-testable, should be scheduled: (1) same-name delete→immediate-recreate of a Search (the UID adopt/reset path is a headline layer-1 behavior with zero e2e today); (2) disable-reconciliation annotation honored during delete-cleanup (fail-open if wrong). P1: upgrade e2e asserting legacy owner-ref removal + label backfill post-upgrade; CR-delete while operator is scaled to 0 → cleanup on restart; restart mid-deletion (sweep resume); MC member-cluster removal sweep (drop a `clusters[]` entry); sharded full-CR-delete cleanup (only scale-down is covered); single-cluster TLS-on lifecycle (TLS cleanup e2e is MC-only today). Member-unreachable fail-closed cleanup is EKS-only (kind's shared/multi-kind harness can't produce a clean partition) — cover that decision logic at unit level. Full detail in the stack reviewer guide.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
anandsyncs: Addressed. `wait_for_search_owned_resources_deleted` now lists every managed top-level kind by `mongodb.com/search-name` and waits for the selector result to be empty, including names the test did not predict. Callers still check canonical names and PVCs, so both expected and orphaned artifacts are covered. Layer 3 passed 69/69.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: Addressed. The restoration wait is bounded to 90 seconds, then the StatefulSet is re-read once and `assert_search_resource_labels` verifies name, namespace, UID, and component; the custom user label is asserted separately. The authoritative Layer-3 patch passed 69/69.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: Addressed. Recreated StatefulSets, Services, ConfigMaps, Envoy resources, and TLS Secrets are re-read and checked with the complete protected-label assertion in the single-cluster, Q2, and simulated-MC recovery scenarios. All corresponding Layer-3 tasks passed in the 69/69 matrix.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
anandsyncs: Addressed. Before deleting the named Search, the test records the primary Search StatefulSet, Envoy Deployment, and TLS Secret UIDs. After the named cleanup and orphan scan, it re-reads all three and requires those UIDs to be unchanged. The Q2 task passed in the authoritative Layer-3 patch.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_metrics_forwarder_replicaset.py
anandsyncs: Addressed across both metrics-forwarder replica-set scenarios: the affected polls use `sleep_time=10` with diagnostic messages, and the copied `interval=` uses were removed. Both metrics-forwarder tasks passed in the Layer-3 69/69 evidence.

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
```diff
@@ -517,6 +582,114 @@ def test_per_cluster_search_query(
         logger.info(f"cluster_index={cluster_index}: deterministic $search (==4) via local mongod OK")
 
 
+@mark.e2e_search_simulated_mc_rs
+def test_recreate_local_search_resources_and_restart_operator(
```
anandsyncs: Please remove this E2E. Post-restart orphan/self-heal recovery is not a requirement, and the scenario also depends on ConfigMap delete routing that the simplified controller keeps intentionally unsupported.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -92,6 +119,56 @@ def test_verify_per_cluster_status(mdbs: MongoDBSearch):
...
+        search_namespace=namespace,
+        search_uid=mdbs["metadata"]["uid"],
+        component="mongot",
+        where=f"StatefulSet {namespace}/{sts.metadata.name}",
+    )
+    assert sts.metadata.labels[CUSTOM_STATEFULSET_LABEL] == CUSTOM_STATEFULSET_LABEL_VALUE
+
+
+@mark.e2e_search_community_basic
+def test_search_statefulset_restores_removed_owner_label(namespace: str, mdbs: MongoDBSearch):
```
anandsyncs: Please drop this expensive label-drift E2E. Required-label merge ordering can be covered by one compact unit test; live protected-label deletion recovery is not a lifecycle guarantee needed by this stack.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -128,16 +205,55 @@ def test_search_assert_search_query(sample_movies_helper: SampleMoviesSearchHelp
...
-    ``data`` -> ``data-<sts>-<ordinal>``)."""
-    core_v1 = client.CoreV1Api()
-    prefix = f"data-{sts_name}-"
-    return [
-        pvc.metadata.name
-        for pvc in core_v1.list_namespaced_persistent_volume_claim(namespace).items
-        if pvc.metadata.name.startswith(prefix)
-    ]
+@mark.e2e_search_community_basic
+def test_recreate_mongot_top_level_resources(
```
anandsyncs: Please remove or narrow this scenario. ResourcesHandler intentionally ignores ConfigMap/Secret deletion, while StatefulSet/Service recreation is pre-existing behavior rather than new lifecycle coverage. This test currently turns unsupported ConfigMap delete self-healing into a required contract.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -1126,6 +1331,130 @@ def execute_vector_search(_idx: int = cluster_index) -> tuple:
         )
 
 
+@mark.e2e_search_q2_mc_rs_steady
+def test_remove_and_readd_search_cluster_entry(
```
anandsyncs, line 1232: The target only requires best-effort cleanup when a cluster is removed. Please end this test after the removal, retained-cluster, and customer-input assertions; re-adding the entry and proving UID turnover is normal reconciliation and does not need a second lifecycle matrix here.

docker/mongodb-kubernetes-tests/tests/search/search_sharded_scale_shards_managed_lb.py
```diff
@@ -578,3 +629,116 @@ def test_scale_down_verify_search(mdb: MongoDB):
     verify_text_search_query(search_tester)
     verify_search_results_from_all_shards(search_tester)
     logger.info("Search verification passed after scale-down to %d shards", SCALED_DOWN_SHARD_COUNT)
+
+
+@MARKER
+def test_same_name_search_recreation_survives_old_generation_cleanup(
```
anandsyncs: Please remove this same-name delete/recreate scenario. Rapid generation replacement is explicitly out of scope, and the uncached UID/resourceVersion fencing that this test exists to validate is being removed from #1382.

docker/mongodb-kubernetes-tests/tests/search/search_external_metrics_forwarder_replicaset.py
```diff
@@ -179,17 +180,47 @@ def test_create_search_resource(helper: SearchDeploymentHelper, mdb: MongoDB, md
...
-
-    run_periodically(check_metrics_forwarder_status, timeout=120, interval=10)
+    wait_for_metrics_forwarder_phase(mdbs, Phase.Running)
 
     tester = om.get_om_tester(project_name=MDB_RESOURCE_NAME)
     tester.assert_mongot_hosts_converged(mdbs.mongot_pod_hostnames())
 
 
+@mark.e2e_search_external_metrics_forwarder_replicaset
+def test_metrics_forwarder_recreates_deleted_resources(om: MongoDBOpsManager, mdbs: MongoDBSearch):
```
anandsyncs: Please remove the ConfigMap deletion/recreation requirement from this E2E. Metrics should react to central credential/CA updates and keep the existing ConfigMap/Secret delete semantics; mode transitions and CR-deletion cleanup already cover the supported forwarder lifecycle.

docker/mongodb-kubernetes-tests/tests/search/search_sharded_scale_shards_managed_lb.py
anandsyncs: Removed `test_same_name_search_recreation_survives_old_generation_cleanup` together with the fencing it validated (#1382 drops the UID/ResourceVersion precondition machinery; rapid replacement races are out of scope).

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
anandsyncs: Removed `test_recreate_local_search_resources_and_restart_operator` and its now-dead operator-restart/controller-idle helpers; post-restart recovery is out of scope.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
anandsyncs: Truncated right after the removal-cleanup, retained-cluster, and customer-input assertions. A minimal 6-line topology restore (re-add entry, Running, 2 clusters) remains only because downstream fault/delete tests in the module assert a 2-cluster topology; the second lifecycle matrix, re-add label sweep, and UID-turnover proof are gone.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: Removed `test_recreate_mongot_top_level_resources` outright.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: Removed `test_search_statefulset_restores_removed_owner_label`; label precedence is covered by the compact unit test in #1362 (`TestSearchManagedLabelsWinOverOverrides`).

docker/mongodb-kubernetes-tests/tests/search/search_external_metrics_forwarder_replicaset.py
anandsyncs: Narrowed to the Deployment-recreation phase only (renamed `test_metrics_forwarder_recreates_deleted_deployment`); the ConfigMap delete/recreate phase is removed along with the member ConfigMap/Secret delete routing in #1382.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -760,6 +772,198 @@ def test_verify_per_cluster_envoy_deployment(
...
+                f"[{other_cluster.cluster_name}] unexpectedly found {sts_name}; "
+                f"named single-cluster spec should only materialize resources on {target_cluster_name}"
+            )
+        except ApiException as exc:
+            assert exc.status == 404, f"[{other_cluster.cluster_name}] unexpected error while reading {sts_name}: {exc}"
+
+    pvcs_before = mongot_data_pvc_names(namespace, sts_name, api_client=target_cluster.api_client)
+    assert pvcs_before, f"expected at least one mongot data PVC for {sts_name} before delete"
+
+    main_cm_name = search_resource_names.mongot_configmap_name_for_cluster(MDBS_RESOURCE_NAME, target_cluster_index)
```
anandsyncs: Please delete the unrelated self-heal block from `main_cm_name` through `primary_uids` (lines 861-927). ConfigMap/Secret deletion recreation is explicitly unsupported, and Envoy Deployment recreation is covered elsewhere. Keep this as a compact named-single integration case: create on exactly one member, assert no cross-cluster owner refs/placement, delete with the shared cleanup helper, and re-read one primary Search artifact to prove isolation. This cuts roughly half the test while retaining its unique edge.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -912,6 +918,290 @@ def deleted() -> tuple[bool, str]:
...
+    mcc: "MultiClusterClient",
+    namespace: str,
+    names: dict[str, str],
+    search_name: str,
+    *,
+    timeout: int = 600,
+) -> None:
+    resources = (
```
anandsyncs: Now that label-scoped absence is the authoritative cleanup assertion, seven exact-name polling lambdas duplicate the final sweep. Keep at most one representative StatefulSet name check (to catch accidental label stripping), the PVC wait, and `wait_for_search_owned_resources_deleted`; inline/remove the single-caller `assert_search_artifacts_present` helper at the same time. This should save about 40 lines and still checks canonical plus orphaned cleanup.

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
```diff
@@ -571,3 +577,47 @@ def test_cross_cluster_isolation_absence(
     Confirms each simulated operator only materialises its own LocalizeToCluster slice.
     """
     assert_cross_cluster_isolation(namespace, per_cluster_mdbs_search, MDBS_RESOURCE_NAME)
+
+
+@mark.e2e_search_simulated_mc_rs
+def test_delete_search_resources_cleanup_per_cluster(
```
anandsyncs: Please remove this per-cluster delete matrix. Q2 already covers reachable member cleanup and cross-Search isolation, while #1393 proves the stronger local owner-reference GC path with the operator down. This case adds no unavailable-member/finalizer behavior, so it is duplicate coverage across a second topology.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -92,6 +115,21 @@ def test_verify_per_cluster_status(mdbs: MongoDBSearch):
     mdbs.assert_cluster_statuses(expected_count=1, expect_managed_lb=False)
 
 
+@mark.e2e_search_community_basic
+def test_search_statefulset_preserves_custom_and_protected_labels(namespace: str, mdbs: MongoDBSearch):
```
anandsyncs: This is an expensive E2E assertion for pure label-merge precedence, already pinned by #1362 `TestSearchManagedLabelsWinOverOverrides`. It has no lifecycle transition. Remove the test and imports rather than keeping the same invariant at unit and E2E layers.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -169,24 +195,30 @@ def test_scale_mongot_offline_reclaims_pvc(namespace: str, mdbs: MongoDBSearch):
...
+    core = client.CoreV1Api()
 
-    # Presence guard: the STS and its data PVC must exist before delete, otherwise
-    # the post-delete absence checks below could pass vacuously.
-    client.AppsV1Api().read_namespaced_stateful_set(sts_name, namespace)
-    pvcs_before = _mongot_data_pvc_names(namespace, sts_name)
-    assert pvcs_before, f"expected at least one mongot data PVC for {sts_name} before delete"
-    logger.info(f"pre-delete: STS {sts_name} present, data PVCs {pvcs_before}")
+    mdbs.load()
+    assert_search_owned_resources_uid(
```
anandsyncs: Please revert this test to its pre-stack STS+PVC presence/deletion checks. The added all-kind UID scan and label-scoped cleanup duplicate the Q2 cleanup test and the stronger #1393 operator-down GC test; removing them also deletes the single-caller `assert_search_owned_resources_uid` helper. The original local GC/PVC coverage remains intact.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
anandsyncs: Done — the self-heal block is deleted; the compact isolation case is kept exactly as specified (create-on-one-member, no owner refs, cross-cluster 404s, shared deletion wait, one re-read of the main CR's StatefulSet proving isolation).

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
anandsyncs: Done — the seven polling lambdas are one shared helper; label-scoped absence remains the authoritative cleanup assertion (now via the search-name/namespace labels, with the uid label removed from the operator).

docker/mongodb-kubernetes-tests/tests/multicluster_search/simulated_mc_rs.py
anandsyncs: Done — the per-cluster delete matrix is removed; simulated_mc_rs.py no longer carries any delta in this PR.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: Done — the label-merge e2e, its fixture label injection, and the custom-label constants are removed.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: Done — reverted to the pre-stack STS+PVC presence/deletion checks; the only surviving delta is the shared PVC-name helper consolidation.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -912,6 +915,188 @@ def deleted() -> tuple[bool, str]:
...
+def wait_for_search_owned_resources_deleted(
+    apps_v1: Any,
+    core_v1: Any,
+    namespace: str,
+    search_name: str,
+    *,
+    where: str,
+    timeout: int = 600,
+) -> None:
+    def deleted() -> tuple[bool, str]:
```
anandsyncs, line 943: This can false-pass if a resource loses or never receives the owner labels: it disappears from the selector while still existing. Capture the concrete kind/name identities before deletion and poll those identities afterward; label-scoped absence can remain a secondary orphan check.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
anandsyncs: The remaining block is still an unsupported remove/re-add recovery path: it re-adds the removed entry and asserts Running, two-cluster status, and workload readiness. Reorder the downstream fault tests before removal, then make removal plus terminal CR deletion the end of the scenario and delete lines 1310–1323.

anandsyncs:
The lifecycle E2E is much smaller (+627/-75 versus +940/-92) but can target about +280 lines while becoming more trustworthy. Keep one local create/query/delete path with an explicit controller-owner precondition and PVC outcome; one hub remove-member path with retained query/stability followed by terminal CR deletion; one supported Deployment self-heal; shard cleanup/customer inputs; and external metrics behavior. Remove the remaining remove/re-add, the standalone named-single shape matrix, and the three-layer artifact/customer UID helpers. Tighten identity assertions to exact namespace and component values rather than presence-only checks.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
```diff
@@ -177,7 +169,7 @@ def test_delete_search_cr_reclaims_pvc(namespace: str, mdbs: MongoDBSearch):
     # Presence guard: the STS and its data PVC must exist before delete, otherwise
     # the post-delete absence checks below could pass vacuously.
     client.AppsV1Api().read_namespaced_stateful_set(sts_name, namespace)
-    pvcs_before = _mongot_data_pvc_names(namespace, sts_name)
+    pvcs_before = mongot_data_pvc_names(namespace, sts_name)
```
anandsyncs, line 172: Before deleting the CR, capture the StatefulSet and assert that its controller owner reference is this MongoDBSearch UID. Without that precondition the later absence only proves that the object disappeared, not that the local owner-reference GC contract is wired correctly.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
anandsyncs: Done — wait_for_search_artifacts_deleted now takes the captured (kind, name) inventory and polls every concrete identity to absence first (the authoritative check), then PVCs; the label-scoped emptiness List stays as the secondary orphan check exactly as suggested. All q2 call sites pass the pre-delete-asserted inventory; the community-basic suite already proved absence via concrete identities and needed no change.

docker/mongodb-kubernetes-tests/tests/search/search_community_basic.py
anandsyncs: Skipping this one (author decision): the owner-reference GC contract is pinned end-to-end by the upgrade suite — the pre/post owner-reference transition assertions and TestPostUpgradeCleanup (operator scaled to zero, owner-reference GC alone reaps every local artifact) — so the basic community test stays minimal.

anandsyncs:
Round-5: the deletion waits are identity-based now (capture kind/name before delete, poll those to
absence; the label-scoped List is demoted to a secondary orphan check). Three author-accepted e2e
additions also landed — the removed-operator-local sweep test in the RS module, its sharded
counterpart carrying the state-ConfigMap coverage (the first matrix caught that the state CM never
exists for RS sources — only the sharded routing-ready latch creates it — so that coverage moved to
the sharded module where the CM is real), and the unregistered-cluster partial-miss test with a
drift-repair probe — so the delta is +914/−76 vs the #1382 base rather than the ~+280 target, which
predated those acceptances.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_metrics_forwarder_replicaset.py
```diff
@@ -200,13 +201,40 @@ def check_per_cluster_metrics_forwarder_absent():
         cs = mdbs.get_cluster_status(0)
         return cs is not None and not cs.get("metricsForwarder")
 
-    run_periodically(check_per_cluster_metrics_forwarder_absent, timeout=120, interval=10)
+    run_periodically(
```
anandsyncs, line 240: The disable contract is only partially asserted here. After Kubernetes teardown and status cleanup, verify the metrics topology state ConfigMap is still present and Ops Manager still contains the mongot hosts. Disable must preserve both; cluster removal and CR deletion are the paths that deregister hosts. One representative source test is enough—do not duplicate this retention assertion across both enterprise and external suites.

docker/mongodb-kubernetes-tests/tests/multicluster_search/q2_mc_rs_steady.py
```diff
@@ -1119,6 +1228,100 @@ def execute_vector_search(_idx: int = cluster_index) -> tuple:
...
+    retained = clients[retained_name]
+    removed = clients[removed_name]
+    retained_names, retained_uids, _ = snapshots[retained_name]
+    removed_names, _, removed_protected_uids = snapshots[removed_name]
+
+    mdbs["spec"]["clusters"] = [deepcopy(retained_entry)]
+    mdbs.update()
+    mdbs.assert_reaches_phase(Phase.Running, timeout=900)
+    mdbs.assert_cluster_statuses(expected_count=1, expect_managed_lb=True)
+    wait_for_search_artifacts_deleted(removed, namespace, removed_names, mdbs.name)
```
anandsyncs, line 1295: This covers Kubernetes artifact removal but not the required metrics live-cluster-removal behavior. Add one metrics-enabled multi-cluster removal scenario (here or a dedicated test): the removed cluster’s forwarder Deployment/pods must terminate before its Ops Manager host IDs disappear, while the retained cluster’s hosts remain registered. Keep whole-CR deletion as the separate best-effort cleanup case, and do not re-add the removed entry.

anandsyncs: Gap confirmed — no e2e anywhere combines metrics with member-cluster removal, and the MC suites never
enable metrics. But it cannot land in this module: the q2 task runs on the cloud-qa variant, where
the metrics forwarder is Unsupported/Failed by design (external source forces explicit config);
forwarder e2es run only on the self-hosted-OM variant. Real coverage needs a dedicated MC +
self-hosted-OM test (~40–60 min new task). **Deferred to a drafted backlog ticket** with one scope
correction: assert the converged end-state (removed cluster's forwarder artifacts gone AND its OM
hosts deregistered; retained clusters' hosts registered throughout) rather than strict
pods-before-hosts ordering — the ordering window is seconds wide and that contract is already
unit-pinned. Resolving as deferred.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_metrics_forwarder_replicaset.py
anandsyncs: Applied (`9d357924f`): `test_disable_metrics_forwarder` now asserts the retention half of the
contract — after K8s teardown and status cleanup, the metrics topology-state ConfigMap still exists
and the Ops Manager project's mongot hosts remain registered. Verified against the operator first:
the disabled path only sweeps workload resources and never touches the state CM or OM hosts, so the
assertions pin real behavior. The state-CM name is mirrored via a small helper in
`search_resource_names.py` rather than an inline string, and the assertion lives in this one
representative suite per the comment. Resolving.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -912,6 +915,236 @@ def deleted() -> tuple[bool, str]:
...
+
+    run_periodically(
+        reached,
+        timeout=timeout,
+        sleep_time=10,
+        msg=f"metrics forwarder status to reach {phase.name}",
+    )
+
+
+def protected_search_input_uids(
```
viveksinghggits, line 1023: should we have a comment here? I didn't understand why do we have protected here.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
```diff
@@ -912,6 +915,236 @@ def deleted() -> tuple[bool, str]:
...
+        f"Service/{names['svc']}": lambda: mcc.read_namespaced_service(names["svc"], namespace),
+        f"Service/{names['proxy']}": lambda: mcc.read_namespaced_service(names["proxy"], namespace),
+        f"ConfigMap/{names['mongot_cm']}": lambda: mcc.read_namespaced_config_map(names["mongot_cm"], namespace),
+        f"Deployment/{names['envoy_deployment']}": lambda: apps.read_namespaced_deployment(
+            names["envoy_deployment"], namespace
+        ),
+        f"ConfigMap/{names['envoy_cm']}": lambda: mcc.read_namespaced_config_map(names["envoy_cm"], namespace),
+        f"Secret/{names['operator_tls_secret']}": lambda: core.read_namespaced_secret(
+            names["operator_tls_secret"], namespace
+        ),
```
viveksinghggits, line 1089: nit: `namespaced` in the names `read_namespaced_...` id not necessary all of those resources are namespaced.

docker/mongodb-kubernetes-tests/tests/common/search/search_resource_names.py
```diff
@@ -253,3 +276,26 @@ def metrics_forwarder_configmap_name(search_name: str, cluster_index: int = 0) -
...
+    """
+    return f"{search_name}-search-metrics-forwarder-{cluster_index}-agent-key"
+
+
+def metrics_forwarder_ca_configmap_name(search_name: str, cluster_index: int = 0) -> str:
+    """ConfigMap name for the forwarder-owned OM CA replica. Mirrors MetricsForwarderCACertConfigMapNameForCluster().
+
+    cluster_index defaults to 0 for single-cluster callers; pass the cluster
+    position explicitly for multi-cluster tests.
+    """
```
viveksinghggits: these comments don't seem to be needed, we can remove them.

viveksinghggits (approved):

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
anandsyncs: Done — added a docstring explaining "protected": these are customer-provided inputs (source TLS Secret, sync-user Secret, CA ConfigMap) the operator consumes but must never delete or recreate; tests snapshot their UIDs and assert they survive lifecycle transitions unchanged.

docker/mongodb-kubernetes-tests/tests/common/search/connectivity.py
anandsyncs: Kept as-is: these names mirror the official kubernetes Python client API (`CoreV1Api.read_namespaced_secret`, `AppsV1Api.read_namespaced_deployment`) and the pre-existing `MultiClusterClient.read_namespaced_*` helpers in kubetester, so renaming here would break the 1:1 correspondence.

docker/mongodb-kubernetes-tests/tests/common/search/search_resource_names.py
anandsyncs: Done — dropped the repeated `cluster_index` paragraphs from all three forwarder naming helpers; each keeps only its one-line docstring.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_metrics_forwarder_replicaset.py
```diff
@@ -200,13 +237,87 @@ def check_per_cluster_metrics_forwarder_absent():
...
+    finalizers = mdbs["metadata"].get("finalizers") or []
+    assert (
+        METRICS_FORWARDER_FINALIZER in finalizers
+    ), f"finalizer {METRICS_FORWARDER_FINALIZER} must survive a forwarder disable, got {finalizers}"
+
 
 @mark.e2e_search_enterprise_metrics_forwarder_replicaset
 def test_deleteing_search_resource_deletes_hosts(om: MongoDBOpsManager, mdbs: MongoDBSearch):
+    ensure_nested_objects(mdbs, ["spec", "observability", "metricsForwarder"])
+    mdbs["spec"]["observability"]["metricsForwarder"]["mode"] = "enabled"
```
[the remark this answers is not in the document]
anandsyncs: Done — renamed to `test_deleting_search_resource_deletes_hosts`.

docker/mongodb-kubernetes-tests/tests/search/search_enterprise_metrics_forwarder_replicaset.py
```diff
@@ -200,13 +237,87 @@ def check_per_cluster_metrics_forwarder_absent():
...
+
+    deployment_name = metrics_forwarder_deployment_name(MDB_RESOURCE_NAME)
+    configmap_name = metrics_forwarder_configmap_name(MDB_RESOURCE_NAME)
+    agent_key_secret = metrics_forwarder_agent_key_secret_name(MDB_RESOURCE_NAME)
+    ca_configmap = metrics_forwarder_ca_configmap_name(MDB_RESOURCE_NAME)
+    apps = client.AppsV1Api()
+    core = client.CoreV1Api()
+    apps.read_namespaced_deployment(deployment_name, mdbs.namespace)
+    core.read_namespaced_config_map(configmap_name, mdbs.namespace)
+    core.read_namespaced_secret(agent_key_secret, mdbs.namespace)
```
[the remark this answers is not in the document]
anandsyncs: Done — the CA ConfigMap is now read once alongside the other forwarder artifacts before `mdbs.delete()`, so its deletion poll can no longer pass vacuously.

fealebenpae (approved):

PR #1385, opened 2026-07-16
CLOUDP-400891: Remove installClusterRole Helm Value

roothorp:
# Summary

Remove installClusterRole helm value, and also modify a test that checked if it worked to now check the preferred value.

This functionality is now exclusively configured using `operator.webhook.registerConfiguration`

## Checklist

---


josvazg (commented):
**q**: isn't this missing the changelog entry? it is a breaking change on its own AFAIU

.evergreen.yml
```diff
@@ -1952,16 +1942,6 @@ buildvariants:
...
 
-  - name: e2e_operator_no_webhook_roles_cloudqa
-    display_name: e2e_operator_no_webhook_roles_cloudqa
-    tags: [ "pr_patch_e2e", "staging", "e2e_test_suite", "cloudqa", "cloudqa_non_static" ]
-    <<: *production_code_paths
-    run_on:
-      - ubuntu2404-large
-    <<: *base_no_om_image_dependency
-    tasks:
-      - name: e2e_mdb_kind_no_webhook_roles_cloudqa_task_group
```
m1kola, line 1872: Currently in master there are two things that gate webhook RBAC:

```
if and .Values.operator.webhook.registerConfiguration .Values.operator.webhook.installClusterRole
```

We've decided that `operator.webhook.installClusterRole` is redundant, but `operator.webhook.registerConfiguration` still stays in place and still gates RBAC.

IMO we should keep testing that there is a way to not deploy webhook RBAC and keep the operator functional. So the question is: do we have another test that does the same, but using `operator.webhook.registerConfiguration`? If yes - then we can remove it. If no - then we should update the test to use `operator.webhook.registerConfiguration` instead of `operator.webhook.installClusterRole`.

P.S: I think it is an overkill to run a whole E2E test just to test this. IMO it should be covered by Helm chart tests and unit tests (that the operator can handle absent RBAC well). That's an alternative you can explore, but not worth spinning wheels on it too much. Making existing approach work is good enough.

m1kola (changes requested):

.evergreen.yml
roothorp: The removed e2e test solely tested `operator.webhook.installClusterRole` (via `MDB_HELM_OPERATOR_WEBHOOK_INSTALL_CLUSTER_ROLE`), and as far as I understand did not interact with `registerConfiguration`. I don't think there is an e2e test for `registerConfiguration` -  there is an equivalent env var (`MDB_WEBHOOK_REGISTER_CONFIGURATION`) but as far as I can see we don't use that anywhere.

m1kola: @roothorp you are right about these two settings. But the test doesn't necessary test a specific setting. It is more about testing being able to deploy the operator without webhook RBAC: the operator should still function.

So today we can deploy the operator without webhook RBAC by setting `operator.webhook.registerConfiguration` or `operator.webhook.installClusterRole` to `false`. Note that both are set to `true` by default and it is enough to set only one to `false` to disable webhook RBAC.

If we remove this test test task, do we test "I can deploy the operator without webhook RBAC and it still works" in some other way" scenario? If we have no other test like this, then we should keep this test and trigger this scenario via the remaining setting.

Does it make sense?

m1kola (approved):
LGTM, assuming tests are green.

docker/mongodb-kubernetes-tests/kubetester/operator.py
```diff
@@ -209,7 +208,22 @@ def wait_for_operator_ready(self, retries: int = 60):
...
+            return value.lower() != "false"
+        return bool(value)
+
     def wait_for_operator_webhook_ready(self, retries: int = 10, multi_cluster: bool = False):
+        if not self.webhook_registration_enabled():
+            logger.info(
+                "Skipping waiting for the operator webhook: the operator was installed with "
+                "operator.webhook.registerConfiguration=false, so no webhook service is served"
+            )
+            return
```
m1kola, line 227: @roothorp at the first glance it makes sense, but I did not expect test code changes here. 

The fact that it worked before the removal is slightly alarming. I do not see how switching over from two settings to just `operator.webhook.registerConfiguration` could cause us to change tests. My intuition is that tests must still pass assuming that we correctly changed the setup.

Could you please explain?

roothorp: This is my understanding of this test:

The old version of the test ran with `registerConfiguration=true` (default), which means the full webhook setup ran: https://github.com/mongodb/mongodb-kubernetes/blob/29b79cdac5049cbd988fca403cb70a35530a7598/pkg/webhook/setup.go#L196-L198

Then, later in the setup (after creating the service and certs), the creation errors because it doesn't have permissions: https://github.com/mongodb/mongodb-kubernetes/blob/29b79cdac5049cbd988fca403cb70a35530a7598/pkg/webhook/setup.go#L241-L245

This was testing "does the Operator work with missing webhook RBAC".

Now that this test is instead setting `registerConfiguration=false`, the aforementioned conditional is entered, which deletes the webhook and cleans everything up. This is now testing "does the Operator work with webhooks totally disabled".

Before these most recent changes I made, `Operator.install()` always calls `wait_for_operator_webhook_ready()`, which checks `https://operator-webhook.<ns>.svc/`. The old version of this test was creating that endpoint, which responded (even if it was a 403). The new version doesn't create the endpoint, so the operator installation "fails", although the failure is because it didn't install the thing we asked it not to install.

So I agree that this is a change in what we're testing, but I believe it also accurately reflects how we are changing the actual functionality; you can't just opt out of the RBAC anymore, you either opt in or out of the whole webhook.

Am I correct in my assumptions here? Does this match your understanding?

m1kola: @roothorp thanks for the explanation! I think it makes perfect sense and we should go ahead with this change.

pkg/webhook/setup.go
```diff
@@ -194,6 +194,14 @@ func shouldRegisterWebhookConfiguration() bool {
 }
 
 func Setup(ctx context.Context, client client.Client, serviceLocation types.NamespacedName, certDirectory string, webhookPort int, svcSelector string, log *zap.SugaredLogger) error {
+	// The certificates need to exist even when we don't register the ValidatingWebhookConfiguration.
+	// The webhook server is always added as a manager runnable (in main.go) and will creash the whole
+	// operator if it cannot read them from certDirectory.
```
m1kola, line 199: Would it make sense or not to not add the webhook as a runnable when `operator.webhook.registerConfiguration`/`MDB_WEBHOOK_REGISTER_CONFIGURATION` is `false`?

roothorp: I think that probably makes sense. I will look into it now, and see if it is feasible.

roothorp: The [helm values](https://github.com/mongodb/mongodb-kubernetes/blob/cloudp-400891-remove-installclusterrole-helm/helm_chart/values-openshift.yaml#L12-L17C270) suggest that we actually want to keep the functionality as is, most notably:
> The operator will still expose the webhook's endpoint on port on MDB_WEBHOOK_PORT (if not specified, the operator uses a default 1993) in case the ValidatingWebhookConfigurations is configured externally (e.g. in OLM/OpenShift) or by the administrator manually.

So perhaps we want to leave this alone?

m1kola: Ah, good catch! That makes sense for OLM. Thanks for looking into this.

m1kola (approved):

m1kola (approved):

mircea-cosbuc (approved):

PR #1389, opened 2026-07-16
KUBE-141: Remove stale mms.https.PEMKeyFile from Ops Manager config when TLS is disabled

Julien-Ben:
# Summary

Disabling TLS on a running Ops Manager (removing `spec.security`) left the instance unable to start whenever the conf directory outlives the container — most commonly a user PVC mounted over `/mongodb-ops-manager/conf`, which the operator supports as a StatefulSet override (HELP-84463). The operator builds the new StatefulSet correctly, but the `mmsconfiguration` binary in the init-ops-manager image only ever overwrote or appended `OM_PROP_*` properties in `conf-mms.properties`, never deleting one it stopped setting. The stale `mms.https.PEMKeyFile` then points at a certificate secret that is no longer mounted and Ops Manager crash-loops in preflight. With the default emptyDir layout the conf directory is recreated on every rollout, which is why this never reproduced on kind.

Solution: Operator-set properties not already present in the file are now written inside a BEGIN/END-delimited block that is stripped and rewritten on every container start, the same way custom JVM params are already handled in this binary. A property the operator stops setting disappears from the file instead of lingering forever.

**A note for reviewers**: this is admittedly a bit manual — operator-owned properties are tracked with custom BEGIN/END comment markers and rewritten by string manipulation, mirroring the existing JVM-params handling. The approach deserves a careful look; happy to discuss alternatives.

Deployments already broken by an older image have the stale property as a plain line outside any block; those still need the existing workaround (remove the line, or delete the conf PVC). The fix prevents the state from the first start on a fixed image.

## Proof of Work

- New unit test reproducing the stale-property bug (fails without the fix), plus idempotency coverage.
- The `om_ops_manager_https` e2e now runs with a PVC mounted over the conf directory and ends with a disable-TLS stage:
  - without the fix it fails on exactly that stage, with the customer's crash-loop signature in the OM pod log: [patch](https://spruce.corp.mongodb.com/version/6a58e665524bfc0007bc4d08)
  - with the fix the task is green: [patch](https://spruce.corp.mongodb.com/version/6a58eaf3e0b9f0000767f0ed)

## Checklist

### DIFF
--- a/changelog/20260716_fix_ops_manager_stale_properties_after_disabling_tls.md
+++ b/changelog/20260716_fix_ops_manager_stale_properties_after_disabling_tls.md
@@ -1,0 +1,6 @@
+---
+kind: fix
+date: 2026-07-16
+---
+
+* **MongoDBOpsManager**: Fixed Ops Manager failing to start after TLS is disabled when the configuration directory is persisted (for example a PVC mounted over `/mongodb-ops-manager/conf`). Properties the operator no longer sets, such as `mms.https.PEMKeyFile`, are now removed from `conf-mms.properties` on container start instead of lingering forever.
--- a/docker/mongodb-kubernetes-init-ops-manager/mmsconfiguration/edit_mms_configuration.go
+++ b/docker/mongodb-kubernetes-init-ops-manager/mmsconfiguration/edit_mms_configuration.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"sort"
 	"strings"
 
 	"golang.org/x/xerrors"
@@ -25,6 +26,9 @@ const (
 	appDbConnectionStringFilePath = appDbConnectionStringPath + "/connectionString"
 	// keep in sync with MmsMongoUri constant from github.com/mongodb/mongodb-kubernetes/pkg/util
 	appDbUriKey = "mongo.mongoUri"
+
+	omPropertiesBlockBegin = "## BEGIN properties managed by the MongoDB Kubernetes Operator (rewritten on every container start, do not edit)"
+	omPropertiesBlockEnd   = "## END properties managed by the MongoDB Kubernetes Operator"
 )
 
 func updateConfFile(confFile string) error {
@@ -210,6 +214,13 @@ func getOmPropertiesFromEnvVars() map[string]string {
 }
 
 func updateMmsProperties(lines []string, newProperties map[string]string) []string {
+	// Drop the block written by a previous container start first: a property the
+	// operator no longer sets (e.g. mms.https.PEMKeyFile after TLS is disabled)
+	// must disappear from the file rather than linger forever. The conf directory
+	// can survive container recreation (in-place crash restarts, or a user PVC
+	// mounted over it), so this file cannot be assumed to be freshly templated.
+	lines = removeOperatorPropertiesBlocks(lines)
+
 	seenProperties := map[string]bool{}
 
 	// Overwrite existing properties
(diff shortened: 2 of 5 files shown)


docker/mongodb-kubernetes-init-ops-manager/mmsconfiguration/edit_mms_configuration.go
```diff
@@ -225,13 +236,43 @@ func updateMmsProperties(lines []string, newProperties map[string]string) []stri
 		}
 	}
 
-	// Add new properties
-	for key, val := range newProperties {
-		if _, ok := seenProperties[key]; !ok {
-			lines = append(lines, fmt.Sprintf("%s=%s", key, val))
+	// Add the remaining properties inside a delimited block so the next container
+	// start can strip and rewrite them.
+	var blockKeys []string
```
viveksinghggits, line 241:
```suggestion
	var keysToAppend []string
```

changelog/20260716_fix_ops_manager_stale_properties_after_disabling_tls.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2026-07-16
+---
+
+* **MongoDBOpsManager**: Fixed Ops Manager failing to start after TLS is disabled when the configuration directory is persisted (for example a PVC mounted over `/mongodb-ops-manager/conf`). Properties the operator no longer sets, such as `mms.https.PEMKeyFile`, are now removed from `conf-mms.properties` on container start instead of lingering forever.
```
viveksinghggits, line 6:
```suggestion
* **MongoDBOpsManager**: Fixed the issue where Ops Manager failed to start after disabling the TLS when the configuration directory is persisted (for example a PVC mounted over `/mongodb-ops-manager/conf`).
```

viveksinghggits (approved):

MaciejKaras (commented):
@Julien-Ben I was asking myself why cannot we completely replace the mms properties file instead? This file contains two types of entries: user and MCK owned. We have the full control over this file and could just combine all entries together and replace the file. 

Is there any other process that writes to this file? I might be missing something here, let me know if my understanding is correct.

viveksinghggits:
Hi @MaciejKaras,
> This file contains two types of entries: user and MCK owned.

This doesn't seem entirely true. According to me there are two types of entries 
- ones that are already part of the OM image (OM managed), it has extensive comments as well, example below
- other ones are the entries that are managed by operator either via CR spec or consts

Now the thing is, if we completely replace the entire file we will loose all the comments and the entries that are already part of the properties file in the container image itself and not necessarily managed by MCK.

[Here](https://gist.github.com/viveksinghggits/43bac7eac0b7fb6ee64ad96070c53608) is an example of the properties file.

MaciejKaras:
@viveksinghggits thanks for clarification.

> ones that are already part of the OM image (OM managed), it has extensive comments as well, example below

This is even more problematic, because if we upgrade the OM version, the mms config that is part of the new image, will not be used, right? We will still keep the mms config file from the previous OM version (due to PVC mount), that might not have correct default settings for the new OM version.

I was thinking that maybe we could keep the OM managed, bundled mms conf file as a template and not modify it at all. Then we would read this base OM conf file, add MCK and user settings, merge everything under a new file that is passed to OM? With this approach there is no appending involved and we always can produce the correct configuration from scratch. wdyt @Julien-Ben @viveksinghggits ?

viveksinghggits:
Hi @MaciejKaras,
That's a great point, upgrade would be problematic if the defaults are changed. And it looks like a bigger change that what's done in the PR. I thin, we should wait for @Julien-Ben to get back to continue this change. He is on leave right now.

Julien-Ben:
Hi @MaciejKaras and @viveksinghggits 
These are very good comments. I would say that what you point out Maciej is indeed an issue, but I should merge that one (once I address the commit) and I/you open a follow up ticket.

WDYT ?

MaciejKaras:
> Hi @MaciejKaras and @viveksinghggits These are very good comments. I would say that what you point out Maciej is indeed an issue, but I should merge that one (once I address the commit) and I/you open a follow up ticket.
> 
> WDYT ?

@Julien-Ben I'm worried that we are going to merge this complex logic into the codebase and it will stay that way. Can you try to investigate the effort required to build solution based on templating mechanism? Why I'm pushing with this solution is that it is free of errors, simple and will be much easier to maintain (zero effort) than the one provided in the PR.

PR #1393, opened 2026-07-17
KUBE-181: Validate MongoDBSearch GA operator upgrades

anandsyncs:
# Summary

Validates the pre-GA→current operator **upgrade path for MongoDBSearch** — GA-created artifacts are adopted in place (same kind/name/UID) with the controller owner reference and `component` label stamped post-upgrade — plus a **full 3-cluster hub-and-spoke sharded lifecycle e2e on self-hosted OM** (new `e2e_multi_cluster_om80_search` variant). Re-enables `e2e_operator_upgrade_search` in both operator task groups. Stacked on #1383.

- **Upgrade suite** (`tests/upgrades/operator_upgrade_search.py`): pre-upgrade artifact/UID capture from the GA operator (GA stamps no owner labels — their post-upgrade appearance is the adoption marker; hence name-prefix listing pre-upgrade), post-upgrade UID stability, and GC-with-operator-scaled-down proving owner-reference locality.
- **OM lifecycle module** (`tests/multicluster_search/mc_sharded_om_lifecycle.py`): metrics forwarder on self-hosted OM incl. OM host deregistration + finalizer release; live cluster removal; all six mongot `setParameter` keys drained at unwiring, asserted directly in the automation config (routing keys removed via AC edit, base keys via spec). Forwarder teardown asserts the three kinds that exist on plain-HTTP OM (the CA ConfigMap is never created there).

## Proof of Work

- **Head `ce11184a5`** = the 4 commits rebased byte-identical onto #1383's `eb6da7b25` (range-diff all `=`) + one lint-fix commit (`ty` reader-callable typing). **PR patch [`6a887993`](https://spruce.corp.mongodb.com/version/6a887993eaddf70007d699bb) fully green** — incl. the new `e2e_multi_cluster_om80_search` variant and the re-enabled upgrade task; the single red was an unrelated kind SA-token bootstrap flake, green on restart. The static-operator variant isn't tagged `pr_patch_e2e`, so its `e2e_operator_upgrade_search` run was proven via surgical patch [`6a888bbe`](https://spruce.corp.mongodb.com/version/6a888bbe1688c400073a9c1a) — all green.
- Earlier lineage: 69/69 full matrix incl. both upgrade-task wirings ([`6a67d2fb`](https://spruce.corp.mongodb.com/version/6a67d2fb23f20c000767e00e)); the lifecycle module has five consecutive green runs (first full pass [`6a68e413`](https://spruce.corp.mongodb.com/version/6a68e4136d2d540007a68bb4)).

## Checklist

---

### DIFF
--- a/docker/mongodb-kubernetes-tests/tests/multicluster_search/mc_sharded_om_lifecycle.py
+++ b/docker/mongodb-kubernetes-tests/tests/multicluster_search/mc_sharded_om_lifecycle.py
@@ -26,7 +26,7 @@ the customer unwires them — spec keys via the operator, routing keys via the A
 
 import json
 from copy import deepcopy
-from typing import Callable, Dict, List
+from typing import Any, Callable, Dict, List
 
 import kubernetes
 import pymongo.errors
@@ -288,9 +288,7 @@ def _shard_sts_names(cluster_index: int, shard_names: List[str]) -> List[str]:
     ]
 
 
-def _shard_artifact_readers(
-    mcc: MultiClusterClient, namespace: str, shard_name: str
-) -> Dict[str, Callable[[], object]]:
+def _shard_artifact_readers(mcc: MultiClusterClient, namespace: str, shard_name: str) -> Dict[str, Callable[[], Any]]:
     ci = _idx(mcc)
     core = mcc.core_v1_api()
     sts = search_resource_names.shard_statefulset_name(MDBS_RESOURCE_NAME, shard_name, ci)
@@ -309,13 +307,13 @@ def _shard_artifact_readers(
 
 def _local_artifact_readers(
     mcc: MultiClusterClient, namespace: str, shard_names: List[str]
-) -> Dict[str, Callable[[], object]]:
+) -> Dict[str, Callable[[], Any]]:
     """Direct readers for the concrete (kind, name) identities of one cluster's Search
     artifact set: per-(cluster, shard) mongot resources plus the cluster-level proxy
     Service and Envoy Deployment/ConfigMap."""
     ci = _idx(mcc)
     apps = mcc.apps_v1_api()
-    readers: Dict[str, Callable[[], object]] = {}
+    readers: Dict[str, Callable[[], Any]] = {}
     for shard_name in shard_names:
         readers.update(_shard_artifact_readers(mcc, namespace, shard_name))
     cluster_proxy = search_resource_names.mc_proxy_svc_name(MDBS_RESOURCE_NAME, ci)
@@ -327,7 +325,7 @@ def _local_artifact_readers(
     return readers
 
 
-def _reader_uids(readers: Dict[str, Callable[[], object]]) -> Dict[str, str]:
+def _reader_uids(readers: Dict[str, Callable[[], Any]]) -> Dict[str, str]:
     """UIDs of every captured artifact; reading them doubles as a presence guard."""
     return {what: read().metadata.uid for what, read in readers.items()}
 
@@ -359,7 +357,7 @@ def _customer_input_uids(mcc: MultiClusterClient, namespace: str, cluster_index:
 
 
 def _wait_for_cluster_artifacts_swept(
-    mcc: MultiClusterClient, namespace: str, readers: Dict[str, Callable[[], object]], shard_names: List[str]
+    mcc: MultiClusterClient, namespace: str, readers: Dict[str, Callable[[], Any]], shard_names: List[str]
 ) -> None:
     """Identity 404-polls for every captured artifact, then PVC reap, then the
     label-scoped emptiness backstop for labeled orphans outside the inventory."""
@@ -997,7 +995,7 @@ def test_remove_shard_lifecycle(
     removed_shard = _shard_names(INITIAL_SHARD_COUNT)[-1]
 
     live_uids: Dict[str, Dict[str, str]] = {}
-    removed_readers: Dict[str, Dict[str, Callable[[], object]]] = {}
+    removed_readers: Dict[str, Dict[str, Callable[[], Any]]] = {}
     for mcc in member_cluster_clients:
         live_uids[mcc.cluster_name] = _reader_uids(_local_artifact_readers(mcc, namespace, live_shards))
         readers = _shard_artifact_readers(mcc, namespace, removed_shard)
--- a/docker/mongodb-kubernetes-tests/tests/multicluster_search/mc_sharded_om_lifecycle.py
+++ b/docker/mongodb-kubernetes-tests/tests/multicluster_search/mc_sharded_om_lifecycle.py
@@ -244,7 +244,7 @@ def _assert_lb_cert_proxy_sans(
         )
         missing = expected_sans - sans
         assert not missing, f"LB server cert {secret_name} missing proxy SANs: {sorted(missing)}"
-        logger.info(f"LB server cert {secret_name} covers all {len(expected_sans)} proxy SANs")
+        logger.info(f"LB server cert for cluster {ci} covers all {len(expected_sans)} proxy SANs")
 
 
 def _source_router_hosts(namespace: str) -> List[str]:
--- a/docker/mongodb-kubernetes-tests/tests/multicluster_search/mc_sharded_om_lifecycle.py
+++ b/docker/mongodb-kubernetes-tests/tests/multicluster_search/mc_sharded_om_lifecycle.py
@@ -1,0 +1,1261 @@
+"""Full MongoDBSearch lifecycle on a 3-cluster hub-and-spoke sharded source with self-hosted Ops Manager.
+
+Hub topology: central MC operator on cluster-1 managing clusters 1-3, self-hosted
+Ops Manager 8.0.25 with a multi-cluster AppDB (the metrics forwarder requires
+OM >= 8.0.25, so this cannot run on cloud-qa), a 3-shard TLS+SCRAM MultiCluster
+sharded MongoDB source, and a MongoDBSearch with three named cluster entries,
+managed LB, and the metrics forwarder pointed at the source's OM project.
+
+Routing is per-cluster: each cluster's shard mongods target their own cluster's
+per-shard proxy Services and each cluster's mongos its own cluster's router proxy.
+Spec setParameter is per-shard and uniform across clusters, so the two routing keys
+are wired per process directly in the OM automation config (customer-owned,
+external-source semantics); only the four uniform base keys ride the spec.
+
+Lifecycle under test, in order: baseline per-(cluster, shard) fan-out with exact
+OM MONGOT host registration and the per-cluster routing observed on every pod,
+the $search data plane through every cluster's mongos, trailing-shard removal
+(search shard list first, then the source; survivors keep cluster-local routing),
+cluster-entry removal (cluster-3 fully swept, its still-running source processes
+rewired to a survivor's proxies, while both survivors stay byte-identical and
+serving), and the CR-delete tail (OM host deregistration, forwarder teardown,
+finalizer release, label sweeps on all three clusters, and the search
+setParameters draining out of the still-Running source's automation config once
+the customer unwires them — spec keys via the operator, routing keys via the AC).
+"""
+
+import json
+from copy import deepcopy
+from typing import Callable, Dict, List
+
+import kubernetes
+import pymongo.errors
+import yaml
+from cryptography import x509
+from kubernetes.client import CoreV1Api
+from kubetester import create_or_update_secret, read_secret, try_load
+from kubetester.certs import create_tls_certs
+from kubetester.kubetester import fixture as yaml_fixture
+from kubetester.kubetester import run_periodically
+from kubetester.mongodb import MongoDB
+from kubetester.mongodb_search import MongoDBSearch
+from kubetester.mongodb_user import MongoDBUser
+from kubetester.multicluster_client import MultiClusterClient
+from kubetester.operator import Operator
+from kubetester.opsmanager import MongoDBOpsManager
+from kubetester.phase import Phase
+from pytest import fixture, mark
+from tests import test_logger
+from tests.common.multicluster.multicluster_utils import (
+    assert_deployment_ready_in_cluster,
+    assert_workload_ready_in_cluster,
+)
+from tests.common.search import search_resource_names
+from tests.common.search.connectivity import (
+    CLUSTER_LOCATION_TAG_KEY,
+    mongot_data_pvc_names,
+    protected_search_input_uids,
+    wait_for_metrics_forwarder_phase,
+    wait_for_mongot_pvcs_deleted,
+    wait_for_resource_deleted,
+    wait_for_search_deleted,
+    wait_for_search_owned_resources_deleted,
+)
+from tests.common.search.mc_search_helper import (
+    assert_sharded_mongot_host_observed,
+    patch_per_cluster_sharded_mongot_host_via_om,
+    remove_mongot_host_via_ac,
+)
+from tests.common.search.movies_search_helper import SampleMoviesSearchHelper
+from tests.common.search.search_tester import SearchTester
+from tests.common.search.sharded_search_helper import (
+    create_issuer_ca,
+    create_lb_certificates,
+    create_per_shard_search_tls_certs,
+)
+from tests.conftest import get_issuer_ca_filepath
+from tests.multicluster.conftest import cluster_spec_list
+from tests.multicluster_search.conftest import (
+    ADMIN_USER_NAME,
+    ADMIN_USER_PASSWORD,
+    ENVOY_PROXY_PORT,
+    MDBS_TLS_CERT_PREFIX,
+    MONGOT_USER_NAME,
+    SOURCE_CERT_PREFIX,
+    USER_NAME,
+    USER_PASSWORD,
+    create_search_users,
+    install_central_mc_operator,
+)
+from tests.search.om_deployment import get_ops_manager
+
+logger = test_logger.get_test_logger(__name__)
+
+MDB_RESOURCE_NAME = "mdb-mc-sh-om"
+MDBS_RESOURCE_NAME = "mdb-mc-sh-om-search"
+
+INITIAL_SHARD_COUNT = 3
+REDUCED_SHARD_COUNT = 2
+
+MEMBERS_PER_CLUSTER: List[int | None] = [1, 1, 1]
+MONGOS_PER_CLUSTER: List[int | None] = [1, 1, 1]
+CONFIG_SRV_PER_CLUSTER: List[int | None] = [1, 1, 1]
+
+CLUSTER_COUNT = len(MEMBERS_PER_CLUSTER)
+MONGOT_REPLICAS_PER_CLUSTER = 1
+# The trailing entry is removed; its source processes keep running and get rewired
+# (direct AC) to the REWIRE_TARGET_INDEX survivor's proxies for the same shard.
+REMOVED_CLUSTER_INDEX = 2
+REWIRE_TARGET_INDEX = 0
+
+# 9 mongot JVMs share one kind host with OM's JVM and 15 source mongods; the default
+# mongot request (cpu=2, memory=4Gi => 2Gi heap) does not fit, and a 1Gi heap serves
+# sample_mflix fine.
+MONGOT_RESOURCE_REQUESTS = {"cpu": "500m", "memory": "2Gi"}
+
+CA_CONFIGMAP_NAME = f"{MDB_RESOURCE_NAME}-ca"
+
+# setParameter keys wiring the source to mongot. On an external source the operator
+# never writes them: the four base keys are customer-set spec (additionalMongodConfig,
+# uniform across shards and clusters) and the two routing keys are set per process
+# directly in the OM automation config, the only layer that can express per-cluster
+# endpoints. The customer unwires both halves on Search delete (see the delete test).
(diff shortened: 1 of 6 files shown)


docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -130,8 +158,8 @@ def sample_movies_helper(mdb: MongoDB, namespace: str) -> SampleMoviesSearchHelp
 @mark.e2e_operator_upgrade_search
 class TestDeployOnOfficialOperator:
 
-    def test_install_latest_official_operator(self, namespace: str, official_operator: Operator):
-        official_operator.wait_for_operator_ready()
+    def test_install_ga_operator(self, namespace: str, ga_operator: Operator):
```
anandsyncs, line 331: The 1.9.0 pin is an input, not an asserted invariant: nothing checks the running operator is actually 1.9.0 after install. If `GA_OPERATOR_VERSION` silently floated to another valid version (or the param were dropped), every later check still passes because all version assertions are relative. Suggest asserting the operator Deployment's image tag `== GA_OPERATOR_VERSION` right after `wait_for_operator_ready()`.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -202,7 +230,7 @@ def test_verify_mongot_version_after_upgrade(self, namespace: str, operator_inst
         assert_mongot_version_matches_operator(namespace, operator, "post-upgrade")
```
anandsyncs: Strict `==`, but both sides come from the *same* operator (its `MDB_SEARCH_VERSION` env vs its own mongot image) — this validates internal consistency, so a no-op 'upgrade' (same image both sides) passes. Capture `MDB_SEARCH_VERSION` pre-upgrade and assert post-upgrade it differs (or pre == the 1.9.0-expected tag). Related, pre-existing: `get_mongot_image_tag` (~line 48) swallows all exceptions with `except Exception: continue`, which can mask a transient API error as 'not found' — worth tightening since it's the oracle these assertions rest on.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Addressed. The supported baseline is now pinned to MCK `1.9.1`, and immediately after readiness the test asserts the running operator container image is exactly `quay.io/mongodb/mongodb-kubernetes-operator:1.9.1`. The exact 1.9.1 to 1.10.0 upgrade task passed in authoritative patch `6a5fff820b0a4b00074e8d5d`.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Addressed. The test records the pre-upgrade `MDB_SEARCH_VERSION` and mongot pod UID, asserts the target operator image differs from the GA image, requires the post-upgrade Search version to differ from the recorded value, and waits for a new ready mongot pod with the target tag. `get_mongot_image_tag` now treats only a genuine 404 as absence and propagates other API errors. The authoritative Layer-4 matrix passed 69/69 on execution 0.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -30,48 +50,346 @@
 USER_NAME = "mdb-user"
 USER_PASSWORD = f"{USER_NAME}-password"
 
+MONGOT_STATEFULSET_NAME = mongot_statefulset_name(MDB_RESOURCE_NAME)
+MONGOT_SERVICE_NAME = mongot_service_name(MDB_RESOURCE_NAME)
+MONGOT_CONFIGMAP_NAME = mongot_configmap_name(MDB_RESOURCE_NAME)
+PROXY_SERVICE_NAME = proxy_service_name(MDB_RESOURCE_NAME)
+
+# Pre-1.9 unindexed names; neither the GA operator nor the upgraded one may create these.
+UNINDEXED_MONGOT_STATEFULSET_NAME = f"{MDB_RESOURCE_NAME}-search"
```
anandsyncs: Please remove the pre-1.9 unindexed-name constants and assert_unindexed_artifacts_absent helper. This upgrade is pinned to 1.9.1, which already uses indexed names, and multi-cluster Search has no legacy named-cluster migration to validate.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -30,48 +50,346 @@
...
+    result = {
+        f"{kind}/{resource.metadata.name}": resource.metadata.uid
+        for kind, resources in list_search_artifacts(namespace).items()
+        for resource in resources
+    }
+    assert all(result.values()), f"Search artifacts missing UIDs: {result}"
+    return result
+
+
+def source_artifact_specs(namespace: str) -> dict[str, Any]:
```
anandsyncs: Please remove this full artifact-spec snapshot and its equality assertion. Stable artifact UIDs, PVC UID preservation, mongot pod rollout, and the successful query already prove adoption without recreation. Exact ClusterIP/selectors/ConfigMap key equality is not an upgrade contract and adds a large helper surface.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Removed the `UNINDEXED_*` constants and `assert_unindexed_artifacts_absent` with its only call site - the upgrade is pinned to 1.9.1, so pre-1.9 naming never occurs.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Removed `source_artifact_specs`, the `UpgradeState.artifact_specs` snapshot, the equality assertion, and the snapshot-derived ClusterIP check. Kept: artifact/PVC UID stability (the churn proof), the two-sided owner-reference assertions, mongot pod rollout, the post-upgrade search query, and live selector/port wiring checks.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -30,48 +50,293 @@
...
+    assert owner.kind == "MongoDBSearch"
+    assert owner.name == MDB_RESOURCE_NAME
+    assert owner.uid == search_uid
+    if expect_controller:
+        assert owner.controller is True
+    else:
+        assert owner.controller is not True
+
+
+def assert_single_cluster_artifact_set(
```
anandsyncs: Even after dropping exact specs, this remains a full artifact-shape contract: exact names plus every label and owner ref on every kind, both before and after upgrade. `search_owned_artifact_uids` already proves no recreation, and the operator-down cleanup later proves owner refs work across all top-level artifacts. Keep one representative pre/post owner-ref transition (for example the StatefulSet), then delete `expected_single_cluster_artifact_names` and this bulk assertion helper. That retains the upgrade contract while removing roughly 50-70 lines.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -216,9 +537,30 @@ def test_enable_multi_mongot_and_managed_lb(self, mdbs: MongoDBSearch):
         mdbs["spec"]["clusters"] = [{"replicas": 2, "loadBalancer": {"managed": {}}}]
         mdbs.update()
 
-    def test_search_running_after_scale(self, mdbs: MongoDBSearch):
+    def test_search_running_after_scale(self, namespace: str, mdbs: MongoDBSearch):
         mdbs.assert_reaches_phase(phase=Phase.Running, timeout=600)
 
+        def mongot_scaled() -> tuple[bool, str]:
```
anandsyncs: The base test already waited for Search Running after scale. The added per-counter StatefulSet readiness loop, pod reads, exact proxy selector/port checks, second query, and another full artifact-set assertion create a second convergence matrix unrelated to 1.9.1 adoption. Revert this class to the preexisting phase/LB/Envoy checks; the upgrade section already proves rollout and query. This removes roughly 35-45 added lines.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -231,15 +573,116 @@ def check_envoy_ready():
...
+                namespace,
+                label_selector=f"app.kubernetes.io/name={OPERATOR_NAME}",
+            ).items
+            done = spec_replicas == current_replicas == ready_replicas == 0 and not operator_pods
+            pod_names = [pod.metadata.name for pod in operator_pods]
+            return done, f"spec={spec_replicas}, current={current_replicas}, ready={ready_replicas}, pods={pod_names}"
+
+        run_periodically(scaled_down, timeout=120, sleep_time=5, msg=f"Operator Deployment {OPERATOR_NAME} scale-down")
+
+    def test_delete_search_resource(self, mdbs: MongoDBSearch):
```
anandsyncs, line 557: Reuse the lifecycle helpers introduced in #1383 here: `wait_for_search_deleted`, `wait_for_search_owned_resources_deleted`, and the existing PVC wait. The local `search_deleted`, all-artifacts, and mongot-pod polling closures duplicate those helpers; PVC deletion already proves the pod no longer mounts the volume. Keep the operator-down step and customer-input UID assertion, but delete the custom polling scaffolding.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -192,18 +483,48 @@ def test_upgrade_operator(self, namespace: str, operator_installation_config: di
...
+    def test_adopts_existing_single_cluster_artifacts(
+        self,
+        namespace: str,
+        mdbs: MongoDBSearch,
+        upgrade_state: UpgradeState,
+    ):
+        assert_single_cluster_artifact_set(namespace, mdbs)
+        mdbs.load()
+        assert mdbs["metadata"]["uid"] == upgrade_state.search_uid
+        assert mdbs["spec"]["clusters"] == [{}], "upgrade must not project the Search to member clusters"
```
anandsyncs: Please remove both `spec.clusters == [{}]` assertions (here and in the pre-upgrade snapshot). Preserving this legacy representation is not part of the supported upgrade contract; UID/PVC continuity, rollout, query, customer-input, and operator-down GC checks do not depend on it.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Done — the full artifact-shape comparison helper is deleted; the retained adoption probe is a single StatefulSet owner-reference transition (GA's non-controller reference and no component label before the upgrade; a controller reference plus component=mongot after).

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Done — reverted wholesale to the pre-stack body.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Done — the three bespoke deletion-polling closures now reuse #1383's lifecycle helpers (wait_for_search_deleted / wait_for_search_owned_resources_deleted plus the existing PVC wait).

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Done — both asserts removed.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -30,48 +54,234 @@
...
+
+def expected_current_operator_image(operator_installation_config: dict[str, str]) -> str:
+    registry = operator_installation_config["registry.operator"].rstrip("/")
+    version = operator_installation_config["operator.version"]
+    build = operator_installation_config.get("operator.build", "")
+    return f"{registry}/{OFFICIAL_OPERATOR_IMAGE_NAME}:{version}{build}"
 
-def get_operator_search_version(namespace: str, operator: Operator) -> str:
+
+def assert_operator_image(operator: Operator, expected_image: str) -> None:
```
anandsyncs: This verifies only the Deployment template, not the operator pod that actually became ready. Reuse the pod lookup already used for MDB_SEARCH_VERSION and assert the selected ready pod container image for both GA and target operators.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -192,18 +432,52 @@ def test_upgrade_operator(self, namespace: str, operator_installation_config: di
...
+        assert mdbs["metadata"]["uid"] == upgrade_state.search_uid
+        sts = get_statefulset(namespace, MONGOT_STATEFULSET_NAME)
+        assert_search_owner_reference(sts, upgrade_state.search_uid, expect_controller=True)
+        labels = sts.metadata.labels or {}
+        assert (
+            labels.get("component") == "mongot"
+        ), f"upgraded operator must stamp component=mongot on the adopted StatefulSet: {labels}"
+
+        pod, pvc = read_mongot_pod_and_pvc(namespace)
+        assert search_owned_artifact_uids(namespace) == upgrade_state.artifact_uids
```
anandsyncs: Exact map equality rejects legitimate artifacts introduced by the target operator. Assert that every pre-upgrade kind/name still exists with the same UID, while allowing additional target-version artifacts.

anandsyncs:
Blocking coverage gap: e2e_operator_upgrade_search exists in .evergreen-tasks.yml but remains commented out in both operator task groups in .evergreen.yml, so normal PR/mainline builds never run this test. Because the stack is intentionally moving user-visible continuity coverage into E2E, this task must be scheduled before the coverage can be treated as persistent. The current exact-head patch ran zero upgrade-task executions.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Done — both operator-version assertions now resolve the actually-ready operator pod (a shared lookup that asserts the container status is ready — the same one MDB_SEARCH_VERSION uses) and check its container image, for the GA 1.9.1 install and the upgraded operator alike.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
anandsyncs: Done — the post-upgrade check now asserts every pre-upgrade (kind, name) still exists with the same UID (the adoption/no-churn contract) and tolerates additional target-version artifacts; on mismatch it reports the missing and uid-changed sets plus the full post-upgrade inventory.

anandsyncs:
Round-5: wired — `e2e_operator_upgrade_search` is re-enabled in BOTH task groups (the stale KUBE-102
TODOs are removed), so the suite now runs in PR and mainline builds. The matrix at this head runs the
canonical 68 tasks plus the first static-variant execution of the suite
(`e2e_static_operator_kind_ubi_cloudqa`). The ready-pod image and same-UID-superset assertions from
the two threads landed as well.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -192,18 +438,64 @@ def test_upgrade_operator(self, namespace: str, operator_installation_config: di
             namespace, operator_installation_config=operator_installation_config, apply_crds_first=True
         )
         operator.wait_for_operator_ready()
+        assert get_current_operator_release_version() == TARGET_OPERATOR_VERSION
```
anandsyncs: Remove the `release.json` assertion and its helper/imports. The test already waits for the operator and asserts the ready operator container image equals the requested target image; that is the deployed behavior we need. Reading repository release metadata adds a second oracle and environment dependency without proving anything about the running upgrade.

anandsyncs: Applied (`4352719aa`): the `release.json` assertion, its helper, the constant, and the now-unused
json/os/Path imports are removed (−15 lines). The ready-pod container-image assertion — which checks
what is actually deployed — stays as the single oracle. The OLM commons module keeps its own
independent copy, so nothing else is affected. Resolving.

docker/mongodb-kubernetes-tests/tests/multicluster_search/mc_sharded_om_lifecycle.py
```diff
@@ -0,0 +1,1261 @@
...
+    for mcc in member_cluster_clients:
+        ci = _idx(mcc)
+        _assert_mongos_mongot_host(namespace, ci, target(ci))
+
+
+def _assert_lb_cert_proxy_sans(
+    namespace: str, central_cluster_client: kubernetes.client.ApiClient, cluster_indexes: List[int]
+) -> None:
+    """Every cluster's Envoy server cert must SAN every cluster's per-shard proxy and
+    router proxy FQDNs: with per-cluster routing each cluster's mongods/mongos dial their
```
viveksinghggits: `server cert must SAN every cluster's per-shard proxy` this sentence sounds wrong. Should this say `cert must have SAN`?

viveksinghggits: after reading rest of the code it loos like we are loosely using SAN as verb, which seems fine.

viveksinghggits (commented):
Just skimmed over, looks ok.

viveksinghggits (approved):

docker/mongodb-kubernetes-tests/tests/multicluster_search/mc_sharded_om_lifecycle.py
anandsyncs: Reworded to "must carry SAN entries for every cluster's per-shard proxy and router proxy FQDNs" — no more SAN-as-a-verb.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -192,18 +424,63 @@ def test_upgrade_operator(self, namespace: str, operator_installation_config: di
...
+        upgrade_state: UpgradeState,
+    ):
+        operator = Operator(namespace=namespace, helm_args=operator_installation_config)
+        expected = operator_installation_config["search.version"]
+        assert expected != upgrade_state.pre_mongot_version, f"upgrade left MDB_SEARCH_VERSION unchanged at {expected}"
+        operator_version = get_operator_search_version(operator)
+        assert operator_version == expected, f"operator MDB_SEARCH_VERSION {operator_version} != {expected}"
+        wait_for_mongot_rollout(namespace, expected, upgrade_state.mongot_pod_uid)
+        assert get_mongot_image_tag(namespace) != upgrade_state.pre_mongot_version
+
```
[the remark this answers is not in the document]
anandsyncs: Done — the test now always asserts MDB_SEARCH_VERSION and the mongot image tag equal the expected target, but only requires a pod rollout/UID change when `search.version` actually differs from the recorded GA mongot version (it currently does not: both are 1.70.1). Applies to both the rollout wait and the adoption-test pod UID assert.

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -30,48 +50,230 @@
...
+    def search_named(resources: list[Any]) -> list[Any]:
+        return [
+            resource
+            for resource in resources
+            if resource.metadata.name == name_prefix or resource.metadata.name.startswith(f"{name_prefix}-")
+        ]
+
+    return {
+        "StatefulSets": search_named(apps.list_namespaced_stateful_set(namespace).items),
```
Julien-Ben, line 168: Is there anything in the python kubernetes client to perform a search with filters (e.g by name), instead of listing everything first and iterating over later ?

docker/mongodb-kubernetes-tests/tests/upgrades/operator_upgrade_search.py
```diff
@@ -30,48 +50,230 @@
...
 # --- Module-scoped fixtures (persist across upgrade) ---
 
 
+@fixture(scope="module")
+def upgrade_state() -> UpgradeState:
+    return UpgradeState()
+
+
+@fixture(scope="module")
+def ga_operator(
```
Julien-Ben, line 253: Do we need a separate function, if we are just passing arguments to another function directly ?

Julien-Ben (approved):
Approving with two nits.

filipcirtog (approved):

PR #1390, opened 2026-07-16
CLOUDP-423293: Add multicluster generate-member-registration plugin command

m1kola:
# Summary

Adds the `kubectl mongodb multicluster generate-member-registration` command: it connects to one member cluster (via a kubeconfig context), reads the ServiceAccount token that `generate-member-resources` created on it, and emits a **credential Secret** (a single-context kubeconfig) and a **`MemberCluster` CR** referencing it to stdout — for `kubectl apply` or GitOps. Purely additive; the operator stays inert to `MemberCluster` CRs until a later slice.

Also folds in the RBAC-unification-slice note (previously #1388, now closed).

Stacked on #1381.

## Proof of Work

`go test ./pkg/kubectl-mongodb/... ./pkg/resourcenames/...` passes and `make precommit-full` is clean. `kubectl mongodb multicluster generate-member-registration --help` renders the flags/usage.

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title? (CLOUDP-423293)
- [x] Have you checked whether your jira ticket required DOCSP changes? (none needed)
- [x] Have you added changelog file? (`changelog/20260716_feature_generate_member_registration.md`)

---

PR #1394, opened 2026-07-17
CLOUDP-400899: Wire operator MemberCluster discovery + watch

m1kola:
# Summary

Slice 3 of the multi-cluster installation-UX epic. The operator now discovers member clusters by reading `MemberCluster` CRs and their per-cluster credential Secrets from its own namespace (keyed by `spec.clusterName`), instead of the MCK 1.x `mongodb-kubernetes-operator-member-list` ConfigMap + monolithic mounted kubeconfig. The legacy path is kept as a fallback (used when no `MemberCluster` CRs exist) so existing `setup`-driven E2E stay green; it's removed in a later slice.

A `MemberCluster` watcher (mirroring the `OperatorConfig` watcher) restarts the operator on CR add/spec-change/delete. No-restart reactivity is deferred to a later slice — see the decision + `multicluster-runtime` evaluation note in `docs/dev/multi-cluster-config-tooling.md`. The base role gains `get/list/watch` on `memberclusters`.

Interim limitations (tagged `TODO(m1kola): slice-3`): in-place credential rotation needs an operator restart; the failover health-checker's file-based cache doesn't apply on the CR path (guarded).

## Proof of Work

- `go build ./...`, `go vet`, `gofmt` clean.
- Unit tests: `pkg/membercluster` (discovery + watcher) and `pkg/kubectl-mongodb/common` pass.

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title? (CLOUDP-400899)
- [x] Have you checked whether your jira ticket required DOCSP changes? (not required)

---

Stacked on top of `iux-multi-cluster-generate-member-registration` (#1390).


pkg/membercluster/discovery.go
```diff
@@ -0,0 +1,82 @@
...
+// It matches the key written by the `generate-member-registration` plugin command.
+const credentialSecretKubeconfigKey = "kubeconfig"
+
+// Discover builds a map of member-cluster REST configs from the MemberCluster CRs in the given
+// namespace. Each CR references a credential Secret whose "kubeconfig" key holds a
+// single-context kubeconfig. The returned map is keyed by spec.clusterName — the logical name
+// the operator uses to resolve clusterSpecList[].clusterName references in workload CRs.
+//
+// The bool return reports whether any MemberCluster CRs were found. false means the caller
+// should fall back to the legacy discovery path (member-list ConfigMap + mounted kubeconfig).
```
m1kola, line 32: Note: this is a transient and will go away once we clean up the legacy path.

pkg/membercluster/discovery.go
```diff
@@ -0,0 +1,82 @@
...
+//
+// A per-cluster failure (missing or unparseable credential Secret) is logged and that cluster
+// is skipped, so one broken cluster does not prevent the operator from managing the others.
+// This mirrors the "don't panic, continue" behaviour of the legacy path in main.go.
```
m1kola, line 36: Note: this will change slightly. At the later stages we will have `RBACValid` condition which will be indicating that a `MemberCluster` CR is not valid. But we will still be skipping clusters with invalid CRs as per design.

pkg/membercluster/watcher.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Watcher watches the MemberCluster CRs scoped by the manager cache (see ByObject in main.go)
+// and initiates a graceful shutdown when a CR is created, its spec changes, or it is deleted,
+// so the operator restarts and rebuilds its member-cluster client map from the current set of
+// MemberCluster CRs.
+//
+// TODO(m1kola): slice-3: restart is used instead of a hot, reactive reload because the
+// per-cluster clients and caches are built once at startup and threaded into every controller,
+// so adding/removing a cluster currently requires re-initialising the manager. A later slice
+// makes membership changes reactive (no restart). This mirrors the OperatorConfig watcher.
```
m1kola, line 25: Note: I'm starting to doubt if we actually need hot reload for `MemberCluster` as initially planned. Restarting in the same way as `OperatorConfig` might be a good enough given that both won't change often.

But it is something that I want to revisit closer to the end of project. I think once I switch E2Es to a new way to define multi-cluster I'll get a better sense of this and it will be easier to make a decision.

pkg/multicluster/memberwatch/memberwatch.go
```diff
@@ -79,6 +80,13 @@ func getClusterCredentials(clustersMap map[string]cluster.Cluster,
 func (m *MemberClusterHealthChecker) populateCache(clustersMap map[string]cluster.Cluster, log *zap.SugaredLogger) {
 	kubeConfigFile, err := multicluster.NewKubeConfigFile(multicluster.GetKubeConfigPath())
 	if err != nil {
+		if os.IsNotExist(err) {
+			// TODO(m1kola): slice-3: expected when member clusters are discovered from
+			// MemberCluster CRs rather than the legacy mounted kubeconfig — there is no
+			// kubeconfig file. File-based health checking does not apply on that path.
+			log.Debugf("No member-cluster kubeconfig file at %s; skipping file-based health cache", multicluster.GetKubeConfigPath())
+			return
+		}
```
m1kola: Because of this we are not doing a health check properly, I believe. Looking into it.

pkg/membercluster/watcher.go
mircea-cosbuc: Is hot-reload not important for dynamically being able to add/remove clusters?

mircea-cosbuc (approved):

pkg/membercluster/watcher.go
m1kola: @mircea-cosbuc currently in this PR we stop the operator (so Kubernetes can restart the pod) when a new `MemberCluster` is added, existing modified or a `MemberCluster` deleted (same as `OperatorConfig`).

It works, but it is disruptive. It would be nice to avoid restarts, if possible.

changelog/20260717_feature_operator_membercluster_discovery.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: feature
+date: 2026-07-17
+---
+
+* **Multi-cluster**: The operator now discovers member clusters by reading `MemberCluster` CRs and their per-cluster credential Secrets from its own namespace, building one API client per cluster. This replaces the need for the monolithic multi-cluster kubeconfig Secret and the `mongodb-kubernetes-operator-member-list` ConfigMap, which are still honoured as a fallback when no `MemberCluster` CRs are present (removed in a later release). The operator restarts to pick up `MemberCluster` additions, spec changes, and removals; rotating a credential Secret in place currently also requires an operator restart.
```
josvazg, line 6: **q**: "the operator restart to pick up ... additions". This reads as if the operator is NOT constantly reconciling the MemberCluster CR set, but only does it at startup time, is that right? Seems weird

pkg/membercluster/watcher.go
m1kola: Ah, forgot to mention here - we will need hot reload (or restart which is smarter) because we will be reconciling `MemberCluster` with status at least to indicate RBAC validity.

changelog/20260717_feature_operator_membercluster_discovery.md
m1kola: @josvazg yes. At the moment we do not watch it, but we will have to in a later stage for RBAC validation as per TD.

Also hot-reload is related: https://github.com/mongodb/mongodb-kubernetes/pull/1394#discussion_r3621157919

pkg/membercluster/discovery.go
```diff
@@ -0,0 +1,82 @@
...
+	restConfigs := map[string]*restclient.Config{}
+	for i := range memberClusterList.Items {
+		mc := &memberClusterList.Items[i]
+		restConfig, err := restConfigFromMemberCluster(ctx, c, mc, namespace)
+		if err != nil {
+			zap.S().Errorf("Skipping member cluster %q (MemberCluster %q): %s", mc.Spec.ClusterName, mc.Name, err)
+			continue
+		}
+		restConfig.Timeout = time.Duration(clientTimeoutSeconds) * time.Second
+		restConfigs[mc.Spec.ClusterName] = restConfig
```
josvazg, line 56: **q**: Should we add some protection here against several CRs using the same clustername, maybe by mistake?

m1kola: @josvazg good point! Unfortunately it won't be possible to validate this on admission to the API server. It is unfortunate we can't use just `metadata.name` due to what we allow in MCK 1.x which is the only way to guarantee uniqueness.

I need to think about it. Probably will have to solve this in a follow up or in the next iteration (once we start adding reconciliation & conditions for `MemberCluster`).

Please leave this comment open for now. I'll make a note out of it.

josvazg: ok, if you note this and you may fix downstream, I do not have further objections here

josvazg (approved):

pkg/membercluster/discovery.go
m1kola: I replaced this with the following in the next PR (https://github.com/mongodb/mongodb-kubernetes/pull/1407/changes#r3639925954):

```go
// TODO(m1kola): We should validate that the clusterName is unique across all MemberCluster CRs, and return an error if not OR better, report it in the status of the MemberCluster CR. For now, we just log a warning and overwrite the previous entry.
if _, exists := restConfigs[mc.Spec.ClusterName]; exists {
	zap.S().Warnf("Duplicate clusterName %q found in MemberCluster CRs; overwriting previous entry", mc.Spec.ClusterName)
}
```

PR #1396, opened 2026-07-20
chore(e2e): Stabilize sharded SCRAM transition fallback

nammn:
# Summary

`e2e_sharded_cluster_x509_to_scram_transition` still flakes after the initial CLOUDP-383102 workaround (#821). This PR fixes the broken fallback rather than re-trying the same recovery path.

**Why this is needed vs. the initial fix (#821).** #821 added `_create_automation_agent_user`: after a 480s wait for `Running`, it pre-creates the `mms-automation-agent` SCRAM user via `kubectl exec` (localhost exception) on `config-0`, then waits up to 600s more. That reduced the ~70% failure rate but left three latent defects in the fallback itself, so the flake returned:

- **Hardcoded `config-0`.** The fixture runs `configServerCount: 2`. When `config-0` is secondary, `db.createUser` returns `not primary` while `config-1` is primary. #821 never tried `config-1`, so the user was never created and the deadlock was never broken.
- **Redundant pymongo probe that added its own bugs.** #821 ran a pymongo "user exists?" probe against `config-0` before the `kubectl exec` loop. The probe omitted `authMechanism="SCRAM-SHA-256"`, surfacing a misleading `MechanismUnavailable` instead of the real state. It also hardcoded `config-0` — the same defect as the create loop. The probe was a fast path whose failure modes silently fell through to the loop anyway, so it added complexity and bugs without correctness.
- **Swallow-and-wait.** #821 caught every `createUser` failure (including `requires authentication`) and continued, so the test burned the full 600s and then timed out with a downstream `StatefulSet not ready` that masked the actual cause — a credential-bootstrap deadlock.

**What this PR does.**

- Derives the config-server pod list from the resource spec (`configServerCount`) via `resource.config_srv_pod_name(i)` instead of hardcoding `config-0` and a count constant. Iterates every pod; advances to the next on `not primary` / `NotWritablePrimary`, stops on success or `already exists`.
- Removes the pymongo user-exists probe entirely — the `kubectl exec` loop already handles `already exists` as a success outcome, making the probe redundant.
- Raises immediately when every candidate pod fails, so the real cause surfaces instead of a 600s timeout.
- Redacts raw mongosh stdout/stderr from the exception (`create_user_js` contains the password; mongosh may echo the eval'd script on error).
- Reuses `SCRAM_AGENT_USER` from `automation_config_tester` instead of hardcoding the username.

## Proof of Work

- `make precommit`: passed
- e2e: the `test_can_enable_scram_sha_256` test itself is the check — if the workaround fails, the cluster never reaches `Running` and the test fails.

## Checklist

    - `skip-changelog` label applies because this is a test-only change


josvazg (commented):
>  The unit tests added in 9920f52b6 (test_create_user_fallback.py) were deleted in 97e594ba0 with no replacement, so the final state has zero test coverage for the pod-iteration/raise logic — the only part of this PR with real branching logic (primary/secondary detection, redaction, exhaustion-raise). Given this logic has already flaked in production once, some unit coverage would reduce risk of a third round of this bug.

josvazg (approved):

helderjs (approved):

PR #1413, opened 2026-07-22
KUBE-179: harden metrics-forwarder and Envoy cleanup error paths

anandsyncs:
[skip-ci]
# Summary

Error-path and ordering hardening for the MongoDBSearch metrics forwarder and Envoy controllers, split out of the lean lifecycle layer #1382 to keep that PR reviewable. This PR is stacked on #1382 at `d91326aca280e943f116159aaa51e96134e76c89`. Everything here closes pre-existing gaps in error paths and write ordering — none of these are regressions introduced by #1382.

Three commits:

1. **Two-phase metrics reconcile with a single diff-gated state write** (`36348f17a`). The forwarder reconcile now runs in two phases over one in-memory topology state: phase 1 updates every cluster's entry and deregisters Ops Manager hosts whose deferral window elapsed; removed-cluster cleanup mutates the same state. The state is persisted exactly once per reconcile — only when it actually changed — and always before phase 2 ensures ConfigMaps/Deployments, so a forwarder never starts pushing metrics for topology that is not durable. Per-cluster failures are aggregated with `errors.Join` instead of surfacing only the first one. Previously each cluster performed its own read-modify-write (N uncached reads + N blind sequential writes per reconcile) and dependency replication ran before the state write.

2. **Pre-deletion cleanup ordering** (`602446f30`). CR-delete cleanup now tracks blocked clusters instead of failing on the first error: a cluster whose Deployment is still terminating, whose client is missing, or that hit an error is skipped for the pass while the other clusters continue through host deregistration and resource deletion. A missing member-cluster client fails closed (previously it was silently skipped, letting the finalizer drop while an unreachable cluster still held forwarder resources and OM hosts). The finalizer is removed only once every cluster is fully clean.

3. **Envoy LB cleanup confirm/status guards** (`946933632`). The managed→unmanaged cleanup path runs off a possibly stale cache. It now re-confirms the live CR with the uncached reader (same UID, same generation, LB still unmanaged) before every precondition delete and before touching `/status/loadBalancer`; a stale cleanup requeues instead of deleting resources or wiping status for a CR that was re-managed or replaced. Status writes on this path use a JSON patch with uid/generation `test` preconditions, and cleanup failures are recorded on the loadBalancer sub-status instead of being silently dropped. The sentinel errors are scoped to the Envoy controller.

All behaviors of the lean base are preserved: deletion-first Reconcile, the unsupported-source persisted-hosts finalizer guard, the removed-cluster cleanup call, the uncached re-add guard before host deregistration, and the `mongodb.com/disable-reconciliation` pause guards (#1382's) sit untouched upstream of these paths.

Diff: 4 files changed, 549 insertions(+), 105 deletions(-) — all in `controllers/operator/` (metrics forwarder + Envoy controllers and their tests). New unit coverage: single/zero state-write assertions, dependency-replication ordering, cross-cluster error aggregation, fail-closed missing clients, and the restored Envoy stale-cleanup/status-patch test family.

## Proof of Work

Local gates on the branch head (restacked onto #1382's rebased head; content byte-identical to the previously reviewed diff modulo line offsets):

- Per-commit `go build ./...` — clean at each of the 3 commits.
- `go test ./controllers/... ./api/...` — all packages ok, zero failures.
- `make fmt` — no churn; `git diff --check` — clean.
- `go tool gotestsum -- -race -count=1 ./...` — **3097 tests, 3 skipped (env-gated), 0 failures**, no data races.
- #1382's pause subtests and this PR's hardening test families verified passing together in one targeted run.

## Checklist

    - `skip-changelog`: internal hardening of unreleased controllers, no user-facing behavior change.

### DIFF
--- a/api/mongodb/v1/search/mongodbsearch_types.go
+++ b/api/mongodb/v1/search/mongodbsearch_types.go
@@ -1046,6 +1046,10 @@ func (s *MongoDBSearch) IsWireprotoEnabled() bool {
 	return ok && val == "true"
 }
 
+func (s *MongoDBSearch) IsReconciliationDisabled() bool {
+	return s.Annotations[DisableReconciliationAnnotation] == "true"
+}
+
 func (s *MongoDBSearch) GetEffectiveMongotPort() int32 {
 	if s.IsWireprotoEnabled() {
 		return s.GetMongotWireprotoPort()
--- a/api/mongodb/v1/search/mongodbsearch_types.go
+++ b/api/mongodb/v1/search/mongodbsearch_types.go
@@ -861,7 +861,7 @@ func (s *MongoDBSearch) GetOwnerReferences() []metav1.OwnerReference {
 	ownerReference := *metav1.NewControllerRef(s, schema.GroupVersionKind{
 		Group:   v1.SchemeGroupVersion.Group,
 		Version: v1.SchemeGroupVersion.Version,
-		Kind:    s.Kind,
+		Kind:    s.GetKind(),
 	})
 	return []metav1.OwnerReference{ownerReference}
 }
--- a/api/mongodb/v1/search/mongodbsearch_types_test.go
+++ b/api/mongodb/v1/search/mongodbsearch_types_test.go
@@ -31,6 +31,26 @@ func TestReplicasOrDefault(t *testing.T) {
 	})
 }
 
+func TestIsReconciliationDisabled(t *testing.T) {
+	tests := []struct {
+		name        string
+		annotations map[string]string
+		want        bool
+	}{
+		{name: "nil annotations"},
+		{name: "missing annotation", annotations: map[string]string{"other": "true"}},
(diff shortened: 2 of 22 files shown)


anandsyncs:
**Parked pending the #1382 descope rework.** The design-change round (2026-07-23) removed the foundations this PR builds on: uncached/APIReader plumbing and UID/generation fencing are gone from #1382, so commit 3's Envoy stale-cleanup confirm/status guards contradict the new design (races are explicitly out of scope), and the two metrics commits (two-phase single state write, pre-deletion cleanup ordering) reference reworked code paths. Per the rework blueprint: the Envoy-guard commit is dropped; the two metrics-ordering commits are candidates for re-authoring against the descoped #1382 after it merges. The branch intentionally stays at `946933632` (based on the pre-rework #1382 head), so this PR shows as conflicting until then.

PR #1437, opened 2026-07-29
KUBE-267: Adopt mongodb-forks/digest + staleRetryTransport for RFC 7616 stale nonce

nammn:
# Why

During an OM upgrade the AppDB restarts. `retryablehttp` retries 5xx with the same digest nonce; the nonce ages past OM's 60s TTL; OM returns `401 + stale=true` (RFC 7616). The operator mapped any 401 to `Failed: "The admin-key secret might be corrupted"` — a false alarm that flakes `e2e_om_ops_manager_upgrade` (~6.7%) and would mislead customers.

# What

1. **Replace hand-rolled digest client with `mongodb-forks/digest`** — deletes ~64 lines of bespoke auth (`digestParts`, `getCnonce`, `getDigestAuthorization`, `authorizeRequest`). MongoDB internal standard, used by `mongosync`, `bia`, `mraautomation`. Fixes a parser that panics on unquoted `stale=true`, adds POST body replay.

2. **Add `staleRetryTransport`** (`controllers/om/api/stale_retry.go`) — a custom `http.RoundTripper` that wraps `digest.Transport` and retries on `401 + stale=true` (RFC 7616 §3.2), up to 5 times. Each retry triggers a fresh digest challenge-response cycle with a new nonce. Genuine credential failures (`stale=false` or absent) are returned immediately without retry. The bound prevents infinite loops against a broken server. 4 unit tests cover: stale retry succeeds, non-stale 401 no retry, body replay on POST, max-bound.

3. **Fix e2e test** — `test_upgrade_om_version` was the only upgrade test without `ignore_errors=True`. Added it (1 line), consistent with every other upgrade test in the file.

# How the retry layers stack

```
retryablehttp.Client     ← 3 retries, exponential backoff, handles 5xx + connection errors
  → staleRetryTransport  ← 5 retries, ONLY on 401 + stale=true (RFC 7616)
    → digest.Transport   ← 1 challenge-response cycle per call
      → http.Transport   ← actual TCP
```

Each error type is handled at the right layer. `staleRetryTransport` does not touch 5xx — that is `retryablehttp`'s job.

Jira: [KUBE-267](https://jira.mongodb.org/browse/KUBE-267)

## Checklist


controllers/om/api/http.go
```diff
@@ -160,22 +177,31 @@ func OptionCAValidate(ca string) func(client *Client) error {
 
 // Request executes an HTTP request, given a series of parameters, over this *Client object.
 // It handles Digest when needed and json marshaling of the `v` struct.
+// On a 401, it retries once with a fresh digest challenge (RFC 7616 stale nonce).
```
nammn: the client doesn't support rfc7616 but might later

controllers/om/api/http.go
```diff
@@ -160,22 +177,31 @@ func OptionCAValidate(ca string) func(client *Client) error {
...
+	// failure will simply 401 again; a stale nonce (RFC 7616) succeeds.
+	var apiErr *apierror.Error
+	if errors.As(err, &apiErr) && apiErr.Status != nil && *apiErr.Status == http.StatusUnauthorized {
+		return client.tryRequest(method, url, path, v)
 	}
 
+	return nil, nil, err
+}
+
+// tryRequest creates and sends a single HTTP request.
```
nammn: this matches what python requests is doing, in later iterations we might use a lib that supports more retries based on the nonce reply

controllers/om/api/digest.go
```diff
@@ -33,38 +28,3 @@ func parseAPIError(statusCode int, method, url string, body []byte) *apierror.Er
 
 	return errorObject
 }
-
-func digestParts(resp *http.Response) map[string]string {
```
nammn, line 37: in theory we could use that and handle stale=true but i don't think thats the way to go

docker/mongodb-kubernetes-tests/tests/opsmanager/om_ops_manager_upgrade.py
```diff
@@ -287,7 +287,7 @@ def test_upgrade_om_version(
         ops_manager.set_appdb_version(custom_appdb_version)
 
         ops_manager.update()
-        ops_manager.om_status().assert_reaches_phase(Phase.Running)
+        ops_manager.om_status().assert_reaches_phase(Phase.Running, ignore_errors=True)
```
MaciejKaras, line 290: wasn't this PR fixing the intermediate error sissue on the upgrades? Do we still need it?

nammn: kind of, so yes we retry once, but upgrade might take longer, proper fix would be having the lib infinitely retry on stale=true nonce here we just retry once on 401 for simplicity

nammn: so i wanted to be double sure and added that ignore_errors=true as well for those long running upgrades on slow clusters

MaciejKaras: ok, so we only retry once? Cannot we infer from the response if the nonce staleness was actually the reason and retry for more time?

MaciejKaras (approved):
LGMT! Just one small not on the nonce staleness. Still it's good direction!

docker/mongodb-kubernetes-tests/tests/opsmanager/om_ops_manager_upgrade.py
nammn: yes i was thinking about that - problem is, you would need to parse the nonce like we manually did here: https://github.com/mongodb/mongodb-kubernetes/pull/1437/changes#diff-e06507c86cda7a8849aab4f17d25d2c1893b8de635f5f61cae15ddfb61494865L37-L48 - feels "hideous" and too much.

i think its better in the lib itself like here: https://github.com/mongodb-forks/digest/pull/60

nammn:

controllers/om/api/stale_retry.go
```diff
@@ -0,0 +1,67 @@
...
+		if resp.StatusCode != http.StatusUnauthorized {
+			return resp, nil
+		}
+		if !isStaleNonce(resp.Header.Get("WWW-Authenticate")) {
+			return resp, nil
+		}
+		if attempt >= maxStaleRetries {
+			return resp, nil
+		}
+		_ = resp.Body.Close()
```
m1kola, line 59: To reuse connection we may also need to read the response body in a whole and only after that close. Something like:

```suggestion
		_, _ = io.Copy(io.Discard, res.Body)
		_ = resp.Body.Close()
```

See comment for the body field on `Response` [here](https://pkg.go.dev/net/http#Response):

```
	// The http Client and Transport guarantee that Body is always
	// non-nil, even on responses without a body or responses with
	// a zero-length body. It is the caller's responsibility to
	// close Body. The default HTTP client's Transport may not
	// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
	// not read to completion and closed.
```

controllers/om/api/stale_retry_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+import (
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestStaleRetrySuccessNoRetry(t *testing.T) {
```
m1kola, line 11: Nit: Seems like there is a lot of boilerplate here. Can these tests be table-tests?

controllers/om/api/http.go
```diff
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
+	"github.com/mongodb-forks/digest"
```
m1kola, line 16: From [readme](https://github.com/mongodb-forks/digest#update-2020):

> Update 2020
>
> This is a fork of the now unmaintained fork of [digest](https://github.com/bobziuchkovski/digest). This implementation now supports the SHA-256 algorithm which was added as part of [rfc 7616](https://tools.ietf.org/html/rfc7616).

PR #1439, opened 2026-07-29
[skip-ci] CLOUDP-429011 - External AppDB: API and admission validation (MongoDB only)

MaciejKaras:
# Summary

Adds the API surface for running the Ops Manager Application Database as an external, user-managed MongoDB resource instead of an MCK-deployed in-cluster one (CLOUDP-429011):

- New `spec.externalApplicationDatabaseRef` on the MongoDBOpsManager CR referencing a MongoDB CR (Kind restricted to `MongoDB` in this PR).
- New `spec.role: AppDB` on the MongoDB CR marking it as the external AppDB, with admission validation (single-cluster topology required; role immutability).
- MongoDBMultiCluster is explicitly rejected for `spec.role` here (dedicated admission validator); multi support is planned for a later iteration.
- CRDs regenerated via `make manifests`; mechanical pointer/signature adaptations in master-owned call sites; changelog entry included.

This PR is the bottom of a 2-PR stack; reconciliation logic arrives in #1440.

## Proof of Work

- `go build ./...` and `go vet ./...` green
- `go test ./api/... ./controllers/... ./pkg/...` green, incl. new table-driven tests exercising the multi rejection through the `ValidateCreate` admission path
- `make manifests` idempotency gate: zero diff; generated ref enum shows only `MongoDB`
- changelog validator: "Changelog entries found and validated"

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you added changelog file?

PR #1440, opened 2026-07-29
CLOUDP-429011 - External AppDB: reconciliation logic and e2e tests

MaciejKaras:
# Summary

Stacked on #1439 (API). Adds the business logic for the external Application Database feature (CLOUDP-429011), MongoDB-only — a `Kind` other than `MongoDB` (e.g. `MongoDBMultiCluster`) in `externalApplicationDatabaseRef` is rejected by the controller until multi support arrives in a later iteration:

- External AppDB reconciliation: fresh start and forward/reverse migration, ownership arbitration between Ops Manager and the MongoDB CR serving as AppDB.
- e2e coverage: fresh-start and forward-migration tests with fixtures + evergreen tasks (single-cluster).

## Proof of Work

- `go build ./...` green; `go test ./api/... ./controllers/... ./pkg/...` green (63 packages)
- e2e files statically verified (py_compile + fixture existence); not run locally

## Checklist

  - [x] (ships in parent #1439; `skip-changelog` label applied here)

### DIFF
--- a/api/mongodb/v1/om/appdb_types.go
+++ b/api/mongodb/v1/om/appdb_types.go
@@ -28,6 +28,21 @@ const (
 	ClusterTopologyMultiCluster = "MultiCluster"
 )
 
+// AppDBUserRoles defines required roles for the AppDB user are outlined in the documentation
+// https://docs.opsmanager.mongodb.com/current/tutorial/prepare-backing-mongodb-instances/#replica-set-security
+var AppDBUserRoles = []authtypes.Role{
+	{Name: "readWriteAnyDatabase", Database: "admin"},
+	{Name: "dbAdminAnyDatabase", Database: "admin"},
+	{Name: "clusterMonitor", Database: "admin"},
+	// Allows user to do db.fsyncLock required by CLOUDP-78890
+	// https://docs.mongodb.com/manual/reference/built-in-roles/#hostManager
+	{Name: "backup", Database: "admin"},
+	{Name: "restore", Database: "admin"},
+	// Allows user to do db.fsyncLock required by CLOUDP-78890
+	// https://docs.mongodb.com/manual/reference/built-in-roles/#hostManager
+	{Name: "hostManager", Database: "admin"},
+}
+
 type AppDBSpec struct {
 	// +kubebuilder:validation:Pattern=^[0-9]+.[0-9]+.[0-9]+(-.+)?$|^$
 	Version string `json:"version"`
@@ -198,8 +213,6 @@ func (m *AppDBSpec) GetAuthOptions() authtypes.Options {
 	}
 }
 
-// GetAuthUsers returns a list of all scram users for this deployment.
-// in this case it is just the Ops Manager user for the AppDB.
 func (m *AppDBSpec) GetAuthUsers() []authtypes.User {
 	passwordSecretName := m.GetOpsManagerUserPasswordSecretName()
 	if m.PasswordSecretKeyRef != nil && m.PasswordSecretKeyRef.Name != "" {
@@ -207,40 +220,9 @@ func (m *AppDBSpec) GetAuthUsers() []authtypes.User {
 	}
 	return []authtypes.User{
 		{
-			Username: util.OpsManagerMongoDBUserName,
-			Database: util.DefaultUserDatabase,
-			// required roles for the AppDB user are outlined in the documentation
-			// https://docs.opsmanager.mongodb.com/current/tutorial/prepare-backing-mongodb-instances/#replica-set-security
-			Roles: []authtypes.Role{
-				{
-					Name:     "readWriteAnyDatabase",
-					Database: "admin",
-				},
-				{
-					Name:     "dbAdminAnyDatabase",
-					Database: "admin",
-				},
-				{
-					Name:     "clusterMonitor",
-					Database: "admin",
-				},
-				// Enables backup and restoration roles
-				// https://docs.mongodb.com/manual/reference/built-in-roles/#backup-and-restoration-roles
-				{
-					Name:     "backup",
-					Database: "admin",
-				},
-				{
-					Name:     "restore",
-					Database: "admin",
-				},
-				// Allows user to do db.fsyncLock required by CLOUDP-78890
-				// https://docs.mongodb.com/manual/reference/built-in-roles/#hostManager
-				{
-					Name:     "hostManager",
-					Database: "admin",
-				},
-			},
+			Username:                   util.OpsManagerMongoDBUserName,
+			Database:                   util.DefaultUserDatabase,
+			Roles:                      AppDBUserRoles,
 			PasswordSecretKey:          m.GetOpsManagerUserPasswordSecretKey(),
 			PasswordSecretName:         passwordSecretName,
 			ScramCredentialsSecretName: m.OpsManagerUserScramCredentialsName(),
--- a/api/mongodb/v1/om/appdb_types.go
+++ b/api/mongodb/v1/om/appdb_types.go
@@ -28,7 +28,7 @@ const (
 	ClusterTopologyMultiCluster = "MultiCluster"
 )
 
-// AppDBUserRoles defines required roles for the AppDB user are outlined in the documentation
+// AppDBUserRoles defines required roles for the AppDB user that are outlined in the documentation
 // https://docs.opsmanager.mongodb.com/current/tutorial/prepare-backing-mongodb-instances/#replica-set-security
 var AppDBUserRoles = []authtypes.Role{
 	{Name: "readWriteAnyDatabase", Database: "admin"},
--- a/api/mongodb/v1/om/opsmanager_types.go
+++ b/api/mongodb/v1/om/opsmanager_types.go
@@ -188,6 +188,9 @@ type ExternalAppDBRef struct {
 	// +kubebuilder:validation:Enum=MongoDB
 	// +kubebuilder:validation:Required
 	Kind string `json:"kind"`
+
+	// Transient fields
+	Namespace string `json:"-"`
 }
 
 type Logging struct {
@@ -677,6 +680,8 @@ func (om *MongoDBOpsManager) InitDefaultFields() {
 		om.Spec.AppDB.Namespace = om.Namespace
 		om.Spec.AppDB.ClusterDomain = om.Spec.GetClusterDomain()
 		om.Spec.AppDB.ResourceType = mdbv1.ReplicaSet
+	} else {
+		om.Spec.ExternalApplicationDatabaseRef.Namespace = om.Namespace
 	}
 }
 
--- a/api/mongodb/v1/om/opsmanager_types.go
+++ b/api/mongodb/v1/om/opsmanager_types.go
@@ -770,6 +770,12 @@ func (om *MongoDBOpsManager) UpdateStatus(phase status.Phase, statusOptions ...s
 }
 
 func (om *MongoDBOpsManager) updateStatusAppDb(phase status.Phase, statusOptions ...status.Option) {
+	if om.Spec.ExternalApplicationDatabaseRef != nil {
+		om.Status.AppDbStatus = AppDbStatus{}
+		om.Status.AppDbStatus.UpdateCommonFields(phase, om.GetGeneration(), statusOptions...)
+		return
+	}
+
 	om.Status.AppDbStatus.UpdateCommonFields(phase, om.GetGeneration(), statusOptions...)
 
 	if option, exists := status.GetOption(statusOptions, status.ReplicaSetMembersOption{}); exists {
--- a/api/mongodb/v1/om/opsmanager_types.go
+++ b/api/mongodb/v1/om/opsmanager_types.go
@@ -311,9 +311,14 @@ func (ms MongoDBOpsManagerSpec) GetOpsManagerCA() string {
 }
 
 func (ms MongoDBOpsManagerSpec) GetAppDbCA() string {
+	if ms.ExternalApplicationDatabaseRef != nil {
+		return ""
+	}
+
 	if ms.AppDB.Security != nil && ms.AppDB.Security.TLSConfig != nil {
 		return ms.AppDB.Security.TLSConfig.CA
 	}
+
 	return ""
 }
 
--- a/api/mongodb/v1/om/opsmanager_types.go
+++ b/api/mongodb/v1/om/opsmanager_types.go
@@ -769,6 +769,12 @@ func (om *MongoDBOpsManager) UpdateStatus(phase status.Phase, statusOptions ...s
 }
 
 func (om *MongoDBOpsManager) updateStatusAppDb(phase status.Phase, statusOptions ...status.Option) {
+	if om.Spec.ExternalApplicationDatabaseRef != nil {
+		om.Status.AppDbStatus = AppDbStatus{}
+		om.Status.AppDbStatus.UpdateCommonFields(phase, om.GetGeneration(), statusOptions...)
+		return
+	}
+
 	om.Status.AppDbStatus.UpdateCommonFields(phase, om.GetGeneration(), statusOptions...)
 
 	if option, exists := status.GetOption(statusOptions, status.ReplicaSetMembersOption{}); exists {
--- a/api/mongodb/v1/om/opsmanager_types.go
+++ b/api/mongodb/v1/om/opsmanager_types.go
@@ -312,7 +312,7 @@ func (ms MongoDBOpsManagerSpec) GetOpsManagerCA() string {
 }
 
 func (ms MongoDBOpsManagerSpec) GetAppDbCA() string {
-	if ms.ExternalApplicationDatabaseRef != nil {
+	if ms.ExternalAppDBRef != nil {
 		return ""
 	}
 
@@ -686,7 +686,7 @@ func (om *MongoDBOpsManager) InitDefaultFields() {
 		om.Spec.AppDB.ClusterDomain = om.Spec.GetClusterDomain()
 		om.Spec.AppDB.ResourceType = mdbv1.ReplicaSet
 	} else {
-		om.Spec.ExternalApplicationDatabaseRef.Namespace = om.Namespace
+		om.Spec.ExternalAppDBRef.Namespace = om.Namespace
 	}
 }
 
@@ -780,7 +780,7 @@ func (om *MongoDBOpsManager) UpdateStatus(phase status.Phase, statusOptions ...s
 }
 
 func (om *MongoDBOpsManager) updateStatusAppDb(phase status.Phase, statusOptions ...status.Option) {
-	if om.Spec.ExternalApplicationDatabaseRef != nil {
+	if om.Spec.ExternalAppDBRef != nil {
 		om.Status.AppDbStatus = AppDbStatus{}
 		om.Status.AppDbStatus.UpdateCommonFields(phase, om.GetGeneration(), statusOptions...)
 		return
--- a/api/mongodb/v1/om/opsmanager_types.go
+++ b/api/mongodb/v1/om/opsmanager_types.go
@@ -187,6 +187,9 @@ type ExternalApplicationDatabaseRef struct {
 	// +kubebuilder:validation:Enum=MongoDB
 	// +kubebuilder:validation:Required
 	Kind string `json:"kind"`
+
+	// Transient fields
+	Namespace string `json:"-"`
 }
 
 type Logging struct {
@@ -676,6 +679,8 @@ func (om *MongoDBOpsManager) InitDefaultFields() {
 		om.Spec.AppDB.Namespace = om.Namespace
 		om.Spec.AppDB.ClusterDomain = om.Spec.GetClusterDomain()
 		om.Spec.AppDB.ResourceType = mdbv1.ReplicaSet
+	} else {
+		om.Spec.ExternalApplicationDatabaseRef.Namespace = om.Namespace
 	}
 }
 
--- a/api/mongodb/v1/om/opsmanager_types.go
+++ b/api/mongodb/v1/om/opsmanager_types.go
@@ -702,10 +702,15 @@ func ensureSecurityWithSCRAM(specSecurity *mdbv1.Security) *mdbv1.Security {
 // AppDBName's AppDB branch is nil-safe only via the admission invariant that at least one of
 // applicationDatabase or externalApplicationDatabaseRef is set.
 func (om *MongoDBOpsManager) AppDBName() string {
-	if om.Spec.ExternalAppDBRef != nil {
-		return om.Spec.ExternalAppDBRef.Name
+	if om.IsInternalAppDB() {
+		return om.Spec.AppDB.Name()
 	}
-	return om.Spec.AppDB.Name()
+
+	return om.Spec.ExternalAppDBRef.Name
+}
+
+func (om *MongoDBOpsManager) IsInternalAppDB() bool {
+	return om.Spec.ExternalAppDBRef == nil
 }
 
 func (om *MongoDBOpsManager) SvcName() string {
--- a/api/mongodb/v1/om/opsmanager_types.go
+++ b/api/mongodb/v1/om/opsmanager_types.go
@@ -312,9 +312,14 @@ func (ms MongoDBOpsManagerSpec) GetOpsManagerCA() string {
 }
 
 func (ms MongoDBOpsManagerSpec) GetAppDbCA() string {
+	if ms.ExternalApplicationDatabaseRef != nil {
+		return ""
+	}
+
 	if ms.AppDB.Security != nil && ms.AppDB.Security.TLSConfig != nil {
 		return ms.AppDB.Security.TLSConfig.CA
 	}
+
 	return ""
 }
 
--- a/controllers/operator/appdbreplicaset_controller.go
+++ b/controllers/operator/appdbreplicaset_controller.go
@@ -686,7 +686,7 @@ func (r *ReconcileAppDbReplicaSet) requestAppDBReverseMigration(ctx context.Cont
 // reclaimAppDBStatefulset transfers the ownership of the AppDB StatefulSet to this OM and clears migration annotations
 func (r *ReconcileAppDbReplicaSet) reclaimAppDBStatefulset(ctx context.Context, opsManager *omv1.MongoDBOpsManager, sts appsv1.StatefulSet) error {
 	sts.OwnerReferences = kube.BaseOwnerReference(opsManager)
-	delete(sts.Annotations, util.AppDBReverseMigrationReadyAnnotation)
+	// stale forward migration annotation cleanup
 	delete(sts.Annotations, util.AppDBMigrationReadyAnnotation)
 	if err := r.client.Update(ctx, &sts); err != nil {
 		return xerrors.Errorf("failed to reclaim StatefulSet %s: %w", sts.GetName(), err)
@@ -870,16 +870,16 @@ func (r *ReconcileAppDbReplicaSet) ReconcileAppDB(ctx context.Context, opsManage
 	}
 	appdbOpts.PrometheusTLSCertHash = prometheusCertHash
 
-	allStatefulSetsExist, err := r.allStatefulSetsExist(ctx, opsManager, log)
+	allStatefulSetsExistAndValid, err := r.allStatefulSetsExistsInValidState(ctx, opsManager, log)
 	if err != nil {
 		return r.updateStatus(ctx, opsManager, workflow.Failed(xerrors.Errorf("failed to check the state of all stateful sets: %w", err)), log, appDbStatusOption)
 	}
 
-	publishAutomationConfigFirst := r.publishAutomationConfigFirst(opsManager, allStatefulSetsExist, log)
+	publishAutomationConfigFirst := r.publishAutomationConfigFirst(opsManager, allStatefulSetsExistAndValid, log)
 
 	workflowStatus = workflow.RunInGivenOrder(publishAutomationConfigFirst,
 		func() workflow.Status {
-			return r.deployAutomationConfigAndWaitForAgentsReachGoalState(ctx, log, opsManager, &podVars, allStatefulSetsExist, appdbOpts)
+			return r.deployAutomationConfigAndWaitForAgentsReachGoalState(ctx, log, opsManager, &podVars, allStatefulSetsExistAndValid, appdbOpts)
 		},
 		func() workflow.Status {
 			return r.deployStatefulSet(ctx, opsManager, log, podVars, appdbOpts)
@@ -2373,27 +2373,7 @@ func (r *ReconcileAppDbReplicaSet) getCurrentStatefulsetHostnames(opsManager *om
 	})
 }
 
-// isReAdoptedStatefulSetPendingReshape returns true when a StatefulSet taken back from a
-// MongoDB CR (reverse migration re-adoption) still carries the CR's pod shape and awaits the
-// rewrite to the internal-AppDB pod template. Internal AppDB pods always run a dedicated
-// "mongod" container (static and non-static architecture, vault or secret config backend);
-// MongoDB CR pods never do.
-func isReAdoptedStatefulSetPendingReshape(sts appsv1.StatefulSet) bool {
-	for _, c := range sts.Spec.Template.Spec.Containers {
-		if c.Name == util.MongodbContainerName {
-			return false
-		}
-	}
-	return true
-}
-
-// allStatefulSetsExist reports whether every member cluster's AppDB StatefulSet exists in its
-// internal-AppDB form. A StatefulSet re-adopted from a MongoDB CR that hasn't been reshaped yet
-// counts as not existing: its pods run no headless agent, so waiting for agent goal state before
-// deployStatefulSet rewrites the pod template would deadlock (the wait is skipped while this
-// returns false).
-func (r *ReconcileAppDbReplicaSet) allStatefulSetsExist(ctx context.Context, opsManager *omv1.MongoDBOpsManager, log *zap.SugaredLogger) (bool, error) {
-	allStsExist := true
+func (r *ReconcileAppDbReplicaSet) allStatefulSetsExistsInValidState(ctx context.Context, opsManager *omv1.MongoDBOpsManager, log *zap.SugaredLogger) (bool, error) {
 	for _, memberCluster := range r.helper.GetHealthyMemberClusters() {
 		stsName := opsManager.Spec.AppDB.NameForCluster(r.helper.getMemberClusterIndex(memberCluster.Name))
 		sts, err := memberCluster.Client.GetStatefulSet(ctx, kube.ObjectKey(opsManager.Namespace, stsName))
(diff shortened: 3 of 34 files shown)


controllers/operator/appdbreplicaset_external_reconciler.go
```diff
@@ -0,0 +1,176 @@
...
+func (e *ReconcileExternalAppDBReplicaSet) ReconcileAppDB(ctx context.Context, opsManager *omv1.MongoDBOpsManager) (reconcile.Result, error) {
+	if err := e.validateExternalAppDBReference(ctx, opsManager); err != nil {
+		return e.updateStatus(ctx, opsManager, workflow.Failed(xerrors.Errorf("Error validating externalApplicationDatabaseRef: %w", err)), e.log, mdbstatus.NewOMPartOption(mdbstatus.OpsManager))
+	}
+
+	if err := e.ensureAppDBStatefulSetOwnership(ctx, opsManager); err != nil {
+		return e.updateStatus(ctx, opsManager, workflow.Failed(xerrors.Errorf("Error detaching internal AppDB: %w", err)), e.log, mdbstatus.NewOMPartOption(mdbstatus.OpsManager))
+	}
+
+	return e.updateStatus(ctx, opsManager, workflow.OK(), e.log, mdbstatus.NewOMPartOption(mdbstatus.AppDb))
```
MaciejKaras: use OpsManager part for ExternalAppDB, not AppDB, check for it everywhere

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -634,6 +738,16 @@ func (r *ReconcileAppDbReplicaSet) ReconcileAppDB(ctx context.Context, opsManage
...
 	log.Infow("ReplicaSet.Status", "status", opsManager.Status.AppDbStatus)
 
+	// Ops Manager must own the AppDB StatefulSet before touching anything: a StatefulSet
+	// still owned by a MongoDB CR (reverse migration) is asked to be released and waited for.
+	opsManagerOwned, err := r.ensureAppDBStatefulSetOwnership(ctx, opsManager)
+	if err != nil {
+		return r.updateStatus(ctx, opsManager, workflow.Failed(err), log, appDbStatusOption)
+	}
+	if !opsManagerOwned {
+		return r.updateStatus(ctx, opsManager, workflow.Pending("waiting for MongoDB controller to release AppDB StatefulSet"), log, appDbStatusOption)
```
MaciejKaras: I would rather split this into two phases:
- if not opsManagerOwned and there is no reverse migration happening, keep Pending and current message
- if not opsManagerOwned and no reverse migration happening, use workflow.Disabled status with information that it is handled by external AppDB and add it's name

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -214,6 +220,20 @@ func (r *ReplicaSetReconcilerHelper) Reconcile(ctx context.Context) (reconcile.R
 		return r.updateStatus(ctx, status)
 	}
 
+	if rs.Spec.Role == mdbv1.RoleAppDB {
+		mongoDBOwned, notMongoDBOwnedMessage, err := r.ensureAppDBStatefulSetOwnership(ctx, rs)
+		if err != nil {
+			return r.updateStatus(ctx, workflow.Failed(err))
+		}
+		if !mongoDBOwned {
+			return r.updateStatus(ctx, workflow.Pending("%s", notMongoDBOwnedMessage))
```
MaciejKaras: I would rather split this into two distinct situations:
- when !mongoDBOwned and no forward migration happening, use Disabled status with correct message
- when !mongoDBOwned and forward migration happening use Pending status with correct message

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_forward.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def test_create_om_with_internal_appdb(self, primary_om: MongoDBOpsManager):
+        primary_om.update()
+        primary_om.om_status().assert_reaches_phase(Phase.Running, timeout=900)
+        primary_om.appdb_status().assert_reaches_phase(Phase.Running, timeout=600)
+
+
+@pytest.mark.e2e_om_external_appdb_forward
+class TestSentinelDocSurvivesForwardMigration:
+    """Procedure 2: start with internal AppDB, write a sentinel doc, create the MongoDB (role: AppDB)
```
MaciejKaras, line 85: check for multicluster usage here

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -702,7 +815,6 @@ func (r *ReconcileAppDbReplicaSet) ReconcileAppDB(ctx context.Context, opsManage
 	appdbOpts := construct.AppDBStatefulSetOptions{
 		InitAppDBImage: images.ContainerImage(r.imageUrls, util.InitDatabaseImageUrlEnv, r.initDatabaseVersion),
 		MongodbImage:   images.GetOfficialImage(r.imageUrls, opsManager.Spec.AppDB.Version, opsManager.GetAnnotations(), r.defaultArchitecture),
-		CustomAgentURL: r.customAgentURL,
```
MaciejKaras, line 705: why is this missing? it's not related to our change I believe. Is it present on master?

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -2263,11 +2371,30 @@ func (r *ReconcileAppDbReplicaSet) getCurrentStatefulsetHostnames(opsManager *om
 	})
 }
 
+// isReAdoptedStatefulSetPendingReshape returns true when a StatefulSet taken back from a
+// MongoDB CR (reverse migration re-adoption) still carries the CR's pod shape and awaits the
+// rewrite to the internal-AppDB pod template. Internal AppDB pods always run a dedicated
+// "mongod" container (static and non-static architecture, vault or secret config backend);
+// MongoDB CR pods never do.
+func isReAdoptedStatefulSetPendingReshape(sts appsv1.StatefulSet) bool {
```
MaciejKaras: q: do we still need this method to be exectuted?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -421,7 +438,15 @@ func publishAutomationConfigFirstRS(ctx context.Context, getter kubernetesClient
 		return false
 	}
 
+	if mdb.Spec.Role == mdbv1.RoleAppDB {
+		return false
+	}
+
 	databaseContainer := container.GetByName(util.DatabaseContainerName, currentSts.Spec.Template.Spec.Containers)
+	// If we are migrating from the internal AppDB to an external AppDB, the database container might not exist
```
MaciejKaras: nit: this is probably not needed since we check the `if mdb.Spec.Role == mdbv1.RoleAppDB` just couple lines earlier

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -804,11 +838,262 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+func (r *ReplicaSetReconcilerHelper) releaseAppDBStatefulsetOwnership(ctx context.Context, sts appsv1.StatefulSet) error {
+	sts.OwnerReferences = nil
+	if err := r.reconciler.client.Update(ctx, &sts); err != nil {
+		return xerrors.Errorf("failed to strip OwnerReferences from StatefulSet %s: %w", sts.GetName(), err)
+	}
+
+	return nil
+}
+
+// ensureAppDBRoleUser mirrors the internal AppDB reconciler's ensureAppDbPassword
```
MaciejKaras: be more brief and just mention what this user is for and why it is computed internally

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -804,11 +838,262 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+			SetField(util.OpsManagerPasswordKey, password).
+			SetOwnerReferences(kube.BaseOwnerReference(mdb)).
+			Build()
+
+		if err := r.reconciler.CreateSecret(ctx, newSecret); err != nil {
+			return xerrors.Errorf("failed to create password secret: %w", err)
+		}
+	}
+
+	// Inject the synthetic mongodb-ops-manager user directly into OM's automation
```
MaciejKaras: be brief with the comment

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -804,11 +838,262 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+		}
+		if err := authentication.ConfigureScramCredentials(&omUser, password, ac); err != nil {
+			return xerrors.Errorf("error generating SCRAM credentials for %s: %w", util.OpsManagerMongoDBUserName, err)
+		}
+		ac.Auth.EnsureUser(omUser)
+		return nil
+	}, r.log)
+}
+
+// ensureAppDBRoleKeyfile keeps the cluster keyfile of an AppDB-role CR's project in sync with
```
MaciejKaras: be brief

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -804,11 +838,262 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+	existingData, err := secret.ReadStringData(ctx, r.reconciler.SecretClient, secretObjectKey)
+	if err != nil && !secret.SecretNotExist(err) {
+		return xerrors.Errorf("failed to check for existing keyfile secret: %w", err)
+	}
+	sharedKey := existingData[constants.AgentKeyfileKey]
+
+	var projectKey string
+	// ReadUpdateAutomationConfig only pushes when the config actually changed, so forcing an
+	// already-matching key is a no-op against Ops Manager.
+	err = conn.ReadUpdateAutomationConfig(func(ac *om.AutomationConfig) error {
```
MaciejKaras: use `if err :=` pattern

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -804,11 +838,262 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+			return xerrors.Errorf("failed to fetch secret %s: %w", name, err)
+		}
+		if err := r.claimSecretForCR(ctx, mdb, name); err != nil {
+			return xerrors.Errorf("failed to claim secret %s: %w", name, err)
+		}
+	}
+	return nil
+}
+
+// claimSecretForCR sets this CR's OwnerReference on a shared handover secret it did not create
```
MaciejKaras: be brief

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -804,11 +838,262 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+	}
+	return nil
+}
+
 func (r *ReplicaSetReconcilerHelper) OnDelete(ctx context.Context, obj runtime.Object, log *zap.SugaredLogger) error {
 	rs := obj.(*mdbv1.MongoDB)
 
-	if err := r.cleanOpsManagerState(ctx, rs, log); err != nil {
-		return err
+	// Deleting an AppDB-role CR is a handover back to internal AppDB management (reverse
```
MaciejKaras: be brief. Mention that it's user responsibility to clean the project after migration. Add this check to the fresh and forward migration e2e tests to verify the project still exists.

controllers/operator/appdbreplicaset_controller.go
MaciejKaras: querying exact diff context

controllers/operator/workflow/disabled.go
```diff
@@ -10,7 +11,7 @@ type disabledStatus struct {
 }
 
 func Disabled() *disabledStatus {
-	return &disabledStatus{okStatus: &okStatus{requeue: false}}
+	return &disabledStatus{okStatus: &okStatus{requeueAfter: util.TWENTY_FOUR_HOURS}}
```
MaciejKaras, line 14: Disabled() must match OK().ReconcileResult() so the OM controller's early-return guard (result != emptyResult) doesn't fire when the external AppDB reconciler returns Disabled. Without this, reconcileOpsManager is skipped and opsManager.observedGeneration stays stale

The 24h requeue is harmless: Disabled.IsOK() is still true, so callers that check !status.IsOK() are unaffected. The daily requeue is a minor benefit — if the subresource gets re-enabled (e.g. backup turned on), the controller picks it up without waiting for an external event.

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -169,13 +169,13 @@ func newAppDBReconcilerHelper(ctx context.Context, opsManager *omv1.MongoDBOpsMa
 		readOnly:        readOnly,
 	}
 
-	appDBSpec := opsManager.Spec.AppDB
```
MaciejKaras: move this to api-changes PR

controllers/operator/appdbreplicaset_controller_multi_test.go
```diff
@@ -75,13 +75,13 @@ func TestAppDB_MultiCluster(t *testing.T) {
 		})
 
 	opsManager := builder.Build()
-	appdb := opsManager.Spec.AppDB
+	appdb := *opsManager.Spec.AppDB
```
MaciejKaras, line 78: this should be part of api-changes PR

controllers/operator/appdbreplicaset_controller_test.go
```diff
@@ -168,7 +169,7 @@ func TestPublishAutomationConfigCreate(t *testing.T) {
 
 	monitoringAutomationConfig, err := buildAutomationConfigForAppDb(ctx, builder, kubeClient, omConnectionFactory.GetConnectionFunc, zap.S())
 	assert.NoError(t, err)
-	version, err = reconciler.publishAutomationConfig(ctx, opsManager, monitoringAutomationConfig, monitoringAutomationConfigSecretName(*appdb), memberCluster.SecretClient)
+	version, err = reconciler.publishAutomationConfig(ctx, opsManager, monitoringAutomationConfig, monitoringAutomationConfigSecretName(appdb), memberCluster.SecretClient)
```
MaciejKaras, line 172: this should be part of api-changes PR

controllers/operator/appdbreplicaset_controller_test.go
```diff
@@ -1855,6 +1856,10 @@ func createRunningAppDB(ctx context.Context, t *testing.T, startingMembers int,
 		SetServiceName(serviceName).
 		AddVolumeClaimTemplates(appDBStatefulSetVolumeClaimTemplates()).
 		SetReplicas(startingMembers).
+		// the operator sets the OM OwnerReference on the AppDB StatefulSet it creates
```
MaciejKaras: remove comment

controllers/operator/appdbreplicaset_controller_test.go
```diff
@@ -2115,10 +2120,232 @@ func TestMonitoringAgentStartupParametersIgnored(t *testing.T) {
 // monitoringAutomationConfigSecretName / monitoringAutomationConfigConfigMapName reproduce the
 // names of the pre-single-agent monitoring automation config resources. The operator no longer
 // creates these; the helpers exist only so tests can assert their absence (or exercise legacy paths).
-func monitoringAutomationConfigSecretName(appdb omv1.AppDBSpec) string {
+func monitoringAutomationConfigSecretName(appdb *omv1.AppDBSpec) string {
```
MaciejKaras, line 2120: should be part of api-changes

controllers/operator/appdbreplicaset_controller.go
MaciejKaras: fetching exact diff context

controllers/operator/appdbreplicaset_controller_multi_test.go
MaciejKaras: fetching exact diff context

controllers/operator/appdbreplicaset_controller_test.go
MaciejKaras: fetching exact diff context

controllers/operator/appdbreplicaset_controller_test.go
MaciejKaras: fetching exact diff context

controllers/operator/workflow/status_test.go
```diff
@@ -20,3 +20,9 @@ func TestOnErrorPrepend(t *testing.T) {
 	failedDecoratedValidationResult := failedValidationResult.OnErrorPrepend("failed wrapper").(*invalidStatus)
 	assert.Equal(t, "failed wrapper my failed validation", failedDecoratedValidationResult.msg)
 }
+
+func TestDisabledReconcileResultMatchesOK(t *testing.T) {
```
MaciejKaras: I would change this test to verify the Disabled() results match what is expected and not that it matches OK result.

controllers/operator/mongodbopsmanager_controller_multi_test.go
MaciejKaras, line 1: Are those changes really necessary?

controllers/operator/mongodbopsmanager_controller.go
```diff
@@ -420,25 +420,14 @@ func (r *OpsManagerReconciler) Reconcile(ctx context.Context, request reconcile.
...
 	}
 
+	r.configureWatchersForDynamicResources(ctx, opsManager, log)
+
 	// 1. Reconcile AppDB
-	emptyResult, _ := workflow.OK().ReconcileResult()
+	emptyResult := reconcile.Result{RequeueAfter: util.TWENTY_FOUR_HOURS}
 	retryResult := reconcile.Result{Requeue: true}
 
-	// TODO: make SetupCommonWatchers support opsmanager watcher setup
```
MaciejKaras: this todo can be removed

controllers/operator/workflow/status_test.go
MaciejKaras: fetching context

controllers/operator/mongodbopsmanager_controller_multi_test.go
MaciejKaras: fetching context

controllers/operator/appdbreplicaset_controller_test.go
MaciejKaras: fetching context

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -1577,7 +1695,7 @@ func configureMonitoring(ac *automationconfig.AutomationConfig, log *zap.Sugared
 				params[k] = v
 			}
 			if requireValidCert {
-				params["sslRequireValidMMSServerCertificates"] = "true"
+				params["sslRequireValidMMSServerCertificates"] = trueString
```
MaciejKaras: Why do we still call ensureAppDbPassword in this PR? All of those changes should be part of the https://github.com/mongodb/mongodb-kubernetes/pull/1459, no?

MaciejKaras: fetching context

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -1719,6 +1837,7 @@ func (r *ReconcileAppDbReplicaSet) generatePasswordAndCreateSecret(ctx context.C
 
 // ensureAppDbPassword will return the password that was specified by the user, or the auto generated password stored in
 // the secret (generate it and store in secret otherwise)
+// TODO is the forward and reverse migration take into account that the user can provide it's own secret with password?
```
MaciejKaras: Investigate if the forward and reverse migration will work properly if user provides it's own credentials through opsManager.Spec.AppDB.PasswordSecretKeyRef. Are the proper credentials will be used when moving forward? For reverse migration same question.

pkg/telemetry/collector.go
Julien-Ben, line 530: I suggest to add a brief comment here saying that if the value is nil it means the AppDB is an external MongoDB, so we report no value from the enum.

Julien-Ben: More generally, what are the telemetry implications of this change ?
When we look at the "resource types" deployed (RS, sharded, AppDB ...), will we be able to distinguish if a database is used as an AppDB, or as a "normal" database ?

Julien-Ben: I noticed after writing that there is a PR higher in the stack for telemetry, I'll take a look at it but feel free to resolve if this is handled

api/mongodb/v1/om/appdb_types.go
```diff
@@ -28,6 +28,21 @@ const (
 	ClusterTopologyMultiCluster = "MultiCluster"
 )
 
+// AppDBUserRoles defines required roles for the AppDB user are outlined in the documentation
```
viveksinghggits:
```suggestion
// AppDBUserRoles defines required roles for the AppDB user that are outlined in the documentation
```
?
this sentence seemed a little off.

pkg/telemetry/collector.go
MaciejKaras: yes yes, it is handled in the https://github.com/mongodb/mongodb-kubernetes/pull/1490

api/mongodb/v1/om/opsmanager_types.go
```diff
@@ -188,6 +188,9 @@ type ExternalApplicationDatabaseRef struct {
 	// +kubebuilder:validation:Enum=MongoDB
 	// +kubebuilder:validation:Required
 	Kind string `json:"kind"`
+
+	// Transient fields
+	Namespace string `json:"-"`
```
viveksinghggits, line 193: for my understanding why do we need this?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -747,11 +773,222 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+
+	passwordSecretName := omv1.OpsManagerUserPasswordSecretName(mdb.Name)
+	keyfileSecretName := fmt.Sprintf("%s-keyfile", mdb.Name)
+
+	for _, name := range []string{passwordSecretName, keyfileSecretName} {
+		s := corev1.Secret{}
+		if err := r.reconciler.client.Get(ctx, kube.ObjectKey(mdb.Namespace, name), &s); err != nil {
+			if errors.IsNotFound(err) {
+				continue
+			}
```
filipcirtog: We fetch this secret here and then fetch it again in claimSecretForCR, and the outer `s` is discarded. Could we collapse to a single Get by moving the NotFound tolerance into claimSecretForCR?

```
if err := r.reconciler.client.Get(ctx, kube.ObjectKey(mdb.Namespace, name), &s); err != nil {
			if errors.IsNotFound(err) {
				continue
}
```

MaciejKaras: Makes sense! Fixing it

filipcirtog (approved):

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: This a convenience way of storing additional information in the resource struct. We use it many places e.g. https://github.com/mongodb/mongodb-kubernetes/blob/92a6af18c9cd5da7b933ab566f29de80fa1a0925/api/mongodb/v1/om/opsmanager_types.go#L644-L649

api/mongodb/v1/om/opsmanager_types.go
```diff
@@ -770,6 +780,12 @@ func (om *MongoDBOpsManager) UpdateStatus(phase status.Phase, statusOptions ...s
 }
 
 func (om *MongoDBOpsManager) updateStatusAppDb(phase status.Phase, statusOptions ...status.Option) {
+	if om.Spec.ExternalApplicationDatabaseRef != nil {
+		om.Status.AppDbStatus = AppDbStatus{}
+		om.Status.AppDbStatus.UpdateCommonFields(phase, om.GetGeneration(), statusOptions...)
+		return
```
lucian-tosa, line 791: shouldn't this set the appdb status to disabled?

api/mongodb/v1/om/opsmanager_types.go
viveksinghggits: I think I understand the point you are making about storing namespace. But if the OM and the external AppDB resource are supposed to be in same namespace always, do we still need this field?

controllers/operator/common_controller.go
```diff
@@ -1035,7 +1034,15 @@ func publishAutomationConfigFirst(ctx context.Context, getter kubernetesClient.C
 		return false
 	}
 
+	if currentSts.Annotations[util.AppDBMigrationReadyAnnotation] == trueString {
+		log.Debugf("Statefulset %s has the forward migration ready annotation set to true.", currentSts.Name)
+		return false
+	}
+
 	databaseContainer := container.GetByName(util.DatabaseContainerName, currentSts.Spec.Template.Spec.Containers)
```
lucian-tosa, line 1042: what if the MongoDB resource was created in such a way that it disables TLS, whereas internal AppDB had TLS enabled?
Normally this procedure would cause this method to return true so that it first disables TLS in the automation config. Now it will return false, meaning that it will update the statefulset first, remove the TLS secrets, and then attempt to reconcile the AC. But now the agents are stuck because there are no certificates

lucian-tosa: another example is scaling down. Same issue, statefulset will be updated to remove the extra pods, but not removed from the AC first

controllers/operator/mongodbopsmanager_controller.go
```diff
@@ -2052,12 +2072,13 @@ func (r *OpsManagerReconciler) OnDelete(ctx context.Context, obj interface{}, lo
 		}
 	}
 
-	if opsManager.Spec.AppDB.IsMultiCluster() {
+	if opsManager.Spec.ExternalAppDBRef == nil && opsManager.Spec.AppDB.IsMultiCluster() {
```
lucian-tosa: nit: maybe worth a `IsInternalAppDB()` method

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -213,6 +219,17 @@ func (r *ReplicaSetReconcilerHelper) Reconcile(ctx context.Context) (reconcile.R
 		return r.updateStatus(ctx, status)
 	}
 
+	if rs.Spec.Role == mdbv1.RoleAppDB {
```
lucian-tosa: nit: same here, maybe `IsAppDB()`

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -747,11 +773,218 @@ func (r *ReplicaSetReconcilerHelper) updateOmDeploymentRs(ctx context.Context, c
...
+	if err := conn.ReadUpdateAutomationConfig(func(ac *om.AutomationConfig) error {
+		if sharedKey != "" {
+			ac.Auth.Key = sharedKey
+			return nil
+		}
+		if err := ac.EnsureKeyFileContents(); err != nil {
+			return xerrors.Errorf("failed to ensure keyfile contents: %w", err)
+		}
+		projectKey = ac.Auth.Key
+		return nil
```
lucian-tosa, line 913: isn't all of this already done in the reconciler when setting up scram auth?

lucian-tosa (changes requested):

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: Yes, you're are right, they share namespace. This is however convenient when you don't have access to OpsManager struct and only ExternalAppDBRef. This is the case for couple methods that receive ExternalAppDBRef as param and need to retrieve the referenced object and OpsManager struct was not passed.

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: well we are using this method `updateStatusAppDb` to set the appdb status to disabled in some calls, but what you are suggesting is overriding the `phase status.Phase` that is passed to this method. I would rather respect the input parameters and just don't set other status fields that only make sense for `spec.applicationDatabase` (internal appDB)

controllers/operator/common_controller.go
MaciejKaras: Hmm, I can test this method with returning true in the migration situation to make sure AC is updated first. I'll check this and let you know

api/mongodb/v1/om/opsmanager_types.go
viveksinghggits: hmmm, I am not really sure if it's good idea/api design to have namespace for external app DB when we know that they are going to be in the same namespace.
I have not generally seen this for example a pod refers to a secret or pvc resource but they don't have their own namespace resource because they are expected to be in the same namespace.
But if you think it's valuable and removing it would not be a good idea then it's ok.

controllers/operator/mongodbreplicaset_controller.go
MaciejKaras: I think creating the keyfile secret for the external AppDB doesn't make sense at all. For MongoDB deployments we are just setting it in the AC, no secret is created. I'll try to drop this functionality, thanks for noticing it.

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: This is unexported field and not part of our API. It's just internal field you and we don't ask users to fill it. And we set it to Ops Manager namespace during the struct initialisation so it cannot change.

viveksinghggits: I understand that it's not part of the CR and user can't set it. But it's still exported (capital N) right? Which doesn't matter btw.
Anyway, I think we can conclude this, I just wanted to highlight that if the field doesn't provide a lot of value, I have not seen this used commonly.

controllers/operator/common_controller.go
MaciejKaras: Ok, so publishing first the AC will not work, because the `updateOmDeploymentRs` calls WaitForReadyState. Processes from the Automation Config will not enter ready state, because the AppDB pods will still have headless agents running. We need to first recreate the Statefulset and then push the AC.

In cases that user has disabled the TLS or changed the member count, voting or whatever we can depend on the automatic recovery after 20 minutes that will push the AC.

Nevertheless I will add information in the Public Docs that we only support migrating with the same configuration as it was in the internal appDB. Does that sound reasonable?

lucian-tosa: I think that most of these things can be part of runtime validation. if spec.members == sts.replicas, if tls is enabled and a secret is mounted, etc.
I think that whatever causes the AC to be published first should be prevented. If it's not a huge change, it would make the UX a lot simpler and a lot less error-prone.
I would say, that if publishACFirst is true, the annotation is there and the role is AppDB than just return Failed.
Would that work?

api/mongodb/v1/om/opsmanager_types.go
lucian-tosa: My bad, didn't see where phase.Disabled is set

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: Yes, that is quite unorthodox, but useful. I'm just reusing existing logic, not creating new one. I will resolve the thread if you don't mind. I believe this is also a nit.

controllers/operator/common_controller.go
MaciejKaras: Yep, I think we should add some more validations to make sure there is nothing forcing the AC change.

controllers/operator/mongodbreplicaset_controller.go
MaciejKaras: Ok, so creating a keyfile and keeping it reclaiming it during the reverse migration is required to not have storedKey mismatch. Some of the old agents (external AppDB) will use the keyfile from the AC while new agents (internal AppDB) will use new, just created keyfile and prevent agents from talking to each other. This is the reason we keep the keyfile secret during the forward migration and create one during fresh start.

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_forward.py
lucian-tosa, line 1: I think the automation agent password will change during this.
It's probably not a problem if the tests are passing, but we should be aware of this.
There are mechanisms to ensure the password is not changed by storing in a secret. However, I think that the secret generated by the appdb reconciler has a different name from the one read by the mongodb reconciler

lucian-tosa, line 1: this should also assert that the keyfile does not change during migrations

MaciejKaras: You mean the `-om-password` that is used to create `mongodb-ops-manager` user in the DB? We are reusing existing one in the `ensureAppDBRoleUser`. But if they specified custom passwordSecretKeyRef we cannot reuse it and will generate a new one.

controllers/operator/common_controller.go
MaciejKaras: I've added those validations in a separate PR, please take a look https://github.com/mongodb/mongodb-kubernetes/pull/1534/changes

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_forward.py
MaciejKaras: @igor-karpukhin we just need to add one more assertion in the test for the keyfile

lucian-tosa: not the password of the OM user, but the password of the automation agent user (auth.autoPwd in the AC). I think AppDB uses `-agent-password` and MongoDB uses `-agent-auth-secret`

igor-karpukhin: Fixed

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_forward.py
```diff
@@ -0,0 +1,220 @@
...
+        # change across the forward migration (and therefore OM pods never roll)
+        assert primary_om.read_appdb_connection_url() == self.connection_string_before
+
+    def test_password_secret_unchanged_after_forward_migration(self, namespace: str):
+        password_secret_now = read_secret(namespace, password_secret_name(OM_NAME))
+        assert password_secret_now == self.password_secret_before
+
+    def test_keyfile_secret_unchanged_after_forward_migration(self, namespace: str):
+        keyfile_secret_now = read_secret(namespace, f"{DB_NAME}-keyfile")
+        assert keyfile_secret_now == self.keyfile_secret_before
```
lucian-tosa, line 141: can we assert the same on reverse migration?

igor-karpukhin: For the reverse migration we allow credentials to be changed. See the test description: https://github.com/mongodb/mongodb-kubernetes/blob/fa2263e3752bf4b6d64dc92cd7009af2eabb6528/docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_forward.py#L151-L152

api/mongodb/v1/om/opsmanager_types.go
```diff
@@ -309,9 +312,14 @@ func (ms MongoDBOpsManagerSpec) GetOpsManagerCA() string {
 }
 
 func (ms MongoDBOpsManagerSpec) GetAppDbCA() string {
+	if ms.ExternalAppDBRef != nil {
```
lucian-tosa, line 315: nit: replace with IsInternalAppdb()

api/mongodb/v1/om/opsmanager_types.go
```diff
@@ -770,6 +785,12 @@ func (om *MongoDBOpsManager) UpdateStatus(phase status.Phase, statusOptions ...s
 }
 
 func (om *MongoDBOpsManager) updateStatusAppDb(phase status.Phase, statusOptions ...status.Option) {
+	if om.Spec.ExternalAppDBRef != nil {
```
lucian-tosa, line 788: nit: IsInternalAppDB()

lucian-tosa (approved):
LGTM, there is one comment about the automation agent's password.

PR #1442, opened 2026-07-29
CLOUDP-400899: Migrate multi-cluster doc snippets to the MemberCluster flow

m1kola:
# Summary

MCK is moving multi-cluster configuration from install time to day 2: instead of running `kubectl mongodb multicluster setup` and passing `--set multiCluster.clusters` at helm install, you install the operator plainly and then register member clusters with `kubectl mongodb multicluster generate-member-resources` / `generate-member-registration`, which emit `MemberCluster` CRs.

This PR migrates the user-facing multi-cluster tutorials to the new flow, ahead of removing the legacy one: the multi-cluster Search scenarios (`docs/search/12-search-rs-multi-cluster`, `docs/search/13-search-sharded-multi-cluster`, run in CI on every PR patch) and the GKE reference architecture (`public/architectures/setup-multi-cluster/ra-02`). Each now installs the operator plainly, then configures member clusters in a separate snippet.

Two things worth knowing:
- The e2e env of the Search snippets strips `multiCluster.clusters` from the helm values the test harness still injects in multi-cluster mode — passing it renders a kubeconfig volume whose Secret is never created and wedges the operator pod in `FailedMount`. Marked with a TODO to remove together with the injection itself.
- `public_gke_code_snippets` (released chart + released plugin, manual-only) cannot pass until MCK 2.x ships — the 1.x artifacts lack the `MemberCluster` CRD and the new plugin commands. It should be validated with the release command against the first 2.x RC.

## Proof of Work

- e2e: N/A (docs/snippets only, no runtime code) — the snippets are validated by targeted Evergreen patches:
- [Patch 474](https://spruce.corp.mongodb.com/version/6a6a1608bfb9970007a335c0): kind Search multi-cluster snippets (replica set + sharded) — **green**
- [Patch 473](https://spruce.corp.mongodb.com/version/6a6a0b3752ec7700073cc519): GKE reference architecture (ra-02 on real GKE), mesh and no-mesh tasks — **green** (its two kind Search tasks are red there because the patch predates the `FailedMount` fix; both pass on patch 474)

## Checklist

- [x] Have you added changelog file? (`skip-changelog` label; the breaking-change entry lands with the removal PR)


docs/search/12-search-rs-multi-cluster/code_snippets/12_0110_configure_member_clusters.sh
```diff
@@ -0,0 +1,37 @@
+# TODO(m1kola): token-wait: Simplify this workflow. We should probably make `generate-member-registration` wait for token.
```
m1kola, line 1: I don't like the fact that we need extra ugly code to wait for the secret to be populated by a token. Not a nice UX. So I'm planning on adding this wait to `generate-member-registration` since it is fetching the secret anyway.

It will simplify UX visible in this snippet and E2E tests as well.

But first I want to complete cleanup of the old commands (needs at least one more PR on top this one).

docs/search/12-search-rs-multi-cluster/code_snippets/12_0110_configure_member_clusters.sh
```diff
@@ -0,0 +1,37 @@
...
+
+echo "Configuring member clusters (RBAC + MemberCluster registration)..."
+
+# Render and apply member-cluster RBAC (ServiceAccount, token Secret, Roles and
```
josvazg, line 5: side comment: if these are instructions for customers, we might need to work on the ergonomics of using our product and tooling, seems more complex than in should.

Claude was flagging some possible corner cases that might trip customers at times, specially around piping commands and how to handle when the tool errors out in between pipes.

josvazg (approved):

docs/search/12-search-rs-multi-cluster/code_snippets/12_0110_configure_member_clusters.sh
m1kola: @josvazg yes, these are for users [per TD](https://docs.google.com/document/d/1twAOo9jG4kHwnnmLC5GSO21sGByKOiMtwOrfkjz2Xpo/edit?tab=t.55vyfwkywh4p).

One thing that I don't like and I didn't foresee on the TD stage is the fact that we need to wait for a token to be populated in the secret between applying `generate-member-resources` and running `generate-member-registration`. This resulted in ugly wait code here and in E2E tests. For this I left a todo above: https://github.com/mongodb/mongodb-kubernetes/pull/1442/changes#r3676080870

Once this is gone - we will end up with 2 loops:
* `generate-member-resources` + `kubectl apply`
* `generate-member-registration` + `kubectl apply`

Alternatively we can drop the loop and duplicate commands with different members. That might be better for the docs.

For now I want to make snippets work with new tooling and keep going with the implementation. And polish later.

mircea-cosbuc (approved):

PR #1448, opened 2026-07-30
Add AGENTS.md and correct stale developer docs

nammn:
# Summary

Adds `AGENTS.md` — a 25-line file listing repo-specific gotchas where the obvious default
is wrong and no existing tool (linter, precommit, CI) catches the mistake. Each entry is a
silent failure: green build, wrong result.

The four gotchas:

1. **`make precommit` rewrites tracked files** — not read-only; rewrites `release.json`,
   `helm_chart/values.yaml`, CRDs, `public/*.yaml` and auto-stages them.
2. **CRDs are sentinel-gated on `api/*.go`** — `make manifests` only regenerates CRDs when
   `api/*.go` changes. Hand-editing CRDs without touching `api/` → precommit passes, wrong
   CRDs ship.
3. **venv is `./venv`, not `.venv`** — `.venv` is silently ignored by repo scripts.
4. **`release.json` agent version overrides** — OM variants read `opsManagerMapping`, not
   top-level `agentVersion`; context files can hardcode `AGENT_VERSION`; empty
   `MDB_CUSTOM_AGENT_URL` silently falls back to production.

The second commit fixes stale developer docs found while verifying the above:

| File | Correction |
|---|---|
| `CONTRIBUTING.md` | Changelog filenames documented with dashes; `changelog.py` requires underscores |
| `scripts/release/create_changelog.py` | `--editor` help said "Default is True" but flag is `store_true` |
| `docker/mongodb-kubernetes-tests/README.md` | Pointed at `evergreen.yaml` and removed `e2e_core_task_group`; tasks now in `.evergreen-tasks.yml`, groups in `.evergreen.yml`. Adds marker/task-name match requirement and `find_test_variants.py` check |

## What we gain

Evidence drawn from two sources: 29 handoff documents (July 9–30) and the OpenCode
conversation database (1,163 sessions, $529 total API cost on this repo).

| Gotcha | Incidents (from conversation DB) | API cost wasted | Agent time wasted |
|---|---|---|---|
| `make precommit` rewrites files | 5 sessions: agent confused by EXIT=2 and re-ran precommit multiple times (KUBE-267, RFC 7616, prek migration); skipped precommit entirely and hand-edited generated files (JVM optimization) | ~$8.76 | ~3-4 rounds across sessions |
| CRD sentinel | Discovered by reviewer subagent during this PR's review — hand-editing CRDs bypasses `make manifests` silently | — | latent — caught by review, not tooling |
| venv `.venv` silently ignored | 1 session: agent created `.venv`, no pip installed, debugged venv activation, user asked "why .venv not venv?", agent had to redo everything | ~$0.00 (0.56M tokens) | ~1 full setup redo |
| `release.json` agent version overrides | 3 incidents: bumped `agentVersion` for OM variants (2 green-but-wrong patches), `om80` context hardcoded `AGENT_VERSION` (1 wasted patch), empty `MDB_CUSTOM_AGENT_URL` fell back to production (2 failed variants) | ~$1.81 | **~5–6h, 3+ wasted Evergreen patches** |

**Total: ~$10.57 API cost + ~6–8h agent time** spent rediscovering these four silent
failures. The file is 25 lines — it pays for itself on the first avoided patch.

## Proof of Work

- `make precommit` passes clean on the branch.
- Each gotcha verified against source: `check_precommit.sh`, `.pre-commit-config.yaml`,
  `Makefile` (sentinel logic), `set_env_context.sh`, `release.json` (`opsManagerMapping`),
  `scripts/dev/contexts/variables/om80` (hardcoded `AGENT_VERSION`).
- The CRD sentinel trap confirmed: `.generated/crd.sentinel` depends only on `$(API_SOURCES)`
  (`Makefile:322`); precommit `generate-manifests` calls `make manifests` which is a no-op
  when `api/` is unchanged.
- e2e: N/A (docs only).

## Checklist


ashortt-mdb:
Love this — especially mining the session history for evidence of what actually burned time. Went through my own accumulated local notes with your filter (silent failure, green-but-wrong, no tooling catches it) and here are some candidates you might consider adding. Offering softly — each would need the same verify-against-source treatment as your existing entries before landing:

1. **EVG task wiring is two-step, and half-wired is silently green.** A task defined in `.evergreen-tasks.yml` but not added to a task group in `.evergreen.yml` never runs — no error anywhere. Your README fix covers the marker↔task-name half; this is the sibling trap. (Related: a task whose tests all `skip` is permanently green.) The `scripts/python/find_test_variants.py` check you added to the README might deserve promotion into AGENTS.md itself.
2. **Never build an `om.Deployment` from scratch** — always read-modify-write via `conn.ReadUpdateDeployment` (`controllers/om/omclient.go`). The automation config is a deliberately untyped map that OM owns; constructing one clobbers OM-side settings and other controllers' processes in shared projects. Looks fine until someone else's deployment breaks.
3. **Two automation-config models — don't conflate them.** Enterprise: untyped `controllers/om.Deployment` pushed over REST. Community/AppDB: typed `pkg/automationconfig` serialized into a Secret the agent reads directly. Grepping "automation config" finds both; picking the wrong one compiles and passes tests.
4. **Two independent StatefulSet builders.** Enterprise `controllers/operator/construct/` vs community `mongodb-community-operator/controllers/construct/` — same conventions, zero shared code. Fixing pod construction in one does not fix the other.
5. **TLS cert rotation may not roll pods on the default k8s-Secret backend** — the Vault path stamps the cert hash into pod annotations (forced roll), the plain Secret path doesn't. Rotating a cert and seeing no restart looks like success. (This one I'd want re-verified against current master before it ships.)
6. **Don't extend the `MongoDBMultiCluster` (mdbmulti) pattern for new multi-cluster work** — it's the most obvious-looking MC reference in the tree but AppDB-MC / sharded-MC are the canonical patterns.

One structural thought: the AGENTS.md convention also supports **nested per-directory files** that agents only load when working in that subtree. Items 2–4 above naturally live next to the code they describe (`controllers/om/`, `pkg/`, …) rather than at root — and nested content is pay-per-use token-wise, while root content is paid on every session. Might be worth keeping the root file for the universal traps (as you have it) and trialling one nested file as a follow-up. Happy to discuss / contribute entries either way.

ashortt-mdb:
To make the nested idea concrete — here's what a couple of nested files could look like, condensed from notes I've been keeping privately (each claim grep-verified against current master today). Same bar as the root file: gotchas where the default is wrong and nothing catches you, kept terse so the token cost stays trivial.

**`controllers/om/AGENTS.md`**

```markdown
# controllers/om — Ops Manager client + automation config

## Never build a `Deployment` from scratch
The automation config is a deliberately untyped `map[string]interface{}` — OM owns the
schema; the operator merges only the fields it manages and must round-trip the rest
untouched. Always read-modify-write via `conn.ReadUpdateDeployment(fn, log)`
(`omclient.go`). Constructing one clobbers OM-side settings and other controllers'
processes in shared projects.

## Auth is NOT a `Deployment` method
It lives on the typed `AutomationConfig` wrapper (`automation_config.go`), applied via
`conn.ReadUpdateAutomationConfig` — mutating the map directly silently drops auth changes.

## This is the *enterprise* AC path only
Community/AppDB use the typed `pkg/automationconfig` model serialized into a Secret
(key `cluster-config.json`) that the agent reads directly — no OM round-trip. Different
model; don't mix. Post-merge, `limitVotingMembers` (`deployment.go`) caps voting members
at 7 — votes beyond that are zeroed operator-side, not by MongoDB config you wrote.
```

**`controllers/operator/AGENTS.md`**

```markdown
# controllers/operator — enterprise reconcilers

## AC-first vs STS-first ordering is load-bearing
Disabling TLS/X509/agent-auth or scaling down must push the automation config *before*
the StatefulSet (`publishAutomationConfigFirst`, `common_controller.go`; executed via
`workflow.RunInGivenOrder`) — wrong order wedges agents on stale mounts. The community
reconciler has its own independent version (`shouldRunInOrder`).

## Two independent StatefulSet builders
`controllers/operator/construct/` (enterprise) and
`mongodb-community-operator/controllers/construct/` (community) share conventions but
zero code. Fixing pod construction in one does not fix the other.

## Deployment-state ConfigMaps are a persisted compat surface
Sharded/AppDB/OM state (`<name>-state` ConfigMap) is migrated on operator upgrade; fields
can be nil on old state. New fields must be nil-tolerant or migrated — the failure is a
panic in the field, not in CI.

## Don't extend the MongoDBMultiCluster (mdbmulti) pattern
It looks like the canonical multi-cluster reference; it isn't. AppDB-MC and sharded-MC
(per-cluster StatefulSets, persisted ClusterMapping) are the patterns for new MC work.
```

Root stays as you have it (universal traps), these load only when an agent is actually in the subtree. If this direction lands, one repo-wide rule is worth committing alongside: **every name asserted in these files must be grep-verifiable at the stated location** — when I fact-checked my own first drafts of exactly this kind of doc, about a quarter of the load-bearing specifics were plausible-but-wrong (paraphrased function names, wrong Secret keys). Stale orientation docs are worse than none.

PR #1454, opened 2026-08-03
CLOUDP-400899, CLOUDP-400894, CLOUDP-400893: Use member-scoped ServiceAccounts for workloads on member clusters

m1kola:
# Summary

Workload pods on member clusters now run under member-scoped ServiceAccounts (`mck-member-<cluster>-{appdb,database-pods,ops-manager}`) instead of the fixed `mongodb-kubernetes-*` names the operator hard-coded; single-cluster and legacy-central clusters keep the Helm-install names, and user pod-template overrides still win. `database-roles.yaml` is now dual-mode: byte-identical when rendered by `helm install`/OLM, member-scoped with `mongodb.com/rbac-version` when rendered by `kubectl mongodb multicluster generate-member-resources` — the generated member RBAC no longer overlaps with Helm/OLM-managed resources. The operator resolves the ServiceAccount per cluster at pod construction for all workload types (AppDB, Ops Manager/backup, MongoDB multi-cluster, `MongoDBMultiCluster`, mongot). The PR also removes the `operator.createResourcesServiceAccountsAndRoles` and `operator.createOperatorServiceAccount` Helm values (base install always deploys RBAC) and adjusts the multi-cluster E2E harness and search snippets accordingly. Context: `docs/dev/multi-cluster-config-tooling.md` (slice 7).

## Proof of Work

- `make precommit`; `go test ./controllers/... ./pkg/...` — incl. new table-driven tests for SA derivation, pod construction, plugin render + `helm template` parity.
- Base-mode Helm render byte-identical (default + OpenShift values); `generate_standalone_yaml` → `config/` and `public/*.yaml` no-diff; `helm unittest` 11/11; `helm lint`.

## Checklist


docker/mongodb-kubernetes-tests/tests/multicluster/multi_2_cluster_clusterwide_replicaset.py
```diff
@@ -184,34 +184,19 @@ def test_create_namespaces(
...
-@pytest.mark.e2e_multi_cluster_2_clusters_clusterwide
-def test_deploy_operator(multi_cluster_operator_clustermode: Operator):
-    multi_cluster_operator_clustermode.wait_for_operator_ready()
+@pytest.fixture(scope="module")
+def install_operator(namespace: str, mdba_ns: str, mdbb_ns: str, member_cluster_names: List[str]) -> Operator:
+    # The operator watches all namespaces, so member RBAC is rendered cluster-wide (default).
+    # The MongoDBMulti workloads run in mdba_ns/mdbb_ns (the central cluster doubles as a member):
+    # an extra narrowed render seeds the member-scoped workload SAs/Roles there.
+    operator = get_multi_cluster_operator_clustermode(namespace)
+    generate_and_apply_member_resources(member_cluster_names, namespace, watched_namespaces=mdba_ns + "," + mdbb_ns)
```
m1kola, line 193: This surfaced an issue with CLI design: how we pass watched namespaces for cluster-wide multi-cluster deployments and how we design RBAC.

This is a workaround. I'm leaving it as is for to avoid this PR growing bigger, but I'm actively working on solving this CLI issue and it will go away once is solved.

controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -2455,6 +2456,7 @@ func (r *ShardedClusterReconcileHelper) getConfigServerOptions(ctx context.Conte
 		Replicas(scale.ReplicasThisReconciliation(r.GetConfigSrvScaler(memberCluster))),
 		StatefulSetNameOverride(r.GetConfigSrvStsName(memberCluster)),
 		ServiceName(r.GetConfigSrvServiceName(memberCluster)),
+		ServiceAccountName(resourcenames.WorkloadDatabasePodsServiceAccount.Name(memberCluster.Name, memberCluster.Legacy)),
```
mircea-cosbuc, line 2459: Is it not possible for a non-base install to still need the standard service names? I'm not sure how a cluster being marked as legacy should have an impact on the construction respecting SA name overrides or not.

mircea-cosbuc (approved):

controllers/operator/mongodbshardedcluster_controller.go
m1kola: I answered in Slack. Adding just for transparency.

Some of the controllers are implemented in a such way they are supposed to handle single- and multi-cluster in a similar way. So we have a "cluster" with name `__default` in a single-cluster setup which simulates multi-cluster for these controllers. And this flag is how you distinguish between the real multi-cluster and single-cluster basically.

For single-cluster (`Legacy` is `false`) we need to use fixed RBAC names because in single-cluster we do not deploy RBAC for member cluster: we have to use the one installed with the operator.

m1kola: > Is it not possible for a non-base install to still need the standard service names?

And to answer this more directly. Non-base install (real multi-cluster) will need to use member cluster SA provisioned separately from base installation (day 2 config).

roothorp (approved):

docker/mongodb-kubernetes-tests/tests/multicluster/multi_2_cluster_clusterwide_replicaset.py
```diff
@@ -184,34 +184,19 @@ def test_create_namespaces(
...
+    return operator
```
josvazg, line 195: > The second call re-renders both member-cluster-rbac.yaml and database-roles.yaml (per pkg/kubectl-mongodb/memberresources/memberresources.go:30-33), not just the workload SA template. That produces redundant RoleBindings for the operator's own role in mdba_ns/mdbb_ns/namespace, duplicating the ClusterRoleBinding from the first call. Harmless here, but it's dead weight and could mask a future regression where someone expects only the SA template to be affected.

**q** : is this a valid concern?

josvazg (commented):
>  No cleanup/idempotency for those leftover namespaced RoleBindings across reruns — pre-existing pattern in this harness, not new, but worth a TODO.

q: do we have a expected usage pattern or tooling about re-running the generation? I guess not, the user must know what to do

docker/mongodb-kubernetes-tests/tests/multicluster/multi_2_cluster_clusterwide_replicaset.py
```diff
@@ -184,34 +184,19 @@ def test_create_namespaces(
     )
 
 
-@pytest.mark.e2e_multi_cluster_2_clusters_clusterwide
-def test_deploy_operator(multi_cluster_operator_clustermode: Operator):
-    multi_cluster_operator_clustermode.wait_for_operator_ready()
+@pytest.fixture(scope="module")
+def install_operator(namespace: str, mdba_ns: str, mdbb_ns: str, member_cluster_names: List[str]) -> Operator:
```
josvazg, line 188: > Fixture now takes member_cluster_names: List[str] — confirm no other module-scoped fixture in this test file caches an operator instance built with the old signature (didn't find one, but worth a final grep before merging).

Just in case this is relevant

josvazg (approved):
LGTM, treat all my queries as non blockers

docker/mongodb-kubernetes-tests/tests/multicluster/multi_2_cluster_clusterwide_replicaset.py
m1kola: Yes, it is. And it is a poor UX. This is something I'm fighting with right now. Trying to come up with a better option.

[My comment](https://github.com/mongodb/mongodb-kubernetes/pull/1454/changes#r3714311073) above is relevant.

docker/mongodb-kubernetes-tests/tests/multicluster/multi_2_cluster_clusterwide_replicaset.py
m1kola: Not sure that I understand the (potential) issue. `member_cluster_names` is a fixture.

m1kola:
> > No cleanup/idempotency for those leftover namespaced RoleBindings across reruns — pre-existing pattern in this harness, not new, but worth a TODO.
> 
> q: do we have a expected usage pattern or tooling about re-running the generation? I guess not, the user must know what to do

@josvazg not sure that I follow. Is it about cleaning up K8s resources after tests? Each test runs with own cluster/set of clusters which then torn down. Since the clusters are throw away, we normally don't do cleanups in E2Es.

josvazg:
> > > No cleanup/idempotency for those leftover namespaced RoleBindings across reruns — pre-existing pattern in this harness, not new, but worth a TODO.
> > 
> > 
> > q: do we have a expected usage pattern or tooling about re-running the generation? I guess not, the user must know what to do
> 
> @josvazg not sure that I follow. Is it about cleaning up K8s resources after tests? Each test runs with own cluster/set of clusters which then torn down. Since the clusters are throw away, we normally don't do cleanups in E2Es.

I guess is what happens if you generate and then generate again on top of the same file set. As I said probably a non issue anyways. I agree, e2es do not usually need cleanups as the end is reaped at test completion.

m1kola:
> I guess is what happens if you generate and then generate again on top of the same file set. As I said probably a non issue anyways.

Oh, yeah. That should be fine: if you give the same input to the same version of the tool it should render the identical resources. We even have a test (introduced earlier) which compares output of rendering directly via Helm and via CLI (but it is there for an unrelated reason - just to make sure that embeded chart does not drift).

PR #1459, opened 2026-08-04
CLOUDP-429011 - Extract AppDBReconciler interface (refactor)

MaciejKaras:
# Summary

Pure refactor extracting the `AppDBReconciler` interface and related helper methods from the internal AppDB reconciler, plus a fix for `ensureAppDbPassword` when `PasswordSecretKeyRef` is set but the referenced secret doesn't exist.

- Introduces `AppDBReconciler` interface in `controllers/operator/appdbreplicaset_reconciler.go`
- `createNewAppDBReconciler` returns `AppDBReconciler` instead of the concrete `*ReconcileAppDbReplicaSet`
- Extracts `readAppDbPassword` from `ensureAppDbPassword` for reuse
- Extracts watcher setup into `configureWatchersForDynamicResources`
- Changes `emptyResult` to explicit `reconcile.Result{RequeueAfter: util.TWENTY_FOUR_HOURS}`
- **Fix**: `ensureAppDbPassword` now returns an error instead of silently generating a random password when `spec.appDB.passwordSecretKeyRef` is set but the referenced secret does not exist. This prevents the controller from falling back to an auto-generated password when the user has explicitly configured a custom secret — the user must create the secret first, and the reconcile will requeue on the next attempt.

This is the base of a 3-PR stack; API surface arrives in #1439, reconciliation logic in #1440.

## Proof of Work

- `go vet ./...` green
- `go test ./controllers/operator/` green (including new `TestEnsureAppDbPassword_ErrorWhenPasswordSecretKeyRefNotFound`)
- `pre-commit run golangci-lint-full --all-files` Passed

## Checklist

- [x] Have you added changelog file? (skip-changelog — pure refactor, no functional impact)

### DIFF
--- a/changelog/20260804_fix_external_appdb_forward_migration_password.md
+++ b/changelog/20260804_fix_external_appdb_forward_migration_password.md
@@ -1,0 +1,6 @@
+---
+kind: fix
+date: 2026-08-04
+---
+
+* **MongoDBOpsManager**: `spec.applicationDatabase.passwordSecretKeyRef` now fails reconciliation when the referenced secret does not exist, instead of silently generating a random password.
--- a/changelog/20260804_fix_external_appdb_forward_migration_password.md
+++ b/changelog/20260804_fix_external_appdb_forward_migration_password.md
@@ -3,4 +3,4 @@ kind: fix
 date: 2026-08-04
 ---
 
-* **MongoDBOpsManager**: `spec.applicationDatabase.passwordSecretKeyRef` now fails reconciliation when the referenced secret does not exist, instead of silently generating a random password.
+* **MongoDBOpsManager**:  `MongoDBOpsManager` resource now fails reconciliation, when the secret referenced by the field `spec.applicationDatabase.passwordSecretKeyRef` does not exist, instead of silently generating a random password.
--- a/controllers/operator/appdbreplicaset_controller.go
+++ b/controllers/operator/appdbreplicaset_controller.go
@@ -1727,7 +1727,8 @@ func (r *ReconcileAppDbReplicaSet) ensureAppDbPassword(ctx context.Context, opsM
 	if err != nil {
 		if secret.SecretNotExist(err) {
 			if passwordRef != nil && passwordRef.Name != "" {
-				return "", xerrors.Errorf("password secret %s referenced by spec.applicationDatabase.passwordSecretKeyRef does not exist", passwordRef.Name)
+				secretObjectKey := kube.ObjectKey(opsManager.Namespace, passwordRef.Name)
+				return "", xerrors.Errorf("password secret %s referenced by spec.applicationDatabase.passwordSecretKeyRef does not exist", secretObjectKey)
 			}
 			log.Debugf("Generated AppDB password and storing in secret/%s", opsManager.Spec.AppDB.GetOpsManagerUserPasswordSecretName())
 			return r.generatePasswordAndCreateSecret(ctx, opsManager, log)
@@ -1757,13 +1758,14 @@ func (r *ReconcileAppDbReplicaSet) ensureAppDbPassword(ctx context.Context, opsM
 func (r *ReconcileAppDbReplicaSet) readAppDbPassword(ctx context.Context, opsManager *omv1.MongoDBOpsManager, log *zap.SugaredLogger) (string, error) {
 	passwordRef := opsManager.Spec.AppDB.PasswordSecretKeyRef
 	if passwordRef != nil && passwordRef.Name != "" {
-		if passwordRef.Key == "" {
-			passwordRef.Key = "password"
+		passwordKey := "password"
+		if passwordRef.Key != "" {
+			passwordKey = passwordRef.Key
 		}
 
 		secretObjectKey := kube.ObjectKey(opsManager.Namespace, passwordRef.Name)
 		log.Debugf("Reading AppDB password from secret %s", secretObjectKey)
-		return secret.ReadKey(ctx, r.SecretClient, passwordRef.Key, secretObjectKey)
+		return secret.ReadKey(ctx, r.SecretClient, passwordKey, secretObjectKey)
 	}
 
 	secretObjectKey := kube.ObjectKey(opsManager.Namespace, opsManager.Spec.AppDB.GetOpsManagerUserPasswordSecretName())
--- a/controllers/operator/appdbreplicaset_controller.go
+++ b/controllers/operator/appdbreplicaset_controller.go
@@ -840,9 +840,10 @@ func (r *ReconcileAppDbReplicaSet) ReconcileAppDB(ctx context.Context, opsManage
 	return r.updateStatus(ctx, opsManager, workflow.OK(), log, appDbStatusOption, status.AppDBMemberOptions(appDBScalers...), status.NewPVCsStatusOptionEmptyStatus())
 }
 
-// BuildAppDBConnectionURL returns the connection string to the AppDB, ensuring the Ops Manager user password exists.
+// BuildAppDBConnectionURL returns the connection string to the AppDB, reading the Ops Manager user password.
+// It assumes ReconcileAppDB has already been called and the password secret exists.
 func (r *ReconcileAppDbReplicaSet) BuildAppDBConnectionURL(ctx context.Context, opsManager *omv1.MongoDBOpsManager, log *zap.SugaredLogger) (string, error) {
-	password, err := r.ensureAppDbPassword(ctx, opsManager, log)
+	password, err := r.readAppDbPassword(ctx, opsManager)
 	if err != nil {
 		return "", xerrors.Errorf("Error getting AppDB password: %w", err)
 	}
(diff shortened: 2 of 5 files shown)


controllers/operator/appdbreplicaset_controller.go
```diff
@@ -1720,58 +1721,51 @@ func (r *ReconcileAppDbReplicaSet) generatePasswordAndCreateSecret(ctx context.C
...
-	appDbPasswordSecretStringData, err := secret.ReadStringData(ctx, r.SecretClient, secretObjectKey)
+	return password, nil
+}
 
-	if secret.SecretNotExist(err) {
-		// create the password
-		if password, err := r.generatePasswordAndCreateSecret(ctx, opsManager, log); err != nil {
-			return "", err
-		} else {
-			log.Debugf("Using auto generated AppDB password stored in secret/%s", opsManager.Spec.AppDB.GetOpsManagerUserPasswordSecretName())
```
josvazg, line 1765: PasswordSecretKeyRef is *userv1.SecretKeyRef, so passwordRef is a pointer into the opsManager struct. The assignment on line 1762 mutates the object passed in.

Harmless in practice because ensureAppDbPassword (line 1724) calls readAppDbPassword first, so later callers see .Key already populated. But the method name is misleading for anyone calling readAppDbPassword directly.

Do we want to keep such mutation? is so, should we rename the method so that it is clear that it can change data.

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -840,9 +840,10 @@ func (r *ReconcileAppDbReplicaSet) ReconcileAppDB(ctx context.Context, opsManage
 	return r.updateStatus(ctx, opsManager, workflow.OK(), log, appDbStatusOption, status.AppDBMemberOptions(appDBScalers...), status.NewPVCsStatusOptionEmptyStatus())
 }
 
-// BuildAppDBConnectionURL returns the connection string to the AppDB, ensuring the Ops Manager user password exists.
+// BuildAppDBConnectionURL returns the connection string to the AppDB, reading the Ops Manager user password.
+// It assumes ReconcileAppDB has already been called and the password secret exists.
 func (r *ReconcileAppDbReplicaSet) BuildAppDBConnectionURL(ctx context.Context, opsManager *omv1.MongoDBOpsManager, log *zap.SugaredLogger) (string, error) {
-	password, err := r.ensureAppDbPassword(ctx, opsManager, log)
+	password, err := r.readAppDbPassword(ctx, opsManager)
```
josvazg: This is no longer a build, does not mutate or create data, so the name is also misleading as with readAppDbPassword. Right?

controllers/operator/mongodbopsmanager_controller.go
```diff
@@ -979,6 +968,17 @@ func (r *OpsManagerReconciler) createBackupDaemonStatefulset(ctx context.Context
...
+	appDBReplicaSet := opsManager.Spec.AppDB
+	r.SetupCommonWatchers(&appDBReplicaSet, nil, nil, appDBReplicaSet.GetName())
+
+	if opsManager.IsTLSEnabled() {
+		r.resourceWatcher.RegisterWatchedTLSResources(opsManager.ObjectKey(), opsManager.Spec.GetOpsManagerCA(), []string{opsManager.TLSCertificateSecretName()})
+	}
+
+	r.watchMongoDBResourcesReferencedByBackup(ctx, opsManager, log)
+}
+
```
josvazg, line 983: The old inline code (pre-commit) had this comment explaining the ordering constraint:
// The order matters here, since appDB and opsManager share the same reconcile ObjectKey being opsmanager crd
// That means we need to remove first, which SetupCommonWatchers does, then register additional watches
That rationale was not carried into the extracted function. The behavior is preserved — SetupCommonWatchers is still called before RegisterWatchedTLSResources and watchMongoDBResourcesReferencedByBackup — but anyone modifying configureWatchersForDynamicResources in the future won't know that reordering these three calls is unsafe.

josvazg (approved):
Nothing major, so OK

changelog/20260804_fix_external_appdb_forward_migration_password.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: fix
+date: 2026-08-04
+---
+
+* **MongoDBOpsManager**: `spec.applicationDatabase.passwordSecretKeyRef` now fails reconciliation when the referenced secret does not exist, instead of silently generating a random password.
```
viveksinghggits: nit:
```suggestion
* **MongoDBOpsManager**:  `MongoDBOpsManager` resource now fails reconciliation, when the secret referenced by the field `spec.applicationDatabase.passwordSecretKeyRef` does not exist, instead of silently generating a random password.
```

controllers/operator/appdbreplicaset_reconciler.go
```diff
@@ -0,0 +1,23 @@
...
+// future implementations can use alternative AppDB backends without changing the
+// Ops Manager controller.
+type AppDBReconciler interface {
+	// ReconcileAppDB brings the AppDB to the desired state.
+	ReconcileAppDB(ctx context.Context, opsManager *omv1.MongoDBOpsManager) (reconcile.Result, error)
+
+	// BuildAppDBConnectionURL returns the MongoDB connection string OpsManager/BackupDaemon
+	// should use to reach the AppDB.
+	BuildAppDBConnectionURL(ctx context.Context, opsManager *omv1.MongoDBOpsManager, log *zap.SugaredLogger) (string, error)
+}
```
viveksinghggits, line 23:
```suggestion
type AppDBReconciler interface {
	// ReconcileAppDB brings the AppDB to the desired state.
	Reconcile(ctx context.Context, opsManager *omv1.MongoDBOpsManager) (reconcile.Result, error)

	// BuildAppDBConnectionURL returns the MongoDB connection string OpsManager/BackupDaemon
	// should use to reach the AppDB.
	BuildConnectionURL(ctx context.Context, opsManager *omv1.MongoDBOpsManager, log *zap.SugaredLogger) (string, error)
}
```

interface name already has AppDB so I think it's not really needed in method names.

```
func (r ApDBReconciler){
r.Reconcile()
}
```

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -1720,58 +1721,51 @@ func (r *ReconcileAppDbReplicaSet) generatePasswordAndCreateSecret(ctx context.C
...
 			}
-			return "", err
+			log.Debugf("Generated AppDB password and storing in secret/%s", opsManager.Spec.AppDB.GetOpsManagerUserPasswordSecretName())
+			return r.generatePasswordAndCreateSecret(ctx, opsManager, log)
 		}
+		return "", err
+	}
 
+	// User-provided password ref path: watch the secret and clean up the auto-generated one.
+	if passwordRef := opsManager.Spec.AppDB.PasswordSecretKeyRef; passwordRef != nil && passwordRef.Name != "" {
```
viveksinghggits: too many if else statements here I was wondering if this can be simplified something like this

```
passwordRef := opsManager.Spec.AppDB.PasswordSecretKeyRef
if passwordRef.name ==""{
return r.generatePasswordAndCreateSecert()
}

password, err := r.readAppDbPassword(ctx, opsManager)
if err != nil {
 if secret.SecretNotExist(err){
  return err("secert doesn't exist)
 }
  return "", err
}
// rest of the code
```

viveksinghggits (commented):
Looks good, left some nits.

controllers/operator/appdbreplicaset_reconciler.go
MaciejKaras: That is true, I also thought about it. I've dropped the idea, because I wanted the existing `ReconcileAppDB()` method to preserve it's name - too basically limit changes. As for the `BuildAppDBConnectionURL` it is replaced by `GetAppDBConfig` in the following https://github.com/mongodb/mongodb-kubernetes/pull/1468 PR.

I would leave it as is if you don't mind, to limit the amount of changes as explained earlier. wdyt?

controllers/operator/appdbreplicaset_controller.go
MaciejKaras: I've tried to flatten it, but the logic is quite complex. For example in case user didn't provide password, we need to generate it only if it does not exits. So first we always should call r.readAppDbPassword().

controllers/operator/mongodbopsmanager_controller.go
MaciejKaras: Good call, let me bring it back.

controllers/operator/appdbreplicaset_controller.go
MaciejKaras: That is true. BuildAppDBConnectionURL it is replaced by GetAppDBConfig in the following https://github.com/mongodb/mongodb-kubernetes/pull/1468 PR. I would like to not change it here if you don't mind.

controllers/operator/appdbreplicaset_controller.go
MaciejKaras: well readAppDbPassword() does not mutate any secrets right now. Only ensureAppDbPassword does. Could you explain differently what you meant?

viveksinghggits (approved):
approving but this comment https://github.com/mongodb/mongodb-kubernetes/pull/1459/changes#r3822621697 is marked resolved but is not addressed.

changelog/20260804_fix_external_appdb_forward_migration_password.md
MaciejKaras: I think the `gh stack sync` command had to remove this change I have applied :/ sorry about it

controllers/operator/appdbreplicaset_controller.go
filipcirtog: I think Jose is referring to the fact that `readAppDbPassword` mutates the `opsManager` object that's passed in. 

**Current code (mutates):**

```go
func (r *ReconcileAppDbReplicaSet) readAppDbPassword(ctx context.Context, opsManager *omv1.MongoDBOpsManager, log *zap.SugaredLogger) (string, error) {
	passwordRef := opsManager.Spec.AppDB.PasswordSecretKeyRef
	if passwordRef != nil && passwordRef.Name != "" {
		if passwordRef.Key == "" {
			passwordRef.Key = "password"  // ← This modifies opsManager.Spec.AppDB.PasswordSecretKeyRef.Key
		}
		
		secretObjectKey := kube.ObjectKey(opsManager.Namespace, passwordRef.Name)
		log.Debugf("Reading AppDB password from secret %s", secretObjectKey)
		return secret.ReadKey(ctx, r.SecretClient, passwordRef.Key, secretObjectKey)
	}
	// ...
}
```

Since `passwordRef` is a pointer (`*userv1.SecretKeyRef`), the assignment on line `passwordRef.Key = "password"` changes the actual `opsManager` struct that was passed in. After calling this method, the caller's `opsManager.Spec.AppDB.PasswordSecretKeyRef.Key` will be set to `"password"` even though it was empty before.

**Non-mutating alternative**

```
func (r *ReconcileAppDbReplicaSet) readAppDbPassword(ctx context.Context, opsManager *omv1.MongoDBOpsManager, log *zap.SugaredLogger) (string, error) {
	passwordRef := opsManager.Spec.AppDB.PasswordSecretKeyRef
	if passwordRef != nil && passwordRef.Name != "" {
		key := passwordRef.Key
		if key == "" {
			key = "password"  // ← Uses a local variable, doesn't modify opsManager
		}
		
		secretObjectKey := kube.ObjectKey(opsManager.Namespace, passwordRef.Name)
		log.Debugf("Reading AppDB password from secret %s", secretObjectKey)
		return secret.ReadKey(ctx, r.SecretClient, key, secretObjectKey)
	}
	// ...
}
```

  
Here, the we use a local `key` variable instead of directly modifying `passwordRef.Key`, so the input `opsManager` remains unchanged.

controllers/operator/appdbreplicaset_controller.go
```diff
@@ -1721,57 +1722,53 @@ func (r *ReconcileAppDbReplicaSet) generatePasswordAndCreateSecret(ctx context.C
...
-		if err != nil {
-			if secret.SecretNotExist(err) {
-				log.Debugf("Generated AppDB password and storing in secret/%s", opsManager.Spec.AppDB.GetOpsManagerUserPasswordSecretName())
-				return r.generatePasswordAndCreateSecret(ctx, opsManager, log)
+
+	password, err := r.readAppDbPassword(ctx, opsManager, log)
+	if err != nil {
+		if secret.SecretNotExist(err) {
+			if passwordRef != nil && passwordRef.Name != "" {
+				return "", xerrors.Errorf("password secret %s referenced by spec.applicationDatabase.passwordSecretKeyRef does not exist", passwordRef.Name)
```
filipcirtog: q: should we also add the `opsManager.Namespace` in the error?

filipcirtog (approved):

controllers/operator/appdbreplicaset_controller.go
MaciejKaras: ah, now I see it. I'll fix it ofc

MaciejKaras:
evergreen refresh

PR #1460, opened 2026-08-04
CLOUDP-429011 - External AppDB: API and admission validation (MongoDB only)

MaciejKaras:
# Summary

Adds the API surface for running the Ops Manager Application Database as an external, user-managed MongoDB resource instead of an MCK-deployed in-cluster one (CLOUDP-429011):

- New `spec.externalApplicationDatabaseRef` on the MongoDBOpsManager CR referencing a MongoDB CR (Kind restricted to `MongoDB` in this PR).
- New `spec.role: AppDB` on the MongoDB CR marking it as the external AppDB, with admission validation (single-cluster topology required; role immutability).
- MongoDBMultiCluster is explicitly rejected for `spec.role` here (dedicated admission validator); multi support is planned for a later iteration.
- CRDs regenerated via `make manifests`; mechanical pointer/signature adaptations in master-owned call sites; changelog entry included.

This PR is the middle of a 3-PR stack; interface extraction in #1459, reconciliation logic in #1440.

## Proof of Work

- `go build ./...` and `go vet ./...` green
- `go test ./api/... ./controllers/... ./pkg/...` green, incl. new table-driven tests exercising the multi rejection through the `ValidateCreate` admission path
- `make manifests` idempotency gate: zero diff; generated ref enum shows only `MongoDB`
- changelog validator: "Changelog entries found and validated"

## Checklist


api/mongodb/v1/mdb/mongodb_validation.go
```diff
@@ -379,6 +383,18 @@ func resourceTypeImmutable(newObj, oldObj MongoDbSpec) v1.ValidationResult {
 	return v1.ValidationSuccess()
 }
 
+// roleImmutable blocks any spec.role transition after creation. Removing role: AppDB would
+// orphan the appdb-detach finalizer and re-enable the full Ops Manager state teardown on
+// deletion, shutting down a database an Ops Manager may still depend on; adding it to an
+// existing resource is an undesigned conversion flow. The only supported way to stop using a
+// resource as AppDB is reverse migration (deleting the resource).
+func roleImmutable(newObj, oldObj MongoDbSpec) v1.ValidationResult {
+	if newObj.Role != oldObj.Role {
```
m1kola: Could you please look into replacing runtime validations with CEL validations?

It will be a much better experience for the users if API server rejects invalid CRs on admission instead of us doing it during reconciliation.

I think these docs are relevant:
* https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules
* https://book.kubebuilder.io/reference/markers/crd-validation.html

MaciejKaras: The problem with moving them to CEL is that I would need to test them with real API server, so running the UnitTests with EnvTest or as part of e2e (which is overkill). Am I right, or there is easier way to test them?

I'm asking, because I don't want to rewrite the whole validation and validation testing engine in this PR.

m1kola: > The problem with moving them to CEL is that I would need to test them with real API server, so running the UnitTests with EnvTest or as part of e2e (which is overkill).

Do you not have the same concern with `kubebuilder:validation`?

But you are right about envtest: we will need to introduce envtest, if we want to test validations. But it is a much better solution than expanding our runtime validation which is poor UX for end users and headache for us.

For us a big issue with runtime validation is the fact that it is really tricky to switch from runtime validation to CEL validation without pain for users (it is arguably a breaking change). Because invalid CRs admitted by an API server are stored in etcd already and they might start failing validation on the different level after introduction of CEL. This could cause issues for user workflows.

So I t think we should not add new runtime validations unless absolutely necessary and use CEL instead.

> I'm asking, because I don't want to rewrite the whole validation and validation testing engine in this PR.

Agreed, it is better to introduce envtest in `master` and then rebase feature branch and make use of CEL validations here.

envtest is super lightweight and setup is very easy. We can add it to our unit test harness and start writing integration tests with it: https://book.kubebuilder.io/reference/envtest.html

It is actually a [recommended](https://github.com/kubernetes-sigs/controller-runtime/blob/6ec95076a056c7a36c0d8acc8316da796426df7e/FAQ.md#q-wheres-the-fake-client--how-do-i-use-it) way for testing:

> Q: Where's the fake client? How do I use it?
>
> A: The fake client [exists](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake), but we generally recommend using [envtest.Environment](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest#Environment) to test against a real API server. In our experience, tests using fake clients gradually re-implement poorly-written impressions of a real API server, which leads to hard-to-maintain, complex test code.

MaciejKaras: Yes, I've used envtest before and I know AKO uses it extensively. Still like I've mentioned it earlier, I don't want to focus on introducing envtest right now, even though I see it as a better approach as you do. 

This is not a priority of my current project and unfortunately I cannot allow myself to spend time on it. If the envtest CEL validation testing framework is introduced on master I can update my PR to use it as well.

m1kola: > This is not a priority of my current project and unfortunately I cannot allow myself to spend time on it.

Lack of envtest forces you to implement, as you agree, considerably worse solution in the project. I understand your are worried about the timeline. Since it is a limiting factor for the user-facing experience and introduces difficult-to-get-rid-of tech debt (further spreads existing bad practice), I think it is reasonable to have a conversation about allowing extra time for introducing envtest to MCK as part of your project.

m1kola: Here is vibe coded PoC for envtest: https://github.com/mongodb/mongodb-kubernetes/pull/1463

Seems like a decent start.

MaciejKaras: >I think it is reasonable to have a conversation about allowing extra time for introducing envtest to MCK as part of your project.

I will disagree here. This requires a separate project/engineering time to rewrite the validation logic. I'm not going to do this as part of OM Backup project - changing the status quo is not part of project goals.

What I agree with is that CLE is better solution and we could schedule a work item for the migration.

m1kola: > I will disagree here. This requires a separate project/engineering time to rewrite the validation logic.

No, it does not. I'm not talking about rewriting existing validation logic. I agree that it is a larger effort and you should not be doing it as part of OM Backup project. Perhaps it will not even be possible to migrate everything to CEL.

I"m saying that we should not be introducing **new runtime validation** logic for OM Backup project and use CEL validation instead. Scope is limited to the project. And if you need envtest to test - this bit should be part of the project/prerequisite.

MaciejKaras: What you're asking me is to pull in to the project scope bootstraping the envtest suite required for CEL validation testing. CEL validation was never a goal of Backup OM project. I don't see the reason why I should break the codebase current status quo (using runtime validation) and introduce CEL validation **in this particular project**. I agree that CEL validation is better, but I'm not degrading the existing validation functionality when using the runtime validations.

It seems that we differ not on the technical level, rather project scope alignment. I'm in the position to keep the current project scope, but we can ask @mircea-cosbuc and @ashortt-mdb for guidance.

m1kola: > What you're asking me is to pull in to the project scope bootstraping the envtest suite required for CEL validation testing.

I did a PoC with an LLM in #1463 in 20 or so minutes yesterday (including some iterations to polish a bit and an example test). I think that PoC is actually good to go as is. I'm happy to mark it as ready for review and address review feedback so you can start using it.

I did that because earlier you said: "If the envtest CEL validation testing framework is introduced on master I can update my PR to use it as well".

Would you like to review this PR?

> I don't see the reason why I should break the codebase current status quo (using runtime validation) and introduce CEL validation **in this particular project**.

It is not a status quo: we do use CEL validation already (grep `kubebuilder:validation:XValidation`). They are just not tested 👎. So adding `envtest` will benefit your project and others.

We are not introducing something new **in this particular project**. We are using the tools already available to us (CEL validation) which gives **better end user experience** and save us from technical debt which is difficult to get rid later. What is missing is testing. I see you are stressed about timelines and I can relate. But I can help you with envtest.

> It seems that we differ not on the technical level, rather project scope alignment.

No, it is technical. My code review feedback simply is: use CEL instead of runtime validation.

Would it take more time because you want to introduce envtest in for testing? Yes, a bit more time will be required. But it is normal - any code review feedback requires additional time to address.

MaciejKaras: In that case my opinion on technical level: Introducing CEL validation without testing is explicit downgrade to what we have right now. I would rather have runtime validations that are testable then have CEL, but not tested. 

If there is will be CEL testing framework available and merged to master, I will happily use it. I can help with reviewing your PR when it is ready for review, no problem.

m1kola: > In that case my opinion on technical level: Introducing CEL validation without testing is explicit downgrade to what we have right now. I would rather have runtime validations that are testable then have CEL, but not tested.

I think you are twisting the argument here. I don't think I ever said that you should add CEL validation without testing. You are right - to do it properly we need tests.

I just don't understand the pushback on the solution you said yourself is a better one.

> If there is will be CEL testing framework available and merged to master, I will happily use it. I can help with reviewing your PR when it is ready for review, no problem.

I just marked #1463 as ready for review. We don't need a special testing framework. Go's testing works really well with envtest.

controllers/operator/workflow/pending.go
MaciejKaras, line 1: @copilot revert changes in this file in this PR

api/mongodb/v1/mdb/mongodb_validation.go
MaciejKaras: I've adopted to CEL in this PR. Please take a look now

api/mongodb/v1/mdb/mongodb_types.go
```diff
@@ -821,6 +836,10 @@ func (d *DbCommonSpec) GetAdditionalMongodConfig() *AdditionalMongodConfig {
 	return d.AdditionalMongodConfig
 }
 
+func (d *DbCommonSpec) GetRole() string {
+	return d.Role
+}
```
viveksinghggits, line 840: nit: are these methods really useful? Can we not just use `d.Role` wherever we need to call `GetRole()`.

api/mongodb/v1/mdb/mongodb_validation.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/stringutil"
 )
 
+// RoleAppDB is the only supported value for spec.role, marking a MongoDB/MongoDBMultiCluster
```
viveksinghggits: Because we don't support `MongoDBMultiCluster` right now should we remove MDBMC from here?

api/mongodb/v1/om/opsmanager_types.go
```diff
@@ -117,8 +118,19 @@ type MongoDBOpsManagerSpec struct {
...
+	// ExternalApplicationDatabaseRef is set, in which case it may be omitted entirely.
+	// Ignored when ExternalApplicationDatabaseRef is set.
+	// +optional
+	AppDB *AppDBSpec `json:"applicationDatabase,omitempty"`
+
+	// ExternalApplicationDatabaseRef references a MongoDB resource
+	// to use as this Ops Manager's AppDB, instead of the internally-managed one.
+	// Takes precedence over applicationDatabase when both are set.
+	// +optional
+	ExternalApplicationDatabaseRef *ExternalApplicationDatabaseRef `json:"externalApplicationDatabaseRef,omitempty"`
```
viveksinghggits, line 133: very nit: Because the field above is called `AppDB` should w call this field `ExternalAppDBRef`, just for consistency.

api/mongodb/v1/om/opsmanager_types.go
```diff
@@ -163,6 +175,21 @@ type MongoDBOpsManagerSpec struct {
 	OpsManagerURL string `json:"opsManagerURL,omitempty"`
 }
 
+// ExternalApplicationDatabaseRef references the MongoDB resource
+// playing the AppDB role for this Ops Manager instance.
```
viveksinghggits:
```suggestion
// ExternalApplicationDatabaseRef references the MongoDB resource
// playing the AppDB role for an Ops Manager instance.
```

viveksinghggits: Because in this comment we don't really know which OM instance we are talking about.

api/mongodb/v1/om/opsmanager_types.go
```diff
@@ -163,6 +175,21 @@ type MongoDBOpsManagerSpec struct {
...
+type ExternalApplicationDatabaseRef struct {
+	// Name of the MongoDB resource to use as the external AppDB.
+	// Must be in the same namespace as the MongoDBOpsManager resource, and named
+	// <MongoDBOpsManager name>-db.
```
viveksinghggits, line 183: If the external AppDB (mongodb) MUST have name `<om-name>-db`, does it even make sense to accept the name of the using this field? Because we know for the fact that user is always going to provide the value `<om-name>-db` for name here right?

api/mongodb/v1/om/opsmanager_types.go
```diff
@@ -163,6 +175,21 @@ type MongoDBOpsManagerSpec struct {
...
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// Kind of the referenced resource.
+	// +kubebuilder:validation:Enum=MongoDB
+	// +kubebuilder:validation:Required
+	Kind string `json:"kind"`
```
viveksinghggits, line 190: I am assuming kind can either be `MongoDB` or `MongoDBMultiCluster` right?

changelog/20260729_feature_external_application_database_reference.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: feature
+date: 2026-07-29
+---
+
+* **MongoDBOpsManager**: You can now point an Ops Manager resource at an external MongoDB Application Database instead of having MCK deploy the AppDB in-cluster — once AppDB is a standard MongoDB resource it's visible to the same backup, monitoring & lifecycle tooling as the rest of the fleet. MongoDB resources needs to be marked with `spec.role: AppDB` when they serve that role.
```
viveksinghggits, line 6: I think instead of focusing on the fact that an OM's AppDB can be pointed to the an external MongoDB, we should focus on the fact that now OM's AppDB can be backed up and it's enabled by referring an external MongoDB as AppDB, or something in that line. wdyt?

viveksinghggits (commented):
Looks pretty good, left some nits.

api/mongodb/v1/mdb/mongodb_cel_envtest_test.go
```diff
@@ -0,0 +1,229 @@
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			rs := newMongoDB(t, tc.mutate)
+			err := k8sClient.Create(ctx, rs)
+
+			if tc.updateMutate != nil {
+				require.NoError(t, err)
+				tc.updateMutate(rs)
+				err = k8sClient.Update(ctx, rs)
+			}
```
filipcirtog: Is `updateMutate` dead code or will it be used in the following stacks? As I can see no table entry sets `updateMutate` in here.

api/mongodb/v1/om/opsmanager_types.go
```diff
@@ -98,6 +98,7 @@ type MongoDBOpsManagerList struct {
 	Items           []MongoDBOpsManager `json:"items"`
 }
 
+// +kubebuilder:validation:XValidation:rule="has(self.applicationDatabase) || has(self.externalApplicationDatabaseRef)",message="at least one of spec.applicationDatabase or spec.externalApplicationDatabaseRef must be set"
```
filipcirtog, line 101: should this be a XOR?

api/mongodb/v1/mdb/mongodb_types.go
```diff
@@ -441,8 +443,21 @@ type DbCommonSpec struct {
 	// +kubebuilder:validation:Enum=SingleCluster;MultiCluster
 	// +optional
 	Topology string `json:"topology,omitempty"`
+
+	// Role marks this resource as playing a special role for another MongoDB
+	// Kubernetes resource. Currently only AppDB is supported, marking this
+	// resource as the externally-managed Application Database for a
+	// MongoDBOpsManager resource.
+	// +optional
+	// +kubebuilder:default=""
```
filipcirtog: Why do we keep a default here? It looks redundant next to json:"role,omitempty" and it persists role: "" into every MongoDB resource. 

I assume the default is here so the field-level `self == oldSelf` has an oldSelf to compare on create. If so, the fix is to move immutability up to the spec level, where oldSelf is the whole old spec and absence is expressible:

`// +kubebuilder:validation:XValidation:rule="has(self.role) == has(oldSelf.role) && (!has(self.role) || self.role == oldSelf.role)",message="spec.role is immutable: ..."`

api/mongodb/v1/mdb/mongodb_types.go
MaciejKaras: It was unused, remove it

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: I wouldn't want to change it right now. Not because it's not valid, but due to the effort required to do so, mainly resolving git conflicts.

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: This is just so that it is explicitly passed by users and to verify they know what their are doing. If we don't ask them for the particular name, they could create MongoDB/AppDB CR with different name and wait for the switch that will not happen.

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: It will be in the near future, but not yet. We need to wait for the MongoDBMultiCluster integration so that's why it's only MongoDB for now.

changelog/20260729_feature_external_application_database_reference.md
MaciejKaras: I think this is more general featureset and backup is only one of the features available through this integration. I'll let @vinilage decide here how we should advertise it in the Release Notes.

api/mongodb/v1/mdb/mongodb_cel_envtest_test.go
MaciejKaras: Good catch, some leftover I believe. Removed

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: No, in the situation where users migrate from internal AppDB, they can still have existing `spec.applicationDatabase `struct present + `spec.externalApplicationDatabaseRef` set to trigger migration. `spec.externalApplicationDatabaseRef` just takes precedence

api/mongodb/v1/mdb/mongodb_types.go
MaciejKaras: Yeah, we can remove the default. That does not change much, the empty string still is going to be present in the code.

filipcirtog (approved):

api/mongodb/v1/mdb/mongodb_types.go
MaciejKaras: I had to bring it back, it's used in ExternalAppDB interface.

viveksinghggits: even if the method was being called, what I meant was it doesn't provide any value. dRole can directly be used instead of calling method.

MaciejKaras: No, but the `ExternalAppDB` is an interface that is implemented by both `MongoDB` and `MongoDBMultiCluster` structs. That's why this method is required and used.

config/crd/bases/mongodb.com_mongodb.yaml
```diff
@@ -1410,7 +1409,8 @@ spec:
                 - message: 'spec.role is immutable: it cannot be added, removed, or
                     changed after creation; to stop using a resource as AppDB, perform
                     a reverse migration (delete the resource)'
-                  rule: self == oldSelf
+                  rule: has(self.role) == has(oldSelf.role) && (!has(self.role) ||
+                    self.role == oldSelf.role)
```
filipcirtog: I think `spec.role immutability` is at the wrong CEL scope , it declares has(self.role) == has(oldSelf.role) && ... on the Role field, where self is a string, so it can't type-check. 

I think it should be moved onto DbCommonSpec.

MaciejKaras: Thanks! Fixed

api/mongodb/v1/mdb/mongodb_types.go
```diff
@@ -441,8 +444,19 @@ type DbCommonSpec struct {
...
+	// Role marks this resource as playing a special role for another MongoDB
+	// Kubernetes resource. Currently only AppDB is supported, marking this
+	// resource as the externally-managed Application Database for a
+	// MongoDBOpsManager resource.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == '' || self == 'AppDB'",message="spec.role must be 'AppDB' when set"
+	Role string `json:"role,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="!has(self.role) || self.role != 'AppDB' || self.type == 'ReplicaSet'",message="spec.resourceType must be ReplicaSet when spec.role is AppDB"
```
lucian-tosa, line 457: nit: this can be placed on dbcommonspec

api/mongodb/v1/mdb/mongodb_types.go
```diff
@@ -441,8 +444,19 @@ type DbCommonSpec struct {
...
+// +kubebuilder:validation:XValidation:rule="!has(self.role) || self.role != 'AppDB' || !has(self.topology) || self.topology != 'MultiCluster'",message="spec.topology MultiCluster is not supported when spec.role is AppDB"
```
lucian-tosa, line 458: same with this one

lucian-tosa (approved):

api/mongodb/v1/om/opsmanager_types.go
MaciejKaras: Done

PR #1463, opened 2026-08-04
Introduce envtest for Go tests

m1kola:
# Summary

Introduces [envtest](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest) (local etcd + kube-apiserver) for Go tests, so API-server-level behaviour can be tested without a cluster. Binaries are provisioned by `make envtest-assets` (wired into `scripts/evergreen/unit-tests-golang.sh`, so `make golang-tests` / CI get them automatically), with a shared helper package in `test/envtest/env` for use from any package. No existing tests were changed.

As a proof of concept, adds a test verifying the CEL validation rules of the MongoDBSearch CRD against a real API server, co-located with the types in `api/mongodb/v1/search`. See `docs/dev/envtest.md` for usage.

## Proof of Work

- `make golang-tests` green: 2978 tests, including the new `TestMongoDBSearchCELValidation` (5 CEL cases: valid spec accepted, `managed`/`unmanaged` XOR, cluster name required-and-unique, index uniqueness, name immutability via `oldSelf`)
- `make precommit` green
- e2e: N/A (no runtime change — tooling and test files only)

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -351,9 +351,10 @@ kustomize:
 	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.5.4)
 
 # envtest: local Kubernetes control plane (etcd + kube-apiserver) used by Go tests
-# (see test/envtest). ENVTEST_K8S_VERSION tracks the minimum supported Kubernetes
-# version from kubernetes-versions.json.
-ENVTEST_K8S_VERSION ?= 1.34.x
+# (see test/envtest). ENVTEST_K8S_VERSION defaults to the minimum supported Kubernetes
+# version from kubernetes-versions.json, reduced to a minor-version wildcard (e.g.
+# 1.34.3 -> 1.34.x) so that setup-envtest resolves the latest available patch release.
+ENVTEST_K8S_VERSION ?= $(shell jq -r '.kubernetes.min' kubernetes-versions.json | cut -d. -f1-2).x
 SETUP_ENVTEST = $(shell pwd)/bin/setup-envtest
 ENVTEST_ASSETS_DIR = $(shell pwd)/bin/envtest
 # setup-envtest (a downloader for the envtest binaries) is versioned independently of
--- a/Makefile
+++ b/Makefile
@@ -350,6 +350,28 @@ KUSTOMIZE ?= $(shell which kustomize 2>/dev/null || echo $(shell pwd)/bin/kustom
 kustomize:
 	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.5.4)
 
+# envtest: local Kubernetes control plane (etcd + kube-apiserver) used by Go tests
+# (see test/envtest). ENVTEST_K8S_VERSION tracks the minimum supported Kubernetes
+# version from kubernetes-versions.json.
+ENVTEST_K8S_VERSION ?= 1.34.x
+SETUP_ENVTEST = $(shell pwd)/bin/setup-envtest
+ENVTEST_ASSETS_DIR = $(shell pwd)/bin/envtest
+# setup-envtest (a downloader for the envtest binaries) is versioned independently of
+# controller-runtime; the module only publishes tags since v0.24.0.
+SETUP_ENVTEST_VERSION ?= v0.24.1
+
+# Download setup-envtest locally if necessary
+setup-envtest:
+	$(call go-install-tool,$(SETUP_ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest@$(SETUP_ENVTEST_VERSION))
+
+# Download the envtest binaries (etcd, kube-apiserver, kubectl); idempotent
+envtest-assets: setup-envtest
+	$(SETUP_ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(ENVTEST_ASSETS_DIR)
+
+# Print the envtest binaries path on stdout (used to set KUBEBUILDER_ASSETS)
+envtest-assets-path: setup-envtest
+	@$(SETUP_ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(ENVTEST_ASSETS_DIR) -p path
+
 # go-install-tool will 'go get' any package $2 and install it to $1.
 PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
 define go-install-tool
--- a/api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
+++ b/api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
@@ -3,6 +3,7 @@ package search_test
 import (
 	"context"
 	"fmt"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -16,14 +17,20 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/test/envtest/env"
 )
 
+// TestMain boots one envtest control plane shared by all tests in this package
+// (see test/envtest/env). Future envtest-based tests in this package should use
+// env.Shared(t) instead of starting their own environment.
+func TestMain(m *testing.M) {
+	os.Exit(env.RunShared(m, env.WithCRDs("mongodb.com_mongodbsearch.yaml")))
+}
+
 // TestMongoDBSearchCELValidation proves that the CEL validation rules defined on
 // the MongoDBSearch CRD are enforced by a real Kubernetes API server (booted
 // locally via envtest). It covers both create-time rules and the oldSelf-based
 // transition rule, neither of which can be exercised by plain Go unit tests.
 func TestMongoDBSearchCELValidation(t *testing.T) {
 	ctx := context.Background()
-	testEnv := env.Start(t, env.WithCRDs("mongodb.com_mongodbsearch.yaml"))
-	k8sClient := testEnv.Client
+	k8sClient := env.Shared(t).Client
 
 	newSearch := func(name string, clusters ...searchv1.ClusterSpec) *searchv1.MongoDBSearch {
 		return &searchv1.MongoDBSearch{
--- a/api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
+++ b/api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
@@ -2,6 +2,7 @@ package search_test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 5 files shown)


api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
```diff
@@ -0,0 +1,91 @@
...
+			ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: "default"},
+			Spec:       searchv1.MongoDBSearchSpec{Clusters: clusters},
+		}
+	}
+
+	t.Run("valid minimal spec is accepted", func(t *testing.T) {
+		search := newSearch("valid", searchv1.ClusterSpec{})
+		require.NoError(t, k8sClient.Create(ctx, search))
+		require.NoError(t, k8sClient.Delete(ctx, search))
+	})
```
MaciejKaras: can this be part of table test?

api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
```diff
@@ -0,0 +1,91 @@
...
+	t.Run("cluster name is immutable for an existing index", func(t *testing.T) {
+		search := newSearch("immutable-name", searchv1.ClusterSpec{Name: "cluster-a", Index: ptr.To(int32(0))})
+		require.NoError(t, k8sClient.Create(ctx, search))
+
+		search.Spec.Clusters[0].Name = "cluster-b"
+		err := k8sClient.Update(ctx, search)
+		require.Error(t, err)
+		assert.True(t, apierrors.IsInvalid(err), "expected an Invalid error, got: %v", err)
+		assert.Contains(t, err.Error(), "clusters[].name is immutable for an existing cluster index")
+	})
```
MaciejKaras: Same here, this should be integrated in the single table test

scripts/evergreen/unit-tests-golang.sh
```diff
@@ -6,6 +6,12 @@ set -x
 
 source scripts/dev/set_env_context.sh
 
+# Provision the envtest binaries (etcd + kube-apiserver) used by envtest-based Go tests
+# (see test/envtest) and expose them to `go test` via KUBEBUILDER_ASSETS.
+make envtest-assets
```
MaciejKaras, line 11: engineers are not calling this file, how can they have envtests ready for running unit tests locally?

docs/dev/envtest.md
```diff
@@ -0,0 +1,52 @@
...
+## Writing a new envtest test
+
+Use the shared helper in `test/envtest/env` — it starts the control plane, installs the
+CRDs from `config/crd/bases` and returns a client with the `mongodb.com/v1` scheme
+registered. It works from any package — co-locate CEL tests next to the API types that
+define the rules (see below), or use it from controller tests:
+
+```go
+func TestSomething(t *testing.T) {
+    testEnv := env.Start(t, env.WithCRDs("mongodb.com_mongodbsearch.yaml"))
```
MaciejKaras: so we are going to start separate envtest for each test? Can this introduce significant delay in our unit test? Maybe we should start envtest just once and then reuse it?

@josvazg what was your envtest setup in AKO?

Makefile
```diff
@@ -350,6 +350,28 @@ KUSTOMIZE ?= $(shell which kustomize 2>/dev/null || echo $(shell pwd)/bin/kustom
 kustomize:
 	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.5.4)
 
+# envtest: local Kubernetes control plane (etcd + kube-apiserver) used by Go tests
+# (see test/envtest). ENVTEST_K8S_VERSION tracks the minimum supported Kubernetes
+# version from kubernetes-versions.json.
+ENVTEST_K8S_VERSION ?= 1.34.x
```
MaciejKaras: can we infer version from kubernetes-versions.json instead of duplicating it?

scripts/evergreen/unit-tests-golang.sh
m1kola: They would need to run `make envtest-assets` once manually as per `docs/dev/envtest.md`.

`KUBEBUILDER_ASSETS` is helpful, but there is defaulting.

MaciejKaras: Ah yes, makes sense

docs/dev/envtest.md
m1kola: Ideally we should not have an env test per test, but while the number of tests is small we can use it as a start.

I believe AKO uses Ginkgo. with `BeforeSuite` which makes it possible to have it one per package. We can probably use Go's `init` for that purpose. I'll experiment with that.

josvazg: In AKO we use envtest only for int tests. We have 3 types of tests e2e, int and e2e2, all using Ginkgo BTW (which I dislike but it is what it is). The e2e2 are the latest version for new code, the idea is that e2e and int can be dropped with all their complex and buggy helpers once the legacy controllers are phased out.

The int tests in AKO are way too similar to the e2e, hence we plan not to use them for new controllers, at least for now.

My 2 cents here, I would not introduce envtest-ing unless we start wanting to test scenarios that cannot be properly covered with unit tests using the fake controller-runtime client or fake client-go and light mocking, but at the same time are way too expensive to run a full blown e2e test.

m1kola: > My 2 cents here, I would not introduce envtest-ing unless we start wanting to test scenarios that cannot be properly covered with unit tests using the fake controller-runtime client or fake client-go and light mocking, but at the same time are way too expensive to run a full blown e2e test.

@josvazg We want to test CEL validations we already have in the project (and adding more). It is not possible to test with fake clients as far as I know.

And envtest is a [recommended](https://github.com/kubernetes-sigs/controller-runtime/blob/6ec95076a056c7a36c0d8acc8316da796426df7e/FAQ.md#q-wheres-the-fake-client--how-do-i-use-it) way for testing:

> Q: Where's the fake client? How do I use it?
>
> A: The fake client [exists](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake), but we generally recommend using [envtest.Environment](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest#Environment) to test against a real API server. In our experience, tests using fake clients gradually re-implement poorly-written impressions of a real API server, which leads to hard-to-maintain, complex test code.

josvazg: It is possible, we do have CEL tests in AKO that work as unit tests.

Admittedly, this is something that SerguisZ did, we probably would have gone for envtest or something like that.

will link a sample

josvazg: sample test
https://github.com/mongodb/mongodb-atlas-kubernetes/blob/main/api/v1/project_reference_cel_test.go

helper package
https://github.com/mongodb/mongodb-atlas-kubernetes/tree/main/test/helper/cel
feel free to clone it

m1kola: @josvazg Ah, I see. It uses K8s go packages. I think we should go with envtest instead of going low level Kubernetes packages like in AKO. It will be tricky to maintain this without having an API server maintainer on the team :)

@MaciejKaras I updated the PR to have a shared setup per-package. Advantage of it is - well, shred setup which should be faster. Disadvantage is that each test must have an unique namespace.

api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
m1kola: @MaciejKaras done.

api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
m1kola: @MaciejKaras done.

Makefile
m1kola: @MaciejKaras that was the intention, but apparently I missed it. Done now.

test/envtest/env/env.go
```diff
@@ -0,0 +1,159 @@
...
+// binaryAssetsDir returns the first version directory under bin/envtest/k8s
+// provisioned by `make envtest-assets`, so that tests run directly (go test,
+// IDE) work without extra setup. KUBEBUILDER_ASSETS, when set (the unit test
+// entry points export it), takes precedence over this field inside envtest.
+// An empty result leaves envtest's own defaults untouched.
+func binaryAssetsDir() string {
+	k8sDir := filepath.Join(repoRoot(), "bin", "envtest", "k8s")
+	entries, err := os.ReadDir(k8sDir)
+	if err != nil {
+		return ""
```
MaciejKaras, line 144: missing error return

test/envtest/env/env.go
```diff
@@ -0,0 +1,159 @@
...
+		_, _ = fmt.Fprintf(os.Stderr, "failed to start the envtest control plane; run `make envtest-assets` to download the required binaries: %v\n", err)
+		return 1
+	}
+	shared = testEnv
+
+	code := m.Run()
+
+	if err := testEnv.Environment.Stop(); err != nil && code == 0 {
+		_, _ = fmt.Fprintf(os.Stderr, "failed to stop the envtest control plane: %v\n", err)
+		code = 1
```
MaciejKaras: nit: just `return 1` to not modify the variable

MaciejKaras (approved):
LGTM!

test/envtest/env/env.go
m1kola: That's intended: if we can't read the dear - we return an empty string which makes envtest use its own defaults.

test/envtest/env/env.go
m1kola: I'm looking for a second reviewer now. If there will be more feedback to address - I'll include this change. if not, will just merge as is.

filipcirtog (approved):
very nice!

api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+// TestMongoDBSearchCELValidation proves that the CEL validation rules defined on
+// the MongoDBSearch CRD are enforced by a real Kubernetes API server (booted
+// locally via envtest). It covers both create-time rules and the oldSelf-based
+// transition rule, neither of which can be exercised by plain Go unit tests.
+func TestMongoDBSearchCELValidation(t *testing.T) {
+	ctx := context.Background()
+	k8sClient := env.Shared(t).Client
+
+	newSearch := func(name string, clusters ...searchv1.ClusterSpec) *searchv1.MongoDBSearch {
```
filipcirtog: nit: WDYT about generating a unique namespace per subtest using GenerateName?


Something like:
```
ns := &corev1.Namespace{
    ObjectMeta: metav1.ObjectMeta{
        GenerateName: "test-cel-",
    },
}
require.NoError(t, k8sClient.Create(ctx, ns))
```

test/envtest/env/env.go
MaciejKaras: kk

api/mongodb/v1/search/mongodbsearch_cel_envtest_test.go
filipcirtog: this could also allow us to use `t.Parallel()`

m1kola: Oh, that is good solution! I totally forgot about it. I'll update the PR later today with this suggestion.

m1kola: Fixed in 73d1d3fe399e45f8da5d90fd7e5602d13716ef92

test/envtest/env/env.go
m1kola: Fixed in 73d1d3fe399e45f8da5d90fd7e5602d13716ef92

PR #1466, opened 2026-08-05
KUBE-323 - Fix GKE LoadBalancer resource cleanup during teardown

nammn:
# KUBE-323: Fix GKE LoadBalancer resource cleanup during teardown

## Problem

Multi-cluster snippet tests leak GCP resources (forwarding rules, target pools, health checks, addresses, NEGs, disks, DNS zones, service accounts) when runs fail or are interrupted before teardown. The accumulated leftovers were blocking new runs (e.g. forwarding-rule quota).

## Design: two plain rules

**Test teardown deletes what it created, best-effort.** Teardown scripts run with the run's own suffix-derived names and Kubernetes contexts; failures warn and continue (`set +e`, `|| true`).

**The periodic reaper deletes everything older than 24h, minus a tiny denylist.** `scripts/evergreen/periodic-cleanup-gcp.sh` now works on a single criterion — age — because Cloud Audit Logs (verified Aug 2026) show that *all* resource creation in `scratch-kubernetes-team` traces to the `k8s-operator-e2e-tests` service account or the GKE control planes of the clusters it creates. The only exclusions:

- `default-*` VPC firewall rules (shared networking)
- `k8s-operator-e2e-tests@` and `*@developer.gserviceaccount.com` service accounts (CI infrastructure)

Safety properties without any filtering machinery:

- Resources still in use cannot be deleted by gcloud; the delete fails, is tolerated, and is retried on the next periodic run.
- Clusters are deleted first (in parallel) so dependent resources are released before their own delete is attempted.
- Anything younger than 24h is untouched, so running tests are never affected.

## Prior cleanup impact and price summary

Earlier iterations of this cleanup were run against the live test project before this final age-only implementation. The [KUBE-323 Jira ticket](https://jira.mongodb.org/browse/KUBE-323) records a cleanup on 2026-08-05 that removed:

- 2,770 stale PVC disks, representing 39,420 GB of pd-balanced storage
- 371 regional forwarding rules
- 378 target pools

The estimated avoided list-price run rate was:

- pd-balanced storage: **$3,942/month**
- forwarding rules: **$2,763.05/month**
- total: **$6,705.05/month**, or approximately **$80,460.60/year** annualized

These are avoided list-price estimates, not realized invoice savings; billing-export data was unavailable. The 378 target pools are included in the deletion count, but no separate target-pool price estimate is included. This historical cleanup is also distinct from the later PR verification run, which removed 18 stale >24h candidates. The preceding leak and cross-run interference are documented in [KUBE-268](https://jira.mongodb.org/browse/KUBE-268).

## Why not filter by run suffix/tags?

An earlier version of this PR matched run suffixes in names/descriptions/tags. It was dropped because the filters were where the bugs lived: the health-check suffix filter could never match (GKE health-check names are hashes like `k8s-<hash>-node`), and the namespace-suffix regex matched nothing until this PR also changed the namespaces. A filter that silently matches nothing looks safe while doing nothing. The denylist inverts the failure mode: anything unexpected is deleted loudly after 24h, in a project that exists only for ephemeral tests.

As a side effect, the reaper now cleans up the legacy *unsuffixed* leftovers on its first run — no separate one-time cleanup needed.

## Verification

- Evergreen `periodic_gcp_cleanup`: green on 4 submitted patches — [run 1](https://spruce.mongodb.com/version/6a75f8ea53519f00079ecde3), [run 2](https://spruce.mongodb.com/version/6a79b1240afe1e000717ba50), [run 3](https://spruce.mongodb.com/version/6a79b5305f874d0007274072), [post-review final](https://spruce.mongodb.com/version/6a79c45cda28b80007533a9c): it removed 18 stale >24h candidates on the first run.
- Evergreen `public_gke_code_snippets` (no-mesh): [green end-to-end](https://spruce.mongodb.com/version/6a79914e8b4c800007f5ba59), and [green again](https://spruce.mongodb.com/version/6a79daf95d606c0007febde9) after the review-fix round.
- Evergreen `public_gke_code_snippets` (mesh, with the new shared LB-cleanup teardown): [green end-to-end](https://spruce.mongodb.com/version/6a79db0066c19800070b23e6). One earlier [patch](https://spruce.mongodb.com/version/6a79c4630afe1e00071806fd) running both snippet tests concurrently failed mid-test on MongoDB readiness timeouts (`mdb-sh`, `rs-user`) during a transient GKE window — teardown code was not yet reached at failure time, and both tests passed on retry.
- PR CI: [520/520 green on the final commit](https://spruce.mongodb.com/version/6a79c45b82858b000720b2e1). 

## Notes

- Positive deletion path is exercised by the periodic reaper itself; no bespoke test suite is added.
- Historical context: earlier iterations of this PR explored suffix-based ownership filtering; the final design replaces it with age-only cleanup for simplicity


nammn:
augment review

nammn:
Adjudicated the Copilot review comments (verified each against the code before acting):

1. **`periodic-cleanup-gcp.sh:35` (trailing newline / last row skipped) — declined, not a bug.** `rows=$(...)` strips the trailing newline, but the consumer loops over `<<<"${rows}"`, and the herestring re-adds exactly one record terminator. Verified with a constructed test (last unterminated row is read) and empirically: dry-run output of this script is byte-identical to the previous array-based implementation, last row included.
2. **`:96` (serialized cluster deletes) — accepted, fixed.** `reap_clusters` now deletes in parallel and `wait`-joins before the compute reapers run, matching the previous implementation's timeout rationale.
3. **`:155` (keyless service accounts skipped forever) — known gap, documented in the script, no fix.** Keyless accounts have no age signal, and deleting them blindly could hit a test mid-setup (ra-09 creates the SA and its key in separate steps), violating the "running tests are never affected" property. An audit-event age source is deliberately out of scope for this design.
4. **`test_gke_multi_cluster_no_mesh_snippets.sh:28/74` (early return aborts remaining clusters) — accepted, fixed.** The helper is now best-effort across all context/namespace pairs and returns the aggregate status.
5. **`scripts/dev/contexts/funcs/gke:30` (stale comment) — accepted, fixed.** The comment now describes the suffix as a naming/debugging choice; the reaper is age-only.
6. **Suppressed comment (mesh runner keeps the leak) — confirmed valid, fixed.** `test_gke_multi_cluster_snippets.sh` deploys Ops Manager with `type: LoadBalancer` (ra-06) and previously tore down clusters without deleting LB Services. The helper moved to `scripts/code_snippets/tests/cleanup_load_balancer_services.sh`, now sourced and invoked by both runners before cluster teardown (and in the reset path). Both runners' `cleanup` command paths also source the ra-02 env so `OM_NAMESPACE`/`MDB_NAMESPACE` are bound there.

While touching the reaper, three more guards were removed after an adversarial audit found them either unreachable (`filter_old` leading-tab guard, blank-line skip — both now fail closed via the numeric check instead of silently skipping) or redundant (`run_inventory`'s stderr tempfile dance; `pipefail` already propagates inventory failures). Dry-run output remains byte-identical at two age thresholds against the live project.

Verification: Evergreen `periodic_gcp_cleanup` green (multiple runs); `public_gke_code_snippets` green for the no-mesh runner; a patch covering both runners is running now.

PR #1467, opened 2026-08-05
CLOUDP-407997: Fix signature publishing

josvazg:
# Summary

Signatures are computer at build time and need to be copied over both at promotion and at final publish time. This PR ensures that behavior.

## Proof of Work

CI

## Checklist

### DIFF
--- a/ci/internal/release/registry.go
+++ b/ci/internal/release/registry.go
@@ -50,6 +50,16 @@ type cRegistry struct {
 	insecure bool
 }
 
+// signatureTagFor returns the cosign "simple signing" sibling tag name for an
+// image digest: the digest with ":" replaced by "-", plus a ".sig" suffix.
+// This is cosign's default tag-based signature storage scheme, used when
+// COSIGN_REPOSITORY points at the image's own repository (see
+// scripts/release/build/image_signing.py) rather than the OCI 1.1 referrers
+// API.
+func signatureTagFor(digest v1.Hash) string {
+	return strings.ReplaceAll(digest.String(), ":", "-") + ".sig"
+}
+
 func (t *cRegistry) CopyWithTags(srcRef string, dstRepo string, tags []string) error {
 	src, err := name.ParseReference(srcRef, t.nameOpts()...)
 	if err != nil {
@@ -88,6 +98,44 @@ func (t *cRegistry) CopyWithTags(srcRef string, dstRepo string, tags []string) e
 				return fmt.Errorf("write image %s: %w", tag, err)
 			}
 		}
+	}
+	if err := t.copySignature(src, desc.Digest, dstRepo); err != nil {
+		return fmt.Errorf("copy signature for %s: %w", srcRef, err)
+	}
+	return nil
+}
+
+// copySignature copies the cosign signature sibling tag (if the source image
+// was signed) from src's repository to dstRepo, under the same
+// digest-derived tag name (see signatureTagFor). The signature tag encodes
+// the image's digest, so one copy covers every tag CopyWithTags applied to
+// that same digest (e.g. both :{version} and :latest). If the source has no
+// signature tag (the image wasn't signed — signing is optional per image,
+// see build_info.json's per-image "sign" field), this is a silent no-op.
+func (t *cRegistry) copySignature(src name.Reference, digest v1.Hash, dstRepo string) error {
+	sigTag := signatureTagFor(digest)
+
+	srcSigRef := src.Context().Tag(sigTag)
+	sigDesc, err := remote.Get(srcSigRef, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	if err != nil {
+		var terr *transport.Error
+		if errors.As(err, &terr) && terr.StatusCode == http.StatusNotFound {
+			return nil
+		}
+		return fmt.Errorf("get signature %s: %w", srcSigRef, err)
+	}
+
+	sigImg, err := sigDesc.Image()
+	if err != nil {
+		return fmt.Errorf("read signature image %s: %w", srcSigRef, err)
+	}
+
+	dstSigRef, err := name.NewTag(fmt.Sprintf("%s/%s:%s", t.host, dstRepo, sigTag), t.nameOpts()...)
+	if err != nil {
+		return fmt.Errorf("parse signature target tag %s: %w", sigTag, err)
+	}
+	if err := remote.Write(dstSigRef, sigImg, remote.WithAuthFromKeychain(authn.DefaultKeychain)); err != nil {
+		return fmt.Errorf("write signature %s: %w", sigTag, err)
 	}
 	return nil
 }
--- a/ci/internal/release/registry.go
+++ b/ci/internal/release/registry.go
@@ -50,12 +50,6 @@ type cRegistry struct {
 	insecure bool
 }
 
-// signatureTagFor returns the cosign "simple signing" sibling tag name for an
-// image digest: the digest with ":" replaced by "-", plus a ".sig" suffix.
-// This is cosign's default tag-based signature storage scheme, used when
-// COSIGN_REPOSITORY points at the image's own repository (see
-// scripts/release/build/image_signing.py) rather than the OCI 1.1 referrers
-// API.
 func signatureTagFor(digest v1.Hash) string {
 	return strings.ReplaceAll(digest.String(), ":", "-") + ".sig"
 }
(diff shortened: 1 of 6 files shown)


ci/internal/release/registry.go
```diff
@@ -50,6 +50,16 @@ type cRegistry struct {
 	insecure bool
 }
 
+// signatureTagFor returns the cosign "simple signing" sibling tag name for an
```
MaciejKaras: nit: lets reduce the AI comment size or completely remove it. For this small method it's just overkill

ci/internal/release/registry.go
```diff
@@ -89,6 +99,44 @@ func (t *cRegistry) CopyWithTags(srcRef string, dstRepo string, tags []string) e
...
+}
+
+// copySignature copies the cosign signature sibling tag (if the source image
+// was signed) from src's repository to dstRepo, under the same
+// digest-derived tag name (see signatureTagFor). The signature tag encodes
+// the image's digest, so one copy covers every tag CopyWithTags applied to
+// that same digest (e.g. both :{version} and :latest). If the source has no
+// signature tag (the image wasn't signed — signing is optional per image,
+// see build_info.json's per-image "sign" field), this is a silent no-op.
+func (t *cRegistry) copySignature(src name.Reference, digest v1.Hash, dstRepo string) error {
```
MaciejKaras, line 141: nit: similar issue with comment lenght

ci/internal/release/registry.go
```diff
@@ -89,6 +99,44 @@ func (t *cRegistry) CopyWithTags(srcRef string, dstRepo string, tags []string) e
 			}
 		}
 	}
+	if err := t.copySignature(src, desc.Digest, dstRepo); err != nil {
+		return fmt.Errorf("copy signature for %s: %w", srcRef, err)
```
MaciejKaras: nit: probably it should be:
>copy signature **failed** for %s: %w

MaciejKaras (approved):
LGTM, added minor comments

m1kola:
@josvazg our images are multi-arch. Do we only care about signing the index (what we do now) or do we also need to recursively sign each image for each arch in the index?

josvazg:
> @josvazg our images are multi-arch. Do we only care about signing the index (what we do now) or do we also need to recursively sign each image for each arch in the index?

That is a good question. I would avoid having a user trying to check the index or an arch image signature and filing just because one or the other hash is not bound to a signature. So I would probably default too sign everything.

josvazg:
Still will double check if that is actually required

josvazg:
I made signing recursive to raw images as well, and the signature cloning is also recursive now.

scripts/release/build/image_signing.py
```diff
@@ -202,6 +202,7 @@ def sign_image(repository: str, tag: str) -> None:
         f"--use-signing-config=false",
         f"--new-bundle-format=false",
         f"--tlog-upload=false",
+        "--recursive",
```
m1kola, line 205: Do we need to also update `verify_signature` to be recursive?

https://github.com/mongodb/mongodb-kubernetes/blob/4d7e41d78b474bbcbe361a8e94a923eb811f3776/scripts/release/build/image_signing.py#L248

ci/internal/release/registry.go
```diff
@@ -85,10 +89,55 @@ func (t *cRegistry) CopyWithTags(srcRef string, dstRepo string, tags []string) e
...
+func (t *cRegistry) copySignature(src name.Reference, digest v1.Hash, dstRepo string) error {
+	sigTag := signatureTagFor(digest)
+
+	srcSigRef := src.Context().Tag(sigTag)
+	sigDesc, err := remote.Get(srcSigRef, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	if err != nil {
+		var terr *transport.Error
+		if errors.As(err, &terr) && terr.StatusCode == http.StatusNotFound {
+			return nil
+		}
```
m1kola, line 151: q: Could you explain this? Why we silently ignore the error?

ci/internal/release/registry.go
```diff
@@ -85,10 +89,55 @@ func (t *cRegistry) CopyWithTags(srcRef string, dstRepo string, tags []string) e
 			}
 		} else {
 			if err := remote.Write(dst, img, remote.WithAuthFromKeychain(authn.DefaultKeychain)); err != nil {
-				return fmt.Errorf("write image %s: %w", tag, err)
+				return fmt.Errorf("failed to write image %s: %w", tag, err)
```
m1kola, line 97: Nit: This reads better, but now inconsistent with other error messages.

ci/internal/release/registry.go
```diff
@@ -50,6 +50,10 @@ type cRegistry struct {
 	insecure bool
 }
 
+func signatureTagFor(digest v1.Hash) string {
+	return strings.ReplaceAll(digest.String(), ":", "-") + ".sig"
+}
+
 func (t *cRegistry) CopyWithTags(srcRef string, dstRepo string, tags []string) error {
```
m1kola, line 62: So we do `copySignature` for a digest. And promote and publish both run `CopyWithTags` twice (version tag and latest tag).

So if I understand correctly every image gets signature copied 4 times during the release process? Is it intended?

https://github.com/mongodb/mongodb-kubernetes/blob/4d7e41d78b474bbcbe361a8e94a923eb811f3776/ci/internal/release/promote.go#L106-L111

https://github.com/mongodb/mongodb-kubernetes/blob/4d7e41d78b474bbcbe361a8e94a923eb811f3776/ci/internal/release/publish.go#L98-L103

scripts/release/build/image_signing.py
josvazg: **cosign** verify doesn't have a `--recursive` flag (it's sign-only in cosign's CLI). Passing the index tag to verify only checks the signature on the top-level index digest — it won't validate the per-arch signatures we now produce with sign --recursive. To actually verify those too, we'd need to resolve the index's child manifest digests and call cosign verify against each repo@sha256:<child-digest> explicitly, similar to what we now do in mckci's copySignature loop.

I am inclined that is a overkill for this PR, specially as fixing the exiting signature infra before **mckci** was not initially in scope. I can accept it as a follow up.

m1kola: @josvazg I'm leaving this up to you because I'm not well versed in image signing. I raised this because my initial hunch was - verify should be symmetrical to signing.

ci/internal/release/registry.go
josvazg: There are 2 legit cases for a missing signature.

1. Unsigned image (top-level digest) — `build_info.json` has a per-image "sign" field; when false, atomic_pipeline.py:93 skips sign_image/verify_signature entirely at build time, so no .sig tag is ever created for that image's digest. Note: currently every entry in build_info.json is "sign": true, so in practice this case doesn't occur today — worth watching if it ever does start silently swallowing what should be an error.

2. Per-arch child manifest, pre-recursive-signing rollout — images signed before the --recursive fix to cosign sign (image_signing.py:198-206) only have a .sig tag on the index digest, none on the child digests. The new child-manifest loop in CopyWithTags will 404 on each until images are rebuilt/re-signed under the new pipeline.

Still it is worth at least logging a warning here, but not erroring out.

ci/internal/release/registry.go
josvazg: fixed error messages

ci/internal/release/registry.go
josvazg: You are right, on each copy pair the tag changes, but the signature is the same in both cases.
We do need to carry the signature to promotion then to release, but we only need to do it once per step.
Doing it several times is not really a problem, but it is wasteful.

Seems to me the best fix is to make the signature copy a explicit top level call and invoke it only once per move.
So we will have 2 copywithtags and 1 copy signatures in the end.

scripts/release/build/image_signing.py
josvazg: I will already prepare a PR for this, just for this recursive verification. It is a simple concept, but will probably be a bit more involved in code that what we would like.

josvazg:
@m1kola see #1518 for recursive signature verification

ci/internal/release/registry.go
m1kola: @josvazg I'm sorry, still don't understand.

Re case 1: so if every image in `build_info.json` has `"sign": true` - how could it happen that some images do not have signature?

Re case 2: but we haven't released anything before introducing recursive signing rollout. Is it relevant?

m1kola (approved):
LGTM. There is still one questions open: just for my understanding (not blocking).

ci/internal/release/registry.go
josvazg: Case 1: If someone changes anything to `"sign": true` to false, this code supports it without changes. The alternative is brittle, one changes a config, some seemly unrelated code breaks.

Case 2: If we ever point this code with an old release, we would hit this.

Do we want this code to be fastidious? I am not sure about that

PR #1468, opened 2026-08-05
CLOUDP-429011 - Add TLS/CA parity for external AppDB (resolve referenced MongoDB CR's TLS)

igor-karpukhin:
# Summary

When an OpsManager uses `spec.externalApplicationDatabaseRef`, the operator previously ignored the referenced CR's TLS settings — OM/BackupDaemon pods had no AppDB CA volume and no `ssl=true`/`mms.mongoCa` configuration. This PR resolves TLS/CA from the referenced MongoDB/MongoDBMultiCluster CR and threads it through the OM/BackupDaemon StatefulSet construction, configuration, and S3 backup pipeline.

Key changes:
- Replaced `BuildAppDBConnectionURL` with `GetAppDBConfig` on the `AppDBReconciler` interface, returning connection string + TLS enabled flag + CA ConfigMap name (works for both internal and external AppDB).
- Removed `GetAppDbCA()` and the `appDBTlsCAConfigMapName` helper that hardcoded `""` for external-AppDB mode; CA is now resolved from the referenced CR via `WithAppDBTLSCAConfigMapName`.
- E2E tests (`om_external_appdb_forward.py`, `om_external_appdb_fresh.py`) now exercise TLS-enabled external AppDB through forward and reverse migrations.

## Proof of Work

- `TestGetAppDBConfig_ExternalAppDB` — table-driven TLS on/off cases.
- `TestOpsManagerReconcile_ExternalAppDBRef_TLS_MountsAppDBCAVolume` — verifies CA volume on OM StatefulSet.
- E2E: sentinel-doc survives forward and reverse migration with TLS-enabled AppDB.

## Checklist

### DIFF
--- a/controllers/operator/appdbreplicaset_external_reconciler.go
+++ b/controllers/operator/appdbreplicaset_external_reconciler.go
@@ -41,14 +41,9 @@ func (r *OpsManagerReconciler) createNewExternalAppDBReconciler(log *zap.Sugared
 // detach-and-adopt migration of any pre-existing internal AppDB (idempotent, no-op once
 // complete), and establishes a watch on the referenced CR.
 func (e *ReconcileExternalAppDBReplicaSet) ReconcileAppDB(ctx context.Context, opsManager *omv1.MongoDBOpsManager) (reconcile.Result, error) {
-	refObject, err := e.validateExternalAppDBReference(ctx, opsManager)
-	if err != nil {
+	if err := e.validateExternalAppDBReference(ctx, opsManager); err != nil {
 		return e.updateStatus(ctx, opsManager, workflow.Failed(xerrors.Errorf("Error validating externalApplicationDatabaseRef: %w", err)), e.log, mdbstatus.NewOMPartOption(mdbstatus.OpsManager))
 	}
-
-	// Set TLS configuration so that it is easily retrievable later in Ops Manager reconciler
-	opsManager.Spec.ExternalApplicationDatabaseRef.IsTLSEnabled = refObject.IsTLSEnabled()
-	opsManager.Spec.ExternalApplicationDatabaseRef.CAConfigMapName = refObject.GetCAConfigMapName()
 
 	if err := e.ensureAppDBStatefulSetOwnership(ctx, opsManager); err != nil {
 		return e.updateStatus(ctx, opsManager, workflow.Failed(xerrors.Errorf("Error detaching internal AppDB StatefulSet: %w", err)), e.log, mdbstatus.NewOMPartOption(mdbstatus.OpsManager))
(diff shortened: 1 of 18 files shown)


controllers/operator/appdbreplicaset_external_reconciler.go
```diff
@@ -67,7 +67,7 @@ func (e *ReconcileExternalAppDBReplicaSet) validateExternalAppDBReference(ctx co
 
 	objectKey := kube.ObjectKey(opsManager.Namespace, ref.Name)
 
-	refObject, err := e.fetchExternalAppDBRefObject(ctx, ref, objectKey)
+	refObject, err := fetchExternalAppDBRefObject(ctx, e.client, ref, objectKey)
```
MaciejKaras: what is the reason to move those two receiver methods to plain functions?

controllers/operator/appdbreplicaset_external_reconciler.go
```diff
@@ -138,29 +138,67 @@ func (e *ReconcileExternalAppDBReplicaSet) computeExternalAppDBConnectionString(
 	password := string(passwordSecret.Data[util.OpsManagerPasswordKey])
 
 	objectKey := kube.ObjectKey(opsManager.Namespace, ref.Name)
-	refObject, err := e.fetchExternalAppDBRefObject(ctx, ref, objectKey)
+	refObject, err := fetchExternalAppDBRefObject(ctx, e.client, ref, objectKey)
```
MaciejKaras: this is the second one

controllers/operator/appdbreplicaset_external_reconciler.go
```diff
@@ -138,29 +138,67 @@ func (e *ReconcileExternalAppDBReplicaSet) computeExternalAppDBConnectionString(
...
 		return "", xerrors.Errorf("failed to fetch externalApplicationDatabaseRef %s: %w", objectKey, err)
 	}
 
 	return refObject.BuildConnectionString(util.OpsManagerMongoDBUserName, password, connectionstring.SchemeMongoDB, nil), nil
 }
 
+// resolveAppDBTLSConfig returns the CA ConfigMap name and TLS-enabled flag that OpsManager and the
+// Backup Daemon should use to trust the AppDB's TLS certificate. For the internally-managed AppDB it
+// reads opsManager.Spec.AppDB; for an external AppDB it reads the referenced MongoDB CR's security config.
+func resolveAppDBTLSConfig(ctx context.Context, c client.Client, opsManager *omv1.MongoDBOpsManager) (caConfigMapName string, tlsEnabled bool, err error) {
```
MaciejKaras: I would move this method call to the `NewOpsManagerReconcilerHelper()` and initialise helper fields there. This will move all of the initialisation of the helper to the single method.

Also I would split the `resolveAppDBTLSConfig` method into two separate methods: one for TLS enabled, one for TLS CA. This will reduce the parameter list reduced and both methods will be small enough.

controllers/operator/appdbreplicaset_external_reconciler.go
MaciejKaras: I've checkout the code and now I understand it ;)

controllers/operator/construct/opsmanager_construction.go
```diff
@@ -93,6 +93,15 @@ func WithConnectionStringHash(hash string) func(opts *OpsManagerStatefulSetOptio
 	}
 }
 
+// WithAppDBTLSCAConfigMapName injects the effective AppDB CA ConfigMap name (resolved by the
+// controller from either the internal AppDB spec or an external AppDB's referenced CR) so the
+// OpsManager and Backup Daemon pods mount the correct CA to trust the AppDB's TLS certificate.
+func WithAppDBTLSCAConfigMapName(name string) func(opts *OpsManagerStatefulSetOptions) {
```
MaciejKaras: This is not ideal, because we also set this field in `getSharedOpsManagerOptions` method and then overwrite it here. The problem is that `getSharedOpsManagerOptions(opsManager *omv1.MongoDBOpsManager)` only has the opsManager as the parameter which we cannot infer the TLS config for the external AppDB.

I have some idea to fix this, but I need to do some testing.

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_forward.py
```diff
@@ -52,17 +54,41 @@ def meta_om(namespace: str, custom_version: Optional[str], custom_appdb_version:
...
+    custom_appdb_version: str,
+    appdb_ca_configmap: str,
+    appdb_cert_prefix: str,
+) -> MongoDBOpsManager:
     resource = MongoDBOpsManager.from_yaml(yaml_fixture("om_external_appdb_primary_om.yaml"), namespace=namespace)
     resource.set_version(custom_version)
     resource.set_appdb_version(custom_appdb_version)
+    # start with a TLS-enabled internal AppDB so that after forward migration to the (also TLS)
+    # external CR the connection string (ssl=true, same hosts) is unchanged.
+    resource["spec"]["applicationDatabase"]["security"] = appdb_tls_security(appdb_ca_configmap, appdb_cert_prefix)
```
MaciejKaras, line 82: I think we can explicitly move TLS configuration to the .yaml file instead of calling specific method

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_fresh.py
```diff
@@ -164,10 +185,22 @@ def _assert_shared_secrets_claimed_and_unchanged(self, namespace: str):
...
+    ):
         primary_om.load()
         primary_om["spec"]["externalApplicationDatabaseRef"] = None
-        primary_om["spec"]["applicationDatabase"] = {"members": 3, "version": custom_appdb_version}
+        # the internal AppDB keeps TLS with the same CA/certs as the external CR, so the computed
+        # connection string (ssl=true + same hosts) is unchanged and the migration stays graceful.
+        primary_om["spec"]["applicationDatabase"] = {
+            "members": 3,
+            "version": custom_appdb_version,
+            "security": appdb_tls_security(appdb_ca_configmap, appdb_cert_prefix),
```
MaciejKaras: possibly inline

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_tls.py
MaciejKaras, line 1: What do we need this test file for? It seems that the TLS configuration is already present in fresh and forward tests

controllers/operator/appdbreplicaset_external_reconciler.go
igor-karpukhin: We can, but there is one thing here: the `NewOpsManagerReconcilerHelper` is called not only in the main `Reconcile` but also in `OnDelete` for cleanup when the OM is deleted. In case of deletion, the referenced MongoDB CR might not exist, so the `resolveAppDBTLSConfig` will return an error and abort the cleanup. So I vote for leaving it here

igor-karpukhin: If we split the method, both functions will have to `GET` the referenced CR, so this is the reason this is one method, instead of two. We can have one `GET` but we store this info in some helper and use it later in two places, if you want

controllers/operator/construct/opsmanager_construction.go
igor-karpukhin: Would you like to fix it? Because I think this fix also affects the comment above

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_forward.py
igor-karpukhin: This is possible, if you prefer the declarative way. But as you can see this is already shared with all `external-appdb` tests

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_tls.py
igor-karpukhin: This is base test, written before applying TLS changes to your tests. Let me remove it

docker/mongodb-kubernetes-tests/tests/opsmanager/om_external_appdb_fresh.py
igor-karpukhin: Good point, fixed

pkg/telemetry/collector.go
filipcirtog, line 327: What about the external domains for an External AppDB? 
Could something like this work?

```
if item.Spec.ExternalApplicationDatabaseRef == nil {
    appDBClusters = len(item.Spec.AppDB.ClusterSpecList)
    appDBExternalDomains = getExternalDomainPropertyForAppDB(item)
} else {
    appDBExternalDomains = getExternalDomainPropertyForExternalAppDB(ctx, operatorClusterClient, item)
}
```

```
func getExternalDomainPropertyForExternalAppDB(ctx context.Context, operatorClusterClient kubeclient.Client, om omv1.MongoDBOpsManager) string {
    ref := om.Spec.ExternalApplicationDatabaseRef
    key := kubeclient.ObjectKey{Namespace: ref.Namespace, Name: ref.Name}

    mdb := &mdbv1.MongoDB{}
    if err := operatorClusterClient.Get(ctx, key, mdb); err != nil {
        return ExternalDomainNone
    }
    return getExternalDomainProperty(*mdb)
}

```

filipcirtog: I have just seen the `resolveExternalAppDBProperties` method on the next stack. Please ignore the comment.

filipcirtog (approved):

MaciejKaras (approved):

PR #1469, opened 2026-08-05
CLOUDP-407997: Start Recoding Architecture Decisions

josvazg:
# Summary

This is both a proposal to start using ADRs Architecture Decision Records on key design decisions and a first use of that to record the CI release invariants we are adopting from now on.

## Proof of Work

N/A

## Checklist


MaciejKaras (approved):
While I personally agree on having ADRs in the repo itself, we are exposing them publicly when adding to GitHub. I would first ask for consensus decision from the team.

josvazg:
As per discussions, the team does not feel ADRs are helpful

PR #1470, opened 2026-08-05
CLOUDP 407997: Support test releases

josvazg:
# Summary

Add some tweaks in the new release process so we can test releases without publishing for real.

It is split in 3 commits that can be reviewed separately:
1) Add a flag to **mckci** to override the publish repo target.
2) Add a similar flag to OCI helm chart releases code.
3) Add the wiring in the GitHub workflow and evergreen release code to support dry-run releases, leveraging the override options just added in.

## Proof of Work

Needs to merge to test it.

## Checklist

### DIFF
--- a/.evergreen-functions.yml
+++ b/.evergreen-functions.yml
@@ -608,6 +608,8 @@ functions:
   publish_helm_chart:
     - command: subprocess.exec
       params:
+        include_expansions_in_env:
+          - dryrun_registry_override
         working_dir: src/github.com/mongodb/mongodb-kubernetes
         binary: scripts/release/publish_helm_chart.sh
 
--- a/.evergreen-release-publish.yml
+++ b/.evergreen-release-publish.yml
@@ -25,6 +25,10 @@ tasks:
             source .generated/context.export.env
             registry_override=""
             if [[ "${IS_DRYRUN}" == "true" ]]; then
+              if [[ -z "${dryrun_registry_override:-}" ]]; then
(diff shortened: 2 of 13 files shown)


filipcirtog (commented):
Should we error out if we're in dry-run mode but `dryrun_registry_override` is unset? Right now the expansion resolves to empty, we pass` --registry-override` "", and the dry run publishes straight to the production registry. 

The GitHub workflow input already documents the project var as a prerequisite but nothing enforces it. Same in scripts/release/publish_helm_chart.sh.

josvazg:
> Should we error out if we're in dry-run mode but `dryrun_registry_override` is unset? Right now the expansion resolves to empty, we pass` --registry-override` "", and the dry run publishes straight to the production registry.
> 
> The GitHub workflow input already documents the project var as a prerequisite but nothing enforces it. Same in scripts/release/publish_helm_chart.sh.

Yeah, that makes sense as an additional protection.

filipcirtog (approved):
LGTM!

m1kola (approved):
LGTM. But I think it would be great to clean up all these conditions once we switch over to the new workflow.

josvazg:
> LGTM. But I think it would be great to clean up all these conditions once we switch over to the new workflow.

There is planning to do quite a bit of cleanup once we move to the new release process for good. But that basically means to remove the old process and rename things so the result does not hint there were two release processes at some point. The cleanup does not include removing the ability to test releases, which I think it is a good thing to have. It could even be promoted to make internal releases in advanced or ad hoc.

PR #1477, opened 2026-08-10
KUBE-308: Add bounds to replica/shard counts

roothorp:
# Summary

There are multiple fields in our CRDs that do not have upper bounds. This means that overallocation (and the resulting gigantic memory consumption) is entirely possible, and awkward to recover from (as the crash will just result in the operator trying to create more).

Adding bounds via kubebuilder, and testing them, is a fairly straightforward fix for this.

We can additionally update the existing webhook validators to check these ranges (and also check that that the requested size is positive, to avoid trying to create negative size slices).

Some CRDs used the same structs for both Spec and Status, so this change requires them being explicitly separated (to avoid adding the limits/validation to the status fields).

## Proof of Work

New unit tests added.

## Checklist


api/mongodb/v1/mdb/mongodb_validation.go
```diff
@@ -348,6 +349,12 @@ func replicasetMemberIsSpecified(ms MongoDbSpec) v1.ValidationResult {
 	if ms.ResourceType == ReplicaSet && ms.Members == 0 {
 		return v1.ValidationError("'spec.members' must be specified if type of MongoDB is %s", ms.ResourceType)
 	}
+	// The member count is used as an allocation and loop bound (host seed slices, DNS name slices)
+	// before this point on clusters running an older CRD without the kubebuilder bounds, so reject
+	// out-of-range magnitudes here too rather than only ==0.
+	if ms.Members < 0 || ms.Members > constants.MaxReplicaSetMembers {
+		return v1.ValidationError("'spec.members' must be between 1 and %d, got %d", constants.MaxReplicaSetMembers, ms.Members)
```
Julien-Ben, line 356: Why don't we tolerate 0 in these validations (it was already the case before your PR I can see), but we do normally ?

api/mongodb/v1/mdb/mongodb_validation_test.go
```diff
@@ -728,3 +728,35 @@ func TestOIDCProviderConfigUniqueIssuerURIValidation(t *testing.T) {
 		})
 	}
 }
+
+func TestMongoDB_ProcessValidations_MembersMagnitudeIsBounded(t *testing.T) {
```
Julien-Ben, line 732: I suggest to add a test case for 0 too

Julien-Ben:
Generally, I wonder if we should bother duplicate our validations between kubebuilder and runtime checks.

I vaguely remember we discussed moving them gradually to kubebuilder only. So I would say it is enough to have them there, and customers will get the right validations as they upgrade the CRDs in the future.

They are supposed to do it anytime they upgrade the operator anyway, according to our upgrade procedure.

PR #1446, opened 2026-07-30
CLOUDP-400899: Remove the legacy multi-cluster installation path

m1kola:
# Summary

MCK 2.x configures multi-cluster at day 2 via `MemberCluster` CRs; the operator wiring and E2E migration landed earlier in this stack. This PR completes the clean break: it removes the legacy install-time path — the `multicluster setup`/`recover` subcommands, the kubeconfig-Secret + member-list ConfigMap discovery, and the `multiCluster.clusters`/`multiCluster.kubeConfigSecretName` Helm values — so we don't ship two mechanisms. The gitops sample's member-cluster RBAC YAML stays (now generated by the CLI itself), and the upgrade tests' released-1.x baseline is unaffected (it provisions the legacy flow via the released MCK 1.x plugin).

Stacked on #1442.

## Proof of Work

Unit tests and helm unittest green locally. The multi-cluster E2E suite was migrated to the new tooling earlier in this stack and is expected to keep passing unchanged.

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

PR #1435, opened 2026-07-28
CLOUDP-400899: Use a released kubectl plugin for the E2E baseline

m1kola:
# Summary

The multi-cluster upgrade tests start from a **released** operator, which discovers its member clusters from the kubeconfig Secret + member-list ConfigMap.

We will soon remove `multicluster setup` from the plugin, so the branch build CLI will no longer be able to provision that baseline. A released `kubectl-mongodb` plugin is now fetched into the test image for that purpose.

The upgrade tests also register their `MemberCluster` CRs *before* the operator is upgraded — an MCK operator that boots with none falls back to single-cluster and cannot reconcile the pre-existing multi-cluster resources. Resolves the `TODO(m1kola): slice-6` in `meko_mck_upgrade.py`.

## Proof of Work

Covered by the multi-cluster variants of `e2e_meko_mck_upgrade`, `e2e_appdb_tls_operator_upgrade_v1_32_to_mck` and `e2e_multi_cluster_appdb_upgrade_downgrade_v1_27_to_mck`, all of which run on every PR patch. No operator runtime change.

## Checklist

- [x] Have you linked a jira ticket and/or is the ticket in the title?
- [x] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you added changelog file?
    - use `skip-changelog` label if not needed
    - refer to [Changelog files and Release Notes](https://github.com/mongodb/mongodb-kubernetes/blob/master/CONTRIBUTING.md#changelog-files-and-release-notes) section in CONTRIBUTING.md for more details

---

PR #1479, opened 2026-08-10
CLOUDP-400899: Remove stale legacy multi-cluster filters and fix recover-test RBAC cleanup

m1kola:
Follow-ups to the legacy multi-cluster installation path removal (#1446).

# Summary

The legacy multi-cluster installation path was removed in #1446, which dropped the
`multiCluster.clusters` injection from `get_operator_helm_values`. This PR cleans up
the leftovers:

- **Search snippet env files** (`docs/search/12-search-rs-multi-cluster`,
  `13-search-sharded-multi-cluster`): the `grep -v multiCluster.clusters` filter in
  `OPERATOR_ADDITIONAL_HELM_VALUES` is now a no-op — replaced with the plain comma
  join the other snippet env files use.
- **`multi_cluster_recover_clusterwide.py`**: the pre-install cluster-scoped RBAC
  cleanup missed resources the install now renders — the telemetry
  ClusterRole/ClusterRoleBinding on central and members, and the `-pvc-resize`
  ClusterRole. Leftovers make a helm reinstall fail on reused clusters, the very
  failure this cleanup exists to prevent. Central names are now derived from
  `MULTI_CLUSTER_OPERATOR_NAME` instead of hardcoded, and the legacy MCK 1.x
  `setup`-tool names are annotated as such.
- **Epic doc** (`docs/dev/multi-cluster-config-tooling.md`): marks the clean-break
  slice done (all three PRs merged: #1435, #1442, #1446).

## Proof of Work

- `shellcheck` clean and `bash -n` passes on both env files.
- `python -m py_compile` passes on the test file; style within black's 120-col limit.
- e2e: N/A locally (no operator runtime change). The touched flows are exercised on
  Evergreen by `private_kind_multi_cluster_code_snippets` and
  `e2e_multi_cluster_recover_clusterwide`.

## Checklist

- [x] Have you added changelog file? (not needed — `skip-changelog` label applied)


mircea-cosbuc (approved):

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_recover_clusterwide.py
```diff
@@ -155,27 +155,42 @@ def test_create_service_entry(service_entries: List[CustomObject]):
...
+        f"{MULTI_CLUSTER_OPERATOR_NAME}-cluster-telemetry",
+        f"{MULTI_CLUSTER_OPERATOR_NAME}-{namespace}-cluster-telemetry-binding",
+        # Legacy MCK 1.x `multicluster setup` names — nothing in 2.x creates them, but they
+        # linger on long-lived dev clusters. Deletion tolerates 404.
+        f"{MULTI_CLUSTER_OPERATOR_NAME}-role",
+        f"{MULTI_CLUSTER_OPERATOR_NAME}-role-binding",
     ]
     for name in central_role_names:
         delete_cluster_role(name, central_cluster_client)
         delete_cluster_role_binding(name, central_cluster_client)
```
josvazg, line 183: > `docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_recover_clusterwide.py:173` — central_role_names only has f"{MULTI_CLUSTER_OPERATOR_NAME}-pvc-resize", but `helm_chart/templates/operator-roles-pvc-resize.yaml:47` names the binding {{ .Values.operator.name }}-pvc-resize-binding (suffix differs from the role).
> 
> The loop at line 181 calls `delete_cluster_role_binding` with the role's name, which 404s silently, so the real -pvc-resize-binding ClusterRoleBinding is never cleaned up and will leak across clusterwide test runs — the exact failure this test is meant to prevent. Add f"{MULTI_CLUSTER_OPERATOR_NAME}-pvc-resize-binding" to the list.

josvazg (commented):
Claude might have picked an issue here

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_recover_clusterwide.py
m1kola: @josvazg `-pvc-resize-binding` is only a `ClusterRoleBinding` when `operator.watchNamespace="*"` ([see `operator-roles-pvc-resize.yaml:40-44`](https://github.com/mongodb/mongodb-kubernetes/blob/de4eb0fc8acbc751af20c2ae07e83d2d40b45e5d/helm_chart/templates/operator-roles-pvc-resize.yaml#L40-L44)). This test watches two explicit namespaces, so the chart renders per-namespace `RoleBindings` with that name and namespaced resources that are removed with the per-run namespaces at teardown, so they cannot leak. The cluster-scoped piece (`ClusterRole` with suffix `-pvc-resize`) is covered.

josvazg (approved):

PR #1480, opened 2026-08-10
CLOUDP-400899: De-duplicate operator workload RBAC across base install and member render

m1kola:
# Summary

The operator's workload-management RBAC rules lived in two drifted copies: the base install templates (`operator-roles-base.yaml`, `operator-roles-pvc-resize.yaml`) and the member-cluster render (`member-cluster-rbac.yaml`). This PR single-sources them:

- `operator-roles-base.yaml` is now dual-mode: the shared rules are written once, unconditionally; central-only rules (CRDs, `operatorconfigs`/`memberclusters`) render only for the base install. The member role is renamed to `mck-member-<cluster>-role-base` (rules identical to the base install's).
- Multi-cluster-only rules (`deletecollection` cleanup — owner references don't work across clusters — and the `rbac-version` self-read) move to a new `mck-member-<cluster>-role-multicluster` role.
- `operator-roles-pvc-resize.yaml` is dual-mode too, pruned to least privilege (`list`, `watch`, `update`) in both modes. **Only base-install change**: the PVC-resize role drops unused `get`/`patch`/`delete` — see the breaking-change changelog entry.

## Proof of Work

- `helm template` base render diff = only the three removed PVC verbs (all scope modes); goldens regenerated (`config/rbac`, `public/mongodb-kubernetes*.yaml`, gitops samples).
- `go test ./pkg/kubectl-mongodb/...` — updated `TestRender` matrix, new `TestRender_OperatorRoleRules` (pins exact rule content), extended `TestHelmTemplateParity`.
- `make precommit` clean.
- e2e: pending on the CI patch — `e2e_multi_cluster_pvc_resize`, `e2e_multi_cluster_recover_clusterwide` (cleanup list updated here), `replica_set_pv_resize`.

## Checklist

---


helm_chart/templates/operator-roles-base.yaml
```diff
@@ -1,8 +1,58 @@
...
+{{- end -}}
+
+{{- $watchNamespace := list $operatorNamespace -}}
+{{- if $memberMode -}}
+{{- /* Member mode: the member SA is bound in memberCluster.workloadNamespaces plus the
+       operator namespace (which holds the SA and the token Secret). "*" is never a valid
+       entry — cluster-wide permission is expressed only via memberCluster.clusterScoped,
+       which renders a single ClusterRoleBinding. */ -}}
+{{- $workloadNamespaces := $memberCluster.workloadNamespaces | default (list $operatorNamespace) -}}
+{{- if has "*" $workloadNamespaces -}}
```
mircea-cosbuc, line 45: nit: indenting here would help see the nested logic.

helm_chart/templates/operator-roles-pvc-resize.yaml
```diff
@@ -1,8 +1,50 @@
...
+{{- $saName := .Values.operator.name -}}
+{{- if $memberMode -}}
+{{- $memberName := printf "mck-member-%s" (required "memberCluster.name is required when memberCluster.enabled=true" $memberCluster.name) -}}
+{{- $name = printf "%s-pvc-resize" $memberName -}}
+{{- $bindingName = printf "%s-pvc-resize-binding" $memberName -}}
+{{- $saName = printf "%s-sa" $memberName -}}
+{{- end -}}
+
+{{- $watchNamespace := list $operatorNamespace -}}
+{{- if $memberMode -}}
```
mircea-cosbuc, line 32: nit: same as before, I think we can indent here.

public/mongodb-kubernetes-openshift.yaml
```diff
@@ -152,11 +152,8 @@ rules:
     resources:
       - persistentvolumeclaims
     verbs:
-      - get
-      - delete
```
mircea-cosbuc, line 156: Is removing delete on PVC correct? Don't we need to still be able to delete PVCs for the destructive resize?

m1kola: @mircea-cosbuc I've specifically looked into `delete` and `patch` operations on PVC during implementation of this PR, but couldn't find any relevant code. But I might be missing something. In that case - please ping me in the right direction.

There was discrepancy between CLI and helm in master. [CLI provisions](https://github.com/mongodb/mongodb-kubernetes/blob/df5ff5bc7d52a601101e344d79f922d3fabd964b/pkg/kubectl-mongodb/common/common.go#L449-L451) `get`, `list`, `create`, `update`, `watch`, `patch` and [Helm renders](https://github.com/mongodb/mongodb-kubernetes/blob/master/helm_chart/templates/operator-roles-pvc-resize.yaml#L29-L34) `get`, `delete`,`list`, `watch`, `patch`, `update`.

I'm not sure if I checked if we need `create` so I ran an agent to verify what permissions are required and it came back confirming `list`, `watch` and `update`.

helm_chart/templates/operator-roles-base.yaml
m1kola: @mircea-cosbuc yes, helm templates are terrible to read. But managing indenting in a yaml (sensitive to indenting) is a nightmare. I think not having indenting is the lesser of two evils.

public/mongodb-kubernetes-openshift.yaml
mircea-cosbuc: @m1kola You're right, I confused it with the statefulset deletion here: https://github.com/mongodb/mongodb-kubernetes/blob/master/controllers/operator/create/create.go#L145-L147

mircea-cosbuc (approved):

josvazg (approved):

PR #1483, opened 2026-08-11
KUBE-328: Fix MongoDBUser connection string secrets to include tls=false for non-TLS deployments

filipcirtog:
# Summary

SRV connection strings assume TLS by default. For non-TLS clusters, MongoDBUser connection string secrets omitted `ssl=false` in `connectionString.standardSrv`, causing client connections to fail unless the secret was manually edited.

This PR updates the shared connection string builder to add `ssl=false` to URIs when TLS is not enabled on the referenced cluster. TLS-enabled clusters are unchanged (`ssl=true`). Standard (`mongodb://`) URIs are changed as well.

## Why we added mongosh to e2e tests

Previously, connection string coverage relied on pymongo with explicit TLS/auth options (`MongoTester`, `with_scram`, etc.). That did not match how (some) customers use the secret: copy-paste the URI and connect with no extra parameters.

mongosh behaves like a real user-facing client — it uses the connection string as-is, with no separate credential or TLS kwargs. That makes it the right tool to validate that generated secrets are self-contained.

mongosh also differs from drivers in practice: pymongo can still succeed via `client.admin.command("ping")` even when extra TLS/auth options are passed separately, so driver-only tests could pass without proving the secret works on its own.

New/updated e2e coverage:
- mongosh in the test image (mongosh 2.3.8)
- `assert_connection_string_with_mongosh` — runs mongosh with the secret URI only
- `connection_string_without_query_param` helper for SRV TLS regression testing
- SRV test proving `ssl=false` matters: mongosh succeeds with the generated secret, fails when `ssl` is stripped
- `ssl=false` / `authSource` content assertions across replica set, sharded, SHA-1, and multi-cluster SCRAM suites

Existing connectivity patterns (`MongoTester`, `with_scram` in multi-cluster) are kept where they already exercised deployment connectivity; only connection-string-secret validation was tightened.

# Proof of Work

Example output for a non-TLS cluster:

`mongodb+srv://<user>:<password>@<service>.<namespace>.svc.<cluster-domain>/?authSource=<db>&replicaSet=<name>&...&ssl=false`

# Checklist

    - use skip-changelog label if not needed
    - refer to Changelog files and Release Notes section in CONTRIBUTING.md for more details

---


controllers/operator/connectionstring/connectionstring.go
```diff
@@ -169,6 +169,9 @@ func (b *builder) Build() string {
 	}
 	if b.isTLSEnabled {
 		connectionParams["ssl"] = "true"
+	} else if b.scheme == SchemeMongoDBSRV {
+		// SRV connection strings assume TLS by default; explicitly disable when the cluster has no TLS.
+		connectionParams["ssl"] = "false"
 	}
```
lucian-tosa, line 174: Should we move to the equivalent "tls" parameter name. It's functionally the same, but is the "modern" one. "ssl" is legacy and used only for compatibility

filipcirtog: Yes indeed. However, I am adding `ssl` in here for consistency reasons. I am aiming in creating a PR for the MCK 2.0 where we will change all connection strings to use the equivalent `tls` name.

viveksinghggits (approved):

controllers/operator/connectionstring/connectionstring.go
```diff
@@ -169,6 +169,9 @@ func (b *builder) Build() string {
...
 	if b.isTLSEnabled {
 		connectionParams["ssl"] = "true"
+	} else if b.scheme == SchemeMongoDBSRV {
+		// SRV connection strings assume TLS by default; explicitly disable when the cluster has no TLS.
+		connectionParams["ssl"] = "false"
```
lucian-tosa: nit: any reason we shouldn't set this to false regardless of connection string scheme?

docker/mongodb-kubernetes-tests/kubetester/mongotester.py
```diff
@@ -136,6 +138,47 @@ def _wait_for_mongodbuser_reconciliation() -> None:
...
+
+
+def _run_mongosh(connection_string: str, eval_script: str) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(
+        ["mongosh", connection_string, "--quiet", "--norc", "--eval", eval_script],
+        capture_output=True,
+        text=True,
+        timeout=120,
+        check=False,
+    )
```
lucian-tosa, line 154: why do we need this? This is essentially testing the same thing as our existing methods that use the mongo driver

filipcirtog: The `MongoDB drivers` and `mongosh` behave differently, which is why we did not catch this problem earlier. For example, in the next PR (the next stack), if the URI does not specify a default database, mongosh defaults to the `test` database. The drivers, however, do not generally assume test as the default database. In the driver, for example, the application must explicitly select a database by calling `client.db(<database>)` before performing an operation.

controllers/operator/connectionstring/connectionstring.go
filipcirtog: fixed. than you.

lucian-tosa: looks good, but do you foresee any problems with backward compatibility issues for users that were previously using this connection string without the ssl flag?

filipcirtog: In theory, fixing this could break things for customers who built custom workflows to work around the original problem. However, as discussed in our sync, this is a bug fix and bugs like this should be reported and patched at the root, not worked around.

lucian-tosa: agreed for the srv connection string, but regarding my original question for the standard connection string, do you see issues there?

filipcirtog: It should not, it's the same but now explicitly called out. WDYT? link: https://www.mongodb.com/docs/manual/reference/connection-string-options/#mongodb-urioption-urioption.ssl

I could revert this and bring it back in 2.0, but having the default explicitly called out is the same as without.

tls: Enables or disables TLS/SSL for the connection.

true: Initiates the connection with TLS/SSL. Default for SRV Connection Format. | false: Initiates the connection without TLS/SSL. Default for Standard Connection String Format.
-- | --

lucian-tosa: makes sense

lucian-tosa (approved):

PR #1485, opened 2026-08-12
CLOUDP-400899: Read MemberCluster clients from a live provider instead of startup snapshots

m1kola:
# Summary

Stacked PR 1 of 2 for no-restart `MemberCluster` reactivity. Today every multi-cluster controller captures its per-cluster client map once at construction, which is why membership changes currently require an operator restart. Making membership reactive first requires all consumers to read the cluster set live.

This PR is a pure, zero-behaviour-change refactor:

- Adds `multicluster.Provider` — a thread-safe registry of per-member-cluster runtime objects (controller-runtime cluster, client, and the `MemberCluster` CR's metadata.name), populated at startup from the existing `MemberCluster` discovery.
- All multi-cluster reconcilers (MongoDBMultiCluster, OpsManager/AppDB, sharded, MongoDBUser, MongoDBSearch main/Envoy/metrics-forwarder) and telemetry now snapshot the provider once per reconcile and thread that snapshot through, so a single reconcile always works on one consistent membership view.
- The clusterName → CR-metadata.name mapping used for member-scoped ServiceAccount derivation now flows on the provider entry and `multicluster.MemberCluster.ResourceName`, replacing the write-once global registry in `pkg/resourcenames` (removed).

The operator still restarts on `MemberCluster` changes exactly as before: the restart-based watch continues to cancel the manager context, and per-cluster watches are still registered once at setup over the startup-populated entries. Replacing that watch with a level-based `MemberCluster` reconciler — which maintains provider entries at runtime and wires per-cluster watches dynamically — **lands in the follow-up PR**.

## Proof of Work

- Unit tests: `controllers/operator` (incl. mdbmulti/OM/AppDB/sharded/search suites), `controllers/searchcontroller`, `pkg/multicluster`, `pkg/resourcenames`, `pkg/telemetry` all green locally; `make precommit` clean.
- e2e: pending — no behaviour change by construction; the multi-cluster e2e suite will run on this PR's Evergreen patch.

## Checklist

---


controllers/operator/construct/backup_construction.go
```diff
@@ -83,7 +83,7 @@ func backupOptions(memberCluster multicluster.MemberCluster, additionalOpts ...f
 		}
 		opts.Replicas = opsManager.Spec.Backup.Members
```
m1kola, line 84: To help reviewers: the only new files are `pkg/multicluster/provider.go` (+ its test) which is a thread-safe registry of per-cluster runtime objects.

Everything else is plumbing:
* Reconcilers stop capturing client maps at construction and instead snapshot the provider once per reconcile, threading the snapshot down (mechanical signature changes from `map[string]client.Client` to `map[string]multicluster.Entry`
* `Add*` functions take the `*Provider` instead of `map[string]cluster.Cluster`)
* The `pkg/resourcenames` registry is deleted — its `clusterName` to `metadata.name` mapping now flows on `Entry.ResourceName` / `multicluster.MemberCluster.ResourceName`.

No behaviour change. The restart on `MemberCluster` change watch is untouched: will be in the next PR.

pkg/telemetry/collector.go
```diff
@@ -134,6 +137,7 @@ func RunTelemetry(leaderTrace context.Context, mongodbImage, databaseNonStaticIm
 		// we are calling this per "ticker" as customers might enable RBACs after the operator has been deployed
 		operatorUUID := getOrGenerateOperatorUUID(ctx, operatorClusterMgr.GetClient(), namespace)
 
+		cc := memberClusterConfigClients()
```
josvazg, line 140: **nit**: This violates the concept of "no behavioral change" because this remaps the member configs on each call and therefore changes the behaviour.

This is what we want, so I will not change the code, just change the claim. This actually does change behavior, slightly.

m1kola: @josvazg not sure that I follow. At this stage (in this PR) it is still the same set of clusters. We just access it differently.

josvazg: ah ok mutations between calls might only happen after incoming changes. ok!

josvazg (approved):
LGTM

roothorp (approved):

PR #1487, opened 2026-08-12
CLOUDP-400899: Make MemberCluster membership changes reactive without an operator restart

m1kola:
# Summary

Follow-up to #1485 (which introduced the `multicluster.Provider` registry). This flips `MemberCluster` add / spec-change / delete to hot reload: a level-based `membercluster.Reconciler` builds and starts each member cluster's runtime entry with a per-entry context and registers it via `provider.Set`; deletes remove the entry and cancel its context. The restart watcher, the startup `Discover`, and the initial-replay suppression are gone — the reconciler's initial informer replay populates the provider.

Each multi-cluster controller registers an engage hook that attaches its per-cluster watches to a newly added cluster (Watch-after-start) and enqueues all CRs of its type, so expansion to a cluster where a CR owns nothing yet still reconciles. The memberwatch health checker is hook-driven (`AddCluster`/`RemoveCluster`). Removal semantics are unchanged: resources on a removed cluster are abandoned.

E2E: restart-waits replaced with no-restart assertions (operator restartCount unchanged across membership changes); the DR suites' `test_operator_processes_member_removal` no longer skip local runs.

## Proof of Work

- e2e: pending — CI intentionally skipped while the stack is under review
- Unit: `pkg/membercluster` (reconciler + provider hooks), `pkg/multicluster/memberwatch`; envtest for Watch-after-start and teardown (`reconciler_envtest_test.go`)
- `make precommit` green

## Checklist


controllers/operator/mongodbuser_controller.go
```diff
@@ -331,6 +333,20 @@ func AddMongoDBUserController(ctx context.Context, mgr manager.Manager, memberCl
 		return err
 	}
 
+	// The user controller holds no per-cluster watches, but expanding to a new member cluster
+	// must reconcile existing users whose MongoDB resource spans that cluster.
```
m1kola, line 337: This is a bit counterintuitive when looking at the diff: why would we need this if we didn't have member-cluster dependent watches in this controller before?

In MCK 1.x/master we've always known all the member clusters at the operator startup time and there was no logic to dynamically add/remove clusters. 

This controller iterates over member clusters and updates some resources (like connection string secret). But it doesn't watch anything on the member clusters: only writes and that's it. So it didn't need any member-specific watches so far.

Now that we have dynamic cluster registration we need to enqueue all `MongoDBUser` objects on a cluster registration so the controller can write to a new cluster.

pkg/membercluster/reconciler_envtest_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	testEnv := &envtest.Environment{
+		CRDInstallOptions: envtest.CRDInstallOptions{
+			Paths: []string{"../../config/crd/bases/operator.mongodb.com_memberclusters.yaml"},
+		},
+	}
+	cfg, err := testEnv.Start()
+	require.NoError(t, err)
+	defer func() { assert.NoError(t, testEnv.Stop()) }()
```
m1kola: This needs to go be rewritten after a rebase on top of master which now includes https://github.com/mongodb/mongodb-kubernetes/pull/1463

pkg/membercluster/reconciler_envtest_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"github.com/mongodb/mongodb-kubernetes/pkg/multicluster"
+)
+
+// Verifies the hot-reload path end to end against a real API server: a MemberCluster CR
+// registers a live cluster entry whose cache serves watches added after the manager has
+// started (Watch-after-start), and deleting the CR removes the entry and stops the
+// cluster's informers quietly.
+//
+// TODO(m1kola): master recently gained envtest integration helpers/utilities that are not
```
m1kola: Feature branch is rebased on top of master. I can fix it now.

pkg/membercluster/providermanager.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return err
+	}
+
+	entryCtx, cancel := context.WithCancel(m.baseCtx)
+	go func() {
+		if err := memberCluster.Start(entryCtx); err != nil {
+			log.Errorf("Member cluster %q stopped with error: %s", mc.Spec.ClusterName, err)
+		}
+	}()
```
josvazg, line 118: The per-cluster `cluster.Cluster.Start()` runs in a bare goroutine, entirely outside the manager's own runnable bookkeeping. 

Two consequences:
- *No shutdown coordination*. The manager's StopAndWait only waits on runnables it knows about (mgr.Add()-registered ones). This goroutine isn't one of them, so the operator process can exit while a member cluster's cache/informers are mid-teardown — not correctness-fatal, but silent and untracked.

- *No error propagation*. If Start() fails (e.g., the member cluster becomes unreachable after being registered), the only outcome is a log line. Nothing feeds back into Reconcile, so there's no requeue/backoff and no operator-visible signal (status/condition) — the entry just sits dead in the provider until something else happens to bump that CR's generation.

Why not just wrap it in mgr.Add(): that's the obvious idiomatic fix, but it doesn't actually work here — controller-runtime has no mgr.Remove(). Anything added via mgr.Add() only stops when the whole manager stops. Since the entire point of this reactive design is per-cluster teardown (CR delete, spec-change rebuild) without restarting the operator, mgr.Add() is structurally the wrong tool.

*High-level alternative*: keep the per-entry goroutine/cancellable-context model (it's the right primitive for individually removable runnables), but make providerManager itself an mgr.Add()-registered runnable — a single, static aggregate that tracks all live per-cluster goroutines (e.g. via a WaitGroup) and blocks the manager's shutdown until they've all drained. Separately, thread failures from the per-cluster Start() back into the reconcile loop (a retry/requeue signal) instead of only logging, so a dead member cluster gets a real recovery path rather than silently rotting until the next spec change.

pkg/membercluster/providermanager.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	m.mu.Unlock()
+
+	if !exists {
+		return
+	}
+	m.provider.Delete(ctx, state.clusterName)
+	state.cancel()
+	log.Infof("Member cluster %q deregistered", state.clusterName)
+}
```
josvazg, line 152: `remove()` deletes from Provider by cluster name (state.clusterName), but providerManager.entries — the thing that decides what gets removed — is keyed by CR name (mc.Name). Those two identifiers are only guaranteed to line up 1:1 if spec.clusterName is unique across all MemberCluster CRs, and nothing enforces that (confirmed: no CEL/webhook validation, only MinLength=1 on the field).

High-level alternative: give Provider entries (or the internal bookkeeping) a notion of who currently owns this slot — e.g., stamp each Provider entry with the CR name that registered it, and make Delete/remove a conditional operation ("delete this cluster-name entry only if it's still owned by resourceName") rather than an unconditional overwrite-by-key. That turns a silent cross-CR clobber into either a no-op (safe) or a surfaced conflict (e.g., a status condition on the "loser" CR saying its clusterName is already claimed) — closing the gap without needing to add validation at the API/webhook layer.

pkg/membercluster/providermanager.go
```diff
@@ -0,0 +1,152 @@
...
+	restConfig.Timeout = m.clientTimeout
+
+	if exists {
+		m.remove(ctx, mc.Name, log)
+	}
+
+	memberCluster, err := m.newCluster(restConfig)
+	if err != nil {
+		return err
+	}
```
josvazg, line 105: Problem: sync() fetches the rest config before tearing down the old entry (good — protects against transient secret errors), but then calls m.remove() before m.newCluster() (line 85 before 89). If newCluster() fails, the old entry is already gone — no fallback, cluster stays unregistered until the next reconcile succeeds.

Alternative: build the full replacement (config + cluster.Cluster) first, only call remove() once it's ready — atomic swap instead of delete-then-hope-to-rebuild.

docker/mongodb-kubernetes-tests/tests/conftest.py
```diff
@@ -961,11 +934,6 @@ def _install_multi_cluster_operator(
             workload_namespaces=member_clusters_workload_namespaces,
             operator_cluster_scoped=member_clusters_operator_cluster_scoped,
         )
-
-        if wait_for_registration_restart:
-            wait_for_operator_pod_restart(
-                namespace, operator_name, previous_restart_count, api_client=central_cluster_client
-            )
         operator.wait_for_operator_ready()
```
josvazg, line 937: **follow up**, not a change request: 

Since membership is now reactive (no restart), this wait_for_operator_ready() only confirms the pod is up, not that the newly registered MemberCluster CR has actually been engaged by the provider. Downstream reconcilers currently self-heal this via panic+requeue if a workload CR races ahead of engagement, so it's likely harmless, but a real sync point (e.g. a status condition/generation on MemberCluster, written once providerManager.sync() succeeds) would close the gap properly.

Such a fix allows both our tests and end users to properly wait on a cluster being 'registered' (its provider entry created). We could even wait for 'fully online' by doing entry.Cluster.GetCache().WaitForCacheSync(ctx) before marking the cluster ready to go.

m1kola: @josvazg `wait_for_operator_ready` (or an equivalent) already exists on `master`/MCK 1.x where we do not have to worry about waiting for dynamic member cluster registration/sync. It is a wait for operator pod to start running after operator installation and is not relevant to member cluster registration.

Later in our project we will add `RBACValid` status condition to `MemeberCluster` which confirms that the cluster was configured with the latest RBAC version. This in theory can be used as a condition for waiting.

(I have a [draft PR here](https://github.com/mongodb/mongodb-kubernetes/pull/1507) if you are curious, but it is still work in progress and will change drastically so I would ignore it for now)

> Such a fix allows both our tests and end users to properly wait on a cluster being 'registered' (its provider entry created). We could even wait for 'fully online' by doing entry.Cluster.GetCache().WaitForCacheSync(ctx) before marking the cluster ready to go.

However I do not think that users (and we in our E2Es) should be waiting for some condition indicating that member clusters are ready when deploying multi-cluster resources. Users should just express the desired state of the world in a CR and operator should do the rest: wait for registration, reconcile as appropriate, etc. Or error if it's not possible to achieve.

Hence the removal of wait conditions in our E2E: to simulate user experience we delibereately do not wait for registration.

We had these waits on restarts intermediately in our feature branch to avoid flakes: some assertions were failing prematurely because of the operator restarts, if I remember correctly. Now it should not be an issue.

pkg/membercluster/providermanager.go
```diff
@@ -75,24 +78,40 @@ func (m *providerManager) sync(ctx context.Context, mc *operatorv1.MemberCluster
...
+	if exists {
+		m.remove(ctx, mc.Name, log)
+	}
+
 	entryCtx, cancel := context.WithCancel(m.baseCtx)
+	m.wg.Add(1)
 	go func() {
+		defer m.wg.Done()
 		if err := memberCluster.Start(entryCtx); err != nil {
 			log.Errorf("Member cluster %q stopped with error: %s", mc.Spec.ClusterName, err)
```
josvazg, line 116: **q**: I see now we are properly shutting down all members but when they exit on error we still just log here, don't propagate the error.

Is error propagation expected to come later? Or did you rule it out completely for some reason?

This is a pure question, I am not advocating to perform the propagation, want to check your view here.

josvazg (approved):
LGTM so far. I can review again once out of draft in case there were significant changes

pkg/membercluster/providermanager.go
m1kola: Fixed the shutdown part: `providerManager` is now a manager-registered runnable that cancels all entry contexts and drains the per-cluster goroutines via a `WaitGroup` on manager stop.

On error propagation: `cluster.Start` never returns an error for an unreachable cluster (informers retry internally and `Start` only returns on context cancel), so piping its error into a requeue would be dead weight likely. As far as I understand `Start` can only error when informers already started. As far as I see this is also [what multicluster-runtime also does](https://github.com/kubernetes-sigs/multicluster-runtime/blob/087d80cae466d81ea74669620412ab8373b2900d/pkg/clusters/clusters.go#L116-L120) with `Start` (`c.ErrorHandler` is just a error logging by default there).

Reachability of the cluster is currently (even in MCK 1.x) is responsibility of the member health checker and we are preserving this (needed for failover feature).

I currently do not plan a separate condition in `MemberCluster` for cluster reachability in the scope of this epic, but:
1.  We can add it in the future if we see the evidence that a standalone condition is useful/needed. This can be added after MCK 2.0.0.
2.  In the later PRs  we will have `RBACValid` condition which requires connectivity to the API server so it can serve as a proxy to some degree for connectivity.

pkg/membercluster/providermanager.go
m1kola: I think I answered in the original thread here: https://github.com/mongodb/mongodb-kubernetes/pull/1487#discussion_r3821347394

Please let me know if I didn't cover something there.

pkg/membercluster/providermanager.go
m1kola: `sync` now refuses a second registration for an already-claimed `spec.clusterName` (first writer wins; the conflicting CR's reconcile errors and requeues), and `remove` deletes the provider entry only when it's still owned by the deleting CR.

pkg/membercluster/providermanager.go
m1kola: Fixed: the replacement cluster is now built before the old entry is removed.

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -1191,33 +1190,40 @@ func AddMultiReplicaSetController(ctx context.Context, mgr manager.Manager, imag
...
+	// enqueue all MongoDBMultiCluster CRs — watch replay alone cannot reach CRs that own no
+	// resources on the new cluster yet. The health checker follows the same membership events.
+	memberClustersProvider.RegisterHooks(ctx, multicluster.Hooks{
+		OnAdd: func(ctx context.Context, clusterName string, entry multicluster.Entry) {
+			if err := c.Watch(source.Kind[client.Object](entry.Cluster.GetCache(), &appsv1.StatefulSet{}, &khandler.EnqueueRequestForOwnerMultiCluster{}, watch.PredicatesForMultiStatefulSet())); err != nil {
+				zap.S().Errorf("failed to set StatefulSet watch on member cluster %s: %s", clusterName, err)
+			}
+			memberClusterHealthChecker.AddCluster(clusterName, entry.Cluster.GetConfig(), zap.S())
+			if err := multicluster.EnqueueAll(ctx, reconciler.client, &mdbmultiv1.MongoDBMultiClusterList{}, eventChannel); err != nil {
+				zap.S().Errorf("failed to enqueue MongoDBMultiCluster resources on member cluster %s add: %s", clusterName, err)
```
roothorp, line 1219: I might be misunderstanding, but do we not want to 'properly' error here, instead of just logging at an error level and moving on? (same question for the other similar changes in other controllers).

pkg/membercluster/providermanager.go
```diff
@@ -0,0 +1,194 @@
...
+	m.wg.Add(1)
+	go func() {
+		defer m.wg.Done()
+		if err := memberCluster.Start(entryCtx); err != nil {
+			log.Errorf("Member cluster %q stopped with error: %s", mc.Spec.ClusterName, err)
+		}
+	}()
+
+	m.mu.Lock()
+	m.entries[mc.Name] = entryState{clusterName: mc.Spec.ClusterName, cancel: cancel, generation: mc.Generation}
```
roothorp, line 121: Is this not racy? Since we're dropping the lock between checking for something already with this name (ln84-91) to do some other stuff, then picking it up again to register the new cluster, is it not possible that another cluster passes the duplicate check in the meantime?

roothorp (approved):
Just a couple of small questions for my understanding, but otherwise LGTM!

pkg/membercluster/providermanager.go
m1kola: @roothorp yes, this is an excellent catch! It is not a big problem right now because we do not override `MaxConcurrentReconciles` for this controller and default is `1`. So there will be no concurrency. But I think this should be guarded in case this changes.

I also realised there is a gap while working on the next PR in the stack and I'm working on a fix for that right now. Debating weather to include it in this PR (which is already large) or in a follow up. I lean towards doing it as a follow up.

controllers/operator/mongodbmultireplicaset_controller.go
m1kola: @roothorp we can't do much in terms of error handling here apart from logging.

This hook will be called when a user adds a new `MemberCluster` CR to their central cluster and it can happen at any time with new dynamic registration. If error happens here - the reconciler needs to keep working for existing clusters and the rest of the operator should work still.

This call is not critical. If this errors, we simply not do force-trigger reconciliation of relevant CRs (`MongoDBMultiCluster` CRs in case of this controller) and they will be reconciled next time user updates it (e.g. adds a new cluster to `clusterSpecList`) or after periodic resync that controller-runtime does. This exists for an edge case when a user first updated `clusterSpecList` on a CR to say "deploy on clusters `foo`, `bar` and `to-be-deployed-baz`" and only after that decided to add `MemberCluster` CR. The other way around would work just fine even without this `EnqueueAll`.

Also the chance of it returning an error is minimal: most likely error is failing to list CRs form an client/informer and if this happens - you have bigger problem somewhere else.

pkg/membercluster/providermanager.go
m1kola: @roothorp I think at this stage it will be best have a smaller PR dedicated to improving concurrency here. I'm working on it in https://github.com/mongodb/mongodb-kubernetes/pull/1535.

PR #1490, opened 2026-08-13
CLOUDP-429011 - Add telemetry for ExternalAppDB

MaciejKaras:
# Summary

Adds telemetry support for the ExternalAppDB feature. When an OpsManager references an external MongoDB/MongoDBMultiCluster resource as its AppDB, the deployment usage snapshot now reports the referenced resource kind, backup mode, external domain configuration, and cluster count.

**New telemetry fields** (`pkg/telemetry/types.go`):
- `externalAppDB` — the kind of the referenced external AppDB resource (`MongoDB` or `MongoDBMultiCluster`); empty for internally-managed AppDB
- `appDBBackupMode` — the backup mode (`enabled`/`disabled`/`terminated`) read from the referenced CR's `spec.backup.mode`; empty when no backup spec is set

**Updated telemetry fields**:
- `externalDomains` — now `omitempty` (was always sent, even when empty); for external AppDB, resolved from the referenced CR instead of the OM's internal `AppDBSpec`
- `appDBClusters` — for external AppDB of kind `MongoDBMultiCluster`, now reports the referenced CR's `ClusterSpecList` length

**Refactoring** (`pkg/telemetry/collector.go`):
- Extracted `getAppDBProperties` / `resolveExternalAppDBProperties` to centralize AppDB property resolution for both internal and external cases
- Extracted `getExternalDomainGeneric` to deduplicate the uniform-vs-cluster-specific domain logic across MongoDB, MongoDBMulti, and AppDB variants
- Renamed `getExternalDomainProperty` → `getExternalDomainPropertyForMongoDB` for consistency with the other variants
- Replaced `ptr.To(x)` with `&x` (removes `k8s.io/utils/ptr` dependency; fixes staticcheck SA4006)

## Proof of Work

- `pkg/telemetry/collector_test.go` — new `"external appdb test"` table case in `TestCollectDeploymentsSnapshot` covering:
  - MongoDB ref with backup enabled, no external domain
  - MongoDB ref with backup disabled, uniform external domain
  - MongoDBMultiCluster ref with backup enabled, 3 clusters, no external domain
  - MongoDBMultiCluster ref with backup terminated, mixed external domains (uniform + cluster-specific), 2 clusters
  - MongoDB ref not found (error path — only `externalAppDB` set, remaining fields omitted via `omitempty`)
- `pkg/telemetry/types_test.go` — updated `TestSearchDeploymentUsageSnapshotProperties_ConvertToFlatMap` fixture to set `AppDBBackupMode` and `ExternalAppDB` so the reflection-based key check passes for the new `omitempty` fields.
- `go test ./pkg/telemetry/... -v -count=1` — all 9 test functions pass.
- `go test ./...` — full suite, zero failures.
- `make precommit` — clean (golangci-lint, ty, all hooks pass).

## Checklist

### DIFF
--- a/pkg/telemetry/collector.go
+++ b/pkg/telemetry/collector.go
@@ -252,7 +252,7 @@ func getMdbEvents(ctx context.Context, operatorClusterClient kubeclient.Client, 
 			}
 
 			if numberOfClustersUsed > 0 {
-				properties.DatabaseClusters = new(numberOfClustersUsed)
+				properties.DatabaseClusters = &numberOfClustersUsed
 			}
 
 			if event := createEvent(properties, now, Deployments); event != nil {
@@ -279,7 +279,7 @@ func addMultiEvents(ctx context.Context, operatorClusterClient kubeclient.Client
 		clusters := len(item.Spec.ClusterSpecList)
 
 		properties := DeploymentUsageSnapshotProperties{
-			DatabaseClusters:         new(clusters), // cannot be null in mdbmulti
+			DatabaseClusters:         &clusters, // cannot be null in mdbmulti
 			DeploymentUID:            string(item.UID),
 			OperatorID:               operatorUUID,
 			Architecture:             string(architectures.GetArchitecture(item.Annotations, defaultArchitecture)),
@@ -324,11 +324,11 @@ func addOmEvents(ctx context.Context, operatorClusterClient kubeclient.Client, o
 			}
 
 			if omClusters > 0 {
-				properties.OmClusters = new(omClusters)
+				properties.OmClusters = &omClusters
 			}
 
 			if appDBProperties.Clusters > 0 {
-				properties.AppDBClusters = new(appDBProperties.Clusters)
+				properties.AppDBClusters = &appDBProperties.Clusters
 			}
 
 			if event := createEvent(properties, now, Deployments); event != nil {
--- a/pkg/telemetry/collector.go
+++ b/pkg/telemetry/collector.go
@@ -239,6 +239,7 @@ func getMdbEvents(ctx context.Context, operatorClusterClient kubeclient.Client, 
 
 			numberOfClustersUsed := getMaxNumberOfClustersSCIsDeployedOn(item)
 			properties := DeploymentUsageSnapshotProperties{
+				Role:                     item.Spec.Role,
 				DeploymentUID:            string(item.UID),
 				OperatorID:               operatorUUID,
 				Architecture:             string(architectures.GetArchitecture(item.Annotations, defaultArchitecture)),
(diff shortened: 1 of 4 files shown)


pkg/telemetry/collector.go
```diff
@@ -643,3 +656,38 @@ func getAuthenticationAgentMode(security *mdbv1.Security) string {
...
+		if err := operatorClusterClient.Get(ctx, key, mdb); err != nil {
+			Logger.Warnf("Couldn't fetch MongoDB %s/%s for OpsManager external database reference: %v", ref.Namespace, ref.Name, err)
+			return appDBProperties
+		}
+
+		if mdb.Spec.Backup != nil {
+			appDBProperties.BackupMode = string(mdb.Spec.Backup.Mode)
+		}
+		appDBProperties.ExternalDomains = getExternalDomainGeneric(mdb.Spec.GetExternalDomain(), false)
+	} else if ref.Kind == "MongoDBMultiCluster" {
```
Julien-Ben, line 679: IIUC MDBMultiCluster will be part of a future phase 2 ? Let's add a comment it is not supported yet then, even if we keep the code

Julien-Ben:
Can we make it so we can distinguish if a mongodb resource is used as an AppDB or as a normal worload ?
So that in a chart like this we don't count both case as "replicaset" for example

image

Also I suggest to create a followup Jira ticket to update our telemetry pannel once this is released.

pkg/telemetry/collector.go
MaciejKaras: yes, it's phase that @igor-karpukhin is already working on and will be released soon after. I believe it does not make any harm to have telemetry support for it already, even though we don't allow setting MongoDBMultiCluster as a Kind.

Like I said, the support for MongoDBMultiCluster is going to follow shorty and I don't see that much value of the comment right now. We would need to remember to remove this comment, but on the other hand we are not going to touch telemetry anymore because it's finished.

MaciejKaras:
> Can we make it so we can distinguish if a mongodb resource is used as an AppDB or as a normal worload ? So that in a chart like this we don't count both case as "replicaset" for example
> 
> image
> Also I suggest to create a followup Jira ticket to update our telemetry pannel once this is released.

@Julien-Ben yes we can. We can update the diagrams. Can you send me the links so I can modify them?

filipcirtog (approved):

Julien-Ben:
> yes we can. We can update the diagrams. Can you send me the links so I can modify them?

@MaciejKaras I meant in the source data we send, with this PR. Do we provide information that enable us to do that later once we have the data ?

pkg/telemetry/collector.go
Julien-Ben: Okay, thanks !

MaciejKaras:
> > yes we can. We can update the diagrams. Can you send me the links so I can modify them?
> 
> @MaciejKaras I meant in the source data we send, with this PR. Do we provide information that enable us to do that later once we have the data ?

Ah, you mean to add this information in the MongoDB deployment itself? Good idea, let me add it

Julien-Ben (approved):
Thanks for adding the Role field !

lucian-tosa (approved):

PR #1491, opened 2026-08-13
Add ExternalMember API type and migration condition plumbing

lucian-tosa:
# Summary

Introduces the ExternalMember type on the MongoDB replica-set spec (with its
subfields) and threads it through the Ops Manager deployment and replica-set
builders, so members running on VMs can be represented in the automation
config. Also adds the migration condition/phase plumbing under
api/mongodb/v1/status.

Layer 1 in the VM-to-Kubernetes migration stack. Base: master.

**Stacked PR 1** (stack #1505), base `master`. This is one layer of the VM-to-Kubernetes

## Proof of Work

marked ready for review.

## Checklist

- [x] Have you added changelog file? Yes — one feature-level entry for the whole stack (`changelog/20260813_feature_vm_to_kubernetes_migration.md`).


api/mongodb/v1/mdb/mongodb_types.go
```diff
@@ -441,6 +467,11 @@ type DbCommonSpec struct {
 	// +kubebuilder:validation:Enum=SingleCluster;MultiCluster
 	// +optional
 	Topology string `json:"topology,omitempty"`
+
+	// +optional
+	ExternalMembers []ExternalMember `json:"externalMembers,omitempty"`
+
```
lucian-tosa, line 473: // +optional?

.idea/inspectionProfiles/Project_Default.xml
MaciejKaras, line 1: Those three files are not relevant

controllers/operator/common_controller.go
```diff
@@ -1106,12 +1107,12 @@ func ReconcileReplicaSetAC(ctx context.Context, d om.Deployment, spec mdbv1.DbCo
 		return xerrors.Errorf("cannot disable x509 internal cluster authentication")
 	}
 
-	excessProcesses := d.GetNumberOfExcessProcesses(resourceName)
-	if excessProcesses > 0 {
-		return xerrors.Errorf("cannot have more than 1 MongoDB Cluster per project (see https://docs.mongodb.com/kubernetes-operator/stable/tutorial/migrate-to-single-resource/)")
-	}
+	//excessProcesses := d.GetNumberOfExcessProcesses(resourceName)
```
MaciejKaras, line 1110: either uncomment it or remove completely

pkg/util/merge/merge.go
```diff
@@ -51,6 +51,14 @@ func StringToBoolMap(map1, map2 map[string]bool) map[string]bool {
 	return mergedMap
 }
 
+func StringsToSet(slice []string) map[string]struct{} {
```
MaciejKaras, line 54: nit: this is a standard convert method, not a merge method. Better place would be `stringutil.go` or just `util.go`

MaciejKaras (approved):

docker/mongodb-kubernetes-tests/kubetester/kubetester.py
```diff
@@ -295,6 +295,16 @@ def delete_pod(cls, namespace: str, name: str):
     def create_deployment(cls, namespace: str, body: Dict):
         cls.clients("appsv1").create_namespaced_deployment(body=body, namespace=namespace)
 
+    @classmethod
+    def create_or_update_statefulset(cls, namespace: str, body: Dict):
```
mircea-cosbuc, line 299: nit: might be out of scope of this PR, but it looks like we're introducing new methods in the Kubetester class rather than using the pattern of defining the standalone methods in `kubetester/__init__.py`. We really should standardize on one or another.

fealebenpae (approved):

mircea-cosbuc (approved):

controllers/operator/common_controller.go
lucian-tosa: this method is uncommented and fixed upstack

pkg/util/merge/merge.go
lucian-tosa: moved

filipcirtog:
evergreen retry

PR #1492, opened 2026-08-13
Support autoPEMKeyFilePath and X509 agent auth for external members

lucian-tosa:
# Summary

Adds autoPEMKeyFilePath plumbing and the X509 agent-auth operator changes
needed when the agent authenticates against members running on VMs, and resets
Ops Manager user mechanisms during migration so credentials are re-derived for
the target deployment.

Layer 2 in the VM-to-Kubernetes migration stack.

**Stacked PR 2** (stack #1505), base `pr-01-api-types`. This is one layer of the VM-to-Kubernetes

## Proof of Work

marked ready for review.

## Checklist


controllers/operator/agents/upgrade.go
```diff
@@ -94,6 +94,12 @@ type dbCommonWithNamespace struct {
 func doUpgrade(ctx context.Context, cmGetter configmap.Getter, secretGetter secrets.SecretClient, factory om.ConnectionFactory, mdbs []dbCommonWithNamespace) error {
 	for _, mdb := range mdbs {
 		log := zap.S().With(string(mdb.ResourceType), mdb.objectKey)
+		if len(mdb.GetExternalMembers()) > 0 { // TODO: for now lets skip this, later we should revisit, since normal deployment won't need them
```
MaciejKaras, line 97: unresolved TODO comment

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -193,15 +194,16 @@ func (r *ReconcileMongoDbMultiReplicaSet) Reconcile(ctx context.Context, request
 	}
 
 	agentCertSecretName := mrs.GetSecurity().AgentClientCertificateSecretName(mrs.GetName())
-	agentCertHash, agentCertPath := r.agentCertHashAndPath(ctx, log, mrs.Namespace, agentCertSecretName, "")
+	agentCertHash, defaultAgentCertPath := r.agentCertHashAndPath(ctx, log, mrs.Namespace, agentCertSecretName, "")
+	agentCertPath := EffectiveAgentCertPEMPath(defaultAgentCertPath, mrs.GetSecurity())
```
MaciejKaras, line 198: q: I thought that we do not support this new agentCertPath in MongoDBMultiCluster, but I see the changes are also applied here. Maybe instead we can just add validation that this field is not settable in MongoDBMultiCluster?

controllers/operator/mongodbmultireplicaset_controller.go
```diff
@@ -414,6 +416,13 @@ func (r *ReconcileMongoDbMultiReplicaSet) reconcileStatefulSets(ctx context.Cont
 	if err != nil {
 		return workflow.Failed(xerrors.Errorf("Failed to retrieve current automation config: %w", err))
 	}
+	var externalAgentVersion string
+	if len(mrs.Spec.GetExternalMembers()) > 0 {
```
MaciejKaras, line 420: q: same question here, should we support it for MC?

controllers/operator/mongodbstandalone_controller.go
```diff
@@ -242,6 +243,10 @@ func (r *ReconcileMongoDbStandalone) Reconcile(ctx context.Context, request reco
 		databaseSecretPath = r.VaultClient.DatabaseSecretPath()
 	}
 
+	agentCertSecretName := s.GetSecurity().AgentClientCertificateSecretName(s.Name)
```
MaciejKaras, line 246: q: is standalone still actively supported? Maybe we should just ignore this setting here or block?

MaciejKaras (approved):
I've added couple of questions but otherwise LGTM

fealebenpae (approved):

controllers/operator/agents/upgrade.go
lucian-tosa: removed upstack

controllers/operator/mongodbmultireplicaset_controller.go
lucian-tosa: removed upstack

controllers/operator/mongodbstandalone_controller.go
lucian-tosa: removed

controllers/operator/mongodbmultireplicaset_controller.go
lucian-tosa: removed

filipcirtog:
evergreen retry

PR #1494, opened 2026-08-13
Add connectivity validator and migration dry-run

lucian-tosa:
# Summary

Adds the connectivity-validator binary and the migration Job builder, and wires
a dry-run into the replica-set and sharded-cluster reconcilers: setting the
mongodb.com/migration-dry-run annotation launches a validator Job that checks
every external VM member is reachable from inside the cluster, without touching
Ops Manager config or StatefulSets. Covers the X509 path by passing the agent
certificate subject DN to the Job, and supports a configurable mongod TLS CA
path.

Layer 4 in the VM-to-Kubernetes migration stack.

**Stacked PR 4** (stack #1505), base `pr-03-replicaset-external-members`. This is one layer of the VM-to-Kubernetes

## Proof of Work

marked ready for review.

## Checklist


cmd/connectivity-validator/migration/connectivitycheck/validator.go
```diff
@@ -0,0 +1,297 @@
...
+	MongodTLSCAPath string
+	// ClientCertRequired indicates that the mongod requires a client certificate
+	// (clientCertificateMode: REQUIRE). When true, a missing cert file is an error rather
+	// than a silent fallback to CA-only TLS.
+	ClientCertRequired bool
+}
+
+// Validate runs the full connectivity check and returns an exit code.
+func Validate(ctx context.Context, cfg Config) int {
+	log := zap.S()
```
MaciejKaras, line 53: why we create log in each function and not pass it through as argument? Or initialise it as global variable?

cmd/connectivity-validator/main.go
```diff
@@ -0,0 +1,51 @@
...
+	encCfg := zap.NewProductionEncoderConfig()
+	encCfg.EncodeTime = zapcore.ISO8601TimeEncoder
+	core := zapcore.NewCore(
+		zapcore.NewJSONEncoder(encCfg),
+		zapcore.AddSync(os.Stdout),
+		zapcore.DebugLevel, // connectivitycheck logs diagnostics at debug
+	)
+	logger := zap.New(core, zap.AddCaller(), zap.AddStacktrace(zapcore.ErrorLevel))
+	defer func() { _ = logger.Sync() }()
+	zap.ReplaceGlobals(logger) // validator package uses zap.S()
```
MaciejKaras, line 28: why we use different loggers in main and validator.go?

config/manager/manager.yaml
```diff
@@ -55,6 +55,9 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            # Must match the image actually running (OLM/certified installs: set to the same ref the catalog pulls, e.g. Red Hat registry).
```
MaciejKaras, line 58: nit: how is this image and version updated? Manually?

MaciejKaras, line 58: nit: unnecessary comment

controllers/operator/migration/jobrunner.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// jobPodOutcome lists the Job's pods once and returns the connectivity-validator container's exit code and finished time.
+func jobPodOutcome(ctx context.Context, kubeClient client.Client, job *batchv1.Job) (int32, time.Time) {
+	var pods corev1.PodList
+	if err := kubeClient.List(ctx, &pods,
+		client.InNamespace(job.Namespace),
+		client.MatchingLabels{"job-name": job.Name},
+	); err != nil {
+		zap.S().Errorf("listing pods for job %s/%s: %v", job.Namespace, job.Name, err)
```
MaciejKaras, line 98: another way of creating logger, now inline

controllers/operator/common_controller.go
```diff
@@ -1350,3 +1352,76 @@ func ReconcileLogRotateSetting(conn om.Connection, agentConfig mdbv1.AgentConfig
...
+// No StatefulSets or Ops Manager config are modified.
+func (r *ReconcileCommonController) runConnectivityJob(
+	ctx context.Context,
+	mdb *mdbv1.MongoDB,
+	sts *appsv1.StatefulSet,
+	connectionString string,
+	allHostnames []string,
+	agentAuthMode string,
+	agentCertHash string,
+	operatorImage string,
```
MaciejKaras, line 1367: We are passing 9 parameters in this function, thats too much. I was thinking about extracting logic that calculates the `subjectDN` into single method and then we could create `job` object outside of `runConnectivityJob` and just pass it here. This way we will have `runConnectivityJob()` method just utilising job object + ctx and logger

controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -927,12 +928,47 @@ func (r *ShardedClusterReconcileHelper) Reconcile(ctx context.Context, log *zap.
...
+			log.Warnf("Failed to clear feature control from group: %s", conn.GroupID())
+		}
+		operatorImage := r.imageUrls[util.OperatorImageEnv]
+		if operatorImage == "" {
+			return r.updateStatus(ctx, sc, workflow.Failed(fmt.Errorf("cannot run connectivity dry-run: operator image unknown (set %s or deploy operator from Helm chart)", util.OperatorImageEnv)).
+				WithAdditionalOptions(mdbstatus.NewMigrationConditionOption(mdbstatus.MigrationCondition(
+					mdbstatus.MigrationPhaseConnectivityCheckFailed, "OperatorImageUnknown",
+					"Set MDB_OPERATOR_IMAGE or deploy with the Helm chart so the operator image is available for the validation Job.",
+				))), log)
+		}
```
MaciejKaras, line 944: this is duplicated in both controllers, could be exported somewhere.

controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -927,12 +928,47 @@ func (r *ShardedClusterReconcileHelper) Reconcile(ctx context.Context, log *zap.
 
 	r.automationAgentVersion = automationAgentVersion
 
+	// Connectivity dry-run: handle before any reconciliation that modifies state.
+	if sc.Annotations[opMigration.AnnotationDryRun] == "true" {
```
MaciejKaras, line 932: nit: can we have a method in the mdbv1 for this? check?

MaciejKaras (approved):

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -323,6 +327,31 @@ func (r *ReplicaSetReconcilerHelper) Reconcile(ctx context.Context) (reconcile.R
...
 
+	// 5a. Connectivity dry-run: launch a validation Job without touching OM or StatefulSets.
+	if isDryRun {
+		// let's not block OM UI in case the customer needs to fix something to get their
+		// dry-run to pass.
+		if result := controlledfeature.ClearFeatureControls(conn, conn.OpsManagerVersion(), log); !result.IsOK() {
+			result.Log(log)
+			log.Warnf("Failed to clear feature control from group: %s", conn.GroupID())
+		}
+		operatorImage := r.reconciler.imageUrls[util.OperatorImageEnv]
```
fealebenpae: I'm not in love with this approach because users on OLM would need to adjust the environment variable manually. How about using the Downward API to inject the operator pod's name into itself so we can query the API server and get the exact image the operator container is running? Bit circuitous, but more robust because it doesn't need user intervention, I think.

fealebenpae (approved):

controllers/operator/mongodbshardedcluster_controller.go
lucian-tosa: it's done upstack

config/manager/manager.yaml
lucian-tosa: it's using the same source as the image for the container, it's done automatically when regenerating the manifests

cmd/connectivity-validator/migration/connectivitycheck/validator.go
lucian-tosa: done

cmd/connectivity-validator/main.go
lucian-tosa: done

config/manager/manager.yaml
lucian-tosa: done

controllers/operator/migration/jobrunner.go
lucian-tosa: done

controllers/operator/common_controller.go
lucian-tosa: done

controllers/operator/mongodbshardedcluster_controller.go
lucian-tosa: done

controllers/operator/mongodbreplicaset_controller.go
lucian-tosa: done

filipcirtog:
evergreen retry

PR #1496, opened 2026-08-13
Add migrate-to-mck plugin subcommands with connection and validation

lucian-tosa:
# Summary

Adds the migrate-to-mck command group to the kubectl-mongodb plugin with the
mongodb and users subcommands: connects to Ops Manager, reads the source
deployment, and validates that it can be represented as MCK resources before
any CRs are generated.

Layer 6 in the VM-to-Kubernetes migration stack.

**Stacked PR 6** (stack #1505), base `pr-05-sharded-external-members`. This is one layer of the VM-to-Kubernetes

## Proof of Work

marked ready for review.

## Checklist


cmd/kubectl-mongodb/migrate-to-mck/command.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/mongodb/mongodb-kubernetes/controllers/om"
+)
+
+const defaultNamespace = "default"
+
+var MigrateCmd = &cobra.Command{
+	Use:   "migrate-to-mck",
+	Short: "Migrate MongoDB deployments to Kubernetes",
+	Long: `Generates Kubernetes Custom Resources from an Ops Manager/Cloud Manager automation
+config for migrating existing deployments to the operator.
```
viveksinghggits, line 19:
```suggestion
	Long: `Generates Kubernetes Custom Resources from an Ops Manager/Cloud Manager automation
config for migrating existing non-K8s deployments to the operator managed deployments.
```
Also, short and long desc don't seem to match. Short says, the tool migrates, Long says the tools generates the K8s manifests.

cmd/kubectl-mongodb/migrate-to-mck/connection.go
```diff
@@ -0,0 +1,129 @@
...
+	SystemLogRotate  *automationconfig.AcLogRotate
+	AuditLogRotate   *automationconfig.AcLogRotate
+}
+
+var omConnectionFactory om.ConnectionFactory = om.NewOpsManagerConnection
+
+type configReader struct {
+	metav1.ObjectMeta
+	configMapName, secretName, namespace string
+}
```
viveksinghggits, line 34: nit: the name of this type feels like it's an interface. And looks seems like it should be used to read config and would have a method `Read()`. Should we name it something else? I am not able to recommend something because I don't really understand the significance.

cmd/kubectl-mongodb/migrate-to-mck/connection.go
```diff
@@ -0,0 +1,129 @@
...
+	proj, err := project.FindProjectInsideOrganization(conn, config.ProjectName, org, log)
+	if err != nil {
+		return nil, err
+	}
+	if proj == nil {
+		return nil, fmt.Errorf("project %q not found in organization %s (%q)", config.ProjectName, org.ID, org.Name)
+	}
+
+	conn.ConfigureProject(proj)
+	return conn, nil
```
viveksinghggits, line 91: this function doesn't seem to be doing anything related to `ReadOnly`, but the function name says so. Also it seems to be creating the OM connection and configuring the project `conn.ConfigureProject(proj)`. It would be better to just create connection from here and the do the `ConfigureProject` at caller or in other function. This would make sure that the function does exactly what it says.

cmd/kubectl-mongodb/migrate-to-mck/mongodb.go
```diff
@@ -0,0 +1,65 @@
...
+kubectl mongodb migrate mongodb --config-map-name my-project --secret-name my-credentials --namespace mongodb`,
+	RunE: runGenerateMongodb,
+}
+
+func init() {
+	MongodbCmd.Flags().StringVar(&mFlags.configMapName, "config-map-name", "", "Name of the ConfigMap containing the OM connection details (baseUrl, orgId, projectName) (required)")
+	MongodbCmd.Flags().StringVar(&mFlags.secretName, "secret-name", "", "Name of the Secret containing the OM API credentials (publicKey, privateKey) (required)")
+	MongodbCmd.Flags().StringVar(&mFlags.namespace, "namespace", defaultNamespace, "Namespace of the ConfigMap and Secret")
+	MongodbCmd.Flags().StringVarP(&mFlags.outputFile, "output", "o", "", "Write generated CRs to this file instead of stdout")
+	MongodbCmd.Flags().StringVar(&mFlags.resourceNameOverride, "resource-name-override", "", "Kubernetes resource name (metadata.name) for the generated CR. When the replica set name is not a valid Kubernetes name it is auto-normalized, but this flag lets you choose a custom name. spec.replicaSetNameOverride is set automatically")
```
viveksinghggits, line 43: very nit: should this just be called `resource-name`? the passed one is used, if not passed default is used.
we have seen this pattern in kubectl flags for example when you create a deployment using `k create deployment`, the flag for replicas is just called replicas. 
```
-r, --replicas=1:
        Number of replicas to create. Default is 1.
```

cmd/kubectl-mongodb/migrate-to-mck/users.go
```diff
@@ -0,0 +1,63 @@
...
+	usersSecretsFile     string
+	resourceNameOverride string
+}
+
+var uFlags usersFlags
+
+var UsersCmd = &cobra.Command{
+	Use:   "users",
+	Short: "Generate MongoDBUser Kubernetes CRs",
+	Long: `Generates MongoDBUser Kubernetes CRs from an Ops Manager/Cloud Manager automation
```
viveksinghggits, line 24: nit: short starts with `Generate` and Long starts with `Generates`, I think we should have `Generate` at both places. Applies to the other subcommands/commands as well.

cmd/kubectl-mongodb/migrate-to-mck/connection.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	conn, err := resolveProjectReadOnly(config, credentials, log)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error resolving Ops Manager project: %w", err)
+	}
+	return conn, kubeClient, nil
+}
```
viveksinghggits, line 60: I noticed that we are not using `kubecClient` that's returned from this function at the call site. Should we still return it? Also, I was thinking we should create kubecclient separately and pass that to `prepareConnection`, instead of creating it here.

cmd/kubectl-mongodb/migrate-to-mck/validation.go
```diff
@@ -0,0 +1,432 @@
...
+				continue
+			}
+			return &proc, nil
+		}
+	}
+	return nil, fmt.Errorf("no voting+priority member found in replica set. Cannot determine source process")
+}
+
+// validateOneDeploymentPerProject ensures the project has exactly one replica set.
+// Sharded cluster support will be added in a follow-up PR.
```
viveksinghggits, line 97:
```suggestion
```

cmd/kubectl-mongodb/migrate-to-mck/validation.go
```diff
@@ -0,0 +1,432 @@
...
+	if len(proc.NetTLSSections()) == 0 || pkgtls.GetTLSModeFromMongodConfig(proc.Args()) == pkgtls.Disabled {
+		return []ValidationResult{{
+			Severity: SeverityWarning,
+			Message:  fmt.Sprintf("Process %q has no TLS. Set spec.additionalMongodConfig.net.tls.mode to \"disabled\" to match the operator and avoid a deployment change.", proc.Name()),
+		}}
+	}
+	return nil
+}
+
+// validateProcessConfigDrift warns when the source process logRotate/auditLogRotate differs from project-level config.
```
viveksinghggits, line 404:
```suggestion
// validateProcessConfigDrift warns when the source process' logRotate/auditLogRotate differs from project-level config.
```
?

cmd/kubectl-mongodb/migrate-to-mck/validation.go
```diff
@@ -0,0 +1,432 @@
...
+	results = append(results, validateProcessesAreValid(d)...)
+	results = append(results, validateAuthSchemaVersion(d)...)
+	results = append(results, validateTLS(sourceProcess)...)
+	results = append(results, validateProcessConfigDrift(sourceProcess, projectProcessConfigs)...)
+
+	return results
+}
+
+// validateProcessesAreValid checks that all processes are mongod or mongos and not disabled.
+func validateProcessesAreValid(d om.Deployment) []ValidationResult {
```
viveksinghggits, line 342:
```suggestion
func validateProcesses(d om.Deployment) []ValidationResult {
```

viveksinghggits (commented):
Looks pretty good, left some nits.

cmd/kubectl-mongodb/migrate-to-mck/connection.go
fealebenpae: @viveksinghggits compare to the `ReadOrCreateProject()` function in `controllers/operator/project/project.go` - it's "readonly" because it just tries to resolve the project without creating it if it's missing.

@lucian-tosa I struggled with this approach similarly when working on the OM Search Awareness project, but fortunately I was able to cut the scope and narrow down the Ops Manager APIs I was pulling in and avoided the `om.Connection` dependency. Perhaps a small comment at `resolveProjectReadOnly()` could add the context necessary? Or perhaps move it to the `controllers/operator/project` package where context can be inferred from its package siblings?

fealebenpae (approved):

cmd/kubectl-mongodb/migrate-to-mck/connection.go
mircea-cosbuc: this implements a [project.Reader](https://github.com/mongodb/mongodb-kubernetes/blob/master/controllers/operator/project/project.go#L24).

cmd/kubectl-mongodb/migrate-to-mck/connection_test.go
```diff
@@ -0,0 +1,96 @@
+package migratetomck
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
mircea-cosbuc, line 6: Nit: This looks like it could be written as one table test.

cmd/kubectl-mongodb/migrate-to-mck/validation.go
lucian-tosa: will be removed in a PR upstack

mircea-cosbuc (approved):

cmd/kubectl-mongodb/migrate-to-mck/connection.go
lucian-tosa: added an idiom to make it clear as to waht interface it implements. Thanks for the suggestion Mircea

lucian-tosa: renamed it as well to projectRefFromFlags

cmd/kubectl-mongodb/migrate-to-mck/users.go
lucian-tosa: done

cmd/kubectl-mongodb/migrate-to-mck/validation.go
lucian-tosa: done

cmd/kubectl-mongodb/migrate-to-mck/validation.go
lucian-tosa: done

cmd/kubectl-mongodb/migrate-to-mck/connection_test.go
lucian-tosa: done

cmd/kubectl-mongodb/migrate-to-mck/connection.go
lucian-tosa: it will be used upstack

cmd/kubectl-mongodb/migrate-to-mck/connection.go
lucian-tosa: added a comment

cmd/kubectl-mongodb/migrate-to-mck/command.go
lucian-tosa: done

cmd/kubectl-mongodb/migrate-to-mck/mongodb.go
lucian-tosa: I would keep it as is due to docs and enablement relying on this already

filipcirtog:
evergreen retry

viveksinghggits (approved):

PR #1497, opened 2026-08-13
Generate MongoDB and MongoDBUser CRs in migrate-to-mck

lucian-tosa:
# Summary

Implements CR generation for the plugin: translates an Ops Manager deployment
into a MongoDB resource plus MongoDBUser resources, covering the shared spec
(TLS, authentication mechanisms, member options, additional mongod config) and
OIDC/SCRAM credential handling.

Layer 7 in the VM-to-Kubernetes migration stack.

**Stacked PR 7** (stack #1505), base `pr-06-plugin-subcommands`. This is one layer of the VM-to-Kubernetes

## Proof of Work

marked ready for review.

## Checklist

### DIFF
--- a/api/mongodb/v1/mdb/mongodb_types.go
+++ b/api/mongodb/v1/mdb/mongodb_types.go
@@ -1690,38 +1690,96 @@ func (m *MongoDB) ObjectKey() client.ObjectKey {
 	return kube.ObjectKey(m.Namespace, m.Name)
 }
 
+// ldapFieldMapping holds both directions of a single field conversion so they stay co-located.
+type ldapFieldMapping struct {
+	toAC func(*Ldap, *ldap.Ldap)
+	toCR func(*ldap.Ldap, *Ldap)
+}
+
+// ldapField creates a bidirectional mapping for fields that copy directly between CR and AC.
+func ldapField[T any](getCR func(*Ldap) *T, getAC func(*ldap.Ldap) *T) ldapFieldMapping {
+	return ldapFieldMapping{
+		toAC: func(c *Ldap, a *ldap.Ldap) { *getAC(a) = *getCR(c) },
+		toCR: func(a *ldap.Ldap, c *Ldap) { *getCR(c) = *getAC(a) },
+	}
+}
+
+// ldapCustomField creates a bidirectional mapping for fields that require custom conversion logic.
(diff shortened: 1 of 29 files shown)


api/mongodb/v1/mdb/mongodb_types.go
```diff
@@ -1649,38 +1649,96 @@ func (m *MongoDB) ObjectKey() client.ObjectKey {
...
+	),
+}
+
+// GetLDAP converts the CR LDAP spec to the AC representation. See ConvertACLdapToCR for the reverse.
+func (m *MongoDB) GetLDAP(password, caContents string) *ldap.Ldap {
+	if !m.IsLDAPEnabled() {
+		return nil
+	}
+	mdbLdap := m.Spec.Security.Authentication.Ldap
+	ac := &ldap.Ldap{
```
viveksinghggits, line 1760: q: what does ac mean? I read ac above in `toAC` and `toCR` and I thought it means to automation config and to custom resource.
Does this also mean automation config here?

cmd/kubectl-mongodb/migrate-to-mck/generator.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// User credentials — maps "username:database" to a pre-created Secret name; no Secret YAML is written
+	ExistingUserSecrets map[string]string
+
+	// Prometheus credentials
+	PrometheusSecretName string // name of a pre-created Secret; no Secret YAML is written when set
+}
+
+// GenerateMongoDBCR generates a MongoDB CR for the given topology.
+func GenerateMongoDBCR(ac *om.AutomationConfig, opts GenerateOptions) (client.Object, string, error) {
```
viveksinghggits: apart from test at one place we don't seem to be using this second returned value anywhere in the code. and even if we were using it it's name of the resource that can be figured out from the first returned value itself, because that's resource.
So I think we should stop returning second value.

viveksinghggits (approved):

api/mongodb/v1/mdb/mongodb_types.go
lucian-tosa: yes

cmd/kubectl-mongodb/migrate-to-mck/generator.go
lucian-tosa: done

filipcirtog:
evergreen retry

mircea-cosbuc (approved):

PR #1500, opened 2026-08-13
Validate imported tool versions and standalone deployments in migrate-to-mck

lucian-tosa:
# Summary

Adds validation of the MongoDB tools version recorded on the imported
deployment, rejects standalone sources with a clear message instead of
generating an unusable resource, and relocates the migration dry-run annotation
constant into pkg/util so both the plugin and the operator share one definition.

Layer 10 in the VM-to-Kubernetes migration stack.

**Stacked PR 10** (stack #1505), base `pr-09-plugin-sharded`. This is one layer of the VM-to-Kubernetes

## Proof of Work

marked ready for review.

## Checklist


api/mongodb/v1/mdb/mongodb_validation.go
```diff
@@ -352,6 +352,15 @@ func replicasetMemberIsSpecified(ms MongoDbSpec) v1.ValidationResult {
 	return v1.ValidationSuccess()
 }
 
+func generatedResourceUsedCorrectImportToolVersion(m *MongoDB) v1.ValidationResult {
```
viveksinghggits, line 355: I think we should just call it `verify/validateToolVersion` or `verifyImportToolVersion`. `generatedResourceUsedCorrectImportToolVersion` is just too big and `generatedResource` doesn't seem to be adding a lot of value.

cmd/kubectl-mongodb/migrate-to-mck/validation.go
```diff
@@ -172,6 +173,24 @@ func validateEmbeddedConfigServer(d om.Deployment) []ValidationResult {
 	return results
 }
 
+// validateNoStandalones rejects standalone mongod processes, i.e. mongods that are not part of any
+// replica set.
```
viveksinghggits, line 177:
```suggestion
// validateNoStandalones rejects the deployment if it has standalone mongod processes, i.e. mongods that are not part of any
// replica set.
```

viveksinghggits (approved):

cmd/kubectl-mongodb/migrate-to-mck/validation.go
lucian-tosa: done

api/mongodb/v1/mdb/mongodb_validation.go
lucian-tosa: done

filipcirtog:
evergreen retry

mircea-cosbuc (approved):

PR #1501, opened 2026-08-13
Make agent download base and CA file path configurable

lucian-tosa:
# Summary

Makes the agent download base URL and the CA file path configurable rather than
hardcoded, so migrated deployments can point at an internal artifact mirror and
a custom CA bundle. Threads both through the plugin generator, the operator and
the generated resources.

Layer 11 in the VM-to-Kubernetes migration stack.

**Stacked PR 11** (stack #1505), base `pr-10-plugin-validation`. This is one layer of the VM-to-Kubernetes

## Proof of Work

marked ready for review.

## Checklist


controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -1236,25 +1238,13 @@ func (r *ShardedClusterReconcileHelper) doShardedClusterProcessing(ctx context.C
...
 		podEnvVars:           podEnvVars,
 		currentAgentAuthMode: currentAgentAuthMode,
-		certTLSType:          certSecretTypesForSTS,
 	}
 
 	if workflowStatus := validateMongoDBResource(sc, conn); !workflowStatus.IsOK() {
 		return workflowStatus
 	}
 
-	// Ensures that all sharded cluster certificates are either of Opaque type (old design)
```
mircea-cosbuc, line 1246: 🥳

mircea-cosbuc (approved):

viveksinghggits (approved):

filipcirtog:
evergreen retry

PR #1502, opened 2026-08-13
Track migration status on the MongoDB resource

lucian-tosa:
# Summary

Adds the MigrationStatus type and reports migration progress on the resource
status: which phase the migration is in, the members still to be moved, and the
conditions that gate progress. Builds on the dry-run conditions from layer 4 and
the shared annotation constant from layer 10.

Layer 12 in the VM-to-Kubernetes migration stack.

**Stacked PR 12** (stack #1505), base `pr-11-configurable-download-ca`. This is one layer of the VM-to-Kubernetes

## Proof of Work

marked ready for review.

## Checklist


filipcirtog:
evergreen retry

api/mongodb/v1/mdb/mongodb_types.go
```diff
@@ -1577,6 +1586,131 @@ func (m *MongoDB) IsOIDCEnabled() bool {
...
+}
+
+// sumClusterCounts totals a per-member-cluster count map.
+func sumClusterCounts(m map[string]int) int {
+	total := 0
+	for _, v := range m {
+		total += v
+	}
+	return total
+}
```
viveksinghggits, line 1633: This function name/comment can be improved. I am not sure if I understood what exactly the function does even after reading the comment. Is it figuring out all the members across clusters? 
But let's not make change as part of this PR, feel free to note this down and we can improve this later.

api/mongodb/v1/mdb/mongodb_types.go
```diff
@@ -1577,6 +1586,131 @@ func (m *MongoDB) IsOIDCEnabled() bool {
...
+
+// shardedClusterInClusterMemberCount totals the in-Kubernetes mongod/mongos processes described by a
+// reconciled sharded size breakdown. It is override-aware
```
viveksinghggits, line 1636:
```suggestion
// reconciled sharded size breakdown. It is override-aware.
```

viveksinghggits: for later.

api/mongodb/v1/status/migration_status.go
```diff
@@ -4,8 +4,39 @@ import (
...
+	// (status.members from the prior reconcile). The replica set is being extended with new k8s members.
+	MigratingReasonExtending MigratingConditionReason = "Extending"
+	// MigratingReasonPruning is used when spec.externalMembers count decreases — VM members are being removed.
+	MigratingReasonPruning MigratingConditionReason = "Pruning"
+	// MigratingReasonInProgress is the stable state: externalMembers exist, counts are unchanged,
+	// no dry-run active. Migration is active, but nothing is changing right now.
+	MigratingReasonInProgress MigratingConditionReason = "InProgress"
+	// MigratingReasonComplete is set on the Migrating=False condition after all external members are removed.
+	MigratingReasonComplete MigratingConditionReason = "MigrationComplete"
+)
```
viveksinghggits, line 24: I think all of these should be named `Migration` instead of `Migrating`. This properly conveys that these are migration reasons and we are correct using the value `MigrationComplete` but the const name says `MigratingReasonComplete`.

viveksinghggits (approved):

mircea-cosbuc (approved):

PR #1503, opened 2026-08-13
Add VM migration e2e coverage

lucian-tosa:
# Summary

Adds the end-to-end suites covering VM-to-Kubernetes migration for replica sets
and sharded clusters across authentication modes (no auth, SCRAM-SHA-1,
SCRAM-SHA-256, SCRAM with TLS, X509, LDAP, OIDC in its group and workforce
variants) and Prometheus, plus the MCK-to-MCK migration case, their shared
helpers, and the Evergreen task and variant registration for all of them.

Layer 13 in the VM-to-Kubernetes migration stack.

**Stacked PR 13** (stack #1505), base `pr-12-migration-status`. This is one layer of the VM-to-Kubernetes

## Proof of Work

marked ready for review.

## Checklist

### DIFF
--- a/docker/mongodb-kubernetes-tests/Dockerfile
+++ b/docker/mongodb-kubernetes-tests/Dockerfile
@@ -99,8 +99,4 @@ COPY release.json /tests/release.json
 # we use the public directory to automatically test resources samples
 COPY public /mongodb-kubernetes/public
 
-RUN if [ -f "/tests/multi-cluster-kube-config-creator_${TARGETARCH}" ]; then \
-    install -m 755 "/tests/multi-cluster-kube-config-creator_${TARGETARCH}" /usr/local/bin/multi-cluster-kube-config-creator; \
-    fi
-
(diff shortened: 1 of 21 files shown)


docker/mongodb-kubernetes-tests/Dockerfile
```diff
@@ -91,3 +91,5 @@ COPY public /mongodb-kubernetes/public
 RUN if [ -f "/tests/multi-cluster-kube-config-creator_${TARGETARCH}" ]; then \
     install -m 755 "/tests/multi-cluster-kube-config-creator_${TARGETARCH}" /usr/local/bin/multi-cluster-kube-config-creator; \
     fi
```
lucian-tosa, line 93: remove this

filipcirtog:
evergreen retry

viveksinghggits (approved):
Just skimmed through the PR didn't actually review it because it's just E2Es. Approving.

mircea-cosbuc (approved):

PR #1504, opened 2026-08-13
Support externalAccess for migrated deployments

lucian-tosa:
# Summary

Adds externalAccess support for migrated deployments: an effective
external-access configuration on the MongoDB spec that the replica-set and
sharded-cluster controllers use when building services for members reachable
from outside the cluster, with e2e coverage.

Layer 14 in the VM-to-Kubernetes migration stack.

**Stacked PR 14** (stack #1505), base `pr-13-e2e-coverage`. This is one layer of the VM-to-Kubernetes

## Proof of Work

CI runs only on the top PR of the stack rather than once per layer: one PR matrix there exercises
the whole implementation, including the `e2e_vm_migration_task_group` suites (registered in
layer 13). The layers below keep Evergreen suppressed with `[skip-ci]`.

## Checklist


lucian-tosa:
evergreen retry

filipcirtog:
evergreen retry

controllers/operator/construct/database_construction.go
```diff
@@ -104,6 +104,13 @@ type DatabaseStatefulSetOptions struct {
 	StsType                 StsType
 	AdditionalMongodConfig  *mdbv1.AdditionalMongodConfig
 
+	// ExternalAccessConfiguration is the external access configuration in force for this
+	// StatefulSet's tier and member cluster, already resolved by
+	// MongoDbSpec.EffectiveExternalAccessConfiguration. Nil means this tier is not externally
+	// exposed. Consumers must use this rather than reading mdb.Spec.ExternalAccessConfiguration,
```
mircea-cosbuc: Is this true? I wouldn't expect consumers to read this from the statefulset level config.

controllers/operator/mongodbshardedcluster_controller.go
```diff
@@ -1470,10 +1470,15 @@ func (r *ShardedClusterReconcileHelper) ensureSSLCertificates(ctx context.Contex
 // Note, that it doesn't remove any existing shards - this will be done later
 func (r *ShardedClusterReconcileHelper) createKubernetesResources(ctx context.Context, s *mdbv1.MongoDB, opts deploymentOptions, log *zap.SugaredLogger) workflow.Status {
 	if r.sc.Spec.IsMultiCluster() {
-		// for multi-cluster deployment we should create pod-services first, as doing it after is a bit too late
-		// statefulset creation loops and waits for sts to become ready, and it's easier for the replica set to be ready if
-		// it can connect to other members in the clusters
-		// TODO the same should be considered for external services, we should always create them before sts; now external services are created inside DatabaseInKubernetes function;
+		// Deliberate topology split: reconcileServices owns multi-cluster, where per-pod services must
```
mircea-cosbuc, line 1473: While the new comment is useful in describing what's going on in the code, the comment we're removing is still valid, unless we made a decision and implemented a different approach that makes the comment obsolete.

mircea-cosbuc (approved):

api/mongodb/v1/mdb/mongodb_validation.go
```diff
@@ -525,6 +527,120 @@ func noReplicaSetNameOverrideChanges(newObj, oldObj MongoDbSpec) v1.ValidationRe
...
+				domains[loc] = spec.EffectiveExternalDomain(t.component, t.tier, cluster)
+			}
+		}
+		return domains
+	}
+
+	// Replica sets and standalones only ever read the top level field, see MongoDbSpec.GetExternalDomain.
+	domains[externalDomainLocation{}] = spec.GetExternalDomain()
+
+	return domains
```
viveksinghggits, line 595: very nit: does this read better?

```suggestion
func externalDomainsByLocation(spec MongoDbSpec) map[externalDomainLocation]*string {
	domains := map[externalDomainLocation]*string{}

	if spec.ResourceType != ShardedCluster {
	    	// Replica sets and standalones only ever read the top level field, see MongoDbSpec.GetExternalDomain.
	domains[externalDomainLocation{}] = spec.GetExternalDomain()

    	return domains
	}
	tiers := []struct {
			component *ShardedClusterComponentSpec
			tier      ShardedClusterTier
		}{
			{spec.MongosSpec, TierMongos},
			{spec.ConfigSrvSpec, TierConfigSrv},
			{spec.ShardSpec, TierShard},
		}
		for _, t := range tiers {
			for _, cluster := range externalDomainClusterNames(spec, t.component) {
				loc := externalDomainLocation{tier: t.tier, cluster: cluster}
				domains[loc] = spec.EffectiveExternalDomain(t.component, t.tier, cluster)
			}
		}
		return domains

```

changelog/20260803_fix_srv_connection_string_external_domain.md
```diff
@@ -0,0 +1,7 @@
+---
+title: SRV connection strings use the configured external domain
+kind: fix
+date: 2026-08-03
+---
+
+* **MongoDB** (replica sets and sharded clusters): the `connectionString.standardSrv` value now uses `spec.externalAccess.externalDomain` when one is configured, instead of the internal `*.svc.cluster.local` service FQDN which is not reachable from outside the cluster. Provisioning the `_mongodb._tcp` SRV records in that zone is the customer's responsibility, as it already is for the A records `externalAccess` depends on. This completes the change started in CLOUDP-215627, which fixed only the non-SRV connection string. For MongoDBMultiCluster resources, only a top-level `spec.externalAccess.externalDomain` is honoured by this fix; per-cluster `externalDomain` values set in `clusterSpecList` are not yet used when building the SRV connection string.
```
viveksinghggits: I think we need to rethink this. 
1. It mentions CLOUDP ticket that we should remove.
2. We should just make the note smaller to just convey how the uses would be impacted (positively).

changelog/20260807_fix_reject_external_domain_changes.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Reject externalDomain changes on existing MongoDB resources
+kind: fix
+date: 2026-08-07
+---
+
+* **MongoDB**: the validating webhook now rejects adding, changing or removing an external domain on an already deployed resource. This covers `spec.externalAccess.externalDomain`, the per-tier `spec.{mongos,configSrv,shard}.externalAccess.externalDomain` fields and their per-member-cluster equivalents in `clusterSpecList`. Any of these transitions repoints existing members at new hostnames, which is unsupported and will lead to a Failed resource.
```
viveksinghggits:
```suggestion
* **MongoDB**: The validating webhook now rejects adding, changing or removing an external domain on an already deployed resource. This covers `spec.externalAccess.externalDomain`, the per-tier `spec.{mongos,configSrv,shard}.externalAccess.externalDomain` fields and their per-member-cluster equivalents in `clusterSpecList`. Any of these transitions repoints existing members at new hostnames, which is unsupported and will lead to a `Failed` resource.
```

controllers/operator/construct/database_construction.go
viveksinghggits: We need to clearly document the difference between `EffectiveExternalAccessConfiguration` and `ExternalAccessConfiguration`.

controllers/operator/construct/database_construction.go
```diff
@@ -201,6 +208,10 @@ func StandaloneOptions(additionalOpts ...func(options *DatabaseStatefulSetOption
 			StatefulSetSpecOverride: stsSpec,
 			MultiClusterMode:        mdb.Spec.IsMultiCluster(),
 			StsType:                 Standalone,
+			// create.DatabaseInKubernetes decides whether to create or delete the per-pod external
+			// services from this field, so a standalone with spec.externalAccess has to carry it here
+			// exactly as ReplicaSetOptions does; leaving it nil would delete the external service.
+			ExternalAccessConfiguration: mdb.Spec.ExternalAccessConfiguration,
```
viveksinghggits: is this comment correct. It talks about `create.DatabaseInKubernetes` but the field is diff.

viveksinghggits (approved):
We should fix the release note issue before merge and other nits can be worked on later.

PR #1507, opened 2026-08-13
CLOUDP-400899: Remove the RBACValid condition scaffolding

m1kola:
# Summary

It has been decided that the initial MCK 2.x release ships **without** the `RBACValid` condition on `MemberCluster`. This PR removes the condition scaffolding that had pre-landed on the feature branch.

**Removal:**
- Drops the `RBAC Valid` printcolumn and the RBACValid references in the `MemberCluster` status doc comment. The `Conditions` field and the status subresource stay as scaffolding for a future condition (e.g. duplicate clusterName signalling).
- Drops the `serviceaccounts get` self-read rule from member-cluster RBAC — its only consumer was the validation probe — and regenerates the CRDs and the gitops RBAC samples.
- The `mongodb.com/rbac-version` annotation keeps being rendered on member RBAC so validation can be reintroduced later without re-annotating.

Also included: the credential Secret's kubeconfig key is now shared between the plugin (writer) and the operator (reader) via a constant.

## Proof of Work

- `go test ./pkg/membercluster/... ./pkg/kubectl-mongodb/...` — green, envtest included
- `make precommit` — green
- `helm template` diff vs the feature branch: member render loses exactly the `serviceaccounts get` rule; base render unchanged. Plugin render (`generate-member-resources`) smoke-checked: annotations intact, rule gone
- `evergreen validate` — clean

## Checklist

- [x] Have you added changelog file? — no entry needed: nothing here is user-visible on any released branch (the feature branch is unreleased); `skip-changelog` label applied


pkg/membercluster/providermanager.go
```diff
@@ -68,13 +77,54 @@ func newProviderManager(baseCtx context.Context, c client.Reader, namespace stri
...
+		saNamespace:    contextInfo.Namespace,
+		kubeconfigHash: hex.EncodeToString(sum[:]),
+	}, nil
+}
+
+// ensure registers the provider entry for mc, rebuilding it when the CR's generation or
+// the credential hash changed. A resync or a status-only update (both leaving generation
+// and hash unchanged) is a no-op.
+func (m *providerManager) ensure(ctx context.Context, mc *operatorv1.MemberCluster, creds *credentials, log *zap.SugaredLogger) error {
 	m.mu.Lock()
```
m1kola, line 74: Note: there are some concurrency issues in here. Not a problem as long as we have max concurrent reconcile set to default `1` in this controller. But I still want to address this. I'm working on it in #1535.

pkg/membercluster/providermanager.go
```diff
@@ -68,13 +77,54 @@ func newProviderManager(baseCtx context.Context, c client.Reader, namespace stri
...
+	if err != nil {
+		return nil, fmt.Errorf("building REST config from credential secret %q: %v", secretName, err)
+	}
+	restConfig.Timeout = m.clientTimeout
+
+	contextInfo := cfg.Contexts[cfg.CurrentContext]
+	if contextInfo == nil || contextInfo.Namespace == "" {
+		// Our CLI tooling generates a kubeconfig with the operator namespace on the context,
+		// but if the user manually created the Secret, it might be missing.
+		return nil, fmt.Errorf("kubeconfig in credential secret %q has no namespace on context %q: set contexts[].context.namespace to the member cluster's operator namespace", secretName, cfg.CurrentContext)
```
lsierant: is this something that we're requiring currently (1.x) in order to having multi-cluster work correctly?

lsierant: I'm not sure if we should do this - context is not tied to a namespace. The namespace field is just a default namespace for this context. And in which namespace we're doing operations should stem from OperatorConfig and not context. 
What if the operator is watching multiple namespaces?

pkg/membercluster/providermanager_test.go
```diff
@@ -183,35 +220,106 @@ func TestRemoveDeregistersEntry(t *testing.T) {
...
+		{
+			name: "kubeconfig malformed",
+			secret: &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{Name: "mck-credential-cluster-bad", Namespace: testNamespace},
+				Data:       map[string][]byte{util.MemberClusterCredentialSecretKubeconfigKey: []byte("not: [valid")},
+			},
+			wantErrSubstr: `parsing kubeconfig in credential secret`,
+		},
+		{
+			name: "context namespace missing",
```
lsierant: As stated above I think we should remove those namespace validations

pkg/membercluster/rbacvalidation.go
```diff
@@ -0,0 +1,70 @@
...
+		return
+	}
+
+	saName := resourcenames.MemberClusterServiceAccountName(mc.Name)
+	sa := &corev1.ServiceAccount{}
+	if err := reader.Get(ctx, types.NamespacedName{Name: saName, Namespace: saNamespace}, sa); err != nil {
+		switch {
+		case apierrors.IsNotFound(err):
+			setRBACValidConditionFalse(mc, fmt.Sprintf(
+				"ServiceAccount %q not found in namespace %q of the member cluster. Regenerate and reapply member-cluster RBAC with 'kubectl mongodb multicluster generate-member-resources'.", saName, saNamespace))
```
lsierant: can we put the member cluster's name into the error messages here and around?

lsierant: Ha, I just realized it doesn't make sense - we're putting condition in the member cluster CR's status itself 🤦‍♂️

pkg/membercluster/reconciler.go
```diff
@@ -51,11 +72,44 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		// TODO(m1kola): surface a duplicate clusterName refusal on the CR's status via a
+		// dedicated condition (e.g. Accepted) instead of only erroring loudly.
+		return err
+	}
+
+	entry, ok := r.provider.Entries()[mc.Spec.ClusterName]
+	if !ok {
+		return fmt.Errorf("provider entry for MemberCluster %q unexpectedly missing after ensure", mc.Name)
+	}
+	r.validator.validate(ctx, mc, entry.Cluster.GetAPIReader(), creds.saNamespace)
```
lsierant: this looks sus: no return value captured, do we mutate some state here? 
First instinct looking at any "validate" function is that the returning err signify validation error, which probably should be captured.

m1kola: > this looks sus: no return value captured, do we mutate some state here?

@lsierant yes. When working with conditions it is difficult to just return an error because single condition can be `true`, `false`, `unkonown` and have multiple reasons. You need custom error types for that and a lot of boilerplate in the controller to handle all error types. When you have multiple conditions (which we will likely have - we need somehow indicate that multiple resources have same logical cluster name in `.spec.clusterName`) - then it gets more complex.

What we do instead:
1. In `Reconcile` we make a deep copy of the CR we are reconciling and pass it along to other methods.
2. Other methods modify status of the CR: set their respective conditions, etc.
3. At the end in reconcile we do a semantic deep equal of the state to determine if we need to update the status or not.

This is a common pattern when working with conditions in an operator. We need it in OLM v2 (you can see an example [here](https://github.com/operator-framework/operator-controller/blob/bf0266922f1ccf02d3efac324ec025e0574a4605/internal/operator-controller/controllers/clusterextension_controller.go#L137-L141)).

I've spent quite some time experimenting with code structure here: everything looked really too complex. Then I remembered about this pattern and ended up using it.

lsierant (approved):

pkg/membercluster/providermanager.go
m1kola: > What if the operator is watching multiple namespaces?

@lsierant this is not the namespace where we will be deploying workloads (that's `WATCH_NAMESPACE` - no change compared to MCK 1.x in this department). This is the namespace where `ServiceAccount` used by this relevant KubeConfig lives and we need to know that namespace to be able to run validation.

Alternative solution is to have a `namespace` field somewhere in the `spec` of `MemberCluster` CR so the operator knows in which namespace to look for its SA. I've considered that when implementing it, but didn't manage to convince myself that an additional required CRD field is worth adding when kubeconfig can carry this information.

However lately I've been thinking about "bring your own service account" implementation (so users can provision SAs on member clusters as a pre-requisite before doing multi-cluster installation which could make GitOps a bit easier). In that case we would probably need to say "use this SA name from this namespace" in `MemberCluster` CR somewhere. If we do that (**this is not in the scope currently**), then we can drop the validation and use NS from CR.

> is this something that we're requiring currently (1.x) in order to having multi-cluster work correctly?

In MCK 1.x we set namespace on the context too in `setup` command [here](https://github.com/mongodb/mongodb-kubernetes/blob/05089c358a542cad99da3018cd5cc46a228e1ebe/pkg/kubectl-mongodb/common/common.go#L878), but I'm not sure if it is every used. Probably not because there is no RBAC validation in MCK 1.x.

pkg/membercluster/providermanager_test.go
m1kola: Related to [this thread](https://github.com/mongodb/mongodb-kubernetes/pull/1507#discussion_r3878485479).

pkg/membercluster/providermanager.go
lsierant: But it feels wrong. We're using kubeconfig's semantics of a default namespace (which is completely optional) for particular context (SA + server) to store that information where it clearly belongs in the operator's configuration. The operator's configuration should know in which namespace we should look for the service account because the actual context and service account might be allowed to manage multiple namespaces. 

The operator should always look it's service account in the namespace where the operator is installed or it's configured where to look (but not via kubeconfig). The RBACs in member clusters' namespaces (in case of multi-namespace deployment) are just referencing that one service account. There is never a situation that multiple operator's service accounts are created per namespace. Each namespace gets only their workload service accounts.

m1kola: > The operator should always look it's service account in the namespace where the operator is installed or it's configured where to look (but not via kubeconfig). The RBACs in member clusters' namespaces (in case of multi-namespace deployment) are just referencing that one service account. There is never a situation that multiple operator's service accounts are created per namespace. Each namespace gets only their workload service accounts.

@lsierant not sure that I follow. Are you suggesting that when an operator is installed in `mongodb-operator` namespace on central cluster we should be assuming that on member clusters SAs also live in `mongodb-operator` namespace?

m1kola:
Putting back to draft because we've decided to remove RBACValid.

main.go
```diff
@@ -269,6 +269,7 @@ func run() error {
 	// startup-built snapshot.
 	memberClustersProvider := multicluster.NewProvider()
 
+	// TODO(m1kola): This shouldn't be wrapped in this condtion anymore, probably.
```
lsierant, line 272: definitely, not probably ;)

lsierant (approved):

main.go
m1kola: @lsierant yep. I have a PR for this already: https://github.com/mongodb/mongodb-kubernetes/pull/1533

Please take a look.

mircea-cosbuc (approved):

PR #1514, opened 2026-08-17
Support MongoDB Search for migrated deployments

lucian-tosa:
Stacked PR 15 — depends on #1504. Review only the top commit.

Carries MongoDB Search (mongot) wiring through a VM-to-Kubernetes migration.

**`migrate-to-mck`** — shards whose mongod settings differ (as they do when only some shards run mongot) no longer silently inherit shard 0's `additionalMongodConfig`. When all shards agree, the shared config goes in `spec.shard`; when they differ, `spec.shard` is omitted and each shard gets its own complete `shardOverride` entry (the operator replaces `additionalMongodConfig` wholesale rather than merging, so each entry is complete, not a delta).

**`controllers/searchcontroller`** — resolve a shard's external (VM) members through its Automation Config replica set name, so `spec.shardNameOverrides` (which `migrate-to-mck` emits whenever the VM replica set name differs from the Kubernetes shard name) doesn't silently drop them. Include those members in the mongot sync-source seeds, and omit the mongos Service while `mongosCount` is 0, since it resolves to no endpoints.

**Also** — fix the `PreserveUnknownFields` marker typo, make `ShardOverride.Members` `omitempty`, add e2e coverage for replica-set and sharded-cluster search migration, and build the `kubectl-mongodb` plugin for single-cluster local e2e runs (VM migration tests need it and nothing else built it).

## Testing
`go build ./...`, `go vet`, and the full unit suite for the touched packages pass locally. New unit coverage: `sharded_internal_search_source_test.go`, `generator_test.go`, plus two `migrate-to-mck` golden fixtures (`heterogeneous_shards`, `search_per_shard_mongot`). E2E: `e2e_vm_migration_replicaset_search`, `e2e_vm_migration_shardedcluster_search`.

### DIFF
--- a/cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
+++ b/cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
@@ -58,22 +58,29 @@ func generateShardedCluster(ac *om.AutomationConfig, opts GenerateOptions) (clie
 		return nil, fmt.Errorf("failed to build MongoDB spec: %w", err)
 	}
 
-	overrides := buildShardedClusterOverrides(k8sResourceName, acClusterName, configRS, acShards)
-	overrides.ConfigSrvSpec = buildShardedComponentSpec(ac.AgentSSL, processMap, configRS.Members())
-	// Shards may differ in their mongod settings. When they all agree, that shared config
-	// belongs in spec.shard. When they differ, spec.shard is left absent — populating it
-	// from one shard would assert that shard's settings about all the others — and each
-	// shard states its own config in an override instead.
-	shardConfigs := shardAdditionalMongodConfigs(ac.AgentSSL, processMap, shardRSes)
+	shardedClusterSpec := buildShardedClusterOverrides(k8sResourceName, acClusterName, configRS, acShards)
+	shardedClusterSpec.ConfigSrvSpec, err = buildShardedComponentSpec(ac.AgentSSL, processMap, configRS.Members())
+	if err != nil {
+		return nil, fmt.Errorf("config server replica set %q: %w", configRS.Name(), err)
+	}
+	// Shard configs, on K8s, derive from one of two places: either the spec.shard field
+	// set once for the entire ShardedClusterSpec, or from a shard-specific override. We
+	// maintain the invariant that exactly one of these routes is used: either all shards
+	// are identical and spec.shard is present, or spec.shard is absent and there is an
+	// override for each shard.
+	shardConfigs, err := shardAdditionalMongodConfigs(ac.AgentSSL, processMap, shardRSes)
+	if err != nil {
+		return nil, err
+	}
 	if common, uniform := commonShardConfig(shardConfigs); uniform {
 		if common != nil {
-			overrides.ShardSpec = &mdbv1.ShardedClusterComponentSpec{AdditionalMongodConfig: common}
+			shardedClusterSpec.ShardSpec = &mdbv1.ShardedClusterComponentSpec{AdditionalMongodConfig: common}
 		}
 	} else {
-		overrides.ShardOverrides = buildShardOverrides(k8sResourceName, shardConfigs)
+		shardedClusterSpec.ShardOverrides = buildShardOverrides(k8sResourceName, shardConfigs)
 	}
-	overrides.MongosSpec = buildMongosComponentSpec(ac.AgentSSL, mongosProcs)
-	spec.ShardedClusterSpec = overrides
+	shardedClusterSpec.MongosSpec = buildMongosComponentSpec(ac.AgentSSL, mongosProcs)
+	spec.ShardedClusterSpec = shardedClusterSpec
 
 	cr := &mdbv1.MongoDB{
 		TypeMeta:   metav1.TypeMeta{APIVersion: "mongodb.com/v1", Kind: "MongoDB"},
@@ -148,47 +155,60 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
 	}
 }
 
-// shardComponentConfig extracts the additionalMongodConfig for one replica-set component,
-// using its first member as representative. Returns nil when the component has no members,
-// its first member is absent from the process map, or nothing survives stripping the
-// operator-managed infrastructure fields.
-func shardComponentConfig(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) *mdbv1.AdditionalMongodConfig {
-	if len(members) == 0 {
-		return nil
+// shardComponentConfig extracts representative additionalMongodConfig from one component's
+// replica set, using pickSourceProcess to choose the member it is read from so a component
+// of a sharded cluster and a plain replica set pick the same way.
+//
+// Returns a nil config, and no error, when nothing survives the filtering: every field was
+// operator-managed, so there is nothing for the generated resource to carry. Nil rather than
+// an empty config because the field is an omitempty pointer, and omitempty only drops a nil
+// one: returning an empty config would write "additionalMongodConfig: {}" into every
+// generated resource that configures nothing, and the operator fills the field in itself
+// when it is absent.
+//
+// A member list with no usable source process is an inconsistent automation config and
+// returns an error rather than being read as "no config": the settings do exist in Ops
+// Manager, and silently dropping them would generate a resource that quietly disagrees with
+// the deployment it is meant to reproduce.
+func shardComponentConfig(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) (*mdbv1.AdditionalMongodConfig, error) {
+	proc, err := pickSourceProcess(members, processMap)
+	if err != nil {
+		return nil, err
 	}
-	proc, ok := processMap[members[0].Name()]
-	if !ok {
-		return nil
-	}
-	return applyClientCertificateMode(agentSSL, proc.AdditionalMongodConfig())
+	return applyClientCertificateMode(agentSSL, proc.AdditionalMongodConfig()), nil
 }
 
 // buildShardedComponentSpec wraps shardComponentConfig in a component spec, or returns nil
-// when there is no config to carry.
-func buildShardedComponentSpec(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) *mdbv1.ShardedClusterComponentSpec {
-	cfg := shardComponentConfig(agentSSL, processMap, members)
-	if cfg == nil {
-		return nil
+// when there is no config to carry, so the component is left out of the generated resource
+// entirely rather than appearing as an empty section.
+func buildShardedComponentSpec(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) (*mdbv1.ShardedClusterComponentSpec, error) {
+	cfg, err := shardComponentConfig(agentSSL, processMap, members)
+	if err != nil {
+		return nil, err
 	}
-	return &mdbv1.ShardedClusterComponentSpec{AdditionalMongodConfig: cfg}
+	if cfg == nil {
+		return nil, nil
+	}
+	return &mdbv1.ShardedClusterComponentSpec{AdditionalMongodConfig: cfg}, nil
 }
 
 // shardAdditionalMongodConfigs returns each shard's additionalMongodConfig, index-aligned
-// with shardRSes. Entries are nil for shards whose config could not be determined.
-func shardAdditionalMongodConfigs(agentSSL *om.AgentSSL, processMap map[string]om.Process, shardRSes []om.ReplicaSet) []*mdbv1.AdditionalMongodConfig {
+// with shardRSes. Entries are nil for shards whose config is entirely operator-managed.
+func shardAdditionalMongodConfigs(agentSSL *om.AgentSSL, processMap map[string]om.Process, shardRSes []om.ReplicaSet) ([]*mdbv1.AdditionalMongodConfig, error) {
 	configs := make([]*mdbv1.AdditionalMongodConfig, 0, len(shardRSes))
 	for _, rs := range shardRSes {
-		configs = append(configs, shardComponentConfig(agentSSL, processMap, rs.Members()))
+		cfg, err := shardComponentConfig(agentSSL, processMap, rs.Members())
+		if err != nil {
+			return nil, fmt.Errorf("shard replica set %q: %w", rs.Name(), err)
+		}
+		configs = append(configs, cfg)
 	}
-	return configs
+	return configs, nil
 }
 
 // commonShardConfig reports whether every shard shares one additionalMongodConfig, and if
-// so returns it. Comparison is on the marshalled map rather than pointer identity, so two
-// independently built but equivalent configs count as identical and map key ordering cannot
-// manufacture a false difference. When shards agree the caller emits spec.shard; when they
-// differ, spec.shard is left absent because it would otherwise assert one shard's settings
-// about all the others.
+// so returns it. Equality is value-based on the AdditionalMongodConfig records, and
+// insensitive to key orderings.
 func commonShardConfig(configs []*mdbv1.AdditionalMongodConfig) (*mdbv1.AdditionalMongodConfig, bool) {
 	if len(configs) == 0 {
 		return nil, true
@@ -204,14 +224,9 @@ func commonShardConfig(configs []*mdbv1.AdditionalMongodConfig) (*mdbv1.Addition
 	return configs[0], true
 }
 
-// buildShardOverrides emits one ShardOverride per shard that has a config, in ascending
-// shard index. Shards are never grouped by identical config: an explicit entry per shard
-// keeps the reader from having to infer which grouping is the default. Each entry carries
-// the shard's complete config rather than a delta, because the operator replaces
-// additionalMongodConfig wholesale rather than merging it
-// (controllers/operator/mongodbshardedcluster_controller.go, processShardOverride).
-// Shards with a nil config are skipped: an override with no additionalMongodConfig is a
-// no-op in the operator.
+// buildShardOverrides returns one ShardOverride per shard that has a config, in ascending
+// shard index. Each entry carries the shard's complete config, not a delta. Shards with a
+// nil config are skipped.
 func buildShardOverrides(k8sResourceName string, configs []*mdbv1.AdditionalMongodConfig) []mdbv1.ShardOverride {
 	var overrides []mdbv1.ShardOverride
 	for i, cfg := range configs {
--- a/cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
+++ b/cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
@@ -2,6 +2,7 @@ package migratetomck
 
 import (
 	"fmt"
+	"reflect"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -59,7 +60,18 @@ func generateShardedCluster(ac *om.AutomationConfig, opts GenerateOptions) (clie
 
 	overrides := buildShardedClusterOverrides(k8sResourceName, acClusterName, configRS, acShards)
 	overrides.ConfigSrvSpec = buildShardedComponentSpec(ac.AgentSSL, processMap, configRS.Members())
-	overrides.ShardSpec = buildShardedComponentSpec(ac.AgentSSL, processMap, shardRSes[0].Members())
+	// Shards may differ in their mongod settings. When they all agree, that shared config
+	// belongs in spec.shard. When they differ, spec.shard is left absent — populating it
+	// from one shard would assert that shard's settings about all the others — and each
+	// shard states its own config in an override instead.
+	shardConfigs := shardAdditionalMongodConfigs(ac.AgentSSL, processMap, shardRSes)
+	if common, uniform := commonShardConfig(shardConfigs); uniform {
+		if common != nil {
+			overrides.ShardSpec = &mdbv1.ShardedClusterComponentSpec{AdditionalMongodConfig: common}
+		}
+	} else {
+		overrides.ShardOverrides = buildShardOverrides(k8sResourceName, shardConfigs)
+	}
 	overrides.MongosSpec = buildMongosComponentSpec(ac.AgentSSL, mongosProcs)
 	spec.ShardedClusterSpec = overrides
 
@@ -136,8 +148,11 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
 	}
 }
 
-// buildShardedComponentSpec extracts additionalMongodConfig for a replica set component using its first member.
-func buildShardedComponentSpec(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) *mdbv1.ShardedClusterComponentSpec {
+// shardComponentConfig extracts the additionalMongodConfig for one replica-set component,
+// using its first member as representative. Returns nil when the component has no members,
+// its first member is absent from the process map, or nothing survives stripping the
+// operator-managed infrastructure fields.
+func shardComponentConfig(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) *mdbv1.AdditionalMongodConfig {
 	if len(members) == 0 {
 		return nil
 	}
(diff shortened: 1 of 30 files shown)


cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -59,7 +60,18 @@ func generateShardedCluster(ac *om.AutomationConfig, opts GenerateOptions) (clie
 
 	overrides := buildShardedClusterOverrides(k8sResourceName, acClusterName, configRS, acShards)
 	overrides.ConfigSrvSpec = buildShardedComponentSpec(ac.AgentSSL, processMap, configRS.Members())
-	overrides.ShardSpec = buildShardedComponentSpec(ac.AgentSSL, processMap, shardRSes[0].Members())
+	// Shards may differ in their mongod settings. When they all agree, that shared config
+	// belongs in spec.shard. When they differ, spec.shard is left absent — populating it
+	// from one shard would assert that shard's settings about all the others — and each
```
christopherthompsonwalsh-mongo: I find this comment confusing, coming to it cold. We're saying that if we have a uniform config for _all shards_, then we will populate something that calls itself "spec.shard", but if we have specs for _individual shards_, the thing that _sounds like_ it is the name for a shard spec is in fact empty... and instead populate an override.

The issue is that the concept of a "spec"'s "shard" config is reused in two different "places" I think, but we're not explicit about places.

I _think_ what we're really trying to say is "this field is either populated once for everyone, here, and individual shard configs inherit from it, **or** we include it in full for each specific shard" --- ie. we're not using any funky "set in both places so merge/invert/add-up-and-divide-by-𝜋" override-semantics?

I'd suggest the following small rewording:

```
    // Shard configs, on K8s, derive from one of two places: either the `spec.shard` field
    // set once for the entire ShardedClusterSpec, or from a shard-specific override. We
    // maintain the invariant that _exactly one_ of these routes is used: either all shards
    // are identical and `spec.shard` is present, or `spec.shard` is absent and there
    // is an override for each shard.
```

christopherthompsonwalsh-mongo: One thing that makes all this **extra** confusing is that the thing we are building, as the "non-shard-specific-overrides" path, is itself called... `overrides`, IIUC :-/

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -59,7 +60,18 @@ func generateShardedCluster(ac *om.AutomationConfig, opts GenerateOptions) (clie
 
 	overrides := buildShardedClusterOverrides(k8sResourceName, acClusterName, configRS, acShards)
```
christopherthompsonwalsh-mongo: I wonder if this might be better named as "shardedClusterSpecs" or "shardedClusterConfigs".

Calling it "overrides" is focussing on the _mechanism_ ("we merge this with K8s defaults") rather than the _semantics_ ("this carries all cluster-specific configuration information we have extracted").

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,22 +148,88 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
 	}
 }
 
-// buildShardedComponentSpec extracts additionalMongodConfig for a replica set component using its first member.
-func buildShardedComponentSpec(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) *mdbv1.ShardedClusterComponentSpec {
+// shardComponentConfig extracts the additionalMongodConfig for one replica-set component,
+// using its first member as representative. Returns nil when the component has no members,
+// its first member is absent from the process map, or nothing survives stripping the
```
christopherthompsonwalsh-mongo: This feels like smuggled semantics --- we didn't mention this when giving a positive description above, but now we're suddenly saying "oh yeah operator-related stuff doesn't count/we return nil if that's all-we-got." --- it leaves me, the reader, wondering if, when the result is non-`nil`, we actually get the operator-controlled fields in the result.

Suggest:

```
// shardComponentConfig extracts representative additionalMongodConfig from one shard's replica-set.
// The result will filtered to contain only significant fields, ie. those not managed by the K8s operator
// after migration. The implementation takes the config from the _first_ member of the set.
//
// Returns nil when the component has no members, its first member is absent from the process map, 
// or the result would be empty
```

(Always assuming I've understood the code!)

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,22 +148,88 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
...
+// operator-managed infrastructure fields.
+func shardComponentConfig(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) *mdbv1.AdditionalMongodConfig {
```
christopherthompsonwalsh-mongo: I wonder a little about returning `nil` here, and no `err`.

Is an _empty_ `AdditionalMongodConfig` really a sin? If not, should we return that in the case where all the fields are operator-managed, rather than have an out-of-band nil?

If the first member is absent from the process map --- well, is that healthy and normal? It _feels_ weird. Would we trust the results of this algorithm in that condition? Is this actually an error?

If the replica set is empty, same question really --- something deeply odd is happening, is it safe to proceed? Here, especially, my instinct would be strongly no --- we're looking to take signal about config from this set of things, maybe if the set is transiently empty the config still "exists", when the set becomes non-empty we suddenly have config live in OM that we didn't reflect in the results here?

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,22 +148,88 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
 	}
 }
 
-// buildShardedComponentSpec extracts additionalMongodConfig for a replica set component using its first member.
-func buildShardedComponentSpec(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) *mdbv1.ShardedClusterComponentSpec {
+// shardComponentConfig extracts the additionalMongodConfig for one replica-set component,
```
christopherthompsonwalsh-mongo: Do we think, someday, someone might be confused as to why there are so many shard overrides? (I hear tell of customers with on the order of 1000 shards...)

Should we (can we even?) emit a log line or some other sideband info saying "we are exploding this sharded clusters into NNN overrides because *this* shard is naughty and different"?

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,22 +148,88 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
...
+	}
+	return configs
+}
+
+// commonShardConfig reports whether every shard shares one additionalMongodConfig, and if
+// so returns it. Comparison is on the marshalled map rather than pointer identity, so two
+// independently built but equivalent configs count as identical and map key ordering cannot
+// manufacture a false difference. When shards agree the caller emits spec.shard; when they
+// differ, spec.shard is left absent because it would otherwise assert one shard's settings
+// about all the others.
```
christopherthompsonwalsh-mongo: "When shards agree the caller emits spec.shard; when they differ, spec.shard is left absent because it would otherwise assert one shard's settings about all the others."

Delete this sentence: it's describing behaviour defined by the _caller_, so it's in the wrong place.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,22 +148,88 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
...
+	for _, rs := range shardRSes {
+		configs = append(configs, shardComponentConfig(agentSSL, processMap, rs.Members()))
+	}
+	return configs
+}
+
+// commonShardConfig reports whether every shard shares one additionalMongodConfig, and if
+// so returns it. Comparison is on the marshalled map rather than pointer identity, so two
+// independently built but equivalent configs count as identical and map key ordering cannot
+// manufacture a false difference. When shards agree the caller emits spec.shard; when they
```
christopherthompsonwalsh-mongo: "Comparison is on the marshalled map rather than pointer identity, so two independently built but equivalent configs count as identical and map key ordering cannot manufacture a false difference."

This is implementation detail: I don't want to know that we marshal maps to solve Go's lack of an "Equals" trait/typeclass/interface. It may be jolly clever but if I want to know _how_ we got equality working I'll read the code. Delete it! Or move it to be a comment in the actual implementation. (*) 

It **is** probably relevant to say we mean what I'd call "real" semantic/value equality, and (sigh!) it's probably contentious what "real" means, so I'd suggest the following:

"Equality is value-based on the AdditionalMongodConfig records, and insensitive to key orderings."

(*) ".. a quote regularly attributed to the American novelist William Faulkner "In writing, you must kill all your darlings".(https://en.wikipedia.org/wiki/Kill_Your_Darlings_(magazine)#cite_note-:0-4)" ;-) ;_)_

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,22 +148,88 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
...
+		if !reflect.DeepEqual(first, cfg.ToMap()) {
+			return nil, false
+		}
+	}
+	return configs[0], true
+}
+
+// buildShardOverrides emits one ShardOverride per shard that has a config, in ascending
+// shard index. Shards are never grouped by identical config: an explicit entry per shard
+// keeps the reader from having to infer which grouping is the default. Each entry carries
```
christopherthompsonwalsh-mongo: "Shards ... infer which grouping is the default".

This is arguing for the design of the _higher level_ functionality. If it belongs anywhere, it belongs on the caller.

Here, it should be deleted: this is a helper function that does one job. No shame!

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,22 +148,88 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
...
+// the shard's complete config rather than a delta, because the operator replaces
+// additionalMongodConfig wholesale rather than merging it
+// (controllers/operator/mongodbshardedcluster_controller.go, processShardOverride).
```
christopherthompsonwalsh-mongo: "because .... processShardOverride)"

Delete this entire phrase. I don't _care_ why, it's a design decision meant elsewhere, and we shouldn't be entangling this bit of code with that other piece of code. Especially since we haven't explained what on earth it might be a delta _against_ --- we have to do nontrivial intellectual work even to establish there _is_ "another leading brand" (ie. another possible solution).

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,22 +148,88 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
...
+func buildShardOverrides(k8sResourceName string, configs []*mdbv1.AdditionalMongodConfig) []mdbv1.ShardOverride {
+	var overrides []mdbv1.ShardOverride
+	for i, cfg := range configs {
+		if cfg == nil {
+			continue
+		}
+		overrides = append(overrides, mdbv1.ShardOverride{
+			// Must be the K8s StatefulSet name; sharded-cluster validation rejects any
+			// other form. Same formula as ShardNameOverrides above.
+			ShardNames: []string{fmt.Sprintf("%s-%d", k8sResourceName, i)},
```
christopherthompsonwalsh-mongo, line 233: Not for this PR but it might be nice to factor out our naming schemes to a common library someday.

controllers/om/process.go
```diff
@@ -41,6 +41,18 @@ const (
 // arguments present on the mongod config level. AdditionalMongodConfig
 // strips these to isolate user-supplied config. When a new
 // infrastructure-managed field is introduced, add its path here.
+//
+// MongoDB Search setParameters (mongotHost, searchIndexManagementHostAndPort,
+// searchTLSMode, useGrpcForSearch, skipAuthenticationToMongot,
+// skipAuthenticationToSearchIndexManagementServer) are deliberately NOT listed. They were
```
christopherthompsonwalsh-mongo: Please condense this --- it's an interesting story, but belongs in a JIRA ticket not in the codebase.

We just want a comment describing the behaviour **at this revision**, none of the story about "expectations" or "originally". And I definitely don't want to know about "applySearchOverrides" returning early! That's implementation details.

I'd suggest that the story we actually need to tell is something like this --- the below will of course require filling out with the context I am missing (but would like to get from a comment ;-) )

- There are two kinds of config parameter (?), and they are distinguished by PQR (whether the controller manages them?)
- We handle them differently: set (a) is blah, set (b) is frobnitz
- There are three system states to consider behaviourally:
   1. **pre-migration**: everything flows through ??? state
   2. **during-migration**: the MongoDBSearch CR is populated from the frobnitz, and is blah
   3. **at migration completion**: turtles all the way down.

(TBH, even that story _might_ be too much detail here: the comment on this field _could_ be just "here are fields that come under the control of infrastructure during a migration, and which are therefore not propagated from X to Y". And maybe link to wherever you want to tell the story. **But tbh I think it's fine to tell the story here**.)

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -329,12 +329,13 @@ func (r *ReplicaSetReconcilerHelper) Reconcile(ctx context.Context) (reconcile.R
 		return r.updateStatus(ctx, workflow.Failed(err))
 	}
 
+	isDryRun := rs.IsMigrationDryRun()
```
christopherthompsonwalsh-mongo: This edit looks like a null op --- but maybe it makes a subsequent merge easier?

controllers/searchcontroller/sharded_internal_search_source.go
```diff
@@ -37,24 +37,59 @@ func (r *ShardedInternalSearchSource) GetShardCount() int {
 	return r.Spec.ShardCount
 }
 
+// externalMembersForK8sShardName maps the Kubernetes shard name the search plan works with
+// (MongoDB.ShardName(i)) to the Automation Config replica set name that external members are
+// recorded under (MongoDB.ShardACRsName(i), which spec.shardNameOverrides can change), and
+// returns that shard's external mongod members. Filtering by the Kubernetes name directly
+// would silently return nothing whenever an override is set -- which is exactly what
+// migrate-to-mck emits when the VM replica set name differs from the Kubernetes shard name.
```
christopherthompsonwalsh-mongo: Delete "Filtering by the Kubernetes operator .... shard name" 

I do not care about what _would_ have happened in a totally parallel reality where I do something which a) has not been contextualised properly b) sounds a bit .... debatably rational ... to me.

Additionally, the latter half of the comment is discussing what an unrelated non-code-entity ("migrate-to-mck") "emits"... totally unhelpful, almost completely irrelevant. Claude?

controllers/searchcontroller/sharded_internal_search_source.go
```diff
@@ -37,24 +37,59 @@ func (r *ShardedInternalSearchSource) GetShardCount() int {
...
+func (r *ShardedInternalSearchSource) externalMembersForK8sShardName(shardName string) []mdbv1.ExternalMember {
+	for i := 0; i < r.Spec.ShardCount; i++ {
+		if r.ShardName(i) == shardName {
+			return r.Spec.GetExternalMembersForRS(r.ShardACRsName(i))
+		}
+	}
+	return nil
+}
+
+// HostSeeds returns the sync-source seed list for one shard: the Kubernetes mongods first, then
```
christopherthompsonwalsh-mongo: What is a sync-source seed list? Mostly out of curiousity, it feels like a term of art.

The ':' should be a ".", and then the next bit of stuff should be its own sentence. (The first sentence should tell me declaratively what this thing gets me, the next sentence is detail about what rule has been chosen.)

controllers/searchcontroller/sharded_internal_search_source.go
```diff
@@ -37,24 +37,59 @@ func (r *ShardedInternalSearchSource) GetShardCount() int {
...
+// any external (VM) members still in the shard. During a VM migration mongodsPerShardCount may be
```
christopherthompsonwalsh-mongo: This ("During a VM migration ...") feels like a totally obvious corollary from the previous sentence.

Unless you are deeply attached, delete.

controllers/searchcontroller/sharded_internal_search_source.go
```diff
@@ -37,24 +37,59 @@ func (r *ShardedInternalSearchSource) GetShardCount() int {
...
+// 0, in which case the external members are the only seeds.
 func (r *ShardedInternalSearchSource) HostSeeds(shardName string) ([]string, error) {
 	members := r.Spec.MongodsPerShardCount
 	clusterDomain := r.Spec.GetClusterDomain()
 	port := r.Spec.GetAdditionalMongodConfig().GetPortOrDefault()
 
-	seeds := make([]string, members)
```
christopherthompsonwalsh-mongo, line 45: This is a lot of lines of code... shouldn't this just be some `append` calls using a couple of helper functions?

(I mean, structurally it's just `concat(members.map(getSeedName), externalMembers.map(m -> m.Hostname))` which is IMHO a darn sight clearer. But even in Go we could do:
```
var seeds []string
append(seeds, getMemberSeedNames(members)...)
append(seeds, getExtenalSeedNames(externalMembers))
```

If you were feeling avant-garde, you could even use the `iter` package.

controllers/searchcontroller/sharded_internal_search_source.go
```diff
@@ -37,24 +37,59 @@ func (r *ShardedInternalSearchSource) GetShardCount() int {
...
 
+// MongosHostsAndPorts returns the routers mongot should talk to: the Kubernetes mongos Service
+// plus any external (VM) mongos still in the cluster. The Service entry is omitted while
+// mongosCount is 0 (a valid mid-migration state) because it would resolve to no endpoints.
 func (r *ShardedInternalSearchSource) MongosHostsAndPorts() []string {
 	clusterDomain := r.Spec.GetClusterDomain()
 	port := r.Spec.GetAdditionalMongodConfig().GetPortOrDefault()
-	return []string{fmt.Sprintf("%s.%s.svc.%s:%d", r.ServiceName(), r.Namespace, clusterDomain, port)}
+
+	hosts := make([]string, 0, 1+len(r.Spec.ExternalMembers))
```
christopherthompsonwalsh-mongo: Same comment as above.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: Took your wording. And you're right about the naming clash. The local variable is now `shardedClusterSpec`.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: Renamed to `shardedClusterSpec`. Agreed, the old name described how it gets merged rather than what it holds. I kept the function called buildShardedClusterOverrides because that one really does only work out the name override fields.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: Used your wording. The filtering is stated up front now instead of turning up as an exception at the end. One change to it: the config comes from the first voting member that resolves to a process, not simply the first member, because I reused pickSourceProcess for the choice. More on that in the next reply.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: You were right to push here and this turned out to be a real incosistency rather than a style question.

On a missing member or an empty replica set, it now errors. Before, the shard was skipped outright by if cfg == nil { continue } in buildShardOverrides, so its Ops Manager settings just disappeared from the generated resource. That is exactly the "would we trust the results" case you were worried about.

On which member to read from, I reused the existing `pickSourceProcess`. So a shard and a plain replica set now pick the same way, including `skipping non voting` and `zero priority members`. 

I also found that `ValidateMigration` only ran `pickSourceProcess` on replicaSets[0], which for a sharded cluster is just one shard. Added `validateSourceProcessPerReplicaSet` so every shard and the config server get checked before generation starts, with the failing replica set named in the message.

On the empty config, I kept nil on purpose. The field is an `omitempty` pointer and `omitempty` only drops `nil`, so an empty struct would write `additionalMongodConfig: {}` into every generated resource that configures nothing.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: Good point and I have not done it. The generator has no logger passed into it at the moment, so it is more than a one liner. If ok, I can do that in a later PR after this whole stack gets merged.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: Deleted. You are right, it was describing what the caller does.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: Gone, replaced with your line:`Equality is value-based on the AdditionalMongodConfig records, and insensitive to key orderings`.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: Deleted.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: Deleted the whole phrase.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: Agreed, but treating it as out of scope a you said

controllers/om/process.go
filipcirtog: Cut from eleven lines to four. It now says what the fields are, that the generated resource carries them, and that the search controller overwrites the whole subtree once it targets the resource. Dropped the "originally" and "was expected to" story along with the applySearchOverrides internals.

controllers/operator/mongodbreplicaset_controller.go
filipcirtog: It was. Pure line movement with nothing reading the variable in between. Reverted, so that file drops out of this PR entirely.

controllers/searchcontroller/sharded_internal_search_source.go
filipcirtog: Both sentences deleted. It now just says what the mapping does. yes, claude xd 👍

controllers/searchcontroller/sharded_internal_search_source.go
filipcirtog: Now reads `HostSeeds returns the hosts mongot syncs one shard's data from. Kubernetes mongods come first, then any external (VM) members still in the shard.` Colon swapped for a full stop and the declarative sentence goes first.

controllers/searchcontroller/sharded_internal_search_source.go
filipcirtog: Deleted, not attached to it :) .

controllers/searchcontroller/sharded_internal_search_source.go
filipcirtog: Pulled out externalHostnames and externalMongosHostnames, so both functions end in a single append.

controllers/searchcontroller/sharded_internal_search_source.go
filipcirtog: same as above.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,20 +155,94 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
...
 
-// buildShardedComponentSpec extracts additionalMongodConfig for a replica set component using its first member.
-func buildShardedComponentSpec(agentSSL *om.AgentSSL, processMap map[string]om.Process, members []om.ReplicaSetMember) *mdbv1.ShardedClusterComponentSpec {
-	if len(members) == 0 {
-		return nil
+// shardComponentConfig extracts representative additionalMongodConfig from one component's
+// replica set, using pickSourceProcess to choose the member it is read from so a component
+// of a sharded cluster and a plain replica set pick the same way.
+//
+// Returns a nil config, and no error, when nothing survives the filtering: every field was
```
christopherthompsonwalsh-mongo: One more small tweak --- this para getting into detail very fast, what I need to know is what nil means/broadly when it happens, and when errors occur.

So I suggest:
```
The result is`nil` when there is no config for the generated additionalMongodConfig 
resource to carry. This can happen if all fields were operator-managed (and results in the
additionalMongodConfig being omitted, in preference to `additionalMongodConfig: {}`).
```

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,20 +155,94 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
...
+// operator-managed, so there is nothing for the generated resource to carry. Nil rather than
+// an empty config because the field is an omitempty pointer, and omitempty only drops a nil
+// one: returning an empty config would write "additionalMongodConfig: {}" into every
+// generated resource that configures nothing, and the operator fills the field in itself
+// when it is absent.
+//
+// A member list with no usable source process is an inconsistent automation config and
```
christopherthompsonwalsh-mongo: This is another Claude-esque "I did XYZ rather than PQR because PQR has cooties" -- I think this kind of comment is _usually_ an antipattern, it only works for the reader if they were secretly already imagining PQR as a baseline, which often isn't the case --- there's no accounting for readers.

I'd suggest instead:

```
Inconsistent automation configs result in a non-nil `error` being returned; in particular
if there is no usable source process for the member list.
```

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
```diff
@@ -136,20 +155,94 @@ func buildShardedClusterOverrides(k8sResourceName, acClusterName string, configR
...
 	}
-	proc, ok := processMap[members[0].Name()]
-	if !ok {
-		return nil
+	return applyClientCertificateMode(agentSSL, proc.AdditionalMongodConfig()), nil
+}
+
+// buildShardedComponentSpec wraps shardComponentConfig in a component spec, or returns nil
+// when there is no config to carry, so the component is left out of the generated resource
+// entirely rather than appearing as an empty section.
```
christopherthompsonwalsh-mongo: Nit: delete "so the component is left out of the generated resource entirely...".

It's describing what the surrounding code is doing, not what this fn does.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
christopherthompsonwalsh-mongo: SG :-)

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
christopherthompsonwalsh-mongo: LG, thanks.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
christopherthompsonwalsh-mongo: Makes sense.

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
christopherthompsonwalsh-mongo: SG.

controllers/searchcontroller/sharded_internal_search_source.go
christopherthompsonwalsh-mongo: So we've still got the for-loop --- I think probably it would be clearer if it went into a helper too. Then you could even make this entire function:

`return append(r.kubernetesMongodHosts(members), externalHostnames(externalMembers)), nil`

... which, I dunno, maybe Go finds embarrassingly short (well, it would get longer if either helper could err), but I really think makes this function incredibly readable.

christopherthompsonwalsh-mongo (approved):
Approval subject to addressing small nits.

viveksinghggits (approved):

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: fixed! ty!

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: fixed! ty!

cmd/kubectl-mongodb/migrate-to-mck/sharded_cluster_generator.go
filipcirtog: fixed! ty!

PR #1517, opened 2026-08-18
CLOUDP-400899: Wait for the member SA token in generate-member-registration

m1kola:
# Summary

`generate-member-registration` previously failed immediately when the member cluster's ServiceAccount token Secret (`mck-member-<cluster>-token`) was not yet populated by Kubernetes's token controller, so every caller hand-rolled a wait loop (the search tutorial snippets and the E2E harness each carried one). The command now polls the Secret's `token`/`ca.crt` keys for up to a minute (`memberregistration.DefaultTokenWaitTimeout`) before failing, and the caller-side wait loops (snippets `12_`/`13_`/`ra-02`, `conftest.py`'s `_wait_for_member_sa_token`) are removed. Timeout errors keep the original diagnostics, prefixed with `timed out after 1m0s waiting for the member ServiceAccount token:`.

## Proof of Work

- Unit: `pkg/kubectl-mongodb/memberregistration` — new `TestGenerate_WaitsForTokenPopulation` and table-driven `TestGenerate_TokenWaitTimeout`; pre-existing tests green unchanged. `go test ./pkg/kubectl-mongodb/...` + `make precommit` pass locally.
- e2e: covered in CI by the multi-cluster code-snippet tasks and the multi-cluster E2E harness (`configure_multi_cluster_members`).

## Checklist

    - No new file: amended the existing `changelog/20260716_feature_generate_member_registration.md` entry (the command is unreleased); `skip-changelog` label applied.

---


cmd/kubectl-mongodb/multicluster/generatememberregistration/generate_member_registration.go
```diff
@@ -89,5 +91,6 @@ func parseFlags() (memberregistration.Options, error) {
 		MemberClusterNamespace:   flags.memberClusterNamespace,
 		OperatorNamespace:        flags.operatorNamespace,
 		MemberClusterLogicalName: memberClusterLogicalName,
+		TokenWaitTimeout:         memberregistration.DefaultTokenWaitTimeout,
```
m1kola, line 94: Plumbing is there if we decide to add a flag to make it customisable later (if there is need for that), but I deliberately did not not add it straight away.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
```diff
@@ -91,16 +97,52 @@ func Generate(ctx context.Context, memberClusterClient kubernetes.Interface, mem
...
+// not exist yet, and transient API errors should not abort the wait.
+func waitForTokenSecret(ctx context.Context, memberClusterClient kubernetes.Interface, tokenSecretName string, opts Options) ([]byte, []byte, error) {
+	var token, ca []byte
+	var lastErr error
+	cond := func(ctx context.Context) (bool, error) {
+		var err error
+		token, ca, err = readTokenSecret(ctx, memberClusterClient, opts.MemberClusterNamespace, tokenSecretName)
+		lastErr = err
+		return err == nil, nil
+	}
```
josvazg, line 138: **q**: This `cond` function treats all errors as retry-able, so RBAC/Forbidden errors retry for full TokenWaitTimeout (1 min) instead of failing fast. Is that intended?

pkg/kubectl-mongodb/memberregistration/memberregistration.go
```diff
@@ -40,30 +50,26 @@ type Options struct {
...
-		return "", xerrors.Errorf("token secret %s/%s has no %q key yet; wait for Kubernetes to populate the ServiceAccount token", opts.MemberClusterNamespace, tokenSecretName, corev1.ServiceAccountRootCAKey)
+	token, ca, err := waitForTokenSecret(ctx, memberClusterClient, tokenSecretName, opts)
+	if err != nil {
+		return "", err
 	}
 
 	kubeconfig, err := buildKubeConfig(opts.MemberClusterName, memberClusterServerURL, opts.MemberClusterNamespace, ca, token)
 	if err != nil {
-		return "", xerrors.Errorf("building kubeconfig: %w", err)
+		return "", xerrors.Errorf("building kubeconfig: %v", err)
```
josvazg, line 72: **q**: This was a mistake right?

```suggestion
		return "", xerrors.Errorf("building kubeconfig: %w", err)
```

m1kola: No, there is no point in wrapping this error here and making it part of the API surface: we do not check it.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
m1kola: @josvazg you are right. We can distinguish errors and fail fast, but it will make things slightly more complicated here and I think the juice is not worth the squeeze.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
josvazg: If we ever check it later or if we add any intermediate layer, %w will already be there. Basically %v for error on a Errorf is kind of a linting notice already.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
josvazg: This wait is customer facing, might be annoying for end users to wait on something they now is broken. Even more so if they do not see why they were waiting, and later find out.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
m1kola: > If we ever check it later or if we add any intermediate layer, %w will already be there.

If we ever check it later - we can modify it later. Currently there is no consumer and absolutely no motivation to wrap it.

You do not add a return value to a function before you have consumers for it, right? Same logic here.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
m1kola: I don't think optimising for this edge case is really worth it. I propose that we go with the simplest solution for now and if we see the need for a fail fast - we can do that in the future versions.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
josvazg: I do not understand, this is a return already.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
josvazg: Hopefully this is not the kind of things that generated flaky tests hard to figure out later.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
m1kola: @josvazg while working in this area I noticed that I previously introduced wrapping in this command and decided to change it to `%v` because there is no point in wrapping here.

I can revert this change to slightly reduce the diff in the PR and submit `%w` -> `%v` replacement separately. I think there are more in other places too.

This is all new code introduced in our feature branch: not touching pre-exiting code.

josvazg: My point was that, since `%w` and wrapping is introduced, using it instead of `%v` within errors is normally a no brainer. Asked Gemini just to double check and these are the "good" reasons not to wrap it gave me:
- **API Encapsulation**: you do not want to expose internal error types to users. Might be a thing in some cases, on other might be exactly what you want to they can distinguish failure scenarios better. Does not apply here.
- **Selective Error Wrapping**: you select what errors get wrapped. If you are selective then you end up with errors that do not trace properly how they came to be other than the concatenated txt, which might or might not be enough. I guess this could be applicable here.

Anyway, it is a nitpick, so proceed as you see fit.
The real error is to use %w when that is not supported, such a plain printfs or logs, I guess I was conflating the two.

m1kola: @josvazg [here](https://go.dev/blog/go1.13-errors) is a good read about errors in general and it has some guidance on when to wrap or not [here](https://go.dev/blog/go1.13-errors#whether-to-wrap). I really like this summary at the end:

> It’s important to remember that whether you wrap or not, the error text will be the same. A person trying to understand the error will have the same information either way; the choice to wrap is about whether to give programs additional information so they can make more informed decisions, or to withhold that information to preserve an abstraction layer.

m1kola:
Tests are still red, but it is due to infra issues. Marking as ready for review. I'll ensure that tests are green before merging.

pkg/kubectl-mongodb/memberregistration/memberregistration.go
```diff
@@ -91,16 +97,52 @@ func Generate(ctx context.Context, memberClusterClient kubernetes.Interface, mem
...
 	}
 
 	return string(secretYAML) + "---\n" + string(memberClusterYAML), nil
 }
 
+// readTokenSecret Gets the token Secret once and returns the complete, user-facing error for a
+// failed read: a Get error (including NotFound) or a missing/empty token or ca.crt data key.
+// waitForTokenSecret's poll condition uses the error verbatim.
+func readTokenSecret(ctx context.Context, memberClusterClient kubernetes.Interface, namespace, name string) (token, ca []byte, err error) {
+	secret, err := memberClusterClient.CoreV1().Secrets(namespace).Get(ctx, name, metav1.GetOptions{})
```
lsierant, line 114: Could you ensure that looking at the error string in the log we'll be able to tell which exactly member cluster it's giving those errors? Quickly looking around I don't see puttng member cluster names. Also, if the name is only in the secret name that doesn't count as we'll be probably removing cluster names from the resource names.

lsierant (approved):

pkg/kubectl-mongodb/memberregistration/memberregistration.go
m1kola: @lsierant this command is per-member clusters so the context should already be clear.

lsierant: OK, just wanted to ensure we'll have a trace in the logs

josvazg (approved):
LGTM

PR #1518, opened 2026-08-19
CLOUDP-407997: Add recursive verification of signatures

josvazg:
# Summary

Add recursive signature verification after #1467 added recursive signing beyond the multi arch image.

This is a separate PR as cosign does not include a recursive verification option, so we needed to compute the extra manifests to sign one by one.

Recursive signatures means that **cosing** does not sign only the top level manifest of a multi-arch image, but also each of the per-platform images as well. This is a better UX for customers as they can verify any hash signature, not just the top index manifest. But for us it means for good test coverage we also need to verify we properly ssigned not just the top index manifest, but also the dependant images.

## Proof of Work

CI added python unit tests.

## Checklist

### DIFF
--- a/scripts/release/build/image_signing.py
+++ b/scripts/release/build/image_signing.py
@@ -1,3 +1,4 @@
+import json
 import os
 import random
 import subprocess
@@ -138,6 +139,58 @@ def get_image_digest(image_name: str) -> Optional[str]:
         logger.error(f"Failed to get digest for {image_name}: {e.stderr}")
 
 
+def get_platform_manifest_digests(image_name: str) -> List[str]:
+    """
+    Returns the per-platform manifest digests if image_name is a multi-arch manifest list/OCI index, or an
+    empty list otherwise. Buildx attestation manifests (platform "unknown/unknown") are skipped since they
+    are not signed container images.
+
+    :param image_name: The full image name with its tag or digest.
+    :return: A list of platform-specific manifest digests.
+    """
+
+    transport_protocol = "docker://"
+    raw_command = [
+        "docker",
+        "run",
+        "--rm",
+        f"--volume={os.path.expanduser('~')}/.aws:/root/.aws:ro",
+        "quay.io/skopeo/stable:latest",
+        "inspect",
+        "--raw",
+    ]
+
+    if is_ecr_registry(image_name):
+        aws_region = os.environ.get("AWS_DEFAULT_REGION", "us-east-1")
+        ecr_password = get_ecr_login_password(aws_region)
+        raw_command.append(f"--creds=AWS:{ecr_password}")
+
+    raw_command.append(f"{transport_protocol}{image_name}")
+
+    try:
+        result = run_command_with_retries(raw_command)
+    except subprocess.CalledProcessError as e:
+        logger.error(f"Failed to get raw manifest for {image_name}: {e.stderr}")
+        return []
+
+    manifest = json.loads(result.stdout)
+    manifest_list_media_types = (
+        "application/vnd.docker.distribution.manifest.list.v2+json",
+        "application/vnd.oci.image.index.v1+json",
+    )
+    if manifest.get("mediaType") not in manifest_list_media_types:
+        # Single-arch image: nothing to recurse into.
+        return []
+
+    digests = []
+    for entry in manifest.get("manifests", []):
+        platform = entry.get("platform", {})
+        if platform.get("architecture") == "unknown" or platform.get("os") == "unknown":
+            continue
+        digests.append(entry["digest"])
+    return digests
+
+
 def build_cosign_docker_command(additional_args: List[str], cosign_command: List[str]) -> List[str]:
     """
     Common logic to build a cosign command with the garasign cosign image provided by DevProd.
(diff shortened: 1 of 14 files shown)


scripts/release/build/image_signing.py
```diff
@@ -138,6 +139,58 @@ def get_image_digest(image_name: str) -> Optional[str]:
...
+        ecr_password = get_ecr_login_password(aws_region)
+        raw_command.append(f"--creds=AWS:{ecr_password}")
+
+    raw_command.append(f"{transport_protocol}{image_name}")
+
+    try:
+        result = run_command_with_retries(raw_command)
+    except subprocess.CalledProcessError as e:
+        logger.error(f"Failed to get raw manifest for {image_name}: {e.stderr}")
+        return []
```
m1kola: Here (and below) I think if skopeo errors we return and empty list. Would that meant that verification would happily report "all good" despite not being able to actually verify?

scripts/release/build/image_signing.py
```diff
@@ -138,6 +139,58 @@ def get_image_digest(image_name: str) -> Optional[str]:
         logger.error(f"Failed to get digest for {image_name}: {e.stderr}")
 
 
+def get_platform_manifest_digests(image_name: str) -> List[str]:
```
m1kola: Do we still need `get_image_digest`? It apparently does the same thing, but only for a single image?

scripts/release/build/image_signing.py
```diff
@@ -246,15 +299,46 @@ def verify_signature(repository: str, tag: str):
...
+        try:
+            verify_ref(platform_ref)
+        except subprocess.CalledProcessError as e:
+            stderr = e.stderr or ""
+            if "no signatures found" in stderr:
+                logger.warning(
+                    f"No recursive signature found for platform image {platform_ref}. "
+                    "This is expected for images signed before recursive signing was introduced."
+                )
+                continue
```
m1kola: 1. A bit odd to depend on the error wording from cosign.
2. Is it a concern? I know [I asked it before](https://github.com/mongodb/mongodb-kubernetes/pull/1467#discussion_r3811131856) and you said that we might point it to old releases. Do we need to do that? What would be the use case? I'm just wondering if we can get rid of this exception handling.

scripts/release/build/image_signing.py
```diff
@@ -138,6 +139,58 @@ def get_image_digest(image_name: str) -> Optional[str]:
         logger.error(f"Failed to get digest for {image_name}: {e.stderr}")
 
 
+def get_platform_manifest_digests(image_name: str) -> List[str]:
+    """
+    Returns the per-platform manifest digests if image_name is a multi-arch manifest list/OCI index, or an
+    empty list otherwise. Buildx attestation manifests (platform "unknown/unknown") are skipped since they
```
m1kola: Nit from my agent:

> It says `buildx` attestation manifests "are not signed container images", but cosign `sign --recursive` walks every child manifest in the index with no platform filter, so attestations are signed — and #1467's `CopySignatures` copies their signatures too.

scripts/release/build/image_signing.py
josvazg: can error out instead

scripts/release/build/image_signing.py
josvazg: will double check

scripts/release/build/image_signing.py
josvazg: will check

scripts/release/build/image_signing.py
josvazg: will check

scripts/release/build/image_signing.py
```diff
@@ -138,6 +139,59 @@ def get_image_digest(image_name: str) -> Optional[str]:
         logger.error(f"Failed to get digest for {image_name}: {e.stderr}")
 
 
+def get_platform_manifest_digests(image_name: str) -> List[str]:
```
viveksinghggits: nit:
```suggestion
def per_platform_manifests_digests(image_name: str) -> List[str]:
```
I think this conveys the meaning better, what do you think about this?

scripts/release/build/image_signing.py
```diff
@@ -246,15 +300,46 @@ def verify_signature(repository: str, tag: str):
...
+    # have a valid signature.
+    try:
+        verify_ref(image)
     except subprocess.CalledProcessError as e:
         # Fail the pipeline if verification fails
         raise Exception(f"Failed to verify signature for image {image}")
 
+    # cosign verify has no --recursive flag. Since sign_image signs the top-level
+    # multi-arch image index AND each per-platform image individually (via
+    # --recursive), also verify every platform digest here. A *missing* signature
```
viveksinghggits: The word `--recursive` is a little out of the place. We are just talking at two levels index digest and per platform image digest.

viveksinghggits: or do these tool actually use this term to denote index digest and individual image digest?

viveksinghggits (commented):
Looks pretty good. Left some nits.

scripts/release/build/image_signing.py
josvazg: The `cosing` tool uses that flag name specifically, see:
https://github.com/mongodb/mongodb-kubernetes/blob/2ac66765074e04c40dd33520ec2e7c9202e37e95/scripts/release/build/image_signing.py#L198-L207

scripts/release/build/image_signing.py
josvazg: While `get_image_digest` is used to get the hash of the given tag to be signed when signing, `get_platform_manifest_digests` is called when verifying the signatures to get the list of manifests to be verified. They are complementary uses, not overlapping

scripts/release/build/image_signing.py
josvazg: will avoid the error string match by checking in advance if there is a signature to verify.

scripts/release/build/image_signing.py
josvazg: Ok, we will actually list and therefore verify all dependant manifests

scripts/release/build/image_signing.py
josvazg: In the end I renamed to `get_manifest_digests` because due to @m1kola earlier comment, we dropped all filtering so we actually list all digests no matter the type.

josvazg:
@m1kola I finally came around to make the recursive signature support fastidious/strict by default.
For the Python code, given that is only used by automation, the strict behavior is fixed.
For the `mckci` cloning of signatures, the strict behavior is the default, but a flag allows manual usages to turn missing sub-signatures into warnings.

This seems to fit best, so the automated behavior is strict and screams loudly on anything unexpected. But the manual CLI allows a relaxed check on demand.

viveksinghggits (approved):

m1kola (approved):

PR #1521, opened 2026-08-19
KUBE-335: Support providing custom CA for member clusters in CLI

mircea-cosbuc:
# Summary

This patch adds support for passing a custom CA for member clusters when using the `kubectl mongodb` plugin to handle cases where the operator should use a different CA when connecting to member clusters than the originator of the `kubectl mongodb` caller.

## Proof of Work

New unit tests and e2e test.

## Checklist

---


config/rbac/database-roles.yaml
```diff
@@ -5,20 +5,26 @@ kind: ServiceAccount
 metadata:
   name: mongodb-kubernetes-appdb
   namespace: mongodb
+
```
m1kola: I think linter is not happy with these extra newlines (here and in other files).

cmd/kubectl-mongodb/multicluster/recover/recover.go
```diff
@@ -26,14 +26,15 @@ func init() {
 	RecoverCmd.Flags().StringVar(&RecoverFlags.SourceCluster, "source-cluster", "", "The source cluster for recovery. This has to be one of the healthy member cluster that is the source of truth for new cluster configuration. [required]")
 	RecoverCmd.Flags().BoolVar(&RecoverFlags.CreateServiceAccountSecrets, "create-service-account-secrets", true, "Create service account token secrets. [optional default: true]")
 	RecoverCmd.Flags().StringVar(&common.MemberClustersApiServers, "member-clusters-api-servers", "", "Comma separated list of api servers addresses. [optional, default will take addresses from KUBECONFIG env var]")
+	RecoverCmd.Flags().StringArrayVar(&common.MemberClusterCAFiles, "member-cluster-ca", nil, "Custom CA certificate for a member cluster, in the form <member-cluster-name>=<path-to-pem-file>, repeatable once per cluster. [optional, default will take the CA from that cluster's ServiceAccount token secret]")
```
m1kola, line 29: 1. Does this mean that we re-use CAs from existing secret when recovering? If so, "Any --member-cluster-ca values passed to `setup` must be passed here too." below seems to contradict this.

    It gives a (wrong?) idea that you have to manually pass CAs provided to `setup` command again.

2. If we do not re-use CAs, is there a way to ensure that if `--member-cluster-ca` is committed, but we previously had custom CAs we do not let the user run the command without `--member-cluster-ca`?

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_cli_custom_ca.py
```diff
@@ -0,0 +1,201 @@
...
+):
+    # the Operator builds its member cluster clients at startup, so the override lands only on restart
+    multi_cluster_operator.restart_operator_deployment()
+
+    # deleting the StatefulSet forces a read and a write through the client that trusts the custom CA
+    member_cluster_client = next(c for c in member_cluster_clients if c.cluster_name == custom_ca_cluster)
+    sts_name = f"{mongodb_multi.name}-{member_cluster_names.index(custom_ca_cluster)}"
+    old_uid = get_statefulset(namespace, sts_name, api_client=member_cluster_client.api_client).metadata.uid
+
+    delete_statefulset(namespace=namespace, name=sts_name, api_client=member_cluster_client.api_client)
```
m1kola: q: so this is us just forcing reconciliation of a multi-cluster CR by deleting a stateful set the member cluster?

m1kola (commented):
Still need to review some changes. Posting comments I've got so far.

config/rbac/database-roles.yaml
m1kola: Looks like it just addressed. Resolving.

public/mongodb-kubernetes.yaml
```diff
@@ -312,7 +312,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: mongodb-kubernetes-operator
-  namespace: mongodb
+  namespace: mongodb
```
m1kola: Another space at the end for some reason.

cmd/kubectl-mongodb/multicluster/recover/recover.go
mircea-cosbuc: > Does this mean that we re-use CAs from existing secret when recovering? 

No, it means we use the CA from the cluster's SA token secret so the same overrides need to be passed for the recover command.

mircea-cosbuc: > If we do not re-use CAs, is there a way to ensure that if --member-cluster-ca is committed, but we previously had custom CAs we do not let the user run the command without --member-cluster-ca?

I think it's a good idea to warn the user that the CA in the current kubeconfig is different than what's in the SA token secret, which could be because the custom CA was passed or because the k8s root CA changed. I wouldn't stop the user from running the command without `--member-cluster-ca`. WDYT?

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_cli_custom_ca.py
mircea-cosbuc: Yes. I'm debating whether this e2e test is worth keeping since it's basically validating that passing a custom CA still ends up with a valid CA in the kubeconfig.

public/mongodb-kubernetes.yaml
mircea-cosbuc: Fixed the linting test, somehow running pre-commit in my environment added extra spaces.

pkg/kubectl-mongodb/common/kubeconfig.go
```diff
@@ -39,6 +42,46 @@ func GetMemberClusterApiServerUrls(kubeconfig *clientcmdapi.Config, clusterNames
...
+			return nil, xerrors.Errorf("invalid member-cluster-ca value '%s', expected format <member-cluster-name>=<path-to-pem-file>", entry)
+		}
+
+		if !slices.Contains(memberClusters, clusterName) {
+			return nil, xerrors.Errorf("member-cluster-ca refers to cluster '%s' which is not one of the member clusters %v", clusterName, memberClusters)
+		}
+
+		if _, ok := cas[clusterName]; ok {
+			return nil, xerrors.Errorf("member-cluster-ca specified more than once for cluster '%s'", clusterName)
+		}
```
m1kola, line 99: I think we are mixing two things in this function: flag parsing and CA validation and it is a bit confusing. It also makes unit tests difficult to follow because there are so many test cases.

Could you split this please?

docker/mongodb-kubernetes-tests/tests/multicluster/multi_cluster_cli_custom_ca.py
m1kola: Yes, I think this could be just a unit test or an integration test.

cmd/kubectl-mongodb/multicluster/recover/recover.go
m1kola: Makes sense!

cmd/kubectl-mongodb/multicluster/setup/setup.go
```diff
@@ -30,6 +30,7 @@ func init() {
 	SetupCmd.Flags().BoolVar(&setupFlags.CreateServiceAccountSecrets, "create-service-account-secrets", true, "Create service account token secrets. [optional default: true]")
 	SetupCmd.Flags().StringVar(&setupFlags.ImagePullSecrets, "image-pull-secrets", "", "Name of the secret for imagePullSecrets to set in created service accounts")
 	SetupCmd.Flags().StringVar(&common.MemberClustersApiServers, "member-clusters-api-servers", "", "Comma separated list of api servers addresses. [optional, default will take addresses from KUBECONFIG env var]")
+	SetupCmd.Flags().StringArrayVar(&common.MemberClusterCAFiles, "member-cluster-ca", nil, "Custom CA certificate for a member cluster, in the form <member-cluster-name>=<path-to-pem-file>, repeatable once per cluster. [optional, default will take the CA from that cluster's ServiceAccount token secret]")
```
m1kola, line 33: Could you please create a new ticket under https://jira.mongodb.org/browse/CLOUDP-260547 epic so we can port it to new CLI commands in for MCK 2.x?

cmd/kubectl-mongodb/multicluster/recover/recover.go
m1kola: @mircea-cosbuc after looking at `replacedCustomCAWarnings` I'm wondering if we are actually doing more harm (in form of confusion) then good here.

I just realised that when you are recovering a cluster, it is very likely that you are spinning up a new cluster, right? So it is very likely that CA won't match and we will confuse/alarm the user with the warning.

m1kola (approved):
Looks good overall. Just one minor thing around custom CA warning.

helderjs (approved):

cmd/kubectl-mongodb/multicluster/setup/setup.go
mircea-cosbuc: Created https://jira.mongodb.org/browse/KUBE-339 .

cmd/kubectl-mongodb/multicluster/recover/recover.go
mircea-cosbuc: @m1kola For new clusters, their CA would not already be present in the Kubeconfig, so no warning would propagate. Is this the case you were concerned about?

PR #1525, opened 2026-08-21
KUBE-329: Reject externalDomain on operator-managed MongoDB + Search (single-cluster)

anandsyncs:
# Summary

`spec.externalAccess.externalDomain` makes the operator advertise external hostnames as replica-set member identities, while mongot's internal-source path always dials internal service DNS — so MongoDB Search on an externalDomain-configured, operator-managed MongoDB never worked and failed only at runtime. This PR rejects the combination at reconcile time: a shared `validateSearchSourceExternalDomain` check, called from `EnterpriseResourceSearchSource.Validate()` and `ShardedInternalSearchSource.Validate()`, moves the `MongoDBSearch` to `Failed` with:

> MongoDB Search is not supported on operator-managed MongoDB resources with spec.externalAccess.externalDomain set

**Scope:** the check applies regardless of topology — multi-cluster internal mongod sources are not supported anyway, and multi-cluster search requires an external source. `MongoDBCommunity` has no `externalAccess`/`externalDomain` field, so despite the ticket's "Enterprise or Community" wording there is no community-side check to add.

**Intended side effect:** the same `Validate()` calls gate mongot parameter injection in the replica-set and sharded-cluster controllers, so the rejected combination also stops receiving mongot config (`mongotHost` setParameter) there — consistent, since a rejected Search should not wire mongot. Verified at both call sites that a `Validate()` failure only skips search wiring and never fails the MongoDB reconcile. This side effect is described here only; the changelog entry covers the user-facing rejection.

## Proof of Work

- Unit tests assert against the shared error value: `TestEnterpriseResourceSearchSource_Validate` gains a single-cluster + externalDomain rejection row; new `TestShardedInternalSearchSource_Validate` covers the rejection and the passing baseline.
- `go build ./...`, `go test ./controllers/searchcontroller/... ./api/... ./controllers/operator/...`, and `golangci-lint run ./controllers/searchcontroller/...` clean locally.
- e2e: not applicable — reconcile-time validation only.

## Checklist

### DIFF
--- a/controllers/searchcontroller/enterprise_search_source_test.go
+++ b/controllers/searchcontroller/enterprise_search_source_test.go
@@ -61,6 +61,7 @@ func TestEnterpriseResourceSearchSource_Validate(t *testing.T) {
 		resourceType        mdbv1.ResourceType
 		authModes           []string
 		internalClusterAuth string
+		externalDomain      *string
 		expectError         bool
 		expectedErrMsg      string
 	}{
@@ -160,6 +161,17 @@ func TestEnterpriseResourceSearchSource_Validate(t *testing.T) {
 			resourceType: mdbv1.ReplicaSet,
 			authModes:    []string{},
 			expectError:  false,
+		},
+		// External domain validation tests
+		{
+			name:           "SingleCluster with externalDomain",
+			version:        "8.2.0",
+			topology:       mdbv1.ClusterTopologySingleCluster,
+			resourceType:   mdbv1.ReplicaSet,
+			authModes:      []string{},
+			externalDomain: ptr.To("example.com"),
+			expectError:    true,
+			expectedErrMsg: errExternalDomainNotSupported.Error(),
 		},
 		// Authentication mode tests
 		{
@@ -280,6 +292,9 @@ func TestEnterpriseResourceSearchSource_Validate(t *testing.T) {
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
 			src := newEnterpriseSearchSource(c.version, c.topology, c.resourceType, c.authModes, c.internalClusterAuth)
+			if c.externalDomain != nil {
+				src.Spec.ExternalAccessConfiguration = &mdbv1.ExternalAccessConfiguration{ExternalDomain: c.externalDomain}
+			}
 			err := src.Validate()
 
 			if c.expectError {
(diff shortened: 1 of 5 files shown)


controllers/searchcontroller/search_construction.go
```diff
@@ -90,6 +91,20 @@ type SearchSourceShardedDeployment interface {
...
 
+var errExternalDomainNotSupported = xerrors.New("spec.externalAccess.externalDomain is not yet supported together with MongoDB Search: remove externalDomain from the MongoDB resource or delete the MongoDBSearch resource")
+
+// validateSearchSourceExternalDomain rejects single-cluster sources with externalDomain set:
+// externalDomain makes the operator advertise external hostnames as replica set member
+// identities while mongot's host seeds use the internal service DNS, which breaks mongot's
+// server discovery. Multi-cluster is deliberately exempt — multi-cluster search requires an
+// external source and is validated elsewhere.
+func validateSearchSourceExternalDomain(spec *mdb.MongoDbSpec) error {
+	if !spec.IsMultiCluster() && spec.GetExternalDomain() != nil {
```
anandsyncs: is the multicluster condition neccessary here?

controllers/searchcontroller/search_construction.go
```diff
@@ -90,6 +91,20 @@ type SearchSourceShardedDeployment interface {
 	MongosHostsAndPorts() []string
 }
 
+var errExternalDomainNotSupported = xerrors.New("spec.externalAccess.externalDomain is not yet supported together with MongoDB Search: remove externalDomain from the MongoDB resource or delete the MongoDBSearch resource")
+
+// validateSearchSourceExternalDomain rejects single-cluster sources with externalDomain set:
```
anandsyncs: make this comment concise

controllers/searchcontroller/enterprise_search_source_test.go
```diff
@@ -371,3 +386,48 @@ func TestShardedEnterpriseSearchSource_GetShardNames(t *testing.T) {
...
+			expectError: false,
+		},
+		{
+			name:           "SingleCluster with externalDomain",
+			topology:       mdbv1.ClusterTopologySingleCluster,
+			externalDomain: ptr.To("example.com"),
+			expectError:    true,
+		},
+		{
+			name:           "MultiCluster with externalDomain",
```
anandsyncs: multicluster intenal mongod is not supported yet

controllers/searchcontroller/search_construction.go
```diff
@@ -90,6 +91,20 @@ type SearchSourceShardedDeployment interface {
 	MongosHostsAndPorts() []string
 }
 
+var errExternalDomainNotSupported = xerrors.New("spec.externalAccess.externalDomain is not yet supported together with MongoDB Search: remove externalDomain from the MongoDB resource or delete the MongoDBSearch resource")
```
anandsyncs: the message is too long. just clearly state that this is not supported for operator managed mongodb resource. make it concise

controllers/searchcontroller/search_construction.go
anandsyncs: Removed the multicluster condition — the check now applies regardless of topology, since multi-cluster internal mongod sources are not supported anyway.

controllers/searchcontroller/search_construction.go
anandsyncs: Shortened the doc comment to two lines.

controllers/searchcontroller/search_construction.go
anandsyncs: Replaced with a concise message: "MongoDB Search is not supported on operator-managed MongoDB resources with spec.externalAccess.externalDomain set".

controllers/searchcontroller/enterprise_search_source_test.go
anandsyncs: Updated — with the multicluster condition removed, this case now expects the rejection as well.

lucian-tosa (approved):

Julien-Ben (approved):
This LGTM, but I wonder if multi cluster cases can have the same issue ? If they consider external access per-cluster instead of at top level.

PR #1534, opened 2026-08-27
CLOUDP-429011 - Block AppDB config changes during forward migration

MaciejKaras:
# Summary

During forward migration the AppDB pods still run headless agents that read the mounted automation config and never poll Ops Manager, so for cluster-visible settings (TLS, CA, members) no publish ordering can apply a change safely — rotated members become incompatible with unrotated ones and the rolling update wedges (see [this discussion](https://github.com/mongodb/mongodb-kubernetes/pull/1440#discussion_r3862882368)). This PR adds runtime validations in the replicaset controller that reject a `role: AppDB` CR spec diverging from the running AppDB while the handover is in progress: TLS must stay enabled, the CA ConfigMap must stay identical, and the member count must match the StatefulSet. The migration window is detected from the adopted StatefulSet itself (pod template still in AppDB form, or the migration-ready annotation present), so it self-closes once the controller replaces the StatefulSet spec. Violations are aggregated into a single validation failure (`Failed` phase, no requeue) telling the user to complete the migration with a matching spec first.

## Proof of Work

- `TestValidateAppDBForwardMigration` — table-driven coverage of every rule (TLS disable, CA change, scale up/down, aggregation, matching spec, closed window)
- Existing `TestAdoptionGate_*` tests pass through the new validation path
- e2e: pending — validation-only change, negative paths covered by unit tests; `e2e_om_external_appdb_*` to be run via Evergreen patch

## Checklist

### DIFF
--- a/controllers/operator/mongodbreplicaset_controller.go
+++ b/controllers/operator/mongodbreplicaset_controller.go
@@ -838,13 +838,28 @@ func (r *ReplicaSetReconcilerHelper) ensureAppDBStatefulSetOwnership(ctx context
 
 const appDBForwardMigrationViolationFmt = "cannot change AppDB configuration during forward migration: %s"
 
-// validateAppDBForwardMigration rejects spec changes that cannot be applied while the AppDB
-// StatefulSet handover (forward migration) is in progress. The window is open until this
-// controller replaces the StatefulSet spec with the enterprise form: until then the pods run
-// headless agents that read the mounted automation config and never poll Ops Manager, so for
-// cluster-visible settings (TLS, CA, members) no publish ordering works - rotated members become
-// incompatible with the unrotated ones. The handover must be config-identical; changes apply
-// once the migration completes.
+// validateAppDBForwardMigration rejects spec changes that are unsafe to apply during
+// AppDB forward migration.
+//
+// There is a window of time during which some spec changes are not safe to reconcile.
+// The window runs from the start of forward migration until AppDBMigrationReadyAnnotation
+// becomes true, at which point this controller replaces the StatefulSet spec with the
+// enterprise form.
+//
+// Changes are unsafe during this window because, for some configurations, they could
+// create a situation where two mutually incompatible populations of pods are running
+// during the rolling update. Transitioning cluster-visible settings (TLS, CA, members)
+// from state A to state Z normally goes through intermediate states (A -> N1 -> N2 ->
+// ... -> Z); a population spanning two neighbouring states (e.g. A and N1) is fine, but
+// a population made up of both A and Z pods is not.
+//
+// In normal operation we handle this in the operator (see e.g. publishAutomationConfigFirst),
+// which has enough visibility to sequence the transition safely. But before
+// AppDBMigrationReadyAnnotation is true, the AppDB pods run headless agents that read
+// the automation config from a locally mounted file and never poll Ops Manager. Since
+// they never poll, the operator has no way to roll a cluster-visible config change out
+// pod-by-pod without creating an incompatible mixed population mid-rollout. The handover
+// must therefore be config-identical; changes apply once the migration completes.
 func (r *ReplicaSetReconcilerHelper) validateAppDBForwardMigration(ctx context.Context, mdb *mdbv1.MongoDB, sts appsv1.StatefulSet) workflow.Status {
 	if sts.Annotations[util.AppDBMigrationReadyAnnotation] != trueString {
 		return workflow.OK()
--- a/controllers/operator/mongodbreplicaset_controller.go
+++ b/controllers/operator/mongodbreplicaset_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"slices"
+	"strings"
 
 	"github.com/hashicorp/go-multierror"
 	"go.uber.org/zap"
@@ -53,8 +54,10 @@ import (
 	"github.com/mongodb/mongodb-kubernetes/pkg/kube"
 	"github.com/mongodb/mongodb-kubernetes/pkg/kube/annotations"
 	"github.com/mongodb/mongodb-kubernetes/pkg/kube/configmap"
+	"github.com/mongodb/mongodb-kubernetes/pkg/kube/container"
 	"github.com/mongodb/mongodb-kubernetes/pkg/kube/secret"
 	"github.com/mongodb/mongodb-kubernetes/pkg/statefulset"
+	"github.com/mongodb/mongodb-kubernetes/pkg/tls"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/architectures"
 	"github.com/mongodb/mongodb-kubernetes/pkg/util/constants"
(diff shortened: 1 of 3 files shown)


lucian-tosa (approved):

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -831,6 +836,44 @@ func (r *ReplicaSetReconcilerHelper) ensureAppDBStatefulSetOwnership(ctx context
 	return workflow.OK()
 }
 
+const appDBForwardMigrationViolationFmt = "cannot change AppDB configuration during forward migration: %s"
+
+// validateAppDBForwardMigration rejects spec changes that cannot be applied while the AppDB
+// StatefulSet handover (forward migration) is in progress. The window is open until this
```
christopherthompsonwalsh-mongo: I think the comment from "The window is open..." onwards could be made clearer --- phrases like "the StatefulSet spec" and "the enterprise form" are inherently allusive: they're not explicitly identifying the thing they name, but rather alluding to ambient knowledge (like, eg, "what the 'enterprise' form of a StatefulSet is"). Similarly "The handover must be config-identical; changes apply once the migration completes" is unclear, and actually splices two quite independent observations together using a ";".

I *think* what we're trying to say is something like:

```
There is a window of time during which some intent states are not safe to reconcile. The window 
runs from the  [the beginning of a forward migration? the creation of a stateful set not managed
 via MCL CRDs?] until the point when AppDBMigrationReadyAnnotation becomes true.

Changes in this period are unsafe for some configurations because they could create a 
situation where two, mutually incompatible populations of pods were running during the 
rolling updates of pods. This typically arises because to transit from state A --> Z, there 
are a series of intermediate states that we must pass through (A --> N1 --> N2 --> ... --> Z);
any population of pods that spans two neighbouring states (eg, A, N1, or N1, N2) is ok, but
a population made up of As and Zs would not be ok.

In normal operation, we handle this in our operator (see eg. publishAutomationConfigFirst), 
which has sufficient visibility to understand what transitions are safe. But before
AppDBMigrationReadyAnnotation is true, pods continue to use the OM-originated 
automation-config secret, so our Operator does not have sufficient visibility into actual 
Pod state to proceed safely.
```

I hope you'll agree this is clearer?

controllers/operator/mongodbreplicaset_controller.go
```diff
@@ -831,6 +836,44 @@ func (r *ReplicaSetReconcilerHelper) ensureAppDBStatefulSetOwnership(ctx context
...
+// controller replaces the StatefulSet spec with the enterprise form: until then the pods run
+// headless agents that read the mounted automation config and never poll Ops Manager, so for
+// cluster-visible settings (TLS, CA, members) no publish ordering works - rotated members become
+// incompatible with the unrotated ones. The handover must be config-identical; changes apply
+// once the migration completes.
+func (r *ReplicaSetReconcilerHelper) validateAppDBForwardMigration(ctx context.Context, mdb *mdbv1.MongoDB, sts appsv1.StatefulSet) workflow.Status {
```
christopherthompsonwalsh-mongo, line 863: This method attempts to prove a universal negative ("there **are no changes** in the config that are unsafe to try to apply [due to the during-migration issues discussed above]")

At the moment, it does so by encoding some wisdom (from Fabio IIRC?) provided on a previous commit.

But this is inherently dangerous: we're trying to prove that there is **no such** problem P, by checking for **two-specific** sub-types, P1 and P2.

This is fine for now (if P1 and P2 are the only ways P can happen).

But it gives a hostage to fortune, as another P3 might come along later, elsewhere in code, and this code might not be updated.

Would it be safer to simply reject **all** changes until `AppDBMigrationReadyAnnotation` is true?

Or is that a dumb newbie question?

christopherthompsonwalsh-mongo (changes requested):
One comment request, one strategy question.

controllers/operator/mongodbreplicaset_controller.go
christopherthompsonwalsh-mongo: (It could obviously also be _shorter_. But what it should not be is _shorter but less clear_.)

controllers/operator/mongodbreplicaset_controller.go
igor-karpukhin: Hi @christopherthompsonwalsh-mongo. Good point and not a dumb question at all. Yes, this is inherently an allow list of known unsafe sub cases, rather than a proof that nothing else is dangerous. It is whitelisting failure modes we have already identified. But the reason it is built this way is the migration window can be long in practice, and users may need to make unrelated but safe changes like updating labels, etc. without being blocked. The "reject-all" pattern would be simpler but more disruptive. @MaciejKaras is on PTO and I'd vote for not to block this merge on redesigning now. I'll leave a TODO here and file a follow-up to consider either a "deny by default" with an explicit allow-list or the "compare the full spec/hash" against the statefulset rather than enumerating changes

controllers/operator/mongodbreplicaset_controller.go
igor-karpukhin: Good point. I updated the doc-string.

igor-karpukhin:
@christopherthompsonwalsh-mongo could you please take a look one more time? I addressed your comments. We'd like to merge this stack on Monday

controllers/operator/mongodbreplicaset_controller.go
MaciejKaras: I think this is very good observation and we already retrieve the External AppDB object (MongoDB) in the ReconcileExternalAppDBReplicaSet:validateExternalAppDBReference(), but we only check if the MongoDB has the correct role: AppDB. I think the migration validations could be called in both call sites and share the same logic. I think it's valuable for users to notify them about the issues with the MongoDB CR object in it's status, not only in the MongoDBOpsManager CR status.

I would go with the @igor-karpukhin recommendation and merge this as is, but create a followup Jira ticket that will add missing validations on the ReconcileExternalAppDBReplicaSet side. This can be handled shortly in the Phase 2 of this project, so we can land fix very soon.

christopherthompsonwalsh-mongo: SG.

I'd strongly recommend, for now, as a defence-in-depth, a comment on `ReconcileAppDB` (or maybe publishAutomationConfigFirst?) warning people to keep this rule here updated.

Something like "Warning: 1) when possible care to should be taken **not** to assume that AutomationConfig actually gets pushed before other actions take place 2) If you need to ensure that AutomationConfig is pushed first to achieve some reconciliation, *then* the relevant config change should probably be added to `validateAppDBForwardMigration`, since during an AppDB migration we **cannot** actually ensure AutomationConfig is pushed first."

(Or maybe you can see a better warning. But the point is to try to prevent "P3" above.)

christopherthompsonwalsh-mongo (approved):
One more recommendation but, subject to that being done, this lgtm.

controllers/operator/mongodbreplicaset_controller.go
christopherthompsonwalsh-mongo: Ta.

controllers/operator/mongodbreplicaset_controller.go
igor-karpukhin: Thanks @MaciejKaras, @christopherthompsonwalsh-mongo. I updated the comment on `ReconcileAppDB` method to include a warning and created a follow up ticket attached to the same Epic

PR #1540, opened 2026-08-27
KUBE-339,CLOUDP-400899: Add --member-cluster-api-server-ca to generate-member-registration

m1kola:
# Summary

MCK 1.x (#1521) added `--member-cluster-ca` to the legacy `multicluster setup`/`recover` commands for cases where the operator must verify a member cluster's API server against a different CA than the caller's environment — e.g. a TLS-terminating proxy in front of the API server. The new `multicluster generate-member-registration` command always embeds the member cluster's ServiceAccount token Secret `ca.crt` into the emitted credential kubeconfig. This PR adds the optional `--member-cluster-api-server-ca` flag (path to a PEM file, validated before any cluster contact): when set, its bundle replaces the token Secret's CA in the emitted kubeconfig; when unset, behaviour is unchanged. The 1.x flag's warn-before-clobber machinery is not ported: the new command is render-only and never touches the operator cluster.

## Proof of Work

- unit: `go test ./pkg/kubectl-mongodb/... ./cmd/kubectl-mongodb/...` — new `TestGenerate_KubeconfigContents_CertificateAuthorityOverride` (override path; existing test covers the default) and new cmd-level `TestLoadMemberClusterApiServerCA` (PEM validation table: valid cert/bundle, private key rejected, non-PEM rejected, missing file)
- `make precommit` clean
- e2e: N/A (no runtime change — kubectl plugin tooling only)

## Checklist

---

### DIFF
--- a/cmd/kubectl-mongodb/multicluster/generatememberregistration/generate_member_registration.go
+++ b/cmd/kubectl-mongodb/multicluster/generatememberregistration/generate_member_registration.go
@@ -1,6 +1,8 @@
 package generatememberregistration
 
 import (
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"net/url"
 	"os"
@@ -21,6 +23,7 @@ var flags struct {
 	operatorNamespace        string
 	memberClusterLogicalName string
 	memberClusterApiServer   string
+	memberClusterApiServerCA string
 }
 
 func init() {
@@ -30,6 +33,7 @@ func init() {
 	GenerateMemberRegistrationCmd.Flags().StringVar(&flags.operatorNamespace, "operator-namespace", "", "Namespace on the operator's cluster where the MemberCluster CR and credential Secret will be created. Must match the operator's installation namespace. [required]")
 	GenerateMemberRegistrationCmd.Flags().StringVar(&flags.memberClusterLogicalName, "member-cluster-logical-name", "", "Name that workloads use to reference this member cluster. Only needed when that name is not RFC 1123 compliant (e.g. it contains underscores). [optional, default: --member-cluster]")
 	GenerateMemberRegistrationCmd.Flags().StringVar(&flags.memberClusterApiServer, "member-cluster-api-server", "", "API server address of the member cluster; must be reachable from the operator Pod. [optional, default: the server address from --member-cluster-context]")
+	GenerateMemberRegistrationCmd.Flags().StringVar(&flags.memberClusterApiServerCA, "member-cluster-api-server-ca", "", "Path to a PEM file with the CA certificate the operator should use to verify the member cluster's API server. [optional, default: the CA from the member cluster's ServiceAccount token Secret]")
 }
 
 // GenerateMemberRegistrationCmd reads a member cluster's ServiceAccount token and emits the
@@ -43,6 +47,7 @@ single-context kubeconfig) and a MemberCluster CR as multi-document YAML to stdo
 
 If the token Secret is not populated yet, the command waits up to a minute for Kubernetes to provision it.
 By default the credential kubeconfig uses the API server address from --member-cluster-context; pass --member-cluster-api-server to override it.
+By default the credential kubeconfig uses the CA from the member cluster's ServiceAccount token Secret; pass --member-cluster-api-server-ca to override it.
 
 Apply the output to the operator's cluster with kubectl, or commit it to Git for GitOps.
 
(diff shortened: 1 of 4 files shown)


mircea-cosbuc (approved):

cmd/kubectl-mongodb/multicluster/generatememberregistration/generate_member_registration.go
```diff
@@ -30,6 +33,7 @@ func init() {
 	GenerateMemberRegistrationCmd.Flags().StringVar(&flags.operatorNamespace, "operator-namespace", "", "Namespace on the operator's cluster where the MemberCluster CR and credential Secret will be created. Must match the operator's installation namespace. [required]")
 	GenerateMemberRegistrationCmd.Flags().StringVar(&flags.memberClusterLogicalName, "member-cluster-logical-name", "", "Name that workloads use to reference this member cluster. Only needed when that name is not RFC 1123 compliant (e.g. it contains underscores). [optional, default: --member-cluster]")
 	GenerateMemberRegistrationCmd.Flags().StringVar(&flags.memberClusterApiServer, "member-cluster-api-server", "", "API server address of the member cluster; must be reachable from the operator Pod. [optional, default: the server address from --member-cluster-context]")
+	GenerateMemberRegistrationCmd.Flags().StringVar(&flags.memberClusterApiServerCA, "member-cluster-api-server-ca", "", "Path to a PEM file with the CA certificate the operator should use to verify the member cluster's API server. [optional, default: the CA from the member cluster's ServiceAccount token Secret]")
```
lsierant, line 36: I'm not sure why do we need this. Shouldn't custom CA be provided in the kubeconfig itself?

m1kola: @lsierant this is a port of a feature that was recently added in MCK 1.x. You can find more details in [KUBE-335](https://jira.mongodb.org/browse/KUBE-335).

TLDR: operator when running in cluster might have to take a different network path to reach a member cluster and certs might be different from what we've gathered from inside the member cluster.

lsierant: Got it! Thanks for explanation!

lsierant (approved):

cmd/kubectl-mongodb/multicluster/generatememberregistration/generate_member_registration.go
```diff
@@ -101,12 +106,52 @@ func parseFlags() (memberregistration.Options, error) {
...
+		return nil, xerrors.Errorf("invalid --member-cluster-api-server-ca: failed reading CA file %q: %w", path, err)
+	}
+	if err := validateCAPEM(caPEM); err != nil {
+		return nil, xerrors.Errorf("invalid --member-cluster-api-server-ca: %q: %w", path, err)
+	}
+	return caPEM, nil
+}
+
+// validateCAPEM rejects a bundle the Operator could not use, or should never be handed: it builds
+// its cert pool from this value alone, and the bundle is copied verbatim into a Secret.
```
christopherthompsonwalsh-mongo, line 143: I think we can improve this comment.

1) "could not use" isn't right, it's anticipating future codepath... here, we are *establishing the rule* that it **should** not be used. 

2) The comment over "or should never be handed ... Secret." is unhelpful and should be deleted.

- "should never be handed" <-- morally? religiously? for reasons of opsec? This is never explained, and kinda redundant: making something I _shouldn't_ use into something I _can't_ use is the point of validation.

- "the bundle is copied verbatim into a Secret" <-- not by this function it isn't. Comment about wider behaviour belongs elsewhere, or at the very very least not in the first para of a comment.  
- " , and the bundle is copied verbatim into a Secret" is a comma splice, and its relevance is hard to discern without reading the rest of the code.

I suggest: 
```
// validateCAPEM rejects malformed bundles, including ones that erroneously include private key material.
//
// To be considered well-formed, caPEM must be PEM-encoded sequence of blocks a) satisfying the
// the requirements of the x509 CertPool.AppendCertsFromPEM method (ie., containing at least one
// "CERTIFICATE" block), and b) containing no blocks labelled "PRIVATE KEY".
```

If we want to gesture at _why_ we're bothering to protect the user from themselves, I suggest a third informative para in the comment (making it clear this is "colouring of the picture" not strictly part of the function spec), and perhaps we could rename the function something like `validateSafeToDistributeCAPEM`. Or `validatePEMFormednessAndSensitivity` or something.

cmd/kubectl-mongodb/multicluster/generatememberregistration/generate_member_registration.go
```diff
@@ -101,12 +106,52 @@ func parseFlags() (memberregistration.Options, error) {
...
 		MemberClusterApiServer:   memberClusterApiServer,
+		MemberClusterApiServerCA: memberClusterApiServerCA,
 		TokenWaitTimeout:         memberregistration.DefaultTokenWaitTimeout,
 	}, nil
 }
+
+// loadMemberClusterApiServerCA reads and validates the PEM bundle named by
+// --member-cluster-api-server-ca; a blank flag value means no override.
+func loadMemberClusterApiServerCA(path string) ([]byte, error) {
+	path = strings.TrimSpace(path)
```
christopherthompsonwalsh-mongo, line 128: Are we sure about this?

If I pass `--member-cluster-api-server-ca=" "` I'm not sure I want to be or deserve to be saved from myself? What if I have some dumb script which really has created a file called " "(*)?

More realistically, what if I have some dumb script which has accidentally *generated the PEM file as foo.pem but _set the flag to " "_*.  I think the correct behaviour is to fail.

More generally, I think testing for flag _absence_ by testing for value _blankness_ is a probably an antipattern.

christopherthompsonwalsh-mongo (commented):
A couple small things, one of them behavioural.

PR #1493, opened 2026-08-13
Add replica-set external members and connection string generation

lucian-tosa:
# Summary

Wires external members through the replica-set controller and the Ops Manager
process/replica-set builders, introduces k8s/<namespace>/<pod> automation-config
process naming (with legacy-deployment detection so existing deployments keep
bare pod names), generates connection strings that include external members,
and includes external members in Search host seeds.

Adapts the multi-cluster Search e2e helpers to the new process naming.

Layer 3 in the VM-to-Kubernetes migration stack.

**Stacked PR 3** (stack #1505), base `pr-02-operator-auth`. This is one layer of the VM-to-Kubernetes
migration feature, split out of `vm-migration-feature-branch` so each piece can be reviewed on its own.
Review only this layer's diff; the base branch carries the layers below it.

## Proof of Work

The stack was carved from the merged feature branch (tip `6cc6feecd`) rather than replayed commit by
commit, so each layer's content is master's version of each file plus the patches of the commits
assigned to that layer and below.

Verified for this layer, on top of its base:

- `go build ./...` and `go vet ./...` clean (vet type-checks test files too)
- full unit suite green (88 packages; `api/mongodb/v1/search` is excluded — it needs
  `make envtest-assets`, which is unrelated to this change)
- changed Python files pass a syntax check

Verified for the stack as a whole: the layers carved out of `vm-migration-feature-branch` compose
back to it exactly, both as a git tree and as an on-disk worktree comparison — the only differences are the changelog
entry added on #1491 and two zero-byte rename leftovers deliberately dropped.

e2e coverage is not exercised per layer: CI runs only on the top PR of the stack, and Evergreen is
suppressed here with `[skip-ci]`. Targeted patches will be run per layer before any of these is
marked ready for review.

## Checklist

- [ ] Have you linked a jira ticket and/or is the ticket in the title?
- [ ] Have you checked whether your jira ticket required DOCSP changes?
- [x] Have you added changelog file? Not needed for this layer — the feature-level entry lives on PR #1491 at the bottom of the stack, so this PR carries the `skip-changelog` label.

PR #1544, opened 2026-09-01
CLOUDP-400899: Unify member-cluster RBAC names across member clusters

m1kola:
# Summary

`kubectl mongodb multicluster generate-member-resources` previously required a `--member-cluster` flag that stamped every rendered resource name with the cluster (`mck-member-<cluster>-*`), forcing one render per member cluster even though the output differed only in names. This PR removes the flag and unifies the member-side names to fixed `mck-member-*` — credentials `mck-member-sa`/`mck-member-token`, roles `mck-member-role-base`/`mck-member-role-multicluster`/`mck-member-pvc-resize` (+ bindings), telemetry `mck-member-cluster-telemetry` (+ binding), and the `mck-member-appdb`/`mck-member-database-pods`/`mck-member-ops-manager` workload ServiceAccounts — so a single render can be applied unchanged to every member cluster. The Helm chart's member mode drops the `memberCluster.name` requirement, and the operator derives the same fixed workload ServiceAccount names at pod construction instead of per-cluster ones.

`generate-member-registration` keeps its `--member-cluster` flag: its output (credential Secret `mck-credential-<name>` + `MemberCluster` CR) is applied to the operator's central cluster, where per-cluster names remain necessary to avoid conflicts between clusters. Only its token Secret lookup on the member cluster changes to the fixed `mck-member-token`.

Before/after: previously one `generate-member-resources --member-cluster=<cluster> …` render per cluster; now one `generate-member-resources --member-cluster-namespace=mongodb` render applied to every member cluster, including the central one.

## Proof of Work

- unit: `go test ./pkg/resourcenames/... ./pkg/kubectl-mongodb/... ./controllers/...` — updated `memberresources` render tests (fixed-name resource sets, Helm CLI parity), `memberregistration` token-lookup tests, controller ServiceAccount-name expectations
- render: base-install `helm template` byte-identical to `feature/mc-installation-ux` across three value combinations; member render carries `mongodb.com/rbac-version` on every resource; the GitOps sample RBAC YAMLs are regenerated from the CLI and byte-identical to its output
- `make precommit` clean

## Checklist

---


helm_chart/templates/member-cluster-rbac.yaml
```diff
@@ -106,7 +107,7 @@ rules:
       - deployments
     verbs:
       - deletecollection
-  # get on serviceaccounts lets the operator read the mck-member-<cluster-name>-sa
+  # get on serviceaccounts lets the operator read the mck-member-sa
```
m1kola: This will go away once #1507 is merged.

changelog/20260831_breaking_unified_member_cluster_rbac.md
```diff
@@ -0,0 +1,6 @@
+---
+kind: breaking
```
lsierant: this release note is unnecessary as it's about changing something that's part of the feature branch and not merged to master. We should only contain things that make sense for customers in comparison to the last released version.

controllers/operator/construct/multicluster/multicluster_replicaset_test.go
```diff
@@ -319,10 +319,10 @@ func TestMultiClusterStatefulSet_ServiceAccount(t *testing.T) {
 			WithClusterNum(0),
 			WithMemberCount(3),
 			construct.GetPodEnvOptions(),
-			WithServiceAccount("mck-member-foo-database-pods"),
+			WithServiceAccount("mck-member-database-pods"),
```
lsierant, line 322: general naming question: why don't we put the operator.name into the resource names? We would then remove any risk of any clashes should anyone deploy two operators in the same namespace.

lsierant (approved):
I've left comments, but you can ignore them as I assume we'll do the cleanups of release notes and other files like markdowns before merging to master. 
Also the resource naming question should be taken outside of this PR. Let's merge.

changelog/20260831_breaking_unified_member_cluster_rbac.md
m1kola: I spotted this before marking as ready for review, but apparently forgot to remove it 🤦‍♂️  Removed now in fd6174c30d87cf3c86f04bc799e10ef8c0d5a156.

controllers/operator/construct/multicluster/multicluster_replicaset_test.go
m1kola: @lsierant I remember considering this kind of naming, but I rejected it because I didn't want to create an unnecessary link/dependency between installation time setting and day 2 configuration. Also OLM users have never specified `operator.name`.

roothorp (approved):

mircea-cosbuc (approved):

m1kola:
`e2e_operator_upgrade_ops_manager` and `e2e_operator_upgrade_replica_set` are red, but it is not related. I think it is because #1493 got released and our feature branch (older codebase) is not compatible with new Automation Config process naming (`k8s/<namespace>/<pod>`). It should be good once we rebase our feature branch.

PR #1554, opened 2026-09-03
CLOUDP-400899: Render namespaced Roles instead of ClusterRoles for narrowed installs

m1kola:
# Summary

Requirement: the operator and its member clusters must be deployable without any ClusterRoles/ClusterRoleBindings unless the operator watches all namespaces — restoring MCK 1.x CLI plugin parity, where member RBAC was one namespaced Role per namespace. The Helm templates behind both the base install and `kubectl mongodb multicluster generate-member-resources` promoted to a ClusterRole whenever the binding set spanned more than one namespace.

The three RBAC templates (`operator-roles-base.yaml`, `operator-roles-pvc-resize.yaml`, `member-cluster-rbac.yaml`) now render one namespaced Role per watched/binding namespace with a per-namespace RoleBinding (a RoleBinding cannot reference a Role in another namespace, hence the duplication). In the base chart, the cluster-wide `namespaces list/watch` rule is now only rendered when watching all namespaces — the only mode in which the operator exercises it.

Previously `operator.watchNamespace="ns1,ns2"` (or `--workload-namespaces=ns1,ns2`) produced a ClusterRole + per-namespace RoleBindings; now it produces purely namespaced Roles/RoleBindings. Default single-namespace installs and member `--operator-cluster-scoped` renders are identical (verified by render diff; `config/rbac` and the checked-in gitops samples regenerate with no diff). Cluster-wide base installs (`operator.watchNamespace="*"`) keep the same ClusterRole/ClusterRoleBinding and only stop rendering a redundant operator-namespace RoleBinding that duplicated the ClusterRoleBinding. Telemetry, `clustermongodbroles`, and webhook ClusterRoles are unchanged and remain individually opt-out.

## Proof of Work

- `go test ./pkg/kubectl-mongodb/... ./cmd/kubectl-mongodb/...` — green (render tests reworked to assert the per-namespace Role shape)
- `helm unittest helm_chart` — 17/17 green, incl. new scope suites covering default / multi-namespace / cluster-wide renders for `operator-roles-base` and `operator-roles-pvc-resize`
- `helm template` diffs: base default, member default, and member `--operator-cluster-scoped` renders byte-identical; base `*` differs only by the dropped redundant RoleBinding; multi-namespace renders inspected (zero ClusterRoles; Role+RoleBinding per namespace)
- `config/rbac` (OLM) and `public/samples/multi-cluster-cli-gitops` regenerate with no diff; `check_changelog` passes
- e2e: Evergreen patch running on this PR

## Checklist

---


helm_chart/tests/operator_roles_base_scope_test.yaml
```diff
@@ -0,0 +1,92 @@
...
+    set:
+      operator.watchNamespace: "*"
+    asserts:
+      - hasDocuments:
+          count: 2
+      - isKind:
+          of: ClusterRole
+        documentIndex: 0
+      - equal:
+          path: rules[9].resources[0]
```
josvazg, line 87: **nit**
This  uses rules[9] to assert presence/absence of the namespaces rule. If a rule is added before position 9, the assertion silently passes or spuriously fails.  Should we consider checking by content (e.g., rules[*].resources containing namespaces)?

pkg/kubectl-mongodb/memberresources/memberresources_test.go
josvazg, line 188: **nit**
This passes "" as a sentinel for "no namespace" (ClusterRoleBinding). Works but intent would be clearer with an explicit constant.

helm_chart/templates/member-cluster-rbac.yaml
```diff
@@ -76,13 +81,14 @@ metadata:
     kubernetes.io/service-account.name: {{ $name }}-sa
     mongodb.com/rbac-version: {{ $rbacVersion }}
 type: kubernetes.io/service-account-token
+{{- range $idx, $ns := $bindingNamespaces }}
 ---
```
josvazg, line 85: **nit**
Here we use $ns and $namespace in close proximity. The comment explains the model, but potential for confusion in future edits.

josvazg (approved):

PR #1557, opened 2026-09-04
KUBE-415: fix: require ns binding on legacy OM API key Secret fallback

josvazg:
# Summary

The legacy <name>-admin-key Secret fallback in APIKeySecretName resolved by CR name only, allowing a same-named CR in a different namespace to inherit another CR's admin credentials. The fallback now checks a mongodb.com/v1.opsManagerNamespace label on the Secret and rejects mismatches.

BREAKING: legacy-format Secrets without this label (post-upgrade state) will cause a one-time reconciliation failure. Remediation: delete the legacy Secret (operator recreates it) or add the label manually.

H1: #3975503
SECBUG: SECBUG-5144

## Proof of Work

Includes detection/regression unit test.

## Checklist

### DIFF
--- a/controllers/operator/mongodbopsmanager_controller.go
+++ b/controllers/operator/mongodbopsmanager_controller.go
@@ -1392,7 +1392,9 @@ func (r *OpsManagerReconciler) prepareOpsManager(ctx context.Context, opsManager
 				SetNamespace(adminKeySecretName.Namespace).
 				SetName(adminKeySecretName.Name).
 				SetStringMapToData(secretData).
-				SetLabels(map[string]string{}).Build()
+				SetLabels(map[string]string{
+					omv1.OpsManagerNamespaceLabel: opsManager.Namespace,
+				}).Build()
 
 			if err := r.PutSecret(ctx, adminSecret, operatorVaultPath); err != nil {
 				return workflow.Failed(xerrors.Errorf("failed to create a secret for admin public api key. %s. The error : %w",
--- a/api/mongodb/v1/om/opsmanager_types.go
+++ b/api/mongodb/v1/om/opsmanager_types.go
@@ -42,6 +42,10 @@ const (
 	queryableBackupDefaultPort int32  = 25999
 
 	LabelResourceOwner = "mongodb.com/v1.mongodbOpsManagerResourceOwner"
+
+	// OpsManagerNamespaceLabel is set on the admin API key Secret to record
+	// the namespace of the MongoDBOpsManager CR that owns the key.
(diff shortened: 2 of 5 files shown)


controllers/operator/mongodbopsmanager_controller.go
MaciejKaras, line 1368: The /unauth endpoint will be called, but the new admin-key credentials will lack GLOBAL_ADMIN permission

controllers/operator/mongodbopsmanager_controller.go
```diff
@@ -1392,7 +1392,9 @@ func (r *OpsManagerReconciler) prepareOpsManager(ctx context.Context, opsManager
 				SetNamespace(adminKeySecretName.Namespace).
 				SetName(adminKeySecretName.Name).
 				SetStringMapToData(secretData).
-				SetLabels(map[string]string{}).Build()
+				SetLabels(map[string]string{
```
MaciejKaras, line 1395: I want to better understand what happens for users that use legacy admin-key secret. This method is not going to be called when the legacy secret already exists. It means that the label will not be added and the `APIKeySecretName()` method call in the reconciliation will return the new `qualifiedNamespacedName`. The secret with that name does not exist and reconciler will try to create new OM credentials, but they will lack the GLOBAL_ADMIN permissions -> https://github.com/mongodb/mongodb-kubernetes/pull/1557/changes#r3977187536.

I think this will basically break the Ops Manager deployment, because Operator will later use credentials with limited permissions. I would like to see the unit tests that verify what actually happens when the legacy admin credentials are present without the namespace label. I believe this will break stuff.

MaciejKaras (changes requested):
I think the current workaround will not work for the existing users. I've added couple comments to justify my concerns.

I thought about a different approach for this issue that is backward compatible. What if we migrate all the legacy secrets to the new namespaced prefixed secret in the reconciler and delete the legacy secret afterwards? This way we could deprecate the legacy secret later making sure all users were able to migrate to new namespaced prefixed secret.

josvazg:
> I think the current workaround will not work for the existing users. I've added couple comments to justify my concerns.
> 
> I thought about a different approach for this issue that is backward compatible. What if we migrate all the legacy secrets to the new namespaced prefixed secret in the reconciler and delete the legacy secret afterwards? This way we could deprecate the legacy secret later making sure all users were able to migrate to new namespaced prefixed secret.

I agree, we should get existing users a smooth transition.

Do you think we should keep working in this PR? or close and restart the fix completely?

josvazg:
We decided this fix needs a different approach enough that we better close this and start over

PR #1564, opened 2026-09-07
Fix the trigger evg -> evg by installing evergreen before using it

josvazg:
# Summary

We were using `evergreen` binary in **evergreen** without installing it first. This fixed that.

## Proof of Work

Re-run of the trigger docs snippets which exercises these fixes:
```shell
evergreen patch -p mongodb-kubernetes \
-t trigger_public_docs_snippets -v trigger_public_docs_snippets \
--param triggered_by_git_tag=1.12.0 -f -y
```
🚧[Repeat trigger_public_docs_snippets with this fix](https://spruce.corp.mongodb.com/version/6a9ff588438aff0007cb14f3)

## Checklist

### DIFF
--- a/.evergreen-functions.yml
+++ b/.evergreen-functions.yml
@@ -69,6 +69,18 @@ functions:
       working_dir: src/github.com/mongodb/mongodb-kubernetes
       binary: scripts/evergreen/setup_jq.sh
 
+  setup_evergreen_cli: &setup_evergreen_cli
+    command: subprocess.exec
+    type: setup
+    retry_on_failure: true
+    params:
+      include_expansions_in_env:
+        - workdir
+      add_to_path:
+        - ${workdir}/bin
+      working_dir: src/github.com/mongodb/mongodb-kubernetes
+      binary: scripts/evergreen/setup_evergreen_cli.sh
+
   setup_context: &setup_context # Running the first switch is important to fill the workdir and other important initial env vars
     command: shell.exec
     type: setup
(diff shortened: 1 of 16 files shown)


m1kola (approved):

.evergreen-functions.yml
```diff
@@ -69,6 +69,17 @@ functions:
       working_dir: src/github.com/mongodb/mongodb-kubernetes
       binary: scripts/evergreen/setup_jq.sh
 
+  setup_evergreen_cli: &setup_evergreen_cli
```
viveksinghggits, line 72: how is evergreen authenticated?

josvazg: Good catch!

viveksinghggits (approved):
approving, considering the fact that we have verified that `EVERGREEN_USER` and `EVERGREEN_API_KEY`  are set properly and evergreen CLI is authenticated.

josvazg:
> approving, considering the fact that we have verified that `EVERGREEN_USER` and `EVERGREEN_API_KEY` are set properly and evergreen CLI is authenticated.

actually I have to kill this one, it is not a good path. It is getting too complex to activate and I just checked the current creds do not have a way to issue patches, so I will avoid that. I need a different approach instead of asking for more permossions to keep making this more complex.

josvazg:
the current evg creds do not have permissions to issue patches in evg, so this path is unfeasible

PR #1569, opened 2026-09-08
run CVE scanning to staging and release images

helderjs:
# Summary

Adds automated CVE scanning of container images to the build-and-publish pipeline (proactive detection instead of reacting to reports).

## Proof of Work

Test to the scan Python script was added.  Result is generated at the `trivy-reports/` directory.

## Checklist

---


igor-karpukhin (approved):

.evergreen-functions.yml
```diff
@@ -667,6 +668,29 @@ functions:
           AGENT_TOOLS_VERSION: ${agent_tools_version}
           MDB_CUSTOM_AGENT_URL: ${mdb_custom_agent_url}
 
+  # Scans the image that was just built and pushed by the pipeline function for
+  # known CVEs using trivy, and notifies Slack when CRITICAL/HIGH vulnerabilities
+  # are found on mainline (staging/release) builds. The scan is notify-only:
+  # continue_on_err ensures it never blocks the build-and-publish flow.
+  trivy_scan:
```
mircea-cosbuc, line 685: I think we need an additional step to upload the scan results, similar to what we do uploading logs and test artefacts: https://github.com/mongodb/mongodb-kubernetes/blob/master/.evergreen-functions.yml#L521-L534

mircea-cosbuc (changes requested):
Looks great, we just need easy access to the report in the task artefacts.

.evergreen-functions.yml
helderjs: Done

viveksinghggits:
If possible, I think it would be really helpful if we can send a slack notification for the vulnerabilities that have fixes already available so that someone can go ahead and work on those CVEs. What do you think?

mircea-cosbuc (approved):

helderjs:
> If possible, I think it would be really helpful if we can send a slack notification for the vulnerabilities that have fixes already available so that someone can go ahead and work on those CVEs. What do you think?

There's a Slack notification will the full report already. Is that enough? Also, let's see how it goes and we can make improvements in follow up tasks.

viveksinghggits:
Just saw this message https://mongodb.slack.com/archives/C023BA9UKC7/p1789767246427569 and it looks like we should change how the messages are being sent. It's too verbose right now, sending the message only when something with fix is available would be a good idea.

PR #1578, opened 2026-09-11
KUBE-317: Scope PVC resize selection by ownerReference

helderjs:
# Summary

The PVC resize logic selected PVCs by a namespace-wide list + ambiguous name prefix regex (^<template>-<sts>-[0-9]+$), which could match foreign PVCs in the same namespace when the attacker controls the CR name and volumeClaimTemplate name via the STS spec merge.

Fix: verify PVC ownership via ownerReference (Kind=StatefulSet, UID match) before applying the name regex. PVCs not owned by the StatefulSet are rejected regardless of name match.

## Proof of Work

CI Pass

## Checklist

---

### DIFF
--- a/changelog/20260911_fix_pvc_resize_owner_reference_isolation.md
+++ b/changelog/20260911_fix_pvc_resize_owner_reference_isolation.md
@@ -1,0 +1,6 @@
+---
+kind: fix
+date: 2026-09-11
+---
+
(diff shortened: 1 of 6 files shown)


helderjs:
Kubernetes only sets ownerReference when the policy is Delete, which MCK never configures. The resize flow additionally depends on PVCs surviving the orphan-delete + recreate of the StatefulSet (which gets a new UID), so a UID-based ownership check breaks by design. PVC selection uses an ambiguous name-prefix regex over a namespace-wide List. A follow-up will need a different mechanism.

PR #1579, opened 2026-09-11
KUBE-359: Tweak release workflow

josvazg:
# Summary

Hopefully the last fixes to the GitHub workflow part of the release.

Evergreen fixes:
- Fixed YAML heredoc parser mangling ${...} by extracting inline bash to scripts/evergreen/publish_images.sh.
- Switched dry-run detection to `tag_description_override` (vars don't propagate through depends_on wildcards).
- Added `dryrun_registry_override` wire-up to route image lookups to staging.
- Made `release_info.py` apply staging override only to operator images, leaving 3rd-party images on quay.io.
- Fused draft release creation so dry-run patches auto-create a test-{version} draft when none exists.
- Removed RH_DRYRUN/REQUESTER — always force-pushes to forks; Red Hat PR is manual.
- Use `target_commit[:8]` directly for dry-run S3 staging paths instead of `get_commit_from_tag()`.
- Fixed Helm chart re-login with staging credentials during dry-run to push to `quay.io/mongodb/staging`.
- Prevented asset upload collisions by deleting existing GitHub release assets before re-uploading on re-runs.
- Added Quay staging login before `mckci_release_publish_images`.
- Created release_kind_code_snippets and release_gke_code_snippets context stubs sourcing `public_*` contexts.
- Wired snippet variants into the publish pipeline with depends_on on all 7 publishing tasks.
- Added `--force` to chart release PR pushes and preflight submit=false in dry-run mode.

## Proof of Work

```shell
% make precommit && evergreen patch -p mongodb-kubernetes \
  -v release_decider_patch \
  -t decide_release_process \
  -d "Test full release pipeline synthetic (KUBE-359)" \
  -f -y -u \
  --param triggered_by_git_tag=1.12.1 \
  --param 'tag_description_override=[new][dry-run] Testing KUBE-359 release pipeline' \
  --param target_commit=a57d64a77ead6d47778afcfd95d61972e66f9353 \
  --browse
```
✅ [Dry-run release triggered as a evergreen patch with a fake tag description](https://spruce.corp.mongodb.com/version/6ab0dd0811ac6500070a43a6)

## Checklist


scripts/evergreen/publish_images.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+set -euo pipefail
+# Extracted from .evergreen-release-publish.yml inlined shell.exec because the
+# YAML expansion parser mangled ${...} inside heredocs.
```
m1kola, line 4: Nit: I don't think it is useful context. IMO can be dropped.

scripts/release/kubectl_mongodb/utils.py
```diff
@@ -42,9 +42,12 @@ def s3_path(filename: str, version: str) -> str:
...
 
 
-# upload_assets_to_github_release uploads the release artifacts (downloaded notarized/signed staging artifacts) to
-# the GitHub release as assets.
-def upload_assets_to_github_release(asset_paths: list[str], release_version: str):
+# upload_assets_to_github_release uploads artifacts to a GitHub release.
+# During dry-run patches, the GitHub Actions workflow didn't create the draft
+# release, so we auto-create it here if it doesn't exist.
+def upload_assets_to_github_release(
+    asset_paths: list[str], release_version: str, is_dryrun: bool = False, target_commitish: str | None = None
```
m1kola, line 49: TIL: [committish](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-commit-ishalsocommittish).

Not sure that sticking to git vocabulary is necessary. But if we do - then apparently it should be `committish` (two `t`s).

scripts/release/create_chart_release_pr.py
```diff
@@ -94,7 +94,9 @@ def commit_and_push_chart(chart_version):
             # Constructs a URL like https://x-access-token:YOUR_TOKEN@github.com/owner/repo.git
             authenticated_url = f"https://x-access-token:{github_token}@{REPO_URL.split('//')[1]}"
             run_command(["git", "remote", "set-url", "origin", authenticated_url], cwd=helm_repo_path)
-            run_command(["git", "push", "-u", "origin", branch_name], cwd=helm_repo_path)
+            run_command(
+                ["git", "push", "--force", "-u", "origin", branch_name], cwd=helm_repo_path
+            )  # --force so the chart branch is re-pushable on re-runs for the same version.
```
m1kola, line 99: Nit: Might not be necessary (unlikely that we will be manually committing & pushing), but won't hurt.

Image

scripts/release/publish-openshift-bundles-to-rh.sh
```diff
@@ -143,11 +126,7 @@ publish_repo() {
   git -C "${repo_dir}" add "operators/${PACKAGE}/${VERSION}"
   git -C "${repo_dir}" commit --quiet --signoff -m "operator ${PACKAGE} (${VERSION})"
 
-  if [[ "${RH_DRYRUN}" == "false" ]]; then
-    git -C "${repo_dir}" push -fu origin "${branch}"
-  else
-    echo "RH_DRYRUN=${RH_DRYRUN}: skipping push (branch prepared locally at ${repo_dir})"
-  fi
+  git -C "${repo_dir}" push -fu origin "${branch}"
```
m1kola, line 129: Nit: Same as above about force with lease.

m1kola (approved):
Mostly nits.

General observation: dry-run adds quite a lot of complexity, IMO. I'm not sure that complexity justifies the benefits of having a dry run.

scripts/evergreen/publish_images.sh
josvazg: I will either follow up on only fix if I need to add another actual fix here.

scripts/release/kubectl_mongodb/utils.py
josvazg: will also follow up, unless a bug forces to change this

scripts/release/create_chart_release_pr.py
josvazg: will use --force-with-lease on a follow up
we need to force mostly in dryruns to retry, but can also happen in  case a release hits a problem and needs a retry

scripts/release/publish-openshift-bundles-to-rh.sh
josvazg: will do

will apply your suggestions on a local commit, will only apply if a fix is needed. Otherwise I will move that commit to a follow up

josvazg:
> Mostly nits.
> 
> General observation: dry-run adds quite a lot of complexity, IMO. I'm not sure that complexity justifies the benefits of having a dry run.

I disagree, last week I executed no less than a dozen dryruns before I reached a green status. If I had to wait for each release to make a test, I will be releasing this in 2028.

Once we deliver the feature, we can review whether or not we want to keep this dryrun complexity, or if it is worth investing on an alternative release revamp that is more easily testable going forward.

The underlying issue here is we lean way too much in the CI machinery wiring for evergreen or Github, instead of depending on it very lightly so we can test the bulk of the logic outside of those environments. But once you rely on them so deep, it is hard to go back.

scripts/release/kubectl_mongodb/utils.py
josvazg: seems this is a PyGithub API param name, it must be `target_commitish`, so will not fix

m1kola:
> I disagree, last week I executed no less than a dozen dryruns before I reached a green status. If I had to wait for each release to make a test, I will be releasing this in 2028.

@josvazg just to clarify as it is not clear from my initial comment. I meant keeping it long-term. While in active development and testing I can clearly see the benefit. After migration it will likely going to be a dead weight.

lucian-tosa (approved):

PR #1585, opened 2026-09-14
KUBE-315: bound Ops Manager pagination and request timeouts, validate baseUrl

abdu681:
# Summary

When the operator needs a list from Ops Manager (agents, organizations, projects), it fetches it page by page and trusts Ops Manager to say when the list ends. The Ops Manager URL comes from the project ConfigMap, which users can edit. A misconfigured or unresponsive Ops Manager can claim the list is huge and always offer one more page, so the operator keeps fetching forever. The operator handles changes with a small pool of workers, and each endless fetch ties up one of them. Once every worker is stuck, the operator stops responding to any change in every namespace it watches. Nothing crashes, so the crash protection never kicks in. It just waits.

The PR changes:
  - A list fetch now stops after 100 pages or 1 minute, whichever comes first, and reports an error. 100 pages covers 10,000 agents at the default page size.
  - Every Ops Manager request now has a 2-minute timeout, so a server that never responds can't hold the operator indefinitely.
  - If a reconcile is cancelled while a fetch is in progress, the fetch stops right away instead of finishing the current page first.
  - The Ops Manager URL from the project ConfigMap is now checked for a valid scheme and host before it is used.
  - After deploying a cluster, the operator waits for the MongoDB agents to appear in Ops Manager. That wait used to retry up to ten times per component with no time limit on each attempt, so a sharded cluster could wait close to an hour. Each wait now stops after 3 minutes and reports an error.

Tests: the page cap and time limit, the request timeout against a server that never responds, accepted and rejected URL values, and the registration wait deadline.

## Proof of Work

[EVG Patch Link](https://spruce.corp.mongodb.com/version/6aa2eff27d6aa1000763f9b1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

- `gofmt -l` clean, `go build ./...` clean, `go vet` clean.
- `go test ./controllers/om/... ./controllers/operator/agents/... ./pkg/util/`
  all pass.

## Checklist

---

### DIFF
--- a/controllers/om/api/http.go
+++ b/controllers/om/api/http.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
@@ -51,6 +52,9 @@ const (
 	defaultRetryWaitMin = 1 * time.Second
 	defaultRetryWaitMax = 10 * time.Second
 	defaultRetryMax     = 3
+
+	// defaultRequestTimeout bounds a single HTTP attempt (connection, response headers and body) against Ops Manager.
+	defaultRequestTimeout = 2 * time.Minute
 )
 
 type Client struct {
@@ -66,9 +70,9 @@ type Client struct {
 
 // NewHTTPClient is a functional options constructor, based on this blog post:
 // https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis
-// The default clients specifies some important timeouts (some of them are synced with AA one):
-// 10 seconds for connection (TLS/non TLS)
-// 10 minutes for requests (time to get the first response headers)
+// The default client uses http.DefaultTransport and bounds every attempt with defaultRequestTimeout. Failed
+// requests are retried up to defaultRetryMax times, so an overall deadline can only be enforced by the context
+// passed to RequestWithContext.
 func NewHTTPClient(options ...func(*Client) error) (*Client, error) {
 	client := &Client{
 		Client: newDefaultHTTPClient(),
@@ -79,7 +83,7 @@ func NewHTTPClient(options ...func(*Client) error) (*Client, error) {
 
 func newDefaultHTTPClient() *retryablehttp.Client {
 	return &retryablehttp.Client{
-		HTTPClient:   &http.Client{Transport: http.DefaultTransport},
+		HTTPClient:   &http.Client{Transport: http.DefaultTransport, Timeout: defaultRequestTimeout},
 		RetryWaitMin: defaultRetryWaitMin,
 		RetryWaitMax: defaultRetryWaitMax,
 		RetryMax:     defaultRetryMax,
@@ -160,17 +164,23 @@ func OptionCAValidate(ca string) func(client *Client) error {
 
 // Request executes an HTTP request, given a series of parameters, over this *Client object.
 // It handles Digest when needed and json marshaling of the `v` struct.
+// Use RequestWithContext to enforce an overall deadline.
 func (client *Client) Request(method, hostname, path string, v interface{}) ([]byte, http.Header, error) {
+	return client.RequestWithContext(context.Background(), method, hostname, path, v)
+}
+
+// RequestWithContext is like Request, but aborts the whole exchange (including retries and backoff waits) as soon as ctx is done.
+func (client *Client) RequestWithContext(ctx context.Context, method, hostname, path string, v interface{}) ([]byte, http.Header, error) {
 	url := hostname + path
 
-	req, err := createHTTPRequest(method, url, v)
+	req, err := createHTTPRequest(ctx, method, url, v)
 	if err != nil {
 		return nil, nil, apierror.New(err)
 	}
 
 	if client.username != "" && client.password != "" {
 		// Only add Digest auth when needed.
-		err = client.authorizeRequest(method, hostname, path, req)
+		err = client.authorizeRequest(ctx, method, hostname, path, req)
 		if err != nil {
 			return nil, nil, err
 		}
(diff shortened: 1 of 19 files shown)


viveksinghggits:
Hi @abdu681 ,
The PR description is too big right now, do you think we can make it shorter 1-2 paragraphs so that it can easily be understood why exactly the change is being made and other important information.

viveksinghggits:
Thank you @abdu681 , the PR desc is perfect now.

controllers/om/ompaginator.go
```diff
@@ -1,10 +1,30 @@
...
+	"fmt"
+	"time"
+
+	"golang.org/x/xerrors"
+)
+
+const (
+	// maxPages is the hard upper bound on the number of pages TraversePages reads: 10,000 automation agents at Ops
+	// Manager's default of 100 per page, or 50,000 organizations at the itemsPerPage=500 the callers request.
+	maxPages = 100
```
viveksinghggits, line 15: should this limitation (hard upper bound) be mentioned in our docs somewhere?

controllers/om/ompaginator.go
```diff
@@ -26,55 +46,75 @@ func (o OMPaginated) HasNext() bool {
...
+}
+
+func applyPredicate(paginated Paginated, predicate PageItemPredicate) bool {
+	for _, entity := range paginated.Results() {
+		if predicate(entity) {
+			return true
+		}
+	}
+	return false
+}
```
viveksinghggits: nit: is `applyPredicate` supposed to apply predicate for every entity in results or just one? If it's supposed to apply it to every entity then we might have a problem here right? because as soon as first predicate is called successfully for first entity, we return.

controllers/om/omclient.go
```diff
@@ -1034,34 +1036,42 @@ func GetReplicaSetMemberIds(conn Connection) (map[string]map[string]int, error)
...
 }
 
 func (oc *HTTPOmConnection) delete(path string) error {
-	_, err := oc.httpVerb("DELETE", path, nil)
+	_, err := oc.httpVerb(context.Background(), "DELETE", path, nil)
 	return err
 }
 
-func (oc *HTTPOmConnection) httpVerb(method, path string, v interface{}) ([]byte, error) {
+func (oc *HTTPOmConnection) httpVerb(ctx context.Context, method, path string, v interface{}) ([]byte, error) {
```
viveksinghggits, line 1068: I was wondering if the callers of these methods should init the context (by using parent context) but if you think that's bigger change and unrelated don't worry about it.

viveksinghggits (approved):
Looks pretty good. Left some questions/nits.

controllers/om/ompaginator.go
```diff
@@ -26,55 +46,72 @@ func (o OMPaginated) HasNext() bool {
...
-		paginated, e := reader(i)
-		if e != nil {
-			return false, e
+	for pageNum := 2; paginated.HasNext(); pageNum++ {
+		if pageNum > maxPages {
+			return false, xerrors.Errorf("%w: stopped after %d pages", ErrPageLimitExceeded, maxPages)
+		}
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			return false, xerrors.Errorf("timed out traversing Ops Manager pages after %d pages: %w", pageNum-1, ctxErr)
 		}
```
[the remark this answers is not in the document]
abdu681: Addressed.

pkg/util/util.go
```diff
@@ -92,7 +98,13 @@ func DoAndRetry(f func() (string, bool), log *zap.SugaredLogger, count, interval
...
 		}
 		log.Debugf("%s Retrying %d/%d (waiting for %d more seconds)", msg, i+1, count, interval)
-		time.Sleep(time.Duration(interval) * time.Second)
+		timer := time.NewTimer(time.Duration(interval) * time.Second)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return false, msg
+		case <-timer.C:
+		}
```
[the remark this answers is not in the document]
abdu681: Addressed.

controllers/om/ompaginator.go
abdu681: The bound is 100 pages, which is around 10k agents in one project,  which I think is very high and anyone who hits it gets a clear error message naming the limit.

I still think it's a fair point and worth changing the docs, I will raise a DOCSP once this merges.

controllers/om/ompaginator.go
abdu681: The `applypredicate` is a search that returns true when we found what we are looking for, stopping at the first match or returning false if nothing on the page is found. This bit is used to decide whether to fetch the next page or not. 

I do see how the name read like apply to all, so I renamed it to anyItemMatches in latest commit.

controllers/om/omclient.go
abdu681: I agree, it is a bigger change which is why I have chosen to do this in a separate PR #1557.  In that PR the  get/put/patch/delete helpers take the context from the callers and the context.Background() calls are gone.

controllers/om/api/http.go
```diff
@@ -187,7 +197,7 @@ func (client *Client) Request(method, hostname, path string, v interface{}) ([]b
 func (client *Client) RequestWithAgentAuth(method, hostname, path string, agentAuth string, v interface{}) ([]byte, http.Header, error) {
 	url := hostname + path
 
-	req, err := createHTTPRequest(method, url, v)
+	req, err := createHTTPRequest(context.Background(), method, url, v)
```
helderjs: **q**: Shouldn't the context comming from the func parameter?

helderjs (approved):
LGTM! My only point is the same raised by Vivek about the `context.Background()`

